PR #8, opened 2020-02-28
Support local experiments

jonathanmetzman:
This PR will enable running experiments (mostly) locally.
I have used it successfully (with some manual initialization) to execute dispatcher.py directly on my host to run a small experiment of two trials with fairfuzz and libFuzzer on my desktop.

It still requires Google Cloud Buckets. It's possible to replace these with filesystem paths but some work must be done to ensure that the paths are accessible from the various docker images this uses.

Still working on getting run_experiment and the dispatcher-image to work so that this can work regardless of host operating system.

Some sharp edges will remain around certain fuzzers like HonggFuzz which will use more than one core by default, this PR won't try to fix these unless it can be done in a general way.


common/gcloud.py
```diff
@@ -131,3 +138,17 @@ def set_default_project(cloud_project: str):
     """Set default project for future gcloud and gsutil commands."""
     return new_process.execute(
         ['gcloud', 'config', 'set', 'project', cloud_project])
+
+
+# pylint: disable=unused-argument
```
inferno-chromium: don't see a point of keeping these unused arguments, since for local we dont need them.
can even rename function to more clear
run_local_instance (since it is just running an instance of startup script locally).

common/gcloud.py
```diff
@@ -131,3 +138,17 @@ def set_default_project(cloud_project: str):
...
+def local_create_instance(instance_name: str,
+                          instance_type: InstanceType,
+                          config: dict,
+                          metadata: dict = None,
+                          startup_script: str = None,
+                          **kwargs) -> bool:
+    """Does the equivalent of "create_instance" for local experiments, runs
+    |startup_script| in the background."""
+    command = ['/bin/bash', startup_script]
```
inferno-chromium: i dont think /bin/bash is needed.

common/logs.py
```diff
@@ -156,6 +156,8 @@ def log(logger, severity, message, *args, extras=None):
     then |extras| is also logged (in addition to default extras)."""
     message = str(message)
     if utils.is_local():
+        if extras:
+            message += ' extras: ' + str(extras)
```
inferno-chromium: maybe ", extras:"

docker/base-image/Dockerfile
```diff
@@ -29,12 +29,13 @@ RUN apt-get update -y && apt-get install -y \
     libffi-dev \
     virtualenv \
     libbz2-dev \
-    liblzma-dev
+    liblzma-dev \
```
inferno-chromium: What are these for ? Just curious.

docker/build.mk
```diff
@@ -145,6 +146,13 @@ debug-$(1)-$(2): .$(1)-$(2)-runner
     -it \
     $(BASE_TAG)/runners/$(1)/$(2)
 
+else
+
```
inferno-chromium: Maybe add a comment that coverage builds don't need runners.

experiment/build/builder.py
```diff
@@ -22,30 +22,27 @@
...
 from common import filesystem
 from common import fuzzer_utils
 from common import gsutil
 from common import logs
-from common import new_process
 from common import utils
+from experiment.build import build_utils
+
+if not utils.is_local_experiment():
+    import experiment.build.gcb_build as buildlib
```
inferno-chromium: Better to do this via classes and inherit gcb and local via some base Build class.

experiment/build/build_utils.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
```
inferno-chromium: shebang is not required, also other files don't have this.

experiment/build/local_build.py
```diff
@@ -0,0 +1,102 @@
...
+from common import environment
+from common import experiment_path as exp_path
+from common import gsutil
+from common import logs
+from common import new_process
+from common import utils
+from experiment.build import build_utils
+
+# Maximum time to wait for a GCB config to finish build.
+GCB_BUILD_TIMEOUT = 2 * 60 * 60  # 2 hours.
```
inferno-chromium: Why are GCB* var required here ?

experiment/build/local_build.py
```diff
@@ -0,0 +1,102 @@
...
+    result = make(image_name)
+    if result.retcode:
+        return result
+    local_copy_coverage_binaries(benchmark)
+    return result
+
+
+def local_copy_coverage_binaries(benchmark):
+    """Copy coverage binaries in a local experiment."""
+    shared_volume = os.environ['SHARED_VOLUME']
```
inferno-chromium: What is this, needs to come via some experiment config attribute, not a env variable (undocumented).

experiment/build/local_build.py
```diff
@@ -0,0 +1,102 @@
...
+
+    return gsutil.cp(coverage_build_archive_shared_volume_path,
+                     coverage_build_archive_gcs_path)
+
+
+def build_fuzzer_benchmark(fuzzer: str, benchmark: str) -> bool:
+    """Builds |benchmark| for |fuzzer|."""
+    image_name = 'build-{}-{}'.format(fuzzer, benchmark)
+    make(image_name)
+    return False
```
inferno-chromium: Why return False ?

experiment/dispatcher.py
```diff
@@ -92,7 +92,11 @@ def dispatcher_main():
     # Set this here because we get failures if we do it in measurer for some
     # reason.
     multiprocessing.set_start_method('spawn')
-    builder.gcb_build_base_images()
+    db_utils.initialize()
+    if os.getenv('LOCAL_EXPERIMENT'):
```
inferno-chromium: shouldn't this if block be inside db_utils.initialize.

experiment/run_experiment.py
```diff
@@ -41,6 +42,10 @@
 FUZZER_NAME_REGEX = re.compile('^[a-z0-9_]+$')
 EXPERIMENT_CONFIG_REGEX = re.compile('^[a-z0-9-]{0,30}$')
 
+# Is dispatcher code being run manually (useful for debugging)?
+MANUAL_EXPERIMENT = os.getenv('MANUAL_EXPERIMENT')
```
inferno-chromium: Not good to initialize using env var in globals.

experiment/build/local_build.py
jonathanmetzman: The problem with using experiment configs is the way we currently do it, they have to be plumbed deep into the depths of the code to be useful. I may switch this to a model where we have a config module that caches the config and can be queried from anywhere but right now, I think env vars are an acceptable alternaitve.

experiment/dispatcher.py
jonathanmetzman: I don't think so. `initialize` is called by any process using the db. This code is meant to be called only once.

experiment/build/builder.py
jonathanmetzman: those variables weren't needed so essentially functions in *_build won't share state so a class isn't needed. I'm not sure if it would be a cleaner interface, but this seems fine to me (similar to posixpath vs ntpath in os.path).

common/gcloud.py
jonathanmetzman: OK changed.

common/gcloud.py
jonathanmetzman: Why do you think so? I'm running a script that wasn't marked executable.

experiment/build/local_build.py
jonathanmetzman: Removed this.

jonathanmetzman:
@inferno-chromium @mbarbella-chromium @lmrs2 @lszekeres 
Done with all comments. Please take another look everyone.

experiment/run_experiment.py
```diff
@@ -242,15 +266,114 @@ def copy_resources_to_bucket(config_dir: str, config: Dict):
...
+        set_experiment_arg = 'EXPERIMENT={experiment}'.format(
+            experiment=self.config['experiment'])
+        set_cloud_project_arg = 'CLOUD_PROJECT={cloud_project}'.format(
+            cloud_project=self.config['cloud_project'])
+        set_cloud_experiment_bucket_arg = (
+            'CLOUD_EXPERIMENT_BUCKET={cloud_experiment_bucket}'.format(
+                cloud_experiment_bucket=self.config['cloud_experiment_bucket']))
+        docker_image_url = '{base_docker_tag}/dispatcher-image'.format(
+            base_docker_tag=base_docker_tag)
+        volume_arg = '{home}/.config/gcloud:/root/.config/gcloud'.format(
```
inferno-chromium, line 327: Is gcloud credentials for gcs. this is still pretty ugly, we need to remove gcs dependency sometime.

experiment/run_experiment.py
```diff
@@ -242,15 +266,114 @@ def copy_resources_to_bucket(config_dir: str, config: Dict):
...
+            set_cloud_experiment_bucket_arg,
+            '-e',
+            'LOCAL_EXPERIMENT=True',
+            '--cap-add=SYS_PTRACE',
+            '--cap-add=SYS_NICE',
+            '--name=dispatcher-container',
+            docker_image_url,
+            '/bin/bash',
+            '-c',
+            'echo ${CLOUD_EXPERIMENT_BUCKET}/${EXPERIMENT}/input && '
```
inferno-chromium, line 362: Why echo ?

experiment/run_experiment.py
```diff
@@ -242,15 +266,114 @@ def copy_resources_to_bucket(config_dir: str, config: Dict):
...
+            'gsutil -m rsync -r '
+            '"${CLOUD_EXPERIMENT_BUCKET}/${EXPERIMENT}/input" ${WORK} && '
```
inferno-chromium, line 364: This means this requires an experiment to have already happened and these stuff built and then local experiment uses it. i dont think we can call this local experiment then. local experiment should be end-to-end building of things.

common/utils.py
```diff
@@ -48,6 +50,11 @@ def is_local():
     return _is_local
 
 
+def is_local_experiment():
```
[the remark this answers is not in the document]
jonathanmetzman: Done.
Though I think we have too many modules.
Isn't everything in this project "experiment specific"?
The comment for experiment_utils.py is: `"""Config management."""` This doesn't seem correct and I don't really know what `experiment_utils.py` is anymore (it seems to be misc utilities that are too specific to fuzzbench to put in utils.py).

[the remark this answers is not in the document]
inferno-chromium: Ignore this comment.

experiment/run_experiment.py
jonathanmetzman: Yep. I agree about GCS dependancy.

experiment/run_experiment.py
jonathanmetzman: Debug code, sorry.

experiment/run_experiment.py
jonathanmetzman: >This means this requires an experiment to have already happened and these stuff built and then local experiment uses it.

It doesn't mean that. The directory contains things that were synced earlier by this file (which runs on the host).

experiment/run_experiment.py
```diff
@@ -242,15 +266,114 @@ def copy_resources_to_bucket(config_dir: str, config: Dict):
...
+        """Noop in local experiments."""
+
+    def start(self):
+        """Start the experiment on the dispatcher."""
+        shared_volume_dir = os.path.abspath('shared-volume')
+        if not os.path.exists(shared_volume_dir):
+            os.mkdir(shared_volume_dir)
+        shared_volume_volume_arg = '{0}:{0}'.format(shared_volume_dir)
+        shared_volume_env_arg = 'SHARED_VOLUME={}'.format(shared_volume_dir)
+        sql_database_arg = 'SQL_DATABASE_URL=sqlite:///{}'.format(
```
lszekeres, line 306: why hardcode sqlite? why use different type of database in cloud and locally? if we think about not just local machines, but other platforms, continuing to use postgres might be more generic, especially if we run the db in a docker container. for the local case we could have a script that sets up https://hub.docker.com/_/postgres. wdyt?

experiment/run_experiment.py
lszekeres: would it be more work to start with abstracting away the file storage system first? by that i mean adding a generic file interface in front of the current GCS client implementation, and then add a local file system implementation of the same interface. there are also libraries for this, that support GCS, Local FS, and other cloud storage systems as well (e.g., https://pypi.org/project/cloudstorage/).

experiment/dispatcher.py
```diff
@@ -118,7 +122,7 @@ def dispatcher_main():
     scheduler_loop_thread = threading.Thread(target=scheduler.schedule_loop,
                                              args=(experiment.config,))
     scheduler_loop_thread.start()
-    measurer_loop_thread = threading.Thread(
+    measurer_loop_thread = multiprocessing.Process(
```
lszekeres, line 125: When running locally, will this schedule all the trials at once in parallel?

experiment/measurer.py
```diff
@@ -67,6 +67,10 @@ def remote_dir_exists(directory: pathlib.Path) -> bool:
 
 def measure_loop(experiment: str, max_total_time: int):
```
lszekeres, line 68: How many measures will be run in parallel when running locally?

experiment/dispatcher.py
jonathanmetzman: Yes. We can add something to limit this in a later iteration.

experiment/run_experiment.py
jonathanmetzman: >why hardcode sqlite?

This was the easiest way to get things up and running quickly. 
I think a full db is probably overkill for locally run experiments, we may not even need it for the service, I only use it because it makes it easier to deal with persistence, multiple experiments, etc.

>why use different type of database in cloud and locally?

We also use sqlite for testing, it seems to work fine as a drop in replacements because of SQLAlchemy.

experiment/measurer.py
jonathanmetzman: As many as there are cores on the machine. This will interfere with experiments so we probably need to add something to limit this later on.

experiment/run_experiment.py
jonathanmetzman: It's easier for me to assume we can use GCS at first, and then remove this dependancy than doing everything at once. I like that library and we can try using it in a later iteration.

jonathanmetzman:
> Once this PR/branch stabilizes/converges, are you planning to break it up into smaller incremental changes? That'd help code reviews a lot.

I realize and will try to keep this in mind for the future. GitHub doesn't make it easy to break up a PR like gerrit does and I don't think the effort is worth it at this point. Usually I try to develop features with this in mind, but this feature was initially developed very quickly right before the launch to see if I could support local experiments so I didn't have a chance to do many PRs.

experiment/run_experiment.py
lszekeres: sounds good

experiment/run_experiment.py
lszekeres: Sure, whatever is the easiest is fine for this CL, however it might simplify things in the long run if we used the same components in all environments/platforms, including e.g. a small local cluster. Especially when the work is distributed on multiple machines/VMs, a db server sounds more appropriate.

lszekeres:
> I realize and will try to keep this in mind for the future. GitHub doesn't make it easy to break up a PR like gerrit does and I don't think the effort is worth it at this point. Usually I try to develop features with this in mind, but this feature was initially developed very quickly right before the launch to see if I could support local experiments so I didn't have a chance to do many PRs.

Yes, PR chains are def a pain on GH. One thing I sometimes do though is reorganize the commits in the PR itself. I.e., soft reset the 42 "tmp" and "fix" commits, re-create meaningful commits with proper descriptions, and force push the cleaned up branch. GH allows reviewing changes commit by commit, and since it's much easier to review smaller incremental changes, you can get much better feedback. You might want to consider doing that if you think it's worth the effort for getting better reviews.

experiment/build/builder.py
```diff
@@ -22,30 +22,29 @@
...
 # Even though it says queueing happen, we end up exceeding limits on "get", so
 # be conservative.
-GCB_MAX_CONCURRENT_BUILDS = 30
-
-# Maximum time to wait for a GCB config to finish build.
-GCB_BUILD_TIMEOUT = 2 * 60 * 60  # 2 hours.
-
-# High cpu configuration for faster builds.
-GCB_MACHINE_TYPE = 'n1-highcpu-8'
+MAX_CONCURRENT_BUILDS = 30
```
lszekeres, line 47: Can you also add to the comment above why 30?

experiment/build/local_build.py
```diff
@@ -0,0 +1,97 @@
...
+from common import gsutil
+from common import logs
+from common import new_process
+from common import utils
+from experiment.build import build_utils
+
+logger = logs.Logger('builder')  # pylint: disable=invalid-name
+
+
+def make(*args):
```
lszekeres, line 33: why not take a list?

experiment/build/local_build.py
```diff
@@ -0,0 +1,97 @@
...
+    """Invoke |make| with |args| and return the result."""
+    assert args
+    command = ['make', '-j'] + list(args)
+    return new_process.execute(command, cwd=utils.ROOT_DIR)
```
lszekeres, line 37: Might want to parameterize the maximum parallelism too (-j N)

experiment/test_scheduler.py
```diff
@@ -68,11 +69,112 @@ def create_trial(experiment, time_started=None, time_ended=None):
       'fuzz-target'),
      ('bloaty_fuzz_target', 'gcr.io/fuzzbench/oss-fuzz/runners/fuzzer-a/bloaty',
       'fuzz_target')])
+def test_create_trial_instance(benchmark, expected_image, expected_target,
+                               experiment_config):
+    """Test that create_trial_instance invokes create_instance
+    and creates a startup script for the instance, as we expect it to."""
+    expected_format_string = '''#!/bin/bash
+# Copyright 2020 Google LLC
```
lszekeres, line 77: consider putting these into golden files

PR #35, opened 2020-03-02
Add link to raw data in doc.

lszekeres:


lszekeres:
> LGTM.
> I think eventually it would be best to provide a link to the corpora and this data from the reports themselves.
Yes, this was added to the report too in 3d88c98e6e2c46548d235d50f472f94da46d2874.

PR #37, opened 2020-03-02
Add blogpost link

jonathanmetzman:


inferno-chromium:
Do we want to do it this style
https://google.github.io/oss-fuzz/reference/useful-links/#blog-posts
with date, and both blog links ?

PR #39, opened 2020-03-02
AFL++: Radamsa and MOpt are mutually exclusive.

andreafioraldi:
If you try to run afl++ with -L and -R, in the most recent release,
the fuzzer will abort with this message:
"MOpt and Radamsa are mutually exclusive. We accept pull requests that integrates MOpt with the optional mutators (custom/radamsa/redqueen/...)".
Yes this was our fault, not enough time to work on it, MOpt integration with the other mutators is one of our GSoC proposals.

### DIFF
--- a/fuzzers/aflplusplus/builder.Dockerfile
+++ b/fuzzers/aflplusplus/builder.Dockerfile
@@ -21,8 +21,7 @@ FROM $parent_image
 RUN git clone https://github.com/vanhauser-thc/AFLplusplus.git /afl && \
     cd /afl && \
     git checkout 842cd9dec3c4c83d660d96dcdb3f5cf0c6e6f4fb && \
-    AFL_NO_X86=1 make PYTHON_INCLUDE=/ && \
-    make radamsa
+    AFL_NO_X86=1 make PYTHON_INCLUDE=/
 
 # Use afl_driver.cpp from LLVM as our fuzzing library.
 RUN apt-get update && \
--- a/fuzzers/aflplusplus/fuzzer.py
+++ b/fuzzers/aflplusplus/fuzzer.py
@@ -46,6 +46,4 @@ def fuzz(input_corpus, output_corpus, target_binary):
             # is also recommended in a short-time scale evaluation.
             '-L',
             '0',
-            # Enable Radamsa mutator.
-            '-R',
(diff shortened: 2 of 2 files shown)


andreafioraldi:
@googlebot I signed it!

jonathanmetzman:
Thanks!
Is it mopt compatible with aflfast?
If you get a chance, it would be interesting to see what other knobs we should be using.
I think in most cases we would be benchmarking the fuzzer with its defaults but AFL++ seems to be an exception since a lot of features are opt-in.

andreafioraldi:
I'll try to add combinations of features as different fuzzers like aflplusplus_redqueen aflplusplus_lafintel

andreafioraldi:
I'll also change the build script, we are deprecating SanitizerCoverage, we loose a lot of speed using trace-pc-guard instead of inline instrumentation (and all the cool features like laf-intel, whitelist, instrim are not available with SanCov enabled)

jonathanmetzman:
>I'll try to add combinations of features as different fuzzers like aflplusplus_redqueen aflplusplus_lafintel

We have a feature that lets fuzzers support configuration at runtime (not buildtime yet) that libFuzzer supports here: https://github.com/google/fuzzbench/blob/master/fuzzers/libfuzzer/fuzzer.py#L64
This might be better for one off tests rather than ones we want to repeat, still need to figure this out.

> I'll also change the build script, we are deprecating SanitizerCoverage, we loose a lot of speed using trace-pc-guard instead of inline instrumentation 

Right, all AFL based fuzzers are paying a bit of a penalty by using this.

>(and all the cool features like laf-intel, whitelist, instrim are not available with SanCov enabled)

Sounds good!

jonathanmetzman:
> > I'll try to add combinations of features as different fuzzers like aflplusplus_redqueen aflplusplus_lafintel
> 
> We have a feature that lets fuzzers support configuration at runtime (not buildtime yet) that libFuzzer supports here: https://github.com/google/fuzzbench/blob/master/fuzzers/libfuzzer/fuzzer.py#L64
> This might be better for one off tests rather than ones we want to repeat, still need to figure this out.

For now, you can land it how you'd like to implement, I can try to clean afterwards if needed.

PR #49, opened 2020-03-03
Update honggfuzz to 77ea4dc4b499799e20ba33ef5df0152ecd113925.

### LINKED ISSUE #47: Some honggfuzz targets seem to be not instrumented
Hi, I was looking at some poorer honggfuzz results, e.g. for the systemd parser, and this log file

https://console.cloud.google.com/storage/browser/_details/fuzzbench-data/sample/experiment-folders/systemd_fuzz-link-parser-honggfuzz/trial-40413/results/fuzzer-log.txt

might suggest that the binary is not instrumented at all. I.e. it is slightly instrumented, probably for some str*cmp comparisons during final linking, but not really for trace-pc-guard,trace-cmp,trace-div,indirect-calls b/c logs like

```Size:8192 (i,b,hw,ed,ip,cmp): 0/0/0/0/0/24, Tot:0/0/0/0/0/4167```

mean, that there're no pc-guard, nor indirect-pc paths seen. Only few cmp hits.

Looking at https://github.com/google/fuzzbench/blob/master/benchmarks/systemd_fuzz-link-parser/oss-fuzz.yaml the target is somehow shared with oss-fuzz, so probably built in some special way, other than majority of targets.

Could it be some obvious problem, that you can spot here, or will it require some deeper debugging? Which I can do, just hoping you might spot the problem way faster :)

robertswiecki:
It'll allow the systemd target to be fully-instrumented.

This honggfuzz version was tested locally/manually with systemd
and with lcms fuzzers from the fuzzbench project.

Closes #47

### DIFF
--- a/fuzzers/honggfuzz/builder.Dockerfile
+++ b/fuzzers/honggfuzz/builder.Dockerfile
@@ -18,10 +18,10 @@ FROM $parent_image
 # honggfuzz requires libfd and libunwid.
 RUN apt-get update -y && apt-get install -y libbfd-dev libunwind-dev libblocksruntime-dev
 
-# Download honggfuz version 2.0.
+# Download honggfuz version 2.1 + 77ea4dc4b499799e20ba33ef5df0152ecd113925
 # Set CFLAGS use honggfuzz's defaults except for -mnative which can build CPU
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Thanks!

PR #51, opened 2020-03-03
Revert "Update honggfuzz to 77ea4dc4b499799e20ba33ef5df0152ecd113925."

inferno-chromium:
Reverts google/fuzzbench#49

Breaks on multiple benchmarks, e.g. woff2, re2

```
clang++: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated [-Wdeprecated]
+ /honggfuzz/hfuzz_cc/hfuzz-clang++ -stdlib=libc++ -O2 -fno-omit-frame-pointer -gline-tables-only -fsanitize=address backward_references.o bit_reader.o block_splitter.o brotli_bit_stream.o compress_fragment.o compress_fragment_two_pass.o decode.o dictionary.o encode.o encode_parallel.o entropy_encode.o font.o glyph.o histogram.o huffman.o literal_cost.o metablock.o normalize.o state.o static_dict.o streams.o table_tags.o transform.o utf8_util.o variable_length.o woff2_common.o woff2_dec.o woff2_enc.o woff2_out.o /honggfuzz/libhfuzz/persistent.o benchmark/target.cc -I SRC/src -o /out/fuzz-target
/honggfuzz/libhfuzz/persistent.o: In function `HF_ITER':
persistent.c:(.text+0x70): multiple definition of `HF_ITER'
/tmp/libhfuzz.0.71aa8a127231f991.a(persistent.o):persistent.c:(.text+0x70): first defined here
/honggfuzz/libhfuzz/persistent.o: In function `HonggfuzzMain':
persistent.c:(.text+0x80): multiple definition of `HonggfuzzMain'
/tmp/libhfuzz.0.71aa8a127231f991.a(persistent.o):persistent.c:(.text+0x80): first defined here
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
  File "<string>", line 1, in <module>
  File "/src/fuzzer.py", line 40, in build
    utils.build_benchmark()
  File "/src/fuzzers/utils.py", line 42, in build_benchmark
    subprocess.check_call(['/bin/bash', '-ex', build_script], env=env)
  File "/usr/local/lib/python3.7/subprocess.py", line 363, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/bin/bash', '-ex', 'benchmark/build.sh']' returned non-zero exit status 1.
The command '/bin/sh -c python3 -u -c "import fuzzer; fuzzer.build()"' returned a non-zero code: 1
```

### DIFF
--- a/fuzzers/honggfuzz/builder.Dockerfile
+++ b/fuzzers/honggfuzz/builder.Dockerfile
@@ -18,10 +18,10 @@ FROM $parent_image
 # honggfuzz requires libfd and libunwid.
 RUN apt-get update -y && apt-get install -y libbfd-dev libunwind-dev libblocksruntime-dev
 
-# Download honggfuz version 2.1 + 77ea4dc4b499799e20ba33ef5df0152ecd113925
+# Download honggfuz version 2.0.
 # Set CFLAGS use honggfuzz's defaults except for -mnative which can build CPU
 # dependent code that may not work on the machines we actually fuzz on.
 RUN git clone https://github.com/google/honggfuzz.git /honggfuzz && \
     cd /honggfuzz && \
-    git checkout 77ea4dc4b499799e20ba33ef5df0152ecd113925 && \
+    git checkout d1de86d03b2b4e332915ee1eda06e62c43daa9b6 && \
     CFLAGS="-O3 -funroll-loops" make


inferno-chromium:
@robertswiecki - we will have a CI sometime soon, so it can just seen there. i see this break on every make run-honggfuzz-<benchmark_name>

robertswiecki:
Ah.. sorry, I've tested it manually with a couple of targets (as in, I downloaded build scripts and built fuzzers locally), but I should've set up the whole environ.

Will do that later today.

robertswiecki:
Ah.. the reason seems to be this line

https://github.com/google/fuzzbench/blob/master/fuzzers/honggfuzz/fuzzer.py#L38

```C
os.environ['FUZZER_LIB'] = '/honggfuzz/libhfuzz/persistent.o'
```

The solution would be probably to set it to an empty object file. E..g created like this

```shell
$ touch empty.c
$ cc -c -o empty.o empty.c
```
And set ```FUZZER_LIB``` to it. Since hfuzz-clang(++) is used as CC/CXX, there's no need to specify any additional fuzzer libraries with honggfuzz.

I'll test it with full setup, and will send some PR later.

jonathanmetzman:
> The reason for the breakage seems to be this line:
> 
> https://github.com/google/fuzzbench/blob/master/fuzzers/honggfuzz/fuzzer.py#L38
> 
> ```c
> os.environ['FUZZER_LIB'] = '/honggfuzz/libhfuzz/persistent.o'
> ```
> 
> The solution would be probably to set it to an empty object file. E..g created like this
> 
> ```shell
> $ touch empty.c
> $ cc -c -o empty.o empty.c
> ```
> 
> And set `FUZZER_LIB` to it. Since hfuzz-clang(++) is used as CC/CXX, there's no need to specify any additional fuzzer libraries with honggfuzz.
> 
> I'll test it with full setup, and will send some PR later.

Maybe not setting FUZZER_LIB at all or setting it to the empty string works? I think that would look cleaner.

robertswiecki:
> Maybe not setting FUZZER_LIB at all or setting it to the empty string works? I think that would look cleaner.

Possibly it could break things, like if one sets it to empty string, then the compilation/linking steps would be

"cc" "-o" "fuzzer" "fuzzer.c" ""

and ld/cc/clang might complain about empty args?

```shell
$ gcc empty.c -o empty.o ""
gcc: error: : No such file or directory
```

jonathanmetzman:
> > Maybe not setting FUZZER_LIB at all or setting it to the empty string works? I think that would look cleaner.
> 
> Possibly it could break things, like if one sets it to empty string, then the compilation/linking steps would be
> 
> "cc" "-o" "fuzzer" "fuzzer.c" ""
> 
> and ld/cc/clang might complain about empty args?
> 
> ```shell
> $ gcc empty.c -o empty.o ""
> gcc: error: : No such file or directory
> ```
> 
> I mean, it depends on how compilation scripts look like per-project, but I wouldn't discard such problems since these might be make/ninja/cmake scripts

Yeah theyre pretty diverse so I could see it causing problems. How about you use the dummy object file for now?
I might hoist this solution somewhere where it can be used more generally if I see other fuzzers benefiting from it.

inferno-chromium:
> Sorry about that, I've tested it manually with a couple of targets (as in, I downloaded build scripts and built fuzzers locally), but I should've set up the whole environ.
> 
> Will do that later today.

FYI @robertswiecki 
1. our CI works now [https://github.com/google/fuzzbench/pull/52], so every new pull request should get tested with all benchmarks. It is still good to check locally using atleast 1/2 benchmarks.
2. that FETCH_HEAD failure should now be fixed with Jonathan's CL - https://github.com/google/fuzzbench/pull/60

robertswiecki:
Awesome :)

FYI, updated HF with systemd instrumentation "tests itself" in https://github.com/google/fuzzbench/pull/63

PR #52, opened 2020-03-04
Use github actions for CI

ammaraskar:
This will be useful to catch issues like https://github.com/google/fuzzbench/pull/51 early.

Some things of note:
* Building all the fuzzers with all the benchmarks currently can take up to 1h30m: https://github.com/ammaraskar/fuzzbench/actions/runs/49177543
  Instead we might want to consider building a representative subset of the benchmarks (e.g ones that use `.s` files since I remember Angora having trouble with this in the past, ones that are large and complex etc)

* afl++ actually seems to be broken right now: https://github.com/ammaraskar/fuzzbench/runs/483851588?check_suite_focus=true

```
+ clang++ -stdlib=libc++ -O2 -fno-omit-frame-pointer -gline-tables-only -fsanitize=address -fsanitize-coverage=trace-pc-guard -std=c++11 benchmark/libjpeg_turbo_fuzzer.cc -I BUILD BUILD/.libs/libturbojpeg.a /libAFL.a -o /out/fuzz-target
+ cp -r benchmark/seeds /out/
[post_build] Copying afl-fuzz to $OUT directory
[post_build] Copying libradamsa.so to $OUT directory
  File "<string>", line 1, in <module>
  File "/src/fuzzer.py", line 29, in build
    shutil.copy('/afl/libradamsa.so', os.environ['OUT'])
  File "/usr/local/lib/python3.7/shutil.py", line 248, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/local/lib/python3.7/shutil.py", line 120, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: '/afl/libradamsa.so'
The command '/bin/sh -c python3 -u -c "import fuzzer; fuzzer.build()"' returned a non-zero code: 1
make: *** [.aflplusplus-libjpeg-turbo-07-2017-builder] Error 1
docker/build.mk:114: recipe for target '.aflplusplus-libjpeg-turbo-07-2017-builder' failed
##[error]Process completed with exit code 2.
```

so the CI fails because of that. I think this is an oversight from https://github.com/google/fuzzbench/pull/39 /cc @jonathanmetzman @andreafioraldi

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,38 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        fuzzer:
+          - afl
+          - aflfast
+          - aflplusplus
+          - eclipser
+          - entropic
+          - fairfuzz
+          - honggfuzz
+          - libfuzzer
+          - mopt
+          - qsym
+
+    env:
+      FUZZER_NAME: ${{ matrix.fuzzer }}
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Python environment
+      uses: actions/setup-python@v1.1.1
+      with:
+        python-version: 3.7
+    - name: Build benchmarks
+      run: |
+        make install-dependencies
+        source .venv/bin/activate
+        make presubmit
+        make build-$FUZZER_NAME-all
--- a/docs/getting-started/adding_a_new_fuzzer.md
+++ b/docs/getting-started/adding_a_new_fuzzer.md
@@ -269,5 +269,8 @@ make build-$FUZZER_NAME-all
 
 * Run `make presubmit` to lint your code and ensure all tests are passing.
 
+* Add your fuzzer to the list in `.github/workflows/ci.yml` to enable building
+  it on continous integration.
+
 * Submit the integration in a
 [GitHub pull request](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request).


jonathanmetzman:
Great patch Ammar!

I was planning on using [this](https://github.com/google/fuzzbench/blob/master/presubmit.py#L151) in CI. This approach will also be faster because it uses GCB.

But your approach might be better since I guess changing afl can break aflfast etc and the one I implemented doesn't account for this.

I think landing this for now is a good idea regardless of what we go with in the end.
Lemme try fixing the afl++ issue i think I know what is wrong.

inferno-chromium:
We don't need to build all fuzzers, we need to hook into the test_changed_integrations command in presubmit.

Thanks for letting us know on aflplusplus failure, fixing it in https://github.com/google/fuzzbench/pull/53

ammaraskar:
Oh and I guess I should mention the upper limits here: the maximum number of entries we can have in the matrix is 256, after which we'll need to split off into another job. (though it's probably unlikely we'll hit that upper limit). 

The longest a single job can run is 6 hours

https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#usage-limits

ammaraskar:
> We don't need to build all fuzzers, we need to hook into the test_changed_integrations command in presubmit.

Aah gotcha, I think that would be a good idea for pull requests to keep the life-cycle of getting a change in fast but for pushes to master you might want to consider building everything just for a thorough check to make sure nothing breaks.

inferno-chromium:
> Great patch Ammar!
> 
> I was planning on using [this](https://github.com/google/fuzzbench/blob/master/presubmit.py#L151) in CI. This approach will also be faster because it uses GCB.
> 
> But your approach might be better since I guess changing afl can break aflfast etc and the one I implemented doesn't account for this.
> 
> I think landing this for now is a good idea regardless of what we go with in the end.
> Lemme try fixing the afl++ issue first i think I know what is wrong.

Sounds good, i think i forgot about the dependency stuff.

jonathanmetzman:
> > Great patch Ammar!
> > I was planning on using [this](https://github.com/google/fuzzbench/blob/master/presubmit.py#L151) in CI. This approach will also be faster because it uses GCB.
> > But your approach might be better since I guess changing afl can break aflfast etc and the one I implemented doesn't account for this.
> > I think landing this for now is a good idea regardless of what we go with in the end.
> > Lemme try fixing the afl++ issue first i think I know what is wrong.
> 
> Sounds good, i think i forgot about the dependency stuff.

It seems like I did too =)

jonathanmetzman:
some things I noticed from: https://github.com/google/fuzzbench/pull/55
1. I don't think we need push and pull_request.
2. I think the model from OSS-Fuzz with travis where the build show up as one check might be better. The current behavior might make it easier to spot what's failing now, but I think as the list grows it will get unwieldy.
3. We should add a job for unittests, linting, type checking (i.e. presubmit.py).

ammaraskar:
> The current behavior might make it easier to spot what's failing now, but I think as the list grows it will get unwieldy.

There's some things that we can use to control this on pull requests like https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths to only trigger the fuzzers that changed.

> We should add a job for unittests, linting, type checking and formatting (i.e. presubmit.py).

Sounds good!

jonathanmetzman:
>There's some things that we can use to control this on pull requests like https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths to only trigger the fuzzers that changed.

Right this was the approach taken by `test_changed_integrations` but the problem it suffers from is that fuzzers like aflplusplus import things from afl and it doesn't account for that. I'll probably make CI build all fuzzers/benchmarks if any of them change.

jonathanmetzman:
> Oh and I guess I should mention the upper limits here: the maximum number of entries we can have in the matrix is 256, after which we'll need to split off into another job. (though it's probably unlikely we'll hit that upper limit).
> 
> The longest a single job can run is 6 hours
> 
> https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#usage-limits

Another limitation, each time a benchmark is added build time will get worse and worse.

PR #53, opened 2020-03-04
Fix afl++ breakage, don't copy libradamso.so

inferno-chromium:


jonathanmetzman:
LGTM

inferno-chromium:
Since radamsa is not used anymore, we don't have a libradamsa.so.

PR #57, opened 2020-03-04
[CI] Only build on pull_request.

jonathanmetzman:
I think this handles pushing to pull requests as well.


jonathanmetzman:
I don't love these wasted CPU cycles actions is spending building fuzzers, but OK for now :-)

PR #58, opened 2020-03-04
Run presubmit seperately

jonathanmetzman:
This will make test failures fail in one place instead of 10, will make them easier to read and will reduce the amount of time it takes to build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,5 +34,19 @@ jobs:
       run: |
         make install-dependencies
         source .venv/bin/activate
-        make presubmit
         make build-$FUZZER_NAME-all
+
+  presubmit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Python environment
+      uses: actions/setup-python@v1.1.1
+      with:
+        python-version: 3.7
+    - name: Build benchmarks
+      run: |
+        make install-dependencies
+        source .venv/bin/activate
+        make presubmit
+    
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,20 @@ name: CI
 on: [pull_request]
 
 jobs:
+  presubmit:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -3,6 +3,20 @@ name: CI
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Python environment
+      uses: actions/setup-python@v1.1.1
+      with:
+        python-version: 3.7
+    - name: Build benchmarks
+      run: |
+        make install-dependencies
```
inferno-chromium: make install-dependencies and source .venv/bin/activate are not needed with make presubmit 
(as it does it automatically). can remove lines 16 and 17.

.github/workflows/ci.yml
```diff
@@ -3,6 +3,20 @@ name: CI
...
 jobs:
+  presubmit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Python environment
+      uses: actions/setup-python@v1.1.1
+      with:
+        python-version: 3.7
+    - name: Build benchmarks
```
inferno-chromium: This name should "Run presubmit checks"
Maybe even rename one below from "Build benchmark" to Build (that fuzzer name var) with all benchmarks ?

.github/workflows/ci.yml
```diff
@@ -46,6 +44,4 @@ jobs:
         python-version: 3.7
     - name: Build benchmarks
       run: |
-        make install-dependencies
-        source .venv/bin/activate
         make build-$FUZZER_NAME-all
```
inferno-chromium: This still requires the other commands :( @lszekeres might need to fix this.

.github/workflows/ci.yml
jonathanmetzman: Done. Ill try to figure out renaming the one below later.

PR #59, opened 2020-03-04
Make git diff work with fresh repo.

inferno-chromium:
This works with both committed and uncommitted commits.


inferno-chromium:
ah, i give up on this for now, will let you guys figure it out. origin... might fail on this new ci as well.

PR #60, opened 2020-03-04
Warn users if presubmit fails because it is on a new repo.

jonathanmetzman:


jonathanmetzman:
Downsides, this hides other errors in the git command.

Issue #46, opened 2020-03-03
The docker image has a too old LLVM version


andreafioraldi:
Hi,
I'm trying to build AFL++ w/o trace-pc using LLVM 9.
However, you base image is Ubuntu 16 that is too old and has a legacy libc++.
Also installing LLVM 9 from the LLVM repos, I cannot replace libc++ w/o recompiling it (and it is not a good idea, it heavily depends in libc).

I tried 

```
ARG parent_image=gcr.io/fuzzbench/base-builder
FROM $parent_image

# Install LLVM 9
RUN apt-get update && \
    apt-get install software-properties-common -y && \
    apt-get install wget -y && \
    wget https://apt.llvm.org/llvm.sh && \
    chmod +x llvm.sh && \
    ./llvm.sh 9 && \
    apt-get install libc++-dev -y && \
    update-alternatives --install /usr/bin/cc cc /usr/bin/clang-9 100 && \
    update-alternatives --install /usr/bin/c++ c++ /usr/bin/clang++-9 100 && \
    update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-9 100 && \
    update-alternatives --install /usr/bin/clang clang /usr/bin/clang-9 100
```

But nothing works. Can you provide a base-builder with a more recent OS?
I understand that some fuzzers need Ubuntu 16 like QSYM, but LLVM 3.8 is really too old.

vanhauser-thc:
Yes modern fuzzers should have a modern Ubuntu. How else can be use features from new llvm versions that make fuzzing better? That would make a very bad comparison ...

jonathanmetzman:
> Hi,
> I'm trying to build AFL++ w/o trace-pc using LLVM 9.
> However, you base image is Ubuntu 16 that is too old and has a legacy libc++.
> Also installing LLVM 9 from the LLVM repos, I cannot replace libc++ w/o recompiling it (and it is not a good idea, it heavily depends in libc).
> 
> I tried
> 
> ```
> ARG parent_image=gcr.io/fuzzbench/base-builder
> FROM $parent_image
> 
> # Install LLVM 9
> RUN apt-get update && \
>     apt-get install software-properties-common -y && \
>     apt-get install wget -y && \
>     wget https://apt.llvm.org/llvm.sh && \
>     chmod +x llvm.sh && \
>     ./llvm.sh 9 && \
>     apt-get install libc++-dev -y && \
[...]

PR #61, opened 2020-03-04
Install llvm tools when installing clang in base-builder.

jonathanmetzman:
This is one fix (of two) needed to build AFL's llvm_mode (not working in #46).


jonathanmetzman:
I'm not going to land "use llvm_mode instead of trace-pc-guard" in this PR.
I'll do it in another one where I enable this for all AFL fuzzers and figure out what's the issue with requiring libc++. I just want CI to test this.

jonathanmetzman:
I'm going to try to land the LLVM tools copy. I'm not sure why we switched to copying it from OSS-Fuzz instead of building it ourselves (maybe so that base-images doesn't take so long to rebuild), but obviously that approach would be less error prone than copying.

PR #62, opened 2020-03-04
[reports] Add initial implementation of progress reports

jonathanmetzman:
An interesting feature would be comparing the progress of
a fuzzer over time.
This patch allows fuzzers from two experiments to
be compared instead of combined by appending the name of
the experiment to the name.

This does not affect the behavior of fuzzbench running in
production.

### DIFF
--- a/experiment/generate_report.py
+++ b/experiment/generate_report.py
@@ -53,6 +53,13 @@ def get_arg_parser():
         default=False,
         help='If set, plots are created faster, but contain less details.')
     parser.add_argument(
+        '-p',
+        '--track_progress',
+        action='store_true',
+        default=False,
+        help='If set, then the report will track progress made in experiments')
+
+    parser.add_argument(
         '-c',
         '--from_cached_data',
         action='store_true',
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
Sorry not quite ready for review yet. I'm going to make some quick changes and prepare a demo.

jonathanmetzman:
OK, ready for review, PTAL.

experiment/generate_report.py
```diff
@@ -52,6 +52,13 @@ def get_arg_parser():
         action='store_true',
         default=False,
         help='If set, plots are created faster, but contain less details.')
+    parser.add_argument(
+        '-p',
+        '--track_progress',
```
inferno-chromium: progress seems confusing to me, since we can do the comparison at end of experiment too.

--compare_experiments ?
--diff_experiments or something feels a little more clear. @lszekeres - thoughts ?

jonathanmetzman: I switched to the more explicit "label-by-experiment"

lszekeres: label-by-experiment sgtm. perhaps label-fuzzers-by-experiment might be clearer though.

PR #63, opened 2020-03-04
Update honggfuzz to v2.1+

robertswiecki:
Update honggfuzz to v2.1+ af1b7b21dca0bd7ee586c3f17524ee41c855c5a5

This allows some targets (e.g. systemd-fuzz-link-parser), to be
instrumented correctly. See
https://github.com/google/fuzzbench/issues/47 for more on that.

It also sets LIB_FUZZER to an empty object file, to avoid problems
described in https://github.com/google/fuzzbench/pull/51 where pointing
it to an object file (libhfuzz/persistent.o) caused for some symbols
inside persistent.o and inside libhfuzz.a to be defined twice in the
final binary (and sometimes in libraries like libsystemd.so)

Tested with:
make run-honggfuzz-systemd_fuzz-link-parser
make run-honggfuzz-openthread-2019-12-23
make run-honggfuzz-mbedtls_fuzz_dtlsclient
make run-honggfuzz-openssl_x509
make run-honggfuzz-proj4-2017-08-14

### DIFF
--- a/fuzzers/honggfuzz/builder.Dockerfile
+++ b/fuzzers/honggfuzz/builder.Dockerfile
@@ -18,10 +18,14 @@ FROM $parent_image
 # honggfuzz requires libfd and libunwid.
 RUN apt-get update -y && apt-get install -y libbfd-dev libunwind-dev libblocksruntime-dev
 
-# Download honggfuz version 2.0.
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
Thanks robert. I'll try running another experiment tonight.

Issue #64, opened 2020-03-04
Process ```python3 /src/experiment/runner.py``` eats 100% of a single core on a multi-CPU machine


robertswiecki:
When running a local experiment with e.g.:

```shell
make run-honggfuzz-proj4-2017-08-14
```

The following process tree (inside docker) is created

```
root     31517  1.8  0.9 2108508 79024 ?       Ssl  16:07   8:08 /usr/sbin/dockerd -H fd://
root     31529  0.3  0.2 1912212 22624 ?       Ssl  16:07   1:35  \_ docker-containerd --config /var/run/docker/containerd/containerd.toml --log-level info
root      8625  0.0  0.1 773488 10964 ?        Sl   23:29   0:00      \_ docker-containerd-shim -namespace moby -workdir /var/lib/docker/containerd/daemon/io.containerd.runtime.v1.linux/moby/13b1d8df9fcfacc96627cc5d8c6906aa5793fad0b8b5e78935ced70f8c911db4 -address /var/run/docker/containerd/containerd.sock -containerd-binary /usr/bin/docker-containerd -runtime-root /var/run/docker/runtime-runc
root      8642  0.0  0.0   4504   748 pts/0    Ss+  23:29   0:00          \_ /bin/sh -c $ROOT_DIR/docker/benchmark-runner/startup-runner.sh
root      8677  0.0  0.0  18068  2904 pts/0    S+   23:29   0:00              \_ /bin/bash -e /src/docker/benchmark-runner/startup-runner.sh
root      8685 96.4  0.5 730840 44724 pts/0    S<l+ 23:29   6:54                  \_ python3 /src/experiment/runner.py
root      8698  0.0  0.1  44380  9592 ?        Ss   23:29   0:00                      \_ python3 -u -c import fuzzer; fuzzer.fuzz('/out/seeds', '/out/corpus', '/out/fuzz-target')
root      8706  198  0.1 730584 10024 ?        Sl   23:29  14:12                          \_ ./honggfuzz --sanitizers --persistent --input /out/seeds --output /out/corpus -na -- /out/fuzz-target
root      8716 62.9  4.8 21475465676 398192 ?  Rs   23:29   4:29                              \_ /out/fuzz-target
root      8717 63.1  4.8 21475465676 398116 ?  Rs   23:29   4:30                              \_ /out/fuzz-target
root      8718 62.8  4.8 21475465676 395052 ?  Rs   23:29   4:29                              \_ /out/fuzz-target
[...]

PR #65, opened 2020-03-05
[new_process] Use get instead of get_nowait to avoid hogging CPU.

jonathanmetzman:


jonathanmetzman:
@robertswiecki noticed this issue in #64

jonathanmetzman:
@robertswiecki I think this now keeps CPU usage below 1%, if you get a chance could you try to confirm if you see the same on your machine?

Thanks!

jonathanmetzman:
Actually, for other fuzzers it seems to be hovering around 3%. I'll probably try to come up with something better that uses even less CPU time when running in production.

common/new_process.py
```diff
@@ -71,9 +72,7 @@ def _mirror_output(process: subprocess.Popen, output_files: List) -> str:
     while True:
         # See if we can get a line from the queue.
         try:
-            # TODO(metzman): Handle cases where the process does not have utf-8
-            # encoded output.
-            line = out_queue.get_nowait().decode('utf-8', errors='ignore')
+            line = out_queue.get(timeout=WAIT_SECONDS).decode('utf-8', errors='ignore')
```
inferno-chromium: Should we sleep at end of loop (while True) using time.sleep. That way, we dont have a forced 5 sec delay for processes that immediately exit (trigger queue.empty and then exit from thread.is_alive) ?

inferno-chromium: adding @oliverchang for review as well, and how he thinks on this approach ?

jonathanmetzman: As discussed offline, this happens to be fine because runner.py is the only caller that reaches this. I'm going to refactor this tomorrow to not rely on this

jonathanmetzman:
Fixes #64

common/new_process.py
oliverchang: Seems like either approach (sleep vs timeout here) would be prone to a race. Refactor SGTM.

robertswiecki:
lgtm, I see no substantial CPU use from this process, and the logging (mirroring of stderr) looks fine.

PR #66, opened 2020-03-05
[new_process] Refactor and simplify new_process.execute

jonathanmetzman:
new_process previously used threads so that it could write fuzzer
output to stdout and a log file in real time. This allowed users
fuzzing locally to see the output of the fuzzer will still experiencing
the same behavior as in production (where a file was written to).

However, the code that did this was buggy and complex as well as
CPU intensive. This patch removes that code. Instead when fuzzing
locally, fuzzer output will only be written to stdout instead of
a log file.

Also rename LOCAL_EXPERIMENT to the more descriptive
FUZZ_OUTSIDE_EXPERIMENT since it is used for running fuzzers
outside of the context of an experiment.

### DIFF
--- a/experiment/runner.py
+++ b/experiment/runner.py
@@ -177,23 +177,33 @@ def run_fuzzer(max_total_time, log_filename):
     runner_niceness = environment.get('RUNNER_NICENESS', 0)
 
     try:
-        with open(log_filename, 'w') as log_file:
-            # Because the runner is launched at a higher priority,
-            # set it back to the default(0) for fuzzing processes.
-            new_process.execute([
-                'nice', '-n',
-                str(0 - runner_niceness), 'python3', '-u', '-c',
-                ('import fuzzer; '
-                 'fuzzer.fuzz('
-                 "'{input_corpus}', '{output_corpus}', '{target_binary}')"
-                ).format(input_corpus=shlex.quote(input_corpus),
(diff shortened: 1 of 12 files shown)


jonathanmetzman:
I need to fix this because the experiment I ran with this patch didn't have fuzzer logs.

experiment/runner.py
```diff
@@ -177,23 +177,36 @@ def run_fuzzer(max_total_time, log_filename):
...
+        ]
+
+        fuzzer_environment = _get_fuzzer_environment()
+        # Write output to stdout if user is fuzzing from command line.
+        # Otherwise, write output to the log file.
+        if environment.get('FUZZ_OUTSIDE_EXPERIMENT'):
+            new_process.execute(command,
                                 timeout=max_total_time,
-                                output_files=[log_file],
+                                output_file=sys.stdout.fileno(),
```
inferno-chromium: i dont like this, i think output_file should either mean None (no output file) or a file handle.
right now, we are doing
default - PIPE
this one - stdout print
output_file - file handle

i think we should just have another argument
print_output and hide this stuff inside newprocess.execute itself
if print_output:
 output_file = sys.stdout.fileno()
elif not output_file:
  output_file = subprocess.PIPE
then use,

kwargs['stdout'] = output_file
kwargs['stderr'] = subprocess.STDOUT

PR #67, opened 2020-03-05
[CI] Build base images and dispatcher image

jonathanmetzman:


jonathanmetzman:
I don't think this actually needed right now as all of these images (except the rarely modified dispatcher-image) are built in the individual fuzzer runs. I'll reopen when I want to test the dispatcher-image build but for now it's a very low priority.

PR #68, opened 2020-03-05
[CI] Only build fuzzer images when necessary

jonathanmetzman:


.github/workflows/ci.yml
```diff
@@ -14,34 +14,3 @@ jobs:
     - name: Run presubmit checks
       run: |
         make presubmit
-
```
inferno-chromium: maybe call this file presubmit.yml as CI name does not make sense. all stuff in this dir is CI ?

fuzzers/test
```diff
@@ -0,0 +1 @@
+hello
```
inferno-chromium: need to remember to remove this. just a test i think :)

jonathanmetzman:
This change isn't 100% necessary and may actually cause CI not to run when it should if I implemented it incorrectly. But I like it because I'd rather not see so many fuzzer builds unless a fuzzer is actually being changed. WDYT? https://github.com/google/fuzzbench/actions/runs/50224004 is proof that it works when a fuzzer is changed.

inferno-chromium:
> This change isn't 100% necessary and may actually cause CI not to run when it should if I implemented it incorrectly. But I like it because I'd rather not see so many fuzzer builds unless a fuzzer is actually being changed. WDYT? https://github.com/google/fuzzbench/actions/runs/50224004 is proof that it works when a fuzzer is changed.

That is file, might want to think of running that integration test you had in presubmit since that can make sure workflow does not break. i agree we shouldn't be building fuzzers, benchmarks for all cases, but that integration test can catch obvious breakages.

fuzzers/test
jonathanmetzman: Yeah I did.

.github/workflows/ci.yml
jonathanmetzman: Done.

PR #69, opened 2020-03-05
[aflplusplus] Use afl-clang-fast(++) as compiler

jonathanmetzman:
Use afl-clang-fast(++) as compiler instead of using clang with
trace-pc-guard for faster and better instrumentation.

### DIFF
--- a/fuzzers/aflplusplus/builder.Dockerfile
+++ b/fuzzers/aflplusplus/builder.Dockerfile
@@ -15,18 +15,21 @@
 ARG parent_image=gcr.io/fuzzbench/base-builder
 FROM $parent_image
 
+# Install wget to download afl_driver.cpp. Install libstdc++ to use llvm_mode.
+RUN apt-get update && \
+    apt-get install wget libstdc++-5-dev -y
+
 # Download and compile afl++ (v2.60c).
 # Build without Python support as we don't need it.
 # Set AFL_NO_X86 to skip flaky tests.
 RUN git clone https://github.com/vanhauser-thc/AFLplusplus.git /afl && \
     cd /afl && \
     git checkout 842cd9dec3c4c83d660d96dcdb3f5cf0c6e6f4fb && \
-    AFL_NO_X86=1 make PYTHON_INCLUDE=/
+    AFL_NO_X86=1 make PYTHON_INCLUDE=/ && \
+    cd llvm_mode && \
+    CXXFLAGS= make
 
 # Use afl_driver.cpp from LLVM as our fuzzing library.
-RUN apt-get update && \
-    apt-get install wget -y && \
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
If AFLplusplus pases all tests here I'll try doing this for all AFL based fuzzers.

jonathanmetzman:
@andreafioraldi I think this what was intended in #46 right?

andreafioraldi:
Yes, ty, after having this working I can enable the various afl++ features when I'll have time (probably this weekend)

jonathanmetzman:
> Yes, ty, after having this working I can enable the various afl++ features when I'll have time (probably this weekend)

Great! No rush.

fuzzers/aflplusplus/builder.Dockerfile
```diff
@@ -15,18 +15,21 @@
 ARG parent_image=gcr.io/fuzzbench/base-builder
 FROM $parent_image
 
+# Install wget to download afl_driver.cpp. Install libstdc++ to use llvm_mode.
+RUN apt-get update && \
+    apt-get install wget libstdc++-5-dev -y
```
inferno-chromium: Don't we have libstdc++ from OSS-Fuzz's clang image? This feels like we would need to do in every afl builder file, should this move to base-builder image ?

fuzzers/aflplusplus/fuzzer.py
```diff
@@ -17,13 +17,35 @@
...
+    utils.append_flags('CFLAGS', cflags)
+    utils.append_flags('CXXFLAGS', cflags)
+
+    os.environ['CC'] = '/afl/afl-clang-fast'
+    os.environ['CXX'] = '/afl/afl-clang-fast++'
+    os.environ['FUZZER_LIB'] = '/libAFLDriver.a'
+
+    # Some benchmarks like lcms
+    # (see: https://github.com/mm2/Little-CMS/commit/ab1093539b4287c233aca6a3cf53b234faceb792#diff-f0e6d05e72548974e852e8e55dffc4ccR212)
+    # fail to compile if the compiler outputs things to stderr in unexepcted
```
inferno-chromium: unexpected typo

fuzzers/aflplusplus/builder.Dockerfile
jonathanmetzman: No we have libc++.
I don't know why using libc++ doesn't work. I (like @kcc pointed out in our meeting) find it odd that the LLVM pass can't use LLVM's C++ standard library. I may look into this and send a patch as well.

jonathanmetzman: We don't need libstdc++ in every image because using libc++ works in most cases.

fuzzers/aflplusplus/fuzzer.py
jonathanmetzman: Fixed.

PR #71, opened 2020-03-06
[afl++] update to 2.62d

jonathanmetzman:


jonathanmetzman:
I probably @andreafioraldi can decide when to update, I just want to verify that updating works without a build error using CI, since @inferno-chromium mentioned there might is a build error (maybe that's from the top of master because I didn't get a build failure locally).

andreafioraldi:
2.62d is fine, better than 2.60c that has no checks fro incompatible mutators

jonathanmetzman:
OK landing then.

PR #74, opened 2020-03-07
Fix wrong virtualenv regression

inferno-chromium:
This causes Python 2 virtualenv to be created and python3 pip deps to be installed in user's system python packages directory. Just use venv as per https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/#creating-a-virtual-environment which picks python3 automatically.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,6 @@ SHELL := /bin/bash
 VENV_ACTIVATE := .venv/bin/activate
 
 ${VENV_ACTIVATE}:
-	python3 -m pip install --user virtualenv
 	rm -rf .venv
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -20,7 +20,7 @@ VENV_ACTIVATE := .venv/bin/activate
 ${VENV_ACTIVATE}:
 	python3 -m pip install --user virtualenv
```
ammaraskar: If the intent is to just use `venv` instead of `virtualenv` then this line can be removed.

inferno-chromium: Yes, right, removed. Thanks for noticing.

Issue #79, opened 2020-03-09
Support specifying which fuzzers to include in the report


lszekeres:
One often wants to see a report that compares only two or just a few fuzzers. E.g., compare only
- afl with libfuzzer, or
- two different versions of the same fuzzer, or
- only fuzzers with dynamic binary instrumentation, etc.

This can be supported by adding a --fuzzers flag to the [report generator](https://google.github.io/fuzzbench/developing-fuzzbench/custom_analysis_and_reports), where users can list the fuzzers they would like the report to compare. We should allow specifying fuzzers with different versions, by either tagging them with their version number or perhaps with an experiment name. E.g., `generate_report --fuzzers afl:v2.5 afl:v2.4`, or `generate_report --fuzzers afl:experiment-2020-01-15 afl:experiment-2020-02-15`.

This is also useful to have because specifying a smaller/different subset of fuzzers can affect the result of the [top level statistical analysis](https://google.github.io/fuzzbench/reference/report/#experiment-summary) (Friedman test, critical difference plot). Further, when we only compare two fuzzer, we can even do more precise statistical tests with more specific visualizations. On the benchmark level, we don't need a pairwise comparison matrix of statistical significance, since we only have a single pair. On the experiment level, we don't need to use Friedman/Nemenyi test (which compares more than two fuzzers, and is rather conservative), but we can use the [Wilcoxon signed-rank test](https://en.wikipedia.org/wiki/Wilcoxon_signed-rank_test), which is specifically designed for comparing two things (i.e., matched samples for the different benchmarks).

jonathanmetzman:
This issue is fixed.

lszekeres:
> This issue is fixed.

Not everything is fixed yet. There's also:

1) Support version/experiment tagging, e.g.,

>  `--fuzzers afl:v2.5 afl:v2.4`.

and

2) Different report when we're comparing only two things:

> when we only compare two fuzzer, we can even do more precise statistical tests with more specific visualizations. On the benchmark level, we don't need a pairwise comparison matrix of statistical significance, since we only have a single pair. On the experiment level, we don't need to use Friedman/Nemenyi test (which compares more than two fuzzers, and is rather conservative), but we can use the Wilcoxon signed-rank test, which is specifically designed for comparing two things (i.e., matched samples for the different benchmarks).

Should we reopen this issue, or would you prefer creating separate issues for these?

jonathanmetzman:
My bad, reopening is fine.

PR #80, opened 2020-03-09
[generate_report] Support fuzzer selection with -f option

jonathanmetzman:
Allow users of generate_report to specify the fuzzers they want
included in a report using `--fuzzers` (`-f` for short).

### DIFF
--- a/experiment/generate_report.py
+++ b/experiment/generate_report.py
@@ -30,7 +30,7 @@ from common import logs
 def get_arg_parser():
     """Returns argument parser."""
     parser = argparse.ArgumentParser(description='Report generator.')
-    parser.add_argument('experiment', nargs='+', help='Experiment name')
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
This does about half (the easy half) of what is asked for in #79

experiment/generate_report.py
```diff
@@ -52,6 +52,12 @@ def get_arg_parser():
         action='store_true',
         default=False,
         help='If set, plots are created faster, but contain less details.')
+
+    parser.add_argument('-f',
+                        '--fuzzers',
+                        nargs='*',
+                        help='Names of the fuzzers to include in the report. '
```
inferno-chromium: Probably fix "help='Experiment name'" above (as it allows multiple experiments).

Can remove "All of them in the experiments by default.", feels obvious.

PR #81, opened 2020-03-09
Add check for FORCE_LOCAL flag, similarly to FORCE_NOT_LOCAL.

lszekeres:


common/utils.py
```diff
@@ -28,6 +28,9 @@
     # like stackdriver logging to happen when running code locally.
     _is_local = False
 
+if os.getenv('FORCE_LOCAL'):
```
jonathanmetzman: Sorry could you please add a TODO here to get rid of using http requests for `is_local`?
Thanks!

lszekeres: Added and issue: https://github.com/google/fuzzbench/issues/82

lszekeres: Added TODO too

PR #90, opened 2020-03-11
New AFL++ fuzzer.py

andreafioraldi:
This AFL++ fuzzer.py works fine. There is the support to build configurations, but the conf is harcoded ATM.
I added also a small fix that includes -pthread by default because I experienced some errors while linking C++ (libc++ uses pthread).

### DIFF
--- a/docker/base-builder/Dockerfile
+++ b/docker/base-builder/Dockerfile
@@ -52,4 +52,4 @@ COPY --from=base-clang /usr/local/lib/libc++*.a /usr/local/lib/
 COPY --from=base-clang /usr/local/include/ /usr/local/include
 
 # Use libc++ thoughout the build.
-ENV CXXFLAGS -stdlib=libc++
+ENV CXXFLAGS -stdlib=libc++ -pthread
--- a/experiment/generate_report.py
+++ b/experiment/generate_report.py
@@ -58,11 +58,10 @@ def get_arg_parser():
         '--benchmarks',
         nargs='*',
         help='Names of the benchmarks to include in the report.')
-    parser.add_argument(
-        '-f',
-        '--fuzzers',
-        nargs='*',
-        help='Names of the fuzzers to include in the report.')
+    parser.add_argument('-f',
+                        '--fuzzers',
+                        nargs='*',
+                        help='Names of the fuzzers to include in the report.')
     parser.add_argument(
         '-l',
         '--label-by-experiment',
@@ -133,8 +132,8 @@ def main():
     args = parser.parse_args()
 
     generate_report(args.experiments, args.report_dir, args.report_name,
-                    args.label_by_experiment, args.benchmarks, args.fuzzers, args.report_type,
-                    args.quick, args.from_cached_data)
+                    args.label_by_experiment, args.benchmarks, args.fuzzers,
+                    args.report_type, args.quick, args.from_cached_data)
 
 
 if __name__ == '__main__':
--- a/fuzzers/aflplusplus/fuzzer.py
+++ b/fuzzers/aflplusplus/fuzzer.py
@@ -115,7 +115,7 @@ def fuzz(input_corpus, output_corpus, target_binary):
     afl_fuzzer.prepare_fuzz_environment(input_corpus)
     os.environ['AFL_PRELOAD'] = '/afl/libdislocator.so'
 
-    flags = []
+    flags = ["-d"] # FidgetyAFL is better when runnign alone
     if os.path.exists(cmplog_target_binary):
         flags += ["-c", cmplog_target_binary]
     if 'ADDITIONAL_ARGS' in os.environ:
--- a/fuzzers/aflplusplus/fuzzer.py
+++ b/fuzzers/aflplusplus/fuzzer.py
@@ -22,19 +22,38 @@ from fuzzers import utils
 # OUT environment variable is the location of build directory (default is /out).
 
 
+def get_cmplog_build_directory(target_directory):
+    """Return path to CmpLog target directory."""
+    return os.path.join(target_directory, 'cmplog')
+
+
 def build():
     """Build fuzzer."""
+    build_modes = []
+    if "BUILD_MODES" in os.environ:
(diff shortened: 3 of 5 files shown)


experiment/generate_report.py
```diff
@@ -58,11 +58,10 @@ def get_arg_parser():
         '--benchmarks',
```
jonathanmetzman: I think you might need to pull since this undoes a change I made recently.

fuzzers/utils.py
```diff
@@ -61,7 +61,8 @@ def append_flags(env_var, additional_flags, env=None):
     '-pthread', '-Wl,--no-as-needed', '-Wl,-ldl', '-Wl,-lm',
     '-Wno-unused-command-line-argument'
 ]
-NO_SANITIZER_COMPAT_CXXFLAGS = ['-stdlib=libc++'] + NO_SANITIZER_COMPAT_CFLAGS
+NO_SANITIZER_COMPAT_CXXFLAGS = ['-stdlib=libc++', '-pthread'] + \
```
jonathanmetzman: I don't want to make you spend more time debugging linking issues, but I'm curious why this change was necessary.
1. Why don't you use these variables?
2. Is the change to CXXFLAGS necessary given this change?
3. Why does the flag need to be added here? Is what we were doing before to append not sufficient?
4. Can you change this to
```
NO_SANITIZER_COMPAT_CXXFLAGS = (['-stdlib=libc++', '-pthread'] +
   NO_SANITIZER_COMPAT_CFLAGS)
```
We prefer using paren to backslash to break up lines in python.

fuzzers/aflplusplus/fuzzer.py
```diff
@@ -44,25 +71,91 @@ def build():
...
+    cmplog_target_binary_directory = (
+        get_cmplog_build_directory(target_binary_directory))
+    target_binary_name = os.path.basename(target_binary)
+    cmplog_target_binary = os.path.join(cmplog_target_binary_directory,
+                                        target_binary_name)
+
     afl_fuzzer.prepare_fuzz_environment(input_corpus)
+    os.environ['AFL_PRELOAD'] = '/afl/libdislocator.so'
+
+    flags = ["-d"]  # FidgetyAFL is better when runnign alone
```
jonathanmetzman: nit: typo in "running".
Also I'm not sure what the comment means. Does it mean that when starting from no seed fidgety AFL is better (I think this is the case).

docker/base-builder/Dockerfile
```diff
@@ -52,4 +52,4 @@ COPY --from=base-clang /usr/local/lib/libc++*.a /usr/local/lib/
 COPY --from=base-clang /usr/local/include/ /usr/local/include
 
 # Use libc++ thoughout the build.
-ENV CXXFLAGS -stdlib=libc++
\ No newline at end of file
+ENV CXXFLAGS -stdlib=libc++ -pthread
```
jonathanmetzman: do you remember where this failed?
If not, it's fine I can look into this..

experiment/generate_report.py
jonathanmetzman: Actually I think the formatter did this.
Sorry it wasn't applied before.

fuzzers/aflplusplus/fuzzer.py
andreafioraldi: I mean when running alone without other parallel instances of AFL++.

docker/base-builder/Dockerfile
andreafioraldi: I tested libpng and afl-clang-fast++ failed to link the harness binary due to the lack of -pthread.
The printed errors were related to functions in libc++ that use pthread.

andreafioraldi:
Yeah sorry for the late reply but I was at dinner.

jonathanmetzman:
> Yeah sorry for the late reply but I was at dinner.

Don't worry, I landed because I assumed we weren't in the same timezone, so I wasn't expecting a response until tommrrow.

andreafioraldi:
Btw you should use -d also for all other AFL based fuzzers. I noticed that you are performing the deterministic stages, this explains why MOPT results to be the best fuzzer. Deterministic stages are useful, in my experience, when having parallel instances. In AFL++ with didn't change the default behavior for retro compatibility with AFL, but in most cases it performs much better with -d, especially on structured formats like video formats.

fuzzers/aflplusplus/fuzzer.py
jonathanmetzman: Ah got it!

jonathanmetzman:
> Btw you should use -d also for all other AFL based fuzzers. I noticed that you are performing the deterministic stages, this explains why MOPT results to be the best fuzzer. Deterministic stages are useful, in my experience, when having parallel instances. In AFL++ with didn't change the default behavior for retro compatibility with AFL, but in most cases it performs much better with -d, especially on structured formats like video formats.

Interesting! I have to think about doing this because I sort of prefer using the defaults for each fuzzer. I'm fine with individual fuzzers turning on knobs but I'm less sure about doing it for all AFL fuzzers.

andreafioraldi:
AFLSmart for sure will benefit, @thuanpv will agree, they use this configuration by default (not here, probably he didn't notice that -d is missing).

thuanpv:
I support the observation of @andreafioraldi. The -d option seems to work well. It was the reason why in the evaluation of our AFLSmart paper, we used that option for both two selected baselines, AFL and AFLFast, to have a fair comparison. Please see the 3rd paragraph in subsection 5.5 of our TSE'19 paper ((https://thuanpv.github.io/publications/TSE19_aflsmart.pdf) for more information. AFLSmart algorithm is not implemented for the deterministic stage yet so -d is its default option. I think having structure-aware mutations in that stage could be beneficial but there are some technical issues there. For example, since most of our new mutators (e.g., chunk deletion, chunk duplication, chunk swapping) could change the size of the seeds, how do we keep the stage deterministic along with the normal byte/bit-level mutation operators (e.g., bit flipping). It could also be the main reason why "splicing" was not implemented as part of the deterministic stage in AFL.

PR #93, opened 2020-03-11
Show reports folder link instead of sample report.

inferno-chromium:
Also, make experiment data its own section for easy scroll and
visibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,8 @@ See [a sample report](https://www.fuzzbench.com/reports/sample/index.html).
 ## Sample Report
 
 You can view our periodically generated reports
-[here](https://www.fuzzbench.com/reports/index.html).
+[here](https://www.fuzzbench.com/reports/index.html) and a sample report
+[here](https://www.fuzzbench.com/reports/sample/index.html).
 These reports are generated using 10 fuzzers against 24 real-world benchmarks,
 with 20 trials each and over a duration of 24 hours.
 The raw data in compressed CSV format can be found at the end of each report.
--- a/README.md
+++ b/README.md
@@ -27,10 +27,11 @@ See [a sample report](https://www.fuzzbench.com/reports/sample/index.html).
 
 ## Sample Report
 
-You can view a sample report
-[here](https://www.fuzzbench.com/reports/sample/index.html).
-This report is generated using 10 fuzzers against 24 real-world benchmarks,
+You can view our periodically generated reports
+[here](https://www.fuzzbench.com/reports/index.html).
+These reports are generated using 10 fuzzers against 24 real-world benchmarks,
 with 20 trials each and over a duration of 24 hours.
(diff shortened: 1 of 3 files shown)


docs/index.md
```diff
@@ -51,11 +51,11 @@ FuzzBench, submit it, and get benchmark results.
 
 ## Sample report
```
jonathanmetzman: I think we need to change this section name.
Also, while I think it's necessary to link to the list of reports, I don't think we should get rid of the link to the sample. 
The sample shows what fuzzbench is capable of.
Showing people the list of reports may not demonstrate this since the latest report might be for a currently running experiment and/or may not contain all fuzzers/benchmarks.

analysis/report_templates/default.html
```diff
@@ -153,6 +155,7 @@ <h5 class="center-align">Mann-Whitney U test</h4>
             {% for benchmark in experiment.benchmarks %}
             <li><a href="#{{ benchmark.name }}">{{ benchmark.name }}</a></li>
             {% endfor %}
+            <li><a href="#data">experiment data</a></li>
```
lszekeres: consider something like "raw data download"?

README.md
```diff
@@ -27,10 +27,12 @@ See [a sample report](https://www.fuzzbench.com/reports/sample/index.html).
...
 ## Sample Report
 
-You can view a sample report
+You can view our periodically generated reports
+[here](https://www.fuzzbench.com/reports/index.html) and a sample report
 [here](https://www.fuzzbench.com/reports/sample/index.html).
-This report is generated using 10 fuzzers against 24 real-world benchmarks,
+These reports are generated using 10 fuzzers against 24 real-world benchmarks,
 with 20 trials each and over a duration of 24 hours.
+The raw data in compressed CSV format can be found at the end of each report.
```
lszekeres: consider: "can be found" -> "is linked"

analysis/report_templates/default.html
```diff
@@ -142,8 +142,10 @@ <h5 class="center-align">Mann-Whitney U test</h4>
         </div> <!-- id="{{ benchmark.name }}" -->
         {% endfor %}
 
-        <div class="divider"></div>
-        See the <a href="data.csv.gz">raw data</a> for this report.
+        <div id="data" class="section scrollspy">
+            <h2 class="green-text text-darken-3">experiment data</h2>
```
lszekeres: consider: "raw experiment data in csv"

inferno-chromium: feels simple is better, also matches style "experiment summary"

analysis/report_templates/default.html
inferno-chromium: feels simple is better, also matches style "experiment summary"

README.md
inferno-chromium: feels like a nit, can improve in future doc iteration?

PR #97, opened 2020-03-11
Make presubmit work for new repos and fix lint/format/type checking in CI

jonathanmetzman:


jonathanmetzman:
Ping

presubmit.py
```diff
@@ -260,17 +260,30 @@ def license_check(paths: List[Path]) -> bool:
...
+        # This probably won't happen to anyone. It can happen if your copy
+        # of the repo wasn't cloned so give instructions on how to handle.
         pass
-    raise Exception(
-        ('"%s" failed. Please run "git pull origin master --rebase" and try'
-         ' again.') % ' '.join(diff_command))
+    raise Exception((
+        '"%s" failed.\n'
+        'Please run "git fetch origin master && '
+        'git symbolic-ref refs/remotes/origin/HEAD refs/remotes/origin/master" '
```
inferno-chromium: Why not just "git pull" ?

presubmit.py
```diff
@@ -260,17 +260,30 @@ def license_check(paths: List[Path]) -> bool:
 
 def get_changed_files() -> List[Path]:
     """Return a list of absolute paths of files changed in this git branch."""
-    diff_command = ['git', 'diff', '--name-only', 'FETCH_HEAD']
+    uncommitted_diff_command = ['git', 'diff', '--name-only', 'HEAD']
```
inferno-chromium: I dont think this is needed, only origin... worked for me.

.github/workflows/presubmit.yml
```diff
@@ -7,6 +7,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - run: |  # Needed for presubmit to work.
+        git fetch origin master --depth 1
```
inferno-chromium: dont understand this, will let @oliverchang to review.
why not git pull ?

presubmit.py
jonathanmetzman: Even for files that you've changed but not modified?
I think that is why this was originally reverted.

inferno-chromium: this was reverted because it failed in CI (something with origin not existing).

.github/workflows/presubmit.yml
oliverchang: I'm assuming this is for speed?

presubmit.py
jonathanmetzman: Ah, fixed.

.github/workflows/presubmit.yml
jonathanmetzman: The `--depth 1` part is but given the weird way GitHub actions checksout a repo `git pull` won't work.
You can see for yourself [here](https://github.com/google/fuzzbench/pull/97/checks?check_run_id=501465871), but GitHub actions does the following to checkout a pull request:
```
git init
/usr/bin/git remote add origin https://github.com/google/fuzzbench
/usr/bin/git config --local gc.auto 0
/usr/bin/git config --local --name-only --get-regexp http\.https\:\/\/github\.com\/\.extraheader
/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +0c98f7fd8841b8e5c5cb01a8582f0e6a17890e74:refs/remotes/pull/97/merge
/usr/bin/git checkout --progress --force refs/remotes/pull/97/merge
```

As a result, some things that work when a repo is cloned using `git clone` don't hold so I'm working around it.
But there's a fair amount here I don't fully understand.

presubmit.py
jonathanmetzman: See other comment on this.

jonathanmetzman:
> Also don't really understand this and the GitHub actions environment.. @jonathanmetzman could you please explain what these symbolic-refs are meant to achieve?

The symbolic refs part is needed because without `git diff  --name-only origin...` fails with the following:
```
fatal: ambiguous argument 'origin...': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

PR #98, opened 2020-03-11
Move restore_directory into utils so other modules can use it.

jonathanmetzman:
Move restore_directory from qsym/fuzzer.py to fuzzers/utils.py so other fuzzers can use it.
AFLplusplus in #90 is copying it for now.


jonathanmetzman:
In theory they could use it now anyway, but I think putting in utils makes sense rather than importing it from qsym for fuzzers unrelated to qsym.

PR #99, opened 2020-03-11
[aflplusplus] Use COMPAT_CFLAGS and fix comments and formatting.

jonathanmetzman:
Fix comments and formatting issues in pull request #90.

### DIFF
--- a/docker/base-builder/Dockerfile
+++ b/docker/base-builder/Dockerfile
@@ -52,4 +52,4 @@ COPY --from=base-clang /usr/local/lib/libc++*.a /usr/local/lib/
 COPY --from=base-clang /usr/local/include/ /usr/local/include
 
 # Use libc++ thoughout the build.
-ENV CXXFLAGS -stdlib=libc++ -pthread
+ENV CXXFLAGS -stdlib=libc++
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
CC @andreafioraldi

andreafioraldi:
I added -pthread also on the base dockerfile. This is the only way that I had to get it working on my PC due to the linker error with libc++. Cannot reproduce now because I was using the PC in my office in university, tomorrow I'll revert the fix and post here the error.

jonathanmetzman:
> I added -pthread also on the base dockerfile. This is the only way that I had to get it working on my PC due to the linker error with libc++. Cannot reproduce now because I was using the PC in my office in university, tomorrow I'll revert the fix and post here the error.

Don't worry about reproducing it, I was able to.
I'm not 100% sure if it is libc++ that needs it. I think what might be happening is projects need it and asan is linked against libc++ so these benchmarks compile fine without pthread when ASAN is used but fail otherwise.

I can look into this more to determine appropriate fix but I think it is better to call `set_no_sanitizer_compilation_flags` then to add -pthread to CXXFLAGS for two reasons:
1. I think there are other flags that may be needed for some OSS-Fuzz projects that we cover in NO_SANITIZER_COMPAT_CFLAGS.
2. base-builder isn't actually used when building the OSS-Fuzz-integrated benchmarks so if -pthread is needed there.

Sorry you had to deal with this. Non-sanitizer builds are a bit of a sharp edge in this project and I'll try to make it work better (probably starting with documenting this).
We started using ASAN by default in anticipation of counting crashes, but I may disable in other fuzzers until then (right now there is a clear advantage to not using ASAN).

PR #103, opened 2020-03-16
[presubmit] Remove unnecessary git diff invocation

jonathanmetzman:
Remove unnecessary git diff invocation that was accidentally
left out of previous commit.


jonathanmetzman:
PTAL.

I accidentally forgot to push the change you asked for in #97

PR #105, opened 2020-03-16
Updated fastcgs to newer commit.

alifahmed:
Updated to commit 14b7428. This replaces -march=native with -msse4.2 flag.

### DIFF
--- a/fuzzers/fastcgs/builder.Dockerfile
+++ b/fuzzers/fastcgs/builder.Dockerfile
@@ -27,7 +27,7 @@ RUN rm -rf /afl
 
 RUN git clone https://github.com/alifahmed/aflmod /afl && \
     cd /afl && \
-    git checkout 45a7513bbf9890e54e7593449cd6404efdb3ee16 && \
+    git checkout 14b7428ba5a6c5cad05e9737847e80128471e9d5 && \
     export AFL_LLVM_LAF_SPLIT_SWITCHES=1 && \
     export AFL_LLVM_LAF_TRANSFORM_COMPARES=1 && \
     export AFL_LLVM_LAF_SPLIT_COMPARES=1 && \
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
@alifahmed here's your report:
https://www.fuzzbench.com/reports/2020-03-16-fastcgs/index.html

I think there is an important caveat in these results though. afl++ in this experiment doesn't use the mopt mutator but fastcgs does. Maybe we should synchronize the two fuzzers and run another experiment?

jonathanmetzman:
[This](http://commondatastorage.googleapis.com/fuzzbench-reports/2020-03-09-fastcgs/index.html)
compares fastcgs to afl++ when it used mopt (though we made some code changes since 2020-03-09 that gave fuzzers slightly more CPU cycles since then so the comparison isn't perfect).

alifahmed:
> [This](http://commondatastorage.googleapis.com/fuzzbench-reports/2020-03-09-fastcgs/index.html) compares fastcgs to afl++ when it used mopt (though we made some code changes since 2020-03-09 that gave fuzzers slightly more CPU cycles since then so the comparison isn't perfect).

@jonathanmetzman Thanks for the report! Fastcgs does especially bad for the libpcap. I will try to figure out why. I also want to explore some other configurations:
* Check if laf-intel justifies the runtime overhead
* Check if skipping deterministic step is beneficial or not for this short run
* Try out original mopt with longer branch-history

I will soon make a few PRs with the changes :)

jonathanmetzman:
> > [This](http://commondatastorage.googleapis.com/fuzzbench-reports/2020-03-09-fastcgs/index.html) compares fastcgs to afl++ when it used mopt (though we made some code changes since 2020-03-09 that gave fuzzers slightly more CPU cycles since then so the comparison isn't perfect).
> 
> @jonathanmetzman Thanks for the report! Fastcgs does especially bad for the libpcap. I will try to figure out why. I also want to explore some other configurations:

Worth pointing out that all AFL based fuzzers (except afl++ until recently, I assumed because of *laf-intel*, but maybe not given fastcgs's performance) do very badly on this benchmark.
The benchmark has no seeds and i have a hunch this hurts AFL based fuzzers more than others.

Issue #107, opened 2020-03-16
aflsmart build is broken


jonathanmetzman:
@thuanpv 
Running `make build-aflsmart-libpcap_fuzz_both` gives me the following error:
```
Step 5/7 : RUN git clone https://github.com/aflsmart/aflsmart /afl &&     cd afl &&     git checkout df095901ea379f033d4d82345023de004f28b9a7 &&     AFL_NO_X86=1 make
fatal: reference is not a tree: df095901ea379f033d4d82345023de004f28b9a7
The command '/bin/sh -c git clone https://github.com/aflsmart/aflsmart /afl &&     cd afl &&     git checkout df095901ea379f033d4d82345023de004f28b9a7 &&     AFL_NO_X86=1 make' returned a non-zero code: 128
make: *** [docker/build.mk:193: .aflsmart-libpcap_fuzz_both-oss-fuzz-builder-intermediate] Error 128
```

I wonder why this wasn't caught by our CI. I will test this again to check.
What's also weird is how this affected the results of the last experiment: [2020-03-11](https://www.fuzzbench.com/reports/2020-03-11/index.html) 
Some of the benchmarks are missing because it failed to build for them (e.g. bloaty_fuzz_target), but others such as freetype have data.

I'll point out that the aggregate ranking for aflsmart in that report is not really accurate because it is missing benchmarks.

thuanpv:
Hi @jonathanmetzman. Thank you for your analysis of the issue. It seems to be very similar to this issue https://github.com/actions/checkout/issues/23 and the root cause could be that in the builder script I check out a merge commit. I am testing it locally with a normal commit to see how it goes.

jonathanmetzman:
These are the builds that failed:
mbedtls_fuzz_dtlsclient
bloaty_fuzz_target
php_php-fuzz-parser
libpcap_fuzz_both
sqlite3_ossfuzz
openssl_x509
irssi_server-fuzz
zlib_zlib_uncompress_fuzzer
jsoncpp_jsoncpp_fuzzer
curl_curl_fuzzer_http
systemd_fuzz-link-parser

So I think the issue only affects oss-fuzz benchmarks, weird.

I also noticed afl++ failed on opnessl_x509

jonathanmetzman:
[...]

PR #108, opened 2020-03-16
[DO NOT MERGE] Test commit to check aflsmart build

jonathanmetzman:
See #107


jonathanmetzman:
The fact that this passes was fixed in https://github.com/google/fuzzbench/pull/116 
The underlying aflsmart breakage is fixed in #117

PR #111, opened 2020-03-16
add afl++ mopt variant

vanhauser-thc:
MOpt in afl++ has several changes to the original MOpt in stages, configuration settings, etc.
This is to test different settings to - over time - make it the best MOpt setup available :)

### DIFF
--- a/.github/workflows/fuzzers.yml
+++ b/.github/workflows/fuzzers.yml
@@ -16,6 +16,7 @@ jobs:
           - afl
           - aflfast
           - aflplusplus
+          - aflplusplus-mopt
           - aflsmart
           - eclipser
           - entropic
--- a/.github/workflows/fuzzers.yml
+++ b/.github/workflows/fuzzers.yml
@@ -16,7 +16,7 @@ jobs:
           - afl
           - aflfast
           - aflplusplus
-          - aflplusplus-mopt
+          - aflplusplus_mopt
           - aflsmart
           - eclipser
           - entropic
--- a/fuzzers/aflplusplus-mopt/builder.Dockerfile
+++ b/fuzzers/aflplusplus-mopt/builder.Dockerfile
@@ -1,0 +1,35 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+ARG parent_image=gcr.io/fuzzbench/base-builder
+FROM $parent_image
+
+# Install wget to download afl_driver.cpp. Install libstdc++ to use llvm_mode.
+RUN apt-get update && \
+    apt-get install wget libstdc++-5-dev -y
+
+# Download and compile afl++ (v2.62d).
+# Build without Python support as we don't need it.
+# Set AFL_NO_X86 to skip flaky tests.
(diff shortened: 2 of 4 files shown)


vanhauser-thc:
@googlebot I signed it!

alifahmed:
@vanhauser-thc You need to add _aflplusplus-mopt_ to the workflow file:
_.github/workflows/fuzzers.yml_

jonathanmetzman:
@vanhauser-thc I'm a bit confused, could you explain what this tests differently from the existing afl++ integration?

vanhauser-thc:
@alifahmed damn I committed from a subdirectory and missed that. thanks!

@jonathanmetzman It uses MOpt, so the mutation are totally different and not comparable to "normal" afl++. But the MOpt used is different to the "MOpt" fuzzer. Hence it is interesting to see if the changes are beneficial or not.

jonathanmetzman:
> @jonathanmetzman It uses MOpt, so the mutations are totally different and not comparable to "normal" afl++. But the MOpt used is different to the "MOpt" fuzzer. Hence it is interesting to see if the changes are beneficial or not.

Oh sorry, I didn't notice that mopt was removed from afl++ by https://github.com/google/fuzzbench/pull/90. 
The initial fuzzer.py file we had used it.

I think at least the build function can be imported from aflplusplus/fuzzer.py.
What do you think of doing that?

I'm still not entirely sure how we are going to handle different runtime configurations of fuzzers.
AFL++'s existing integration supports using `ADDITIONAL_ARGS` (https://github.com/google/fuzzbench/blob/master/fuzzers/aflplusplus/fuzzer.py#L118). If we go with that route instead of this one, I'll probably need to support a variants/ directory that contains env vars for all variants of a fuzzer. If I do that, I'll replace this integration with the supported one.

vanhauser-thc:
> > @jonathanmetzman It uses MOpt, so the mutations are totally different and not comparable to "normal" afl++. But the MOpt used is different to the "MOpt" fuzzer. Hence it is interesting to see if the changes are beneficial or not.
> 
> Oh sorry, I didn't notice that mopt was removed from afl++ by #90.
> The initial fuzzer.py file we had used it.
> 
> I think at least the build function can be imported from aflplusplus/fuzzer.py.
> What do you think of doing that?

I copied the python script (obviously) and made very little changes. What do you mean by importing? It is not a module (and I am very, very, very basic with python. Give me C, assembly or bash and I am king, but python gets me running for my life ;)

> I'm still not entirely sure how we are going to handle different runtime configurations of fuzzers.
> AFL++'s existing integration supports using `ADDITIONAL_ARGS` (https://github.com/google/fuzzbench/blob/master/fuzzers/aflplusplus/fuzzer.py#L118). If we go with that route instead of this one, I'll probably need to support a variants/ directory that contains env vars for all variants of a fuzzer. If I do that, I'll replace this integration with the supported one.

Yes, either parallel fuzzers (like what I am commiting here) or a subdirectory with variants.
As long as there is no support for variants it would be great to have them as additional fuzzers for now.
We are doing a lot of changes to the fuzzer at the moment so I want to grab the opportunity to have benchmarks on the different features and how to change the results over the weeks.

And what is this issue in the CI?
```
CI / presubmit (pull_request) Failing after 1m — presubmit
```

jonathanmetzman:
> I copied the python script (obviously) and made very little changes. What do you mean by importing? It is not a module (and I am very, very, very basic with python. Give me C, assembly or bash and I am king, but python gets me running for my life ;)

OK. I tried changing this PR myself but the directory rename is going to make it complicated and I only know how to edit PRs directly through web UI. I've made my fixes here: 

https://github.com/google/fuzzbench/pull/111

What I mean is that fuzzers.aflplusplus.fuzzer is a module that could be imported from fuzzer.py
so no need to copy the build() function, we can import the module and call it from there.

> And what is this issue in the CI?

GitHub actions unfortunately doesn't make it obvious but if you click enough nonsense you see that the command `make presubmit` failed on the PR.
it failed for two reasons:
1. A formatting issue (I think this issue might be present in the normal aflplusplus integration because there was an issue I've since fixed where presubmit in CI would pass no matter what).
2. aflplusplus-mopt isn't a valid module name, I switched the hyphen to an underscore.

PR #117, opened 2020-03-17
[aflsmart] Fix issue building aflsmart in OSS-Fuzz

jonathanmetzman:
OSS-Fuzz benchmarks don't use / as CWD. Be consistent about
using absolute paths for the repo and don't rely on CWD being /.
Also, don't assume CFLAGS is unset.

### DIFF
--- a/fuzzers/aflsmart/builder.Dockerfile
+++ b/fuzzers/aflsmart/builder.Dockerfile
@@ -15,6 +15,12 @@
 ARG parent_image=gcr.io/fuzzbench/base-builder
 FROM $parent_image
 
+# Download and compile AFLSmart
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
This should make progress on #107 but there will still be another build error.
None of that will be visible in CI (or from make-aflsmart-all) until #116 lands.

jonathanmetzman:
CC @thuanpv 

Should probably wait for CI fix to land

thuanpv:
Great! Thanks a lot @inferno-chromium and @jonathanmetzman.

jonathanmetzman:
aflsmart builds are passing now!

PR #118, opened 2020-03-17
Update honggfuzz to the newest version, and disable --sanitizers

robertswiecki:
The newest version has some additional "improvements" which can make it
a better fuzzer :).

Not specifying --sanitizers doesn't affect fuzzbench. It's only useful
for off-fuzz/clusterfuzz where it's used to make *san libraries generate
proper reports (so, Clusterfuzz can pick them up and create end-user
reports - e.g with stacktraces). It's not necessary for fuzzbench.

Specifying --sanitizers_del_report=true will actually delete those
report files. With some benchmarks (e.g. with proj4) it's possible to
create testcases which crash the benchmark, and then tons of ASAN log
files will be left around, eating up disk space, and for no good reasons.

Thirdly, it sets the memory limit to 2GB per fuzzed benchmark, in order
to bring it in line with the current libfuzzer config, and to avoid OOMs.

### DIFF
--- a/fuzzers/honggfuzz/builder.Dockerfile
+++ b/fuzzers/honggfuzz/builder.Dockerfile
@@ -18,14 +18,14 @@ FROM $parent_image
 # honggfuzz requires libfd and libunwid.
 RUN apt-get update -y && apt-get install -y libbfd-dev libunwind-dev libblocksruntime-dev
 
-# Download honggfuz version 2.1 + 815ccf8e9580b8c83ee673211da9fa217d354b5f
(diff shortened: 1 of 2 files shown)


inferno-chromium:
@jonathanmetzman mentioned some out of disk space issues sometimes, so thanks for --sanitizers_del_report fix, really appreciate it.

inferno-chromium:
Note that we periodically generate reports and can be seen at https://www.fuzzbench.com/reports/index.html, we will probably start another one sometime in next day or two.

robertswiecki:
Yup, saw them. Thanks :)

PR #121, opened 2020-03-17
Update experiment summary text

thuanpv:
Hi,

As discussed, the rankings shown on the experiment summary diagram may not be correct if a fuzzer does not support all benchmarks.

This pull request is for updating the experiment summary text in the generated FuzzBench report to clarify this. I also add a list of benchmarks which are supported by AFLSmart as an example. The list is short but we are working on more input models to support more benchmarks.

Thanks, 

Thuan


inferno-chromium:
Will let @lszekeres and @jonathanmetzman review the text as well and merge if ok.

PR #123, opened 2020-03-18
Update to newest fastcgs commit

alifahmed:
Other changes:
* Removed laf-intel
* Removed deterministic fuzzing steps
* Removed address sanitizer
* Also, passed '-pthread' flag to solve linker error in a few benchmarks


jonathanmetzman:
I changed this PR to use CFLAGS without ASAN properly.
The frame pointer and line tables stuff is also related to ASAN.

alifahmed:
> I changed this PR to use CFLAGS without ASAN properly (note to self: should document this).
> The frame pointer and line tables stuff is also related to ASAN.

Makes sense. Thanks!

PR #126, opened 2020-03-18
Dont use ASAN for benchmark builds

jonathanmetzman:
We aren't detecting crashes yet so it isn't very useful.

### DIFF
--- a/fuzzers/afl/fuzzer.py
+++ b/fuzzers/afl/fuzzer.py
@@ -26,7 +26,7 @@ def prepare_build_environment():
     """Set environment variables used to build AFL-based fuzzers."""
     utils.set_no_sanitizer_compilation_flags()
 
-    cflags = ['-O2', '-fsanitize-coverage=trace-pc-guard']
+    cflags = ['-O3', '-fsanitize-coverage=trace-pc-guard']
     utils.append_flags('CFLAGS', cflags)
     utils.append_flags('CXXFLAGS', cflags)
 
--- a/fuzzers/afl/fuzzer.py
+++ b/fuzzers/afl/fuzzer.py
@@ -25,6 +25,7 @@ from fuzzers import utils
 def prepare_build_environment():
     """Set environment variables used to build AFL-based fuzzers."""
     utils.set_no_sanitizer_compilation_flags()
+
     cflags = ['-O2', '-fsanitize-coverage=trace-pc-guard']
     utils.append_flags('CFLAGS', cflags)
     utils.append_flags('CXXFLAGS', cflags)
--- a/fuzzers/afl/fuzzer.py
+++ b/fuzzers/afl/fuzzer.py
@@ -24,10 +24,8 @@ from fuzzers import utils
 
 def prepare_build_environment():
     """Set environment variables used to build AFL-based fuzzers."""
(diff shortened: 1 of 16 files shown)


jonathanmetzman:
@inferno-chromium Could you take this over please?

fuzzers/afl/fuzzer.py
```diff
@@ -24,10 +24,8 @@
 
 def prepare_build_environment():
     """Set environment variables used to build AFL-based fuzzers."""
-    cflags = [
-        '-O2', '-fno-omit-frame-pointer', '-gline-tables-only',
-        '-fsanitize=address', '-fsanitize-coverage=trace-pc-guard'
-    ]
+    utils.set_no_sanitizer_compilation_flags()
+    cflags = ['-O3' '-fsanitize-coverage=trace-pc-guard']
```
inferno-chromium: Why some has O3 vs others have O2, we should standardize on one.

robertswiecki:
Should this also change honggfuzz's compilation flags. In https://github.com/google/fuzzbench/pull/129 the build seem to be generated correctly w/o asan.

inferno-chromium:
Yes looks like we missed Honggfuzz change here. i recounted the things, fastcgs is not required, rest all fuzzers are here except Honggfuzz.

jonathanmetzman:
@morehouse @kcc Does not using ASAN hurt libFuzzer? I vaguely remember something about ASAN giving libFuzzer the ability to instrument more of the runtime when ASAN is used. WDYT about not using ASAN for libFuzzer if we aren't going to use it for any other fuzzer?

morehouse:
> @morehouse @kcc Does not using ASAN hurt libFuzzer? I vaguely remember something about ASAN giving libFuzzer the ability to instrument more of the runtime when ASAN is used. WDYT about not using ASAN for libFuzzer if we aren't going to use it for any other fuzzer?

I'm pretty sure some important libFuzzer heuristics rely on ASan hooks for memcmp/strcmp and friends.  https://github.com/llvm-mirror/compiler-rt/blob/master/lib/fuzzer/FuzzerTracePC.cpp#L585

jonathanmetzman:
> > @morehouse @kcc Does not using ASAN hurt libFuzzer? I vaguely remember something about ASAN giving libFuzzer the ability to instrument more of the runtime when ASAN is used. WDYT about not using ASAN for libFuzzer if we aren't going to use it for any other fuzzer?
> 
> I'm pretty sure some important libFuzzer heuristics rely on ASan hooks for memcmp/strcmp and friends. https://github.com/llvm-mirror/compiler-rt/blob/master/lib/fuzzer/FuzzerTracePC.cpp#L585

So I assume using ASAN is better than not using it then? WDYT?

morehouse:
Probably.  We have unit tests that definitely benefit from the strcmp hooks, but I never did an A/B test on fuzzer-test-suite.

jonathanmetzman:
> Probably. We have unit tests that definitely benefit from the strcmp hooks, but I never did an A/B test on fuzzer-test-suite.

OK. I'll add one libfuzzer w/asan and one without. If ASAN has significant benefits we can use it for libFuzzer in the future.

PR #134, opened 2020-03-19
[CI] Miscellaneous improvements

jonathanmetzman:
Fail a fuzzer build as soon as the first benchmark fails.
Use -j when building for parallelism (probably a minor gain since we don't build multiple targets at once).
Cleanup yaml file.

### DIFF
--- a/.github/workflows/build.py
+++ b/.github/workflows/build.py
@@ -67,18 +67,19 @@ def delete_docker_images():
 def make_builds(benchmarks, fuzzer):
     """Use make to build each target in |build_targets|."""
     make_targets = get_make_targets(benchmarks, fuzzer)
-    success = True
     for pull_target, build_target in make_targets:
         # Pull target first.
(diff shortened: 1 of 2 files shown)


.github/workflows/build.py
```diff
@@ -67,18 +67,19 @@ def delete_docker_images():
...
         # Pull target first.
         subprocess.run(['make', '-j', pull_target], check=False)
 
         # Then build.
-        result = subprocess.run(['make', build_target], check=False)
+        print('Building', build_target)
+        build_command = ['make', '-j', build_target]
+        result = subprocess.run(build_command, check=False)
         if not result.returncode == 0:
-            success = False
```
inferno-chromium: Could be simplified to "if result.return_code:"

jonathanmetzman: I think this simplification makes it harder to understand what is going on.

jonathanmetzman:
Unclear if this actually speeds things up at all, the savings are probably quite small.

PR #139, opened 2020-03-20
[make] Fix $BENCHMARK in OSS-Fuzz benchmarks

### LINKED ISSUE #137: inconsistent BENCHMARK value
Hi @jonathanmetzman,

When I tested AFLSmart on libpcap_fuzz_both benchmark, I figured out that the BENCHMARK env. variable value may not be consistent with the name of corresponding benchmark stored in benchmarks folder.

For example, when I run the experiment for libpcap_fuzz_both benchmark, BENCHMARK value is "libpcap". Maybe it uses the benchmark name from OSS-Fuzz project? I don't know whether this is an issue or not.

Thuan

jonathanmetzman:
Production is unaffected because it never had this bug.

Fixes #137

### DIFF
--- a/docker/build.mk
+++ b/docker/build.mk
@@ -226,7 +226,7 @@ run-$(1)-$(2): .$(1)-$(2)-oss-fuzz-runner
     -e FUZZ_OUTSIDE_EXPERIMENT=1 \
     -e TRIAL_ID=1 \
     -e FUZZER=$(1) \
-    -e BENCHMARK=$($(2)-project-name) \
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
This is safe to land now, but I want to wait to land it because I want to CI was sped up by the recent GCB build we did (I'm hoping through caching it will be much faster).

jonathanmetzman:
CI seemed about 20 minutes faster for most fuzzers (except for eclipser I think).

PR #140, opened 2020-03-21
[honggfuzz] update version

robertswiecki:
This version adds splicing
(https://github.com/google/honggfuzz/commit/150cbd741c7346571e5a9c47b7819d2e37500a64)
which rather significantly improves coverage in some tests (e.g. in
sqlite3)

PR #143, opened 2020-03-23
Update honggfuzz (previous update performs badly under KVM).

robertswiecki:
Previous update https://github.com/google/fuzzbench/pull/140 performs
not-so-well under GCE/KVM, because of extensive bus locking (use of
atomics).

Sorry for so quick pace of updates, but I've just tested it under KVM.
On my desktop PC it performed well enough.


jonathanmetzman:
I'll just wait for honggfuzz builds to complete before landing.

PR #146, opened 2020-03-23
Show WIP message to avoid confusion in real-time reports.

inferno-chromium:


analysis/report_templates/default.html
```diff
@@ -16,7 +16,12 @@
 <!DOCTYPE html>
 <html>
     <head>
-        <title>FuzzBench: {{ experiment.name }} report</title>
+        <title>
```
jonathanmetzman: I think the warning needs to be much more prominent than "(wip)" or else no one will notice.

inferno-chromium: That is in title of tab. More prominent one is in big title on page, see change below.

inferno-chromium: Screenshot from 2020-03-23 15-04-04

jonathanmetzman: 1. I wouldn't use the abbreviation, write out "work in progress"
2. I don't think it's obvious to people what it means for a report to be in progress so explaining this somewhere would help.

analysis/report_templates/default.html
```diff
@@ -16,7 +16,12 @@
...
+            FuzzBench: {{ experiment.name }} report
+            {% if in_progress %}
+            (running)
```
inferno-chromium: This line is on tab title, so there is no space for that long message. Rest modified as per your suggestion.

PR #148, opened 2020-03-24
[analysis] Drop "id" column from experiment data

### LINKED ISSUE #133: [reports] Data.csv.gz don't need to contain id column
It has these columns because data.csv.gz contains data from a join query of snapshots on trials.
time_started and time_ended are from trials but they are probably not useful for the analysis people want to do so they just take up space at this point.

jonathanmetzman:
It is the same as trial_id so it just takes up space.
Fixes #133

### DIFF
--- a/analysis/queries.py
+++ b/analysis/queries.py
@@ -26,4 +26,8 @@ def get_experiment_data(experiment_names):
         sqlalchemy.orm.joinedload('trial')).filter(
             models.Snapshot.trial.has(
                 models.Trial.experiment.in_(experiment_names)))
-    return pd.read_sql_query(snapshots_query.statement, db_utils.engine)
+
+    # id must be loaded to do the join but get rid of it now since
+    # trial_id provides the same info.
+    data = pd.read_sql_query(snapshots_query.statement, db_utils.engine)
+    return data.drop(columns=['id'])


analysis/queries.py
```diff
@@ -26,4 +26,8 @@ def get_experiment_data(experiment_names):
         sqlalchemy.orm.joinedload('trial')).filter(
             models.Snapshot.trial.has(
                 models.Trial.experiment.in_(experiment_names)))
-    return pd.read_sql_query(snapshots_query.statement, db_utils.engine)
+
+    # id must be loaded to do the join but get rid of it now since
```
lszekeres: Consider explicitly "SELECT"-ing the columns that we want in the query itself above. I think it'd be `['experiment', 'time_started', 'time_ended', 'benchmark', 'fuzzer', 'trial_id', 'time', 'edges_covered']`. Not sure what's better.

jonathanmetzman: I don't think this is possible given that the join is when id == trial_id.

lszekeres: (It should be. In a SQL query the order of execution looks something like FROM, JOIN, WHERE, ..., SELECT)

jonathanmetzman: Right I tried to load only the columns I wanted from trial but `id` showed up anyway.
I can try to figure out how (if possible, it seems like a common use case so probably is) to force SQLalchemy to do what I want here but I'm not sure it is worth it.

lszekeres: Yes, this is no big deal, it's also fine as it is.

PR #149, opened 2020-03-24
Add support for fuzzer dictionaries.

inferno-chromium:
Use fuzzer dictionaries where supported. Currently, available for
AFL fuzzers, libFuzzer and Honggfuzz.

### DIFF
--- a/benchmarks/libxml2-v2.9.2/build.sh
+++ b/benchmarks/libxml2-v2.9.2/build.sh
@@ -26,8 +26,7 @@ build_lib() {
 }
 
 get_git_tag https://gitlab.gnome.org/GNOME/libxml2.git v2.9.2 SRC
-get_git_revision https://github.com/google/afl f10d601b3f3026461c669251696e6f1328ce6c00 afl
 build_lib
 
 $CXX $CXXFLAGS -std=c++11 $SCRIPT_DIR/target.cc -I BUILD/include BUILD/.libs/libxml2.a $FUZZER_LIB -lz -o $FUZZ_TARGET
-cp afl/dictionaries/xml.dict $FUZZ_TARGET.dict
+wget https://raw.githubusercontent.com/google/AFL/debe27037b9444bbf090a0ffbd5d24889bb887ae/dictionaries/xml.dict -O $FUZZ_TARGET.dict
--- a/benchmarks/libxml2-v2.9.2/build.sh
+++ b/benchmarks/libxml2-v2.9.2/build.sh
@@ -26,6 +26,8 @@ build_lib() {
 }
 
 get_git_tag https://gitlab.gnome.org/GNOME/libxml2.git v2.9.2 SRC
+get_git_revision https://github.com/google/afl f10d601b3f3026461c669251696e6f1328ce6c00 afl
 build_lib
 
 $CXX $CXXFLAGS -std=c++11 $SCRIPT_DIR/target.cc -I BUILD/include BUILD/.libs/libxml2.a $FUZZER_LIB -lz -o $FUZZ_TARGET
+cp afl/dictionaries/xml.dict $FUZZ_TARGET.dict
--- a/fuzzers/afl/fuzzer.py
+++ b/fuzzers/afl/fuzzer.py
@@ -91,6 +91,9 @@ def run_afl_fuzz(input_corpus,
     ]
     if additional_flags:
         command.extend(additional_flags)
+    dictionary_path = utils.get_dictionary_path(target_binary)
+    if dictionary_path:
+        command.extend(['-x', dictionary_path])
     command += [
         '--',
         target_binary,
@@ -98,6 +101,7 @@ def run_afl_fuzz(input_corpus,
         # performs.
         '2147483647'
     ]
+    print('[run_fuzzer] Running command: ' + ' '.join(command))
     output_stream = subprocess.DEVNULL if hide_output else None
     subprocess.call(command, stdout=output_stream, stderr=output_stream)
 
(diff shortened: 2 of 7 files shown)


fuzzers/afl/fuzzer.py
```diff
@@ -91,13 +91,17 @@ def run_afl_fuzz(input_corpus,
     ]
     if additional_flags:
         command.extend(additional_flags)
+    dictionary_path = utils.get_dictionary_path(target_binary)
```
jonathanmetzman: Do you think it's better to make users get their own dictionary or should we pass it as an argument?
@lszekeres WDYT?

benchmarks/libxml2-v2.9.2/build.sh
```diff
@@ -26,6 +26,8 @@ build_lib() {
 }
 
 get_git_tag https://gitlab.gnome.org/GNOME/libxml2.git v2.9.2 SRC
+get_git_revision https://github.com/google/afl f10d601b3f3026461c669251696e6f1328ce6c00 afl
```
jonathanmetzman: nit: can we just curl/wget https://github.com/google/AFL/blob/debe27037b9444bbf090a0ffbd5d24889bb887ae/dictionaries/xml.dict
Probably slightly faster than cloning whole repo.

fuzzers/afl/fuzzer.py
inferno-chromium: I thought about this. dictionary should always be derivable from target_binary, that is why i went with this simpler model. Also, new fuzzers might not even use this feature, so thought to keep api/arguments simple.

benchmarks/libxml2-v2.9.2/build.sh
inferno-chromium: Done!

inferno-chromium:
> Mostly LGTM.
> Thanks for implementing this!
> 
> In a later PR I think it would be nice if get_dictionary_path could respect an env var that says whether to use a dict (if available) or not. that way we could run experiments without dictionaries as well.

Sure, adding SKIP_DICT.

fuzzers/afl/fuzzer.py
inferno-chromium: Not all new fuzzers will support this and this attribute is derived from target_binary, so thought to create a helper function and keep api simpler.

lszekeres: SGTM

fuzzers/utils.py
```diff
@@ -108,3 +109,36 @@ def restore_directory(directory):
             os.getcwd()
         except FileNotFoundError:
             os.chdir(initial_cwd)
+
+
+def get_dictionary_path(target_binary):
+    """Return dictionary path for a target binary."""
+    if os.getenv('SKIP_DICT'):
```
lszekeres: If not all the fuzzers support using dictionaries, would it make sense to make not using them the deafult, and only use them if explicitly requested (eg with env var)? I'm worried about fairness for fuzzers that don't take dicts.

inferno-chromium: This is a tough one, right now all fuzzers have dicts except eclipser [dicts is way more common feature right now, and it can change in future]. And would this mean we don't allow fuzzers to add benchmark specific optimizations other than seed corpus? As an example, AFLSmart provides pits for file formats for some benchmarks, we allow this and i feel this is ok [i think this will let fuzzers try things other than dumb/coverage guided mutations]. We can let community provides ideas here too.

lszekeres: I think we can allow these optimizations and do fair evaluation at the same time. I agree that it's very useful to measure the effect of various benchmark specific optimizations, such as a dictionary file or an input format spec. However, it's also important that we compare apples to apples. So if we have fuzzers that doesn't support dicts or pits, we should only use those as a separate configs. For example, it would be unfair to compare afl-smart that uses input specs with afl that can't use them -- as we'd be giving the former additional information (provided by a human) to work with. Instead, we should compare "afl-smart-pit", "afl-smart", and "afl" in that case. Does that make sense?

jonathanmetzman: This is a fair point. At least in the case of dictionaries though, they are so ubiquitous that I think performance with dictionaries is pretty much as interesting as performance without, it doesn't seem like such a "special case" in my mind.

I don't really know what to do about aflsmart here. I think as long as it uses prewritten pits that aren't tuned once added to Fuzzbench it's probably fine to include in the generic report. It doesn't seem that different from a fuzzer that infers the format of a corpus and does special mutations based on the format (I think radamsa does this for XML). If it starts getting tuned to fuzzbench then it's more of a problem (writing a new pit for a common format would be more OK IMO). Need to think more on this.

fuzzers/afl/fuzzer.py
jonathanmetzman: I'm still thinking about this. Passing arguments explicitly (even if there are more of them) is IMO easier to deal with than using helper functions to deal with implicit behavior.

I'm trying to ensure that these files never get as complicated as the ClusterFuzz launchers.

fuzzers/utils.py
lszekeres: Just to clarify: I wasn't arguing against including any of these. I was arguing for including *both* "with dict"/"without dict" (or "with pit"/"without pit") configs *if* not all fuzzers support it. This allows us to (1) see the effect of the dict or pit itself (eg. "afl-smart-pit" <> "afl-smart"), and (2) compare the raw effectiveness of the engine itself (eg. "afl-smart" <> "afl"). 
In cases when all fuzzers support them, we don't need to include the "without" versions.

PR #151, opened 2020-03-24
Remove afl deterministic fuzzer (default, without -d).

inferno-chromium:
The default deterministic mode shows overall bad performance
in vanilla AFL fuzzer and its fuzzer variants (e.g. AFLSmart).
Now, deprecating it.
See results in
https://www.fuzzbench.com/reports/2020-03-19/index.html


inferno-chromium:
Didn't bother modifying dict stuff in afl deterministic in https://github.com/google/fuzzbench/pull/149, i think we are ready to remove this fuzzer variant.

lszekeres:
+1 on adding it back as a "config", once we have config support.

jonathanmetzman:
> +1 on adding it back as a "config", once we have config support.

Heh, will be hard since configs are usually meant to add flags not remove them.

lszekeres:
> Heh, will be hard since configs are usually meant to add flags not remove them.
Hm, that doesn't sound ideal. How about having a design doc/discussion for supporting configs some time?

PR #154, opened 2020-03-25
[fastcgs] Update commit

alifahmed:


jonathanmetzman:
By the way, I have combined 2020-03-19 and 2020-03-23 so you can see how fastcgs would have performed in the experiment run on the 19th.
http://commondatastorage.googleapis.com/fuzzbench-reports/2020-03-23_and_2020-03-19/index.html

I haven't put this with the other reports because it's hard to read.
I'm probably going to run a full experiment tonight, so results should be ready by the weekend

alifahmed:
> By the way, I have combined 2020-03-19 and 2020-03-23 so you can see how fastcgs would have performed in the experiment run on the 19th.
> http://commondatastorage.googleapis.com/fuzzbench-reports/2020-03-23_and_2020-03-19/index.html
> 
> I haven't put this with the other reports because it's hard to read.
> I'm probably going to run a full experiment tonight, so results should be ready by the weekend or early next week.

@jonathanmetzman: Thanks, the report helped a lot!! I was expecting fastcgs would do better than mopt. This commit should increase the exec/sec for fastcgs. So I am hoping to see better results in the next run :)

Issue #48, opened 2020-03-03
What benchmarks should be used


jonathanmetzman:
Since we can use any OSS-Fuzz target as a benchmark, we should figure out what we should look for in benchmarks. Community feedback here is definitely appreciated!

catenacyber:
One interesting attribute of benchmarks is their kind of input :
- text (such as json)
- binary (such as X86)
- and cases with both such as https://www.fuzzbench.com/reports/sample/index.html#libpcap_fuzz_both

stevenagy:
I'd love to see some benchmark complexity metrics like basic block counts and executable sizes!

ammaraskar:
Just a quick summary of what exists so far:

|          Benchmark          |          Category         | Input Type |
|:---------------------------:|:-------------------------:|------------|
| bloaty_fuzz_target          |     code object parser    |   binary   |
| curl_fuzzer_http            |        HTML client        |    text    |
| freetype2-2017              |        Font parser        |   binary   |
| harfbuzz-1.3.2              |        Font parser        |   binary   |
| irssi_server-fuzz           |         IRC server        |    text    |
| jsoncpp_jsoncpp_fuzzer      |        JSON parser        |    text    |
| lcms-2017-03-21             |       Color manager       |   binary   |
| libjpeg-turbo-07-2017       |        Image parser       |   binary   |
| libpcap_fuzz_both           | Network protocol analyzer |    both    |
| libpng-1.2.56               |        Image parser       |   binary   |
| libxml2-v2.9.2              |         XML parser        |    text    |
| mbedtls_fuzz_dtlsclient     |   TLS handshake handler   |   binary   |
| openssl_x509                |     Crypto data parser    |   binary   |
| openthread-2019-12-23       |   IPV6 protocol handler   |   binary   |
| php_php-fuzz-parser         |        PHP compiler       |    text    |
| proj4-2017-08-14            |  Geographic model parser  |    text    |
| re2-2014-12-09              |        Regex engine       |    text    |
| sqlite3_ossfuzz             |         SQL engine        |    text    |
| systemd_fuzz-link-parser    |       Config parser       |    text    |
| vorbis-2017-12-11           |        Audio codec        |   binary   |
| woff2-2016-05-06            |       Font converter      |   binary   |
[...]

PR #155, opened 2020-03-25
Add reference doc on benchmarks

jonathanmetzman:
Add reference doc that has a table for each benchmark giving:
1. The name of the fuzz target.
2. Whether it has a dictionary.
3. The number of seed inputs
4. The number of program edges (when compiled with LLVM's instrumentation).
5. The size of the target (when linked against libFuzzer.

### DIFF
--- a/docs/reference/benchmarks.md
+++ b/docs/reference/benchmarks.md
@@ -10,31 +10,33 @@ This page describes each benchmark that is part of fuzzbench.
 
 The table below was generated using benchmarks.py.
 
-| Benchmark                   | Fuzz target            | Has dictionary? | Number of seed inputs | Number of progam edges | Size (MB) |
-|:---------------------------:|:----------------------:|:---------------:|:---------------------:|:----------------------:|:---------:|
-| bloaty_fuzz_target          | fuzz_target            | False           | 94                    | 89530                  | 43.74     |
-| curl_curl_fuzzer_http       | curl_fuzzer_http       | False           | 31                    | 62523                  | 20.11     |
-| freetype2-2017              | fuzz-target            | False           | 2                     | 19056                  | 6.76      |
-| harfbuzz-1.3.2              | fuzz-target            | False           | 58                    | 10021                  | 6.24      |
-| irssi_server-fuzz           | server-fuzz            | True            | 895                   | 37455                  | 15.3      |
-| jsoncpp_jsoncpp_fuzzer      | jsoncpp_fuzzer         | True            | 0                     | 5536                   | 5.96      |
-| lcms-2017-03-21             | fuzz-target            | False           | 1                     | 6959                   | 6.11      |
-| libjpeg-turbo-07-2017       | fuzz-target            | False           | 1                     | 9586                   | 6.4       |
-| libpcap_fuzz_both           | fuzz_both              | False           | 0                     | 8149                   | 6.14      |
-| libpng-1.2.56               | fuzz-target            | False           | 1                     | 2991                   | 5.8       |
-| libxml2-v2.9.2              | fuzz-target            | False           | 0                     | 50461                  | 8.23      |
-| mbedtls_fuzz_dtlsclient     | fuzz_dtlsclient        | False           | 1                     | 10942                  | 6.67      |
-| openssl_x509                | x509                   | True            | 2241                  | 45989                  | 18.14     |
-| openthread-2019-12-23       | fuzz-target            | False           | 0                     | 17932                  | 6.72      |
-| php_php-fuzz-parser         | php-fuzz-parser        | True            | 2782                  | 123767                 | 15.57     |
-| proj4-2017-08-14            | fuzz-target            | False           | 44                    | 6156                   | 6.17      |
-| re2-2014-12-09              | fuzz-target            | False           | 0                     | 6547                   | 6.02      |
-| sqlite3_ossfuzz             | ossfuzz                | False           | 1258                  | 45136                  | 7.9       |
-| systemd_fuzz-link-parser    | fuzz-link-parser       | False           | 6                     | 53453                  | 5.91      |
-| vorbis-2017-12-11           | fuzz-target            | False           | 1                     | 5022                   | 6.0       |
-| woff2-2016-05-06            | fuzz-target            | False           | 62                    | 10923                  | 6.72      |
-| wpantund-2018-02-27         | fuzz-target            | False           | 8                     | 28888                  | 7.35      |
-| zlib_zlib_uncompress_fuzzer | zlib_uncompress_fuzzer | False           | 0                     | 875                    | 5.69      |
+| Benchmark                   | Fuzz target            | Has dictionary? | Number of seed inputs | Number of progam edges\* | Binary Size (MB)\*\* |
+|:---------------------------:|:----------------------:|:---------------:|:---------------------:|:------------------------:|:--------------------:|
+| bloaty_fuzz_target          | fuzz_target            | False           | 94                    | 89530                    | 43.74                |
+| curl_curl_fuzzer_http       | curl_fuzzer_http       | False           | 31                    | 62523                    | 20.11                |
+| freetype2-2017              | fuzz-target            | False           | 2                     | 19056                    | 6.76                 |
+| harfbuzz-1.3.2              | fuzz-target            | False           | 58                    | 10021                    | 6.24                 |
+| irssi_server-fuzz           | server-fuzz            | True            | 895                   | 37455                    | 15.3                 |
+| jsoncpp_jsoncpp_fuzzer      | jsoncpp_fuzzer         | True            | 0                     | 5536                     | 5.96                 |
+| lcms-2017-03-21             | fuzz-target            | False           | 1                     | 6959                     | 6.11                 |
+| libjpeg-turbo-07-2017       | fuzz-target            | False           | 1                     | 9586                     | 6.4                  |
+| libpcap_fuzz_both           | fuzz_both              | False           | 0                     | 8149                     | 6.14                 |
+| libpng-1.2.56               | fuzz-target            | False           | 1                     | 2991                     | 5.8                  |
+| libxml2-v2.9.2              | fuzz-target            | False           | 0                     | 50461                    | 8.23                 |
+| mbedtls_fuzz_dtlsclient     | fuzz_dtlsclient        | False           | 1                     | 10942                    | 6.67                 |
+| openssl_x509                | x509                   | True            | 2241                  | 45989                    | 18.14                |
+| openthread-2019-12-23       | fuzz-target            | False           | 0                     | 17932                    | 6.72                 |
+| php_php-fuzz-parser         | php-fuzz-parser        | True            | 2782                  | 123767                   | 15.57                |
+| proj4-2017-08-14            | fuzz-target            | False           | 44                    | 6156                     | 6.17                 |
+| re2-2014-12-09              | fuzz-target            | False           | 0                     | 6547                     | 6.02                 |
+| sqlite3_ossfuzz             | ossfuzz                | False           | 1258                  | 45136                    | 7.9                  |
+| systemd_fuzz-link-parser    | fuzz-link-parser       | False           | 6                     | 53453                    | 5.91                 |
+| vorbis-2017-12-11           | fuzz-target            | False           | 1                     | 5022                     | 6.0                  |
+| woff2-2016-05-06            | fuzz-target            | False           | 62                    | 10923                    | 6.72                 |
+| wpantund-2018-02-27         | fuzz-target            | False           | 8                     | 28888                    | 7.35                 |
+| zlib_zlib_uncompress_fuzzer | zlib_uncompress_fuzzer | False           | 0                     | 875                      | 5.69                 |
 
-Size is determined by the size of the libFuzzer build of the target.
-Number of program edges is the number of "guards" in the libFuzzer build of the target.
+\*Number of program edges is the number of
+["guards"](https://clang.llvm.org/docs/SanitizerCoverage.html#id2) in the
+libFuzzer build of the target.
+\*\*Size is determined by the size of the libFuzzer build of the target.
--- a/docs/reference/benchmarks.md
+++ b/docs/reference/benchmarks.md
@@ -1,0 +1,40 @@
+---
+layout: default
+title: Benchmarks
+nav_order: 4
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
This was inspired by https://github.com/google/fuzzbench/issues/48#issuecomment-594080630

I don't have all the info that table did in this table here because I only have info that can be determined programatically.

Should I add that info to this table?

docs/reference/benchmarks.md
```diff
@@ -0,0 +1,40 @@
...
+nav_order: 4
+permalink: /reference/benchmarks/
+parent: Reference
+---
+
+This page describes each benchmark that is part of fuzzbench.
+
+The table below was generated using benchmarks.py.
+
+| Benchmark                   | Fuzz target            | Has dictionary? | Number of seed inputs | Number of progam edges | Size (MB) |
```
lszekeres: Size -> Binary Size ?

Or footnote the comment under the table?

lszekeres: Do we need the "Fuzz target" column? It doesn't seem very important/useful.

docs/reference/benchmarks.md
```diff
@@ -0,0 +1,40 @@
...
+| re2-2014-12-09              | fuzz-target            | False           | 0                     | 6547                   | 6.02      |
+| sqlite3_ossfuzz             | ossfuzz                | False           | 1258                  | 45136                  | 7.9       |
+| systemd_fuzz-link-parser    | fuzz-link-parser       | False           | 6                     | 53453                  | 5.91      |
+| vorbis-2017-12-11           | fuzz-target            | False           | 1                     | 5022                   | 6.0       |
+| woff2-2016-05-06            | fuzz-target            | False           | 62                    | 10923                  | 6.72      |
+| wpantund-2018-02-27         | fuzz-target            | False           | 8                     | 28888                  | 7.35      |
+| zlib_zlib_uncompress_fuzzer | zlib_uncompress_fuzzer | False           | 0                     | 875                    | 5.69      |
+
+Size is determined by the size of the libFuzzer build of the target.
+Number of program edges is the number of "guards" in the libFuzzer build of the target.
```
lszekeres: consider: "number of instrumented program locations" ?

docs/reference/benchmarks.py
```diff
@@ -0,0 +1,178 @@
...
+import tarfile
+import zipfile
+
+from common import benchmark_utils
+from common import fuzzer_utils
+from common import utils
+from common import oss_fuzz
+
+BUILD_ARCHIVE_EXTENSION = '.tar.gz'
+LEN_BUILD_ARCHIVE_EXTENSION = len(BUILD_ARCHIVE_EXTENSION)
```
lszekeres: Consider just inlining this (it's used at one place).

docs/reference/benchmarks.py
```diff
@@ -0,0 +1,178 @@
...
+COVERAGE_BUILD_PREFIX = 'coverage-build-'
+LEN_COVERAGE_BUILD_PREFIX = len(COVERAGE_BUILD_PREFIX)
```
lszekeres: Consider just inlining this (it's used at one place).

docs/reference/benchmarks.py
```diff
@@ -0,0 +1,178 @@
...
+        seeds = count_standard_seeds(standard_seeds_path)
+    else:
+        seeds = count_oss_fuzz_seeds(fuzz_target_path)
+    return seeds
+
+
+def get_benchmark_info(build_path):
+    """Get BenchmarkInfo for the benchmark in |build_path|."""
+    basename = os.path.basename(build_path)
+    benchmark = basename[LEN_COVERAGE_BUILD_PREFIX:-LEN_BUILD_ARCHIVE_EXTENSION]
```
lszekeres: considering inlining those globals here ie `basename[len(COVERAGE_BUILD_PREFIX):-len(BUILD_ARCHIVE_EXTENSION)]`

docs/reference/benchmarks.py
```diff
@@ -0,0 +1,178 @@
...
+    assert search, benchmark
+    num_guards = int(search.groupdict()['num_guards'])
+
+    size = os.path.getsize(fuzz_target_path)
+    size = round(size / ONE_MB, 2)
+    return BenchmarkInfo(benchmark, fuzz_target, has_dictionary, seeds,
+                         num_guards, size)
+
+
+def infos_to_markdown_table(benchmark_infos):
```
lszekeres: Consider casting to `DataFrame`, which has a `to_markdown()` method

docs/reference/benchmarks.py
```diff
@@ -0,0 +1,178 @@
...
+    """Count the number of seeds in the OSS-Fuzz seed archive for
+    |fuzze_target_path|."""
+    zip_file_name = fuzz_target_path + '_seed_corpus.zip'
+    if not os.path.exists(zip_file_name):
+        return 0
+    seeds = 0
+    with zipfile.ZipFile(zip_file_name) as zip_file:
+        for info in zip_file.infolist():
+            if info.filename.endswith('/'):
+                continue
```
lszekeres: consider something like `len(ZipFile.namelist())` or `len([f for f in ZipFile.namelist() if not f.endswith('/') ])`

docs/reference/benchmarks.py
```diff
@@ -0,0 +1,178 @@
...
+            seeds += 1
+    return seeds
+
+
+def count_standard_seeds(seeds_path):
+    """Count the number of seeds for a standard benchmark."""
```
lszekeres: consider something like `len([p for p in Path(seed_path).glob('**/*') if p.is_file()])`

(`from pathlib import Path`)

docs/reference/benchmarks.md
lszekeres: Should we add a Dictionary size too?

docs/reference/benchmarks.py
```diff
@@ -0,0 +1,178 @@
...
+
+def get_benchmark_info(build_path):
+    """Get BenchmarkInfo for the benchmark in |build_path|."""
+    basename = os.path.basename(build_path)
+    benchmark = basename[LEN_COVERAGE_BUILD_PREFIX:-LEN_BUILD_ARCHIVE_EXTENSION]
+    benchmark = get_real_benchmark_name(benchmark)
+    parent_dir = os.path.dirname(build_path)
+    benchmark_path = os.path.join(parent_dir, benchmark)
+
+    try:
```
lszekeres: `common.filesystem.create_directory()` does this.

docs/reference/benchmarks.md
inferno-chromium: Over time, this will be lot of oss-fuzz benchmarks, so i think it is useful.

docs/reference/benchmarks.md
```diff
@@ -0,0 +1,40 @@
...
+| proj4-2017-08-14            | fuzz-target            | False           | 44                    | 6156                   | 6.17      |
+| re2-2014-12-09              | fuzz-target            | False           | 0                     | 6547                   | 6.02      |
+| sqlite3_ossfuzz             | ossfuzz                | False           | 1258                  | 45136                  | 7.9       |
+| systemd_fuzz-link-parser    | fuzz-link-parser       | False           | 6                     | 53453                  | 5.91      |
+| vorbis-2017-12-11           | fuzz-target            | False           | 1                     | 5022                   | 6.0       |
+| woff2-2016-05-06            | fuzz-target            | False           | 62                    | 10923                  | 6.72      |
+| wpantund-2018-02-27         | fuzz-target            | False           | 8                     | 28888                  | 7.35      |
+| zlib_zlib_uncompress_fuzzer | zlib_uncompress_fuzzer | False           | 0                     | 875                    | 5.69      |
+
+Size is determined by the size of the libFuzzer build of the target.
```
inferno-chromium: Add "* " at start of these points.

docs/reference/benchmarks.py
```diff
@@ -0,0 +1,178 @@
...
+    return pool.map(get_benchmark_info, build_paths)
+
+
+def get_real_benchmark_name(benchmark):
+    """The method we use to infer benchmark names from coverage builds
+    doesn't quite work because the project name is used in OSS-Fuzz
+    builds instead. This function figures out the actual benchmark based on
+    the project name."""
+    benchmarks_dir = os.path.join(utils.ROOT_DIR, 'benchmarks')
+    real_benchmarks = [
```
inferno-chromium: i think it might be better to iterate through directory and calculate the benchmark name based on if is_oss_fuzz or regular benchmark, this two loops - one here and later is more confusing.

docs/reference/benchmarks.py
```diff
@@ -0,0 +1,178 @@
...
+        tar_file.extractall(benchmark_path)
+
+    if benchmark_utils.is_oss_fuzz(benchmark):
+        fuzz_target = oss_fuzz.get_config(benchmark)['fuzz_target']
+    else:
+        fuzz_target = fuzzer_utils.DEFAULT_FUZZ_TARGET_NAME
+
+    fuzz_target_path = fuzzer_utils.get_fuzz_target_binary(
+        benchmark_path, fuzz_target)
+    assert fuzz_target_path, benchmark
```
inferno-chromium: Better error message than just benchmark.

docs/reference/benchmarks.py
```diff
@@ -0,0 +1,178 @@
...
+def count_oss_fuzz_seeds(fuzz_target_path):
+    """Count the number of seeds in the OSS-Fuzz seed archive for
+    |fuzze_target_path|."""
+    zip_file_name = fuzz_target_path + '_seed_corpus.zip'
+    if not os.path.exists(zip_file_name):
+        return 0
+    seeds = 0
+    with zipfile.ZipFile(zip_file_name) as zip_file:
+        for info in zip_file.infolist():
+            if info.filename.endswith('/'):
```
inferno-chromium: if info.isfile():

docs/reference/benchmarks.py
```diff
@@ -0,0 +1,178 @@
...
+    """Count the number of seeds for a standard benchmark."""
+    seeds = 0
+    for _, _, files in os.walk(seeds_path):
+        seeds += len(files)
+    return seeds
+
+
+def get_seed_count(benchmark_path, fuzz_target_path):
+    """Count the number of seeds for a benchmark."""
+    standard_seeds_path = os.path.join(benchmark_path, 'seeds')
```
inferno-chromium: s/_path/_directory ?

docs/reference/benchmarks.py
```diff
@@ -0,0 +1,178 @@
...
+    has_dictionary = os.path.exists(fuzz_target_path + '.dict')
+
+    seeds = get_seed_count(benchmark_path, fuzz_target_path)
+
+    result = subprocess.run([fuzz_target_path, '-runs=0'],
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.STDOUT,
+                            check=True)
+
+    output = result.stdout.decode()
```
inferno-chromium: no need to decode, just add b" in regex.

docs/reference/benchmarks.py
```diff
@@ -0,0 +1,178 @@
...
+    search = GUARDS_REGEX.search(output)
+    assert search, benchmark
```
inferno-chromium: Better error message than just benchmark.

docs/reference/benchmarks.py
inferno-chromium: +1

inferno-chromium: Getting hard to read, split into two lines.

docs/reference/benchmarks.md
jonathanmetzman: Yeah was feeling conflicted about this since the info is somewhat redundant and it's taking up expensive real estate in the second column. But I think it's important to put somewhere.

docs/reference/benchmarks.py
```diff
@@ -0,0 +1,178 @@
...
+    if benchmark_utils.is_oss_fuzz(benchmark):
+        fuzz_target = oss_fuzz.get_config(benchmark)['fuzz_target']
+    else:
+        fuzz_target = fuzzer_utils.DEFAULT_FUZZ_TARGET_NAME
+
+    fuzz_target_path = fuzzer_utils.get_fuzz_target_binary(
+        benchmark_path, fuzz_target)
+    assert fuzz_target_path, benchmark
+
+    has_dictionary = os.path.exists(fuzz_target_path + '.dict')
```
inferno-chromium: This is not complete, please use utils.get_dictionary_path

docs/reference/benchmarks.py
jonathanmetzman: Not available on the Python version on the machine I'm using.

docs/reference/benchmarks.py
jonathanmetzman: I inlined, but I don't really think splitting helps with readability

docs/reference/benchmarks.md
jonathanmetzman: Maybe if someone asks for it but I'd rather not get in the business of dictionary parsing (simple as they are) here.

docs/reference/benchmarks.py
```diff
@@ -0,0 +1,178 @@
...
+def main():
+    """Print a markdown table with important data on each
+    benchmark."""
+    if len(sys.argv) != 2:
+        print('Usage {} <coverage_builds_directory>'.format(sys.argv[0]))
+        return 1
+    coverage_builds_dir = sys.argv[1]
+    infos = get_benchmark_infos(coverage_builds_dir)
+    print(infos_to_markdown_table(infos))
+    print(BENCHMARK_INFO_FIELDS)
```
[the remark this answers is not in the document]
jonathanmetzman: Intuitively yes, but this makes it easier to copy. I prefer it this way.

docs/reference/benchmarks.md
lszekeres: If we could turn them into links to the fuzz target source code, that'd be very useful.

jonathanmetzman: I don't know of an automated way to do this.

docs/reference/benchmarks.md
lmrs2: just say "number of instrumented basic blocks". Edges are only known at fuzzing time given the instrumentation done. 

Seed and dictionary may or may not be used by the fuzzer, so maybe we should add a disclaimer about this. 

If we can provide where the seeds come from, or how they have been generated, that's also useful. Just bonus

PR #158, opened 2020-03-26
afl++ run option changes

vanhauser-thc:
normal aflplusplus did not receive the -d option as compared to all other afl based fuzzer.
additionally the change modifies two things:
  - different schedules / seed selection algorithms to test them against the targets
  - fixed random seeds to compare the results in the future better against each other

### DIFF
--- a/fuzzers/aflplusplus/fuzzer.py
+++ b/fuzzers/aflplusplus/fuzzer.py
@@ -112,7 +112,9 @@ def fuzz(input_corpus, output_corpus, target_binary):
     # os.environ['AFL_ALIGNED_ALLOC'] = '1' # align malloc to max_align_t
     # os.environ['AFL_PRELOAD'] = '/afl/libdislocator.so'
 
-    flags = []
+    flags = ['-d'] # disable deterministic mutations.
+    flags += ['-pmmopt']  # modified MOpt scheduling.
+    flags += ['-s123']  # fixed random seed.
     if os.path.exists(cmplog_target_binary):
         flags += ['-c', cmplog_target_binary]
     if 'ADDITIONAL_ARGS' in os.environ:
--- a/fuzzers/aflplusplus/fuzzer.py
+++ b/fuzzers/aflplusplus/fuzzer.py
@@ -112,8 +112,7 @@ def fuzz(input_corpus, output_corpus, target_binary):
     # os.environ['AFL_ALIGNED_ALLOC'] = '1' # align malloc to max_align_t
     # os.environ['AFL_PRELOAD'] = '/afl/libdislocator.so'
 
-    flags = ['-d'] # disable deterministic mutations.
-    flags += ['-pmmopt']  # modified MOpt scheduling.
+    flags = ['-pmmopt']  # modified MOpt scheduling.
     flags += ['-s123']  # fixed random seed.
     if os.path.exists(cmplog_target_binary):
         flags += ['-c', cmplog_target_binary]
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
and do we need to make changes to the individual fuzzer setups to add the dictionary loading or is this done for every fuzzer?

inferno-chromium:
> and do we need to make changes to the individual fuzzer setups to add the dictionary loading or is this done for every fuzzer?

No it is all automatic, they inherit from afl_fuzzer.run_fuzz which does it automatically.

fuzzers/aflplusplus/fuzzer.py
```diff
@@ -112,7 +112,9 @@ def fuzz(input_corpus, output_corpus, target_binary):
     # os.environ['AFL_ALIGNED_ALLOC'] = '1' # align malloc to max_align_t
     # os.environ['AFL_PRELOAD'] = '/afl/libdislocator.so'
 
-    flags = []
+    flags = ['-d'] # disable deterministic mutations.
```
inferno-chromium: Remove -d, it is already there in afl_fuzzer.run_fuzz and is the default now.

vanhauser-thc:
@inferno-chromium good to know, thanks!

jonathanmetzman:
> actually I want to see how reliable the numbers are that are generated - if the results differ largely than there would be a quality problem in fuzzbench. If you think thats unnecessary I can remove it

I think this will be interesting to try, but won't determinism go out the window if the target programs have any nondeterministic behavior?

vanhauser-thc:
> > actually I want to see how reliable the numbers are that are generated - if the results differ largely than there would be a quality problem in fuzzbench. If you think thats unnecessary I can remove it
> 
> I think this will be interesting to try, but won't determinism go out the window if the target programs have any nondeterministic behavior?

fuzzing nondeterministic programs is difficult anyway (thats where you see stability < 100% in afl).
with the -s setting all seed selections and mutations will be the same for a target, so just the CPU speed will result in different overall executions performed. and that is what would be interesting to see.
However - I think we cannot see it as we only see the final data. you should be able to see the 14 individual runs and see how much they differ

jonathanmetzman:
> > > actually I want to see how reliable the numbers are that are generated - if the results differ largely than there would be a quality problem in fuzzbench. If you think thats unnecessary I can remove it
> > 
> > 
> > I think this will be interesting to try, but won't determinism go out the window if the target programs have any nondeterministic behavior?
> 
> fuzzing nondeterministic programs is difficult anyway (thats where you see stability < 100% in afl).

I haven't looked at the data in a while, but I think most OSS-Fuzz targets (and those from other places) don't have 100% stability (at least when run in persistent mode). The same is probably true for our benchmarks.
You can actually look at one: http://commondatastorage.googleapis.com/fuzzbench-data/2020-03-19/experiment-folders/sqlite3_ossfuzz-aflplusplus/trial-63227/corpus/corpus-archive-0097.tar.gz

This one has 76% stability.

Most often I see instability caused by things like this

```
void f() {
static Foo* foo = new Foo();
}
```

which isn't really nondeterminism, just nondeterminism across runs.

I remember when the stability stat came out, it bothered me a lot to see instability, but overtime I've seen unstable targets find a lot of bugs so I'm not sure how much it matters (at least without determining if the instability is real nondeterminism or the kind I described above).

jonathanmetzman:
>In addition, randomism is part of fuzzing, if you remove it experiments are biased.

+1 to this. I see if I can share the data on my claim about OSS-Fuzz targets are generally unstable if people would like. I'm curious if other people have seen this.

PR #159, opened 2020-03-26
Ensemble Fuzzing of AFL Forks

Leo-Neat:
Ensemble fuzzing using MOpt, Fairfuzz, AFL, AFLPlusPlus, and AFLFast. Each fuzzer is rotated every hour.


fuzzers/ensemble/fuzzer.py
```diff
@@ -0,0 +1,145 @@
...
+import subprocess
+import os
+import tempfile
+
+from fuzzers import utils
+
+# OUT environment variable is the location of build directory (default is /out).
+
+# Fuzz each fuzzer for 1 hour before rotating.
+SECONDS_PER_FUZZER = 3600
```
jonathanmetzman: 60*60 is clearer.

fuzzers/ensemble/fuzzer.py
```diff
@@ -0,0 +1,145 @@
...
+    """Set environment variables used to build AFL-based fuzzers."""
+    utils.set_no_sanitizer_compilation_flags()
+
+    cflags = ['-O3', '-fsanitize-coverage=trace-pc-guard']
+    utils.append_flags('CFLAGS', cflags)
+    utils.append_flags('CXXFLAGS', cflags)
+
+    os.environ['CC'] = 'clang'
+    os.environ['CXX'] = 'clang++'
+    os.environ['FUZZER_LIB'] = '/libAFL.a'
```
jonathanmetzman: Let's call this libAFLDriver.a
It's more correct and a lot less confusing in this case.

fuzzers/ensemble/fuzzer.py
```diff
@@ -0,0 +1,145 @@
...
+def build():
+    """Build benchmark with AFL."""
+    prepare_build_environment()
+
+    utils.build_benchmark()
+
+    print('[post_build] Copying afl-fuzz to $OUT directory')
+    # Copy out the afl-fuzz binary as a build artifact.
+    shutil.copy('/afl/afl-fuzz', os.path.join(os.environ['OUT'], 'afl'))
+    shutil.copy('/aflpp/afl-fuzz', os.path.join(os.environ['OUT'], 'aflpp'))
```
jonathanmetzman: Let's just call it "aflplusplus", chars are cheap!

fuzzers/ensemble/fuzzer.py
```diff
@@ -0,0 +1,145 @@
...
+    shutil.copy('/aflsmart/afl-fuzz', os.path.join(os.environ['OUT'],
+                                                   'aflsmart'))
+    shutil.copy('/fair/afl-fuzz', os.path.join(os.environ['OUT'], 'fair'))
```
jonathanmetzman: call this fairfuzz.

fuzzers/ensemble/fuzzer.py
```diff
@@ -0,0 +1,145 @@
...
+    # Tell AFL to not use its terminal UI so we get usable logs.
+    os.environ['AFL_NO_UI'] = '1'
+    # Skip AFL's CPU frequency check (fails on Docker).
+    os.environ['AFL_SKIP_CPUFREQ'] = '1'
+    # No need to bind affinity to one core, Docker enforces 1 core usage.
+    os.environ['AFL_NO_AFFINITY'] = '1'
+    # AFL will abort on startup if the core pattern sends notifications to
+    # external programs. We don't care about this.
+    os.environ['AFL_I_DONT_CARE_ABOUT_MISSING_CRASHES'] = '1'
+
```
jonathanmetzman: Set the env var AFL_FAST_CAL=1 please. 
It will make it quicker to start up fuzzing (see https://github.com/google/afl/blob/master/docs/env_variables.txt#L148). I suspect 1 hour might be too frequent. Maybe let's try 2 hours?

fuzzers/ensemble/fuzzer.py
```diff
@@ -0,0 +1,145 @@
...
+def build():
+    """Build benchmark with AFL."""
+    prepare_build_environment()
+
+    utils.build_benchmark()
+
+    print('[post_build] Copying afl-fuzz to $OUT directory')
+    # Copy out the afl-fuzz binary as a build artifact.
+    shutil.copy('/afl/afl-fuzz', os.path.join(os.environ['OUT'], 'afl'))
```
jonathanmetzman: I would put the names of all of these variants and then loop over them instead of copying calls to `shutil.copy`

fuzzers/ensemble/fuzzer.py
```diff
@@ -0,0 +1,145 @@
...
+    shutil.copy('/aflpp/afl-fuzz', os.path.join(os.environ['OUT'], 'aflpp'))
+    shutil.copy('/aflsmart/afl-fuzz', os.path.join(os.environ['OUT'],
+                                                   'aflsmart'))
+    shutil.copy('/fair/afl-fuzz', os.path.join(os.environ['OUT'], 'fair'))
+    shutil.copy('/mopt/afl-fuzz', os.path.join(os.environ['OUT'], 'mopt'))
+
+
+def prepare_fuzz_environment(input_corpus):
```
jonathanmetzman: Does this function and run_afl_fuzz need to be copied from afl/fuzzer.py?
Or can you simply import them?

fuzzers/ensemble/fuzzer.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def fuzz(input_corpus, output_corpus, target_binary):
+    """Run afl-fuzz on target."""
+    prepare_fuzz_environment(input_corpus)
+    fuzzer_queue = collections.deque(
+        ['fair', 'mopt', 'afl', 'aflpp', 'aflsmart'], maxlen=5)
+
+    tmp_dir = tempfile.TemporaryDirectory()
+    next_fuzzer_input_corpus = os.path.join(str(tmp_dir), 'input_corpus')
```
jonathanmetzman: I don't think str call is needed.

fuzzers/ensemble/fuzzer.py
```diff
@@ -0,0 +1,145 @@
...
+                         fuzzer_output_dir, target_binary, SECONDS_PER_FUZZER)
+        except subprocess.TimeoutExpired:
+            tmp_dir.cleanup()
+            fuzzer_queue.rotate(1)
+            next_fuzzer_name = fuzzer_name
+            print('Switching to fuzzer to {0}.'.format(next_fuzzer_name))
+            tmp_dir = tempfile.TemporaryDirectory()
+            next_fuzzer_input_corpus = os.path.join(str(tmp_dir),
+                                                    'input_corpus')
+            shutil.copytree(os.path.join(fuzzer_output_dir, 'queue'),
```
jonathanmetzman: What happens the second time a fuzzer is run. Will copytree on the same next_fuzzer_input_corpus twice work?

jonathanmetzman:
This isn't tested in CI either though it probably (maybe not since in theory it should only fail if one of the fuzzers it uses does) should.

Leo-Neat:
@jonathanmetzman Can you make sure the changes to the AFLDriver linking in the Dockerfile make sense?

fuzzers/ensemble/fuzzer.py
jonathanmetzman: ACK

fuzzers/ensemble/builder.Dockerfile
```diff
@@ -0,0 +1,109 @@
+# Copyright 2020 Google LLC
```
jonathanmetzman: Can you please rename this ensemble_afl? Renaming fuzzers will be bad so I think it's better to be descriptive now in case we want to add a regular ensemble that includes other fuzzers.

jonathanmetzman:
LGTM. I'll try this out when the current experiment finishes (probably can start around sunday).
Can you add this to the list of fuzzers in fuzzers.yml github action so we can test this build?

fuzzers/ensemble/builder.Dockerfile
Leo-Neat: Done. Sounds good!

jonathanmetzman:
This is ready to land but I'm not sure we will land it right now since it isn't exactly a fuzzer, it's a group of them. I just finished benchmarking it and I will try to compare it to the other fuzzers soon.

jonathanmetzman:
Marking as draft since I don't think we want to land this yet.

PR #168, opened 2020-03-27
[libfuzzer_asan] Remove fuzzer

jonathanmetzman:
According to https://www.fuzzbench.com/reports/2020-03-19/index.html
the differences between libFuzzer and libFuzzer with ASAN are quite
small (though it says libFuzzer+ASAN is slightly better). Since
we don't need to test both remove libfuzzer+asan. It's slightly better
for me if libFuzzer isn't using ASAN like every other fuzzer, but
we can make asan the default for libFuzzer if that is better.


jonathanmetzman:
CC @kcc @morehouse @Dor1s

PR #170, opened 2020-03-27
rename file .c to .cc since we are using the CXX compiler in build.sh

lmrs2:
This is necessary for lafintel integration


jonathanmetzman:
Build failures seem unrelated to change, let's wait for a few more builds to pass.

PR #172, opened 2020-03-27
Add repo git hash to report.

### LINKED ISSUE #163: Include info on changes to FuzzBench in reports
This is similar to https://github.com/google/fuzzbench/issues/94 but much easier to implement.
When creating an experiment, save the hash of the latest git commit. Then when generating the report for the experiment include this info and a link to the diff of this hash from the hash of the last experiment.

inferno-chromium:
Fixes #163.


inferno-chromium:
Example - http://commondatastorage.googleapis.com/fuzzbench-reports/aarya-2020-03-27/index.html

jonathanmetzman:
Is there a file from alembic that is missing here?

inferno-chromium:
> Is there a file from alembic that is missing here?

Done, added. PTAL.

common/utils.py
```diff
@@ -64,3 +65,9 @@ def file_hash(file_path):
             chunk = file_handle.read(chunk_size)
 
     return digest.hexdigest()
+
+
+def git_hash():
```
jonathanmetzman: nit: `get_git_hash` for clairty.

database/alembic/versions/43dc3aacd80e_git_hash.py
```diff
@@ -0,0 +1,29 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('experiment', sa.Column('git_hash',
+                                          sa.String(),
+                                          nullable=True))
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('experiment', 'git_hash')
```
jonathanmetzman: Shouldn't this be drop_column?

database/models.py
```diff
@@ -25,6 +25,7 @@ class Experiment(Base):
 
     name = Column(String, nullable=False, primary_key=True)
     time_created = Column(DateTime(), server_default=sqlalchemy.func.now())
+    git_hash = Column(String, nullable=True)
```
jonathanmetzman: Do we want to make this nullable?
(If we make it nullable we don't need to deal with migrating existing data and we can handle non-master branches by not including them).
Do we need the branch? maybe get_git_commit should return None if not on master? Maybe it should except if there are uncommitted changes?

common/utils.py
jonathanmetzman: I don't think this should be in utils.
We can't use it anywhere besides run_experiment since we don't copy the `.git/` into dispatcher image when running an experiment, i would just put it there.

database/alembic/versions/43dc3aacd80e_git_hash.py
inferno-chromium: Done!

analysis/report_templates/default.html
```diff
@@ -35,10 +35,15 @@
         <h1 class="center-align brown-text text-darken-3">
             FuzzBench: {{ experiment.name }} report
         </h1>
+        {% if git_hash %}
+        <h5 class="center-align brown-text text-darken-3">
+            Commit hash: <a href="https://github.com/google/fuzzbench/commits/{{ git_hash }}">{{ git_hash }}</a>
+        </h5>
+        {% endif %}
```
lszekeres: As this is not the most important information for most users, consider putting this at the bottom of the report with small letters.

analysis/rendering.py
```diff
@@ -20,7 +20,7 @@
 from common import utils
 
 
-def render_report(experiment_results, template, in_progress):
+def render_report(experiment_results, template, git_hash, in_progress):
```
lszekeres: I think this should come from the database, so it should be in experiment_results.

database/models.py
inferno-chromium: Yes nullable allows for empty columns, in case we need it for experiment without hashes.
only master might not be doable, since we might want to run experiment on an older commit added assert on uncommitted changes.

analysis/rendering.py
inferno-chromium: Moved.

analysis/report_templates/default.html
inferno-chromium: Done, moved to experiment data section.

PR #177, opened 2020-03-28
Avoid redundant database query during report generation.

lszekeres:


inferno-chromium:
Yes the issue here is experiment_df is based on snapshot table, not the experiment table, so it does not have git_hash data. we would need to do another query, maybe you can move query function from data_utils into queries.py and use it directly in generate_report (instead of experiment_results).

lszekeres:
> Yes the issue here is experiment_df is based on snapshot table, not the experiment table, so it does not have git_hash data. we would need to do another query, maybe you can move query function from data_utils into queries.py and use it directly in generate_report (instead of experiment_results).

Ah good catch, thank you! Fixed the query in https://github.com/google/fuzzbench/pull/177/commits/a1436fd2ab4c3e5b46fdca77c58b4b9240fc42d7, also addressing https://github.com/google/fuzzbench/pull/148#discussion_r397541641.

analysis/experiment_results.py
```diff
@@ -42,6 +42,13 @@ def __init__(self,
             # Take name from first row.
             self.name = experiment_df.experiment[0]
 
+        # FuzzBench repo commit hash.
+        self.git_hash = None
+        if 'git_hash' in experiment_df.columns:
```
inferno-chromium: Do we need this if, i think it is better to exception here since this column is always supposed to here.

lszekeres: Yes, we need this to support the use case of [generating reports](https://google.github.io/fuzzbench/developing-fuzzbench/custom_analysis_and_reports#generating-reports) from previously published data.csv.gz-s, as they don't have a git_hash column. Another option is to regenerate the [previous reports](https://www.fuzzbench.com/reports/index.html) and the raw data files.

lszekeres:
> Please do verify that you see a git_hash in report from `aarya-2020-03-27` experiment [git_hash=test].

Done.

analysis/experiment_results.py
inferno-chromium: LGTM then, forgot about that. Thanks for verifying.

analysis/queries.py
```diff
@@ -14,20 +14,22 @@
...
-    return data.drop(columns=['id'])
+    snapshots_query = db_utils.query(
+        Experiment.git_hash,\
+        Trial.experiment, Trial.fuzzer, Trial.benchmark,\
+        Trial.time_started, Trial.time_ended,\
+        Snapshot.trial_id, Snapshot.time, Snapshot.edges_covered)\
+        .select_from(Experiment)\
+        .join(Trial)\
+        .join(Snapshot)\
+        .filter(Experiment.name.in_(experiment_names))
```
jonathanmetzman: Are you sure you don't need to join on trial_id = id?
For example, this is how the query is done in SQL: `SELECT * FROM snapshot JOIN trial ON experiment = '2020-03-23' and id = trial_id ORDER BY time_ended;`
Maybe this isn't needed (as it wasn't in the last query) because SQLAlchemy knows the two fields are connected?

Also, we generally don't use backslashes to break up lines of code.

lszekeres: > Maybe this isn't needed (as it wasn't in the last query) because SQLAlchemy knows the two fields are connected?

Yes, SQLAlchemy automatically constructs the `JOIN` conditions from the relationships encoded in the models. This is the generated SQL query:
```sql
SELECT experiment.git_hash AS experiment_git_hash, trial.experiment AS trial_experiment, trial.fuzzer AS trial_fuzzer, trial.benchmark AS trial_benchmark, trial.time_started AS trial_time_started, trial.time_ended AS trial_time_ended, snapshot.trial_id AS snapshot_trial_id, snapshot.time AS snapshot_time, snapshot.edges_covered AS snapshot_edges_covered 
FROM experiment JOIN trial ON experiment.name = trial.experiment JOIN snapshot ON trial.id = snapshot.trial_id 
WHERE experiment.name IN (%(name_1)s)
```

> Also, we generally don't use backslashes to break up lines of code.

I know, that's why I put it in a separate commit, see https://github.com/google/fuzzbench/pull/177/commits/c70211404c630c8e9b509f850ca91f73749d6d99. I think in this case it largely improves readability and yapf didn't complain either :) I'd keep it unless you feel strongly about it. Wdyt?

jonathanmetzman: OK fine to keep

PR #178, opened 2020-03-28
afl++ temporary fuzzer: no allocs

vanhauser-thc:
This is a major change that removes most memory allocations in afl-fuzz.

In case a new test run of fuzzbench is already running then do not accept this change and we will test it differently in the next run.


jonathanmetzman:
There is one already running. How about you change this and I can combine an experiment with just this with data from the currently running all-fuzzer experiment.

vanhauser-thc:
> There is one already running. How about you change this and I can combine an experiment with just this with data from the currently running all-fuzzer experiment.

that would be great. how do you want me to change this?

jonathanmetzman:
Sorry for the confusion,I don't need it to be changed. I thought you wanted to change it before we run it if we couldn't run it immediately.

jonathanmetzman:
If this is fine to land as is, let me know and I will do so.

vanhauser-thc:
it is fine to land, thanks!

PR #182, opened 2020-03-29
Add Proxygen benchmark.

inferno-chromium:


jonathanmetzman:
In .github/workflows/build.py there's a list of benchmarks. This needs to be added to one.

inferno-chromium:
@mhlakhani - do you think this will be a good benchmark fuzzer ?

inferno-chromium:
> In .github/workflows/build.py there's a list of benchmarks. This needs to be added to one.

Done, thanks!

jonathanmetzman:
> @mhlakhani - do you think this will be a good benchmark fuzzer ?

e.g. does it not do things considered "wrong" for fuzzing like accumulating state, connecting to the network, etc.

mhlakhani:
apologies for not seeing this until now - yeah I think this would be a good benchmark. it does not keep any state across requests and only does parsing.

jonathanmetzman:
Can you update the table of our benchmarks after you run this?
The script I wrote only works if you have the coverage build zip that we create during experiments so feel free to wait until you do an experiment or get the data manually now.

inferno-chromium:
> Can you update the table of our benchmarks after you run this?
> The script I wrote only works if you have the coverage build zip that we create during experiments so feel free to wait until you do an experiment or get the data manually now.

Yes, once oyu LGTM this, we merge, run and then i can update table.

Issue #119, opened 2020-03-17
Dont drop fuzzers from benchmarks where they have few trials


jonathanmetzman:
It's a pretty silent way to show an error, and I've missed it many times.
Instead, include those fuzzers and add some kind of warning.

jonathanmetzman:
I don't think this issue should be quite closed when the next PR lands.

PR #184, opened 2020-03-29
Add warning about fuzzers with low number of trials in the report.

lszekeres:
Fixes https://github.com/google/fuzzbench/issues/119.

### DIFF
--- a/analysis/report_templates/default.html
+++ b/analysis/report_templates/default.html
@@ -100,6 +100,24 @@
 
             <h2 class="green-text text-darken-3">{{ benchmark.name }} summary</h2>
 
+            {% if benchmark.fuzzers_with_not_enough_samples %}
+            <div class="card-panel deep-orange lighten-3">
+                <div class="row valign-wrapper">
+                    <div class="col s4 m2">
+                        <i class="medium material-icons">error</i>
+                    </div>
+                    <div class="col s8 m10">
+                        <span class="black-text">
+                            The following fuzzers do not have enough samples:
+                            <strong>
+                                {{ ', '.join(benchmark.fuzzers_with_not_enough_samples) }}.
+                            </strong>
+                        </span>
+                    </div>
+                </div>
+            </div>
+            {% endif %}
+
             <div class="row">
                 <div class="col s6">
                     <h5 class="center-align">Ranking by median reached coverage</h5>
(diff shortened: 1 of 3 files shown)


analysis/report_templates/default.html
```diff
@@ -100,6 +100,24 @@ <h2 class="green-text text-darken-3">experiment summary</h2>
 
             <h2 class="green-text text-darken-3">{{ benchmark.name }} summary</h2>
 
+            {% if benchmark.fuzzers_with_not_enough_samples %}
```
inferno-chromium: Should this also check the in_progress attribute, since i think we only want to show this at end of experiment and not in real-time [since measurer might be behind].

analysis/report_templates/default.html
```diff
@@ -100,6 +100,24 @@ <h2 class="green-text text-darken-3">experiment summary</h2>
...
+            <div class="card-panel deep-orange lighten-3">
```
inferno-chromium: I think this block should come after the graphs and table of each benchmark since it is not relevant for most users / other fuzzer authors ?

analysis/report_templates/default.html
jonathanmetzman: +1

analysis/report_templates/default.html
jonathanmetzman: Do you mean it is only relevant for the author of the fuzzer that doesn't have enough samples? Will they see this if we put it somewhere hard to see?

inferno-chromium: Yes i mean it is only relevant for that particular fuzzer author. I think since it is card and benchmark section is small, it is still very visible if put at end part of experiment section.

lszekeres: isn't it useful for any reader of the report to make anomalies very visible? (if i recall correctly, one reason we wanted to add this was because we haven't noticed some fuzzers missing / having few trials in some experiments for a long time.)
how about moving it after the plots, but before the summary table? i tried putting it after the table, but there it can be confusing to decide which benchmark it belongs to, as it's right on top of the next benchmark's title. i think having it above the summary table makes more sense, because it also suggests expanding the table to check the exact sample numbers in the table. wdyt?

analysis/report_templates/default.html
lszekeres: Good idea, done.

jonathanmetzman:
LGTM

PR #186, opened 2020-03-30
Set no sanitizer and optimization flags by default, before fuzzer.build.

### LINKED ISSUE #167: Set an optimization level by default
Unless there is a specific reason not to in particular cases, we probably want all fuzzers using the same optimization level to compile benchmarks. This means that we should probably set this in `CFLAGS/CXXFLAGS` before fuzzer.build() is called so that fuzzer integrators don't need to think about this.

inferno-chromium:
Fixes #167

### DIFF
--- a/fuzzers/utils.py
+++ b/fuzzers/utils.py
@@ -23,9 +23,6 @@ import tempfile
 # Keep all fuzzers at same optimization level until fuzzer explicitly needs or
 # specifies it.
 DEFAULT_OPTIMIZATION_LEVEL = '-O3'
-ALLOWED_OPTIMIZATION_FLAGS = [
-    '-O0', '-O1', '-O2', '-O3', '-Ofast', '-Os', '-Oz', '-Og', '-O', '-O4'
-]
 
 OSS_FUZZ_LIB_FUZZING_ENGINE_PATH = '/usr/lib/libFuzzingEngine.a'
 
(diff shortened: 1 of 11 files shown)


fuzzers/utils.py
```diff
@@ -142,3 +149,28 @@ def get_dictionary_path(target_binary):
                                      options_file_path)
                 return dictionary_path
     return None
+
+
+def set_default_optimization_flag_if_needed(env=None):
```
jonathanmetzman: I don't really know why the "if_needed" part is useful.
I think this is good. It changes the behavior on OSS-Fuzz (where O1 is already in CFLAGS) and makes it inconsistent with FTS benchmarks.

inferno-chromium: We were resetting flags anyway in set_no_sanitizer_compilation_flags, so optimization flag was always set to -O3. but fixed this function to just add optimization flags always, you are right, that enforces the optimization always (which we want).

vanhauser-thc:
Unless you set AFL_DONT_OPTIMIZE afl compilers always set -O3 internally

inferno-chromium:
CI had a sporadic network failure, so redoing it just to be sure.

inferno-chromium:
> Unless you set AFL_DONT_OPTIMIZE afl compilers always set -O3 internally

Good to know, this is just to enforce it same across everything.

PR #187, opened 2020-03-30
Improve the look of card panels in report.

lszekeres:
Previously, with smaller screen widths, the icon and the text could overlap in card panels.

### DIFF
--- a/analysis/report_templates/default.html
+++ b/analysis/report_templates/default.html
@@ -43,10 +43,10 @@
 
         <div class="card-panel amber lighten-5">
             <div class="row">
-                <div class="col s1">
-                    <i class="medium material-icons">warning</i>
+                <div class="col s4 m2">
+                    <center><i class="medium material-icons">warning</i></center>
                 </div>
-                <div class="col s11">
+                <div class="col s8 m10">
                     <span class="black-text">
                         Please consider this as a preliminary report to
                         demonstrate the capabilities of FuzzBench. While we have
@@ -128,7 +128,7 @@
             <div class="card-panel deep-orange lighten-3">
                 <div class="row valign-wrapper">
                     <div class="col s4 m2">
-                        <i class="medium material-icons">error</i>
+                        <center><i class="medium material-icons">error</i></center>
                     </div>
                     <div class="col s8 m10">
                         <span class="black-text">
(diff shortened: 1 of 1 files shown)


analysis/report_templates/default.html
```diff
@@ -43,10 +43,10 @@ <h5 class="center-align brown-text text-darken-4">
 
         <div class="card-panel amber lighten-5">
             <div class="row">
-                <div class="col s1">
-                    <i class="medium material-icons">warning</i>
+                <div class="col s4 m2">
+                    <center><i class="medium material-icons">warning</i></center>
```
inferno-chromium: does "center-align" inside class work, i think that is how is done in many places.

analysis/report_templates/default.html
```diff
@@ -128,7 +128,7 @@ <h5 class="center-align">Mean coverage growth over time</h5>
             <div class="card-panel deep-orange lighten-3">
                 <div class="row valign-wrapper">
                     <div class="col s4 m2">
-                        <i class="medium material-icons">error</i>
+                        <center><i class="medium material-icons">error</i></center>
```
inferno-chromium: does "center-align" inside class work, i think that is how is done in many places.

analysis/report_templates/default.html
lszekeres: Yes! That's much better indeed, thanks!

analysis/report_templates/default.html
lszekeres: Done

jonathanmetzman:
LGTM

inferno-chromium:
@lszekeres - warning icon is taking up too much space, can you please fix, see
fuzzbench.com/reports/2020-04-01/index.html

lszekeres:
The problem is that if we go back to the 1:11 grid column ratio, ie. `<div class="col s4 m1 center-align">`, we get this in a ~half screen wide window: image

I was using columns because i was following the card panel example from: https://materializecss.com/media-css.html

Alternatively we could make it floating with `<div class="left">`: 
image

Or use various custom CSS magic. Wdyt?

inferno-chromium:
Can add some min-width on the warning div.

PR #193, opened 2020-03-31
[lafintel] Fix build

jonathanmetzman:


fuzzers/lafintel/fuzzer.py
```diff
@@ -29,7 +29,9 @@ def prepare_build_environment():
     # (see https://github.com/php/php-src/blob/master/Zend/zend_cpuinfo.h).
     # It is not supported by clang-3.8, so we define the MACRO below
     # to replace any __builtin_cpu_supports() with 0, i.e., not supported
-    cflags = ['-O3', '-fPIC', '-D__builtin_cpu_supports\\(x\\)=0']
+    cflags = ['-O3', '-fPIC']
+    if 'php' in os.environ['PWD']:
```
lszekeres: just wondering how robust relying on PWD is. perhaps we should create a BENCHMARK variable later, if benchmark specific options become more common.

fuzzers/lafintel/fuzzer.py
```diff
@@ -29,7 +29,9 @@ def prepare_build_environment():
     # (see https://github.com/php/php-src/blob/master/Zend/zend_cpuinfo.h).
     # It is not supported by clang-3.8, so we define the MACRO below
     # to replace any __builtin_cpu_supports() with 0, i.e., not supported
-    cflags = ['-O3', '-fPIC', '-D__builtin_cpu_supports\\(x\\)=0']
+    cflags = ['-O3', '-fPIC']
```
inferno-chromium: ah, can you remove -O3 and utils.set_no_sanitizer_compilation_flags() call from above.  this is not needed anymore after my yesterday fix.
also, can you do same fix in aflplusplus_noalloc which was checked in after my fix.

fuzzers/lafintel/fuzzer.py
inferno-chromium: This PWD is so hacky, how can we fix this. can we just export BENCHMARK_NAME in builder dockerfile.

lmrs2: yes. I filed a bug today for this.

inferno-chromium:
You probably need to add this in CI as well like https://github.com/google/fuzzbench/pull/196/files

fuzzers/lafintel/fuzzer.py
jonathanmetzman: Yeah we will replace this with something more robust.

jonathanmetzman: Exporting it in the builds is non-trivial. I don't think we should wait for that to land this PR.

jonathanmetzman:
> You probably need to add this in CI as well like https://github.com/google/fuzzbench/pull/196/files

Already done.

jonathanmetzman:
@lmrs2 This fails because it doesn't have your patch to the pass to fix the openssl build. Can you add that to this PR and then merge this please?
I won't make any more changes to this PR since it's difficult to have 2 people working on the same PR at once.

PR #195, opened 2020-03-31
integrate KLEE

lmrs2:


.github/workflows/fuzzers.yml
```diff
@@ -23,6 +23,7 @@ jobs:
           - fairfuzz
           - fastcgs
           - honggfuzz
+          - lafintel
```
jonathanmetzman: Shouldbe klee.

fuzzers/afl/fuzzer.py
```diff
@@ -23,7 +23,8 @@
 
 
 def prepare_build_environment():
-    """Set environment variables used to build AFL-based fuzzers."""
+    """Set environment variables used to build targets for AFL-based
```
jonathanmetzman: Don't know why this changed.

fuzzers/lafintel/fuzzer.py
```diff
@@ -0,0 +1,64 @@
+# Copyright 2020 Google LLC
```
jonathanmetzman: Don't know why this is being committed.

fuzzers/lafintel/runner.Dockerfile
```diff
@@ -0,0 +1,14 @@
+# Copyright 2020 Google LLC
```
jonathanmetzman: Also don't know what this change is doing here.

kren1:
Really glad to see this! I gave it a go a couple of weeks ago (found [here](https://github.com/srg-imperial/fuzzbench/commits/master) if interested), but didn't get this far.

Two notes:
*  You need to compile the `.bc` file with these flags:      
```
       '-O1',
        '-g',
        '-Xclang',
        '-disable-llvm-passes',
        '-D__NO_STRING_INLINES',
        '-D_FORTIFY_SOURCE=0',
        '-U__OPTIMIZE__',
```
This ensures that KLEE performs optimizations it wants (with `optimize` flag) on an unoptimized version of code. It also prevents the compiler using "safe" versions of LLVM functions, which are not supported.

* While running some of the fuzzbench benchmarks, I noticed that the runner docker image has some libraries installed (such as zlib). For fuzzers this is not a problem, because they can link in the systems shared object. For KLEE this is a problem, because it needs the full binary for symbolic execution, so you need to compile zlib yourself and link it into klee with something like `-link-llvm-lib=libz.bca` argument. Note that without this KLEE will still work, but will only explore a single path in the programs that uses zlib on symbolic input.

You don't need  -libcxx` for non C++ benchmarks.

 C++ support is limited to that don't use exceptions. However you also need to compile the benchmarks with `libcxx++` headers (not the stdlibc++ ones that are probably installed).  If you need exceptions you can use this PR https://github.com/klee/klee/pull/966 which works in my experience.

I'm happy to help further.

lmrs2:
Hey, thanks for reaching out! We need help with this integration :)

On Thu, Apr 23, 2020 at 7:28 AM Timotej Kapus <notifications@github.com>
wrote:

> Really glad to see this! I gave it a go a couple of weeks ago (found here
> <https://github.com/srg-imperial/fuzzbench/commits/master> if
> interested), but didn't get this far.
>
> Two notes:
>
>    - You need to compile the .bc file with these flags:
>
>        '-O1',
>         '-g',
>         '-Xclang',
>         '-disable-llvm-passes',
>         '-D__NO_STRING_INLINES',
>         '-D_FORTIFY_SOURCE=0',
>         '-U__OPTIMIZE__',
>
> This ensures that KLEE performs optimizations it wants (with optimize
> flag) on an unoptimized version of code. It also prevents the compiler
> using "safe" versions of LLVM functions, which are not supported.
>

was not aware of this. Can you explain the meaning of each? We use '-g -O0
-Xclang -disable-O0-optnone' as suggested on the klee tutorial
<https://klee.github.io/tutorials/testing-function/>. Will clang
perform some builtin LLVM passes if we don't specify '-disable-llvm-pass'?
Which fortify functions are not supported?
Why do we need to pass -U__OPTIMIZE__?

Sorry, lots of questions :)


>
>    - While running some of the fuzzbench benchmarks, I noticed that the
>    runner docker image has some libraries installed (such as zlib). For
>    fuzzers this is not a problem, because they can link in the systems shared
>    object. For KLEE this is a problem, because it needs the full binary for
>    symbolic execution, so you need to compile zlib yourself and link it into
>    klee with something like -link-llvm-lib=libz.bca argument. Note that
>    without this KLEE will still work, but will only explore a single path in
>    the programs that uses zlib on symbolic input.
>
>
we're in the process of cleaning up some benchmarks (ie removing
dependency) and adding a dependency folder to support these sort of
dependencies. We have. some non-committed code that does exactly what you
suggest for   for libz. Thanks for confirming this is the right way!

>
>
> You don't need -libcxx` for non C++ benchmarks.
>
we're aware of this, but does it hurt to include it anyway? The startup may
be a little slower. Any other side effects? Currently we always include it
because we don't know which programs use C++ vs C. In addition, the driver
program used in benchmark is C++ for all targets.


> C++ support is limited to that don't use exceptions. However you also need
> to compile the benchmarks with libcxx++ headers (not the stdlibc++ ones
> that are probably installed). If you need exceptions you can use this PR
> klee/klee#966 <https://github.com/klee/klee/pull/966> which works in my
> experience.
>
thank you so much for this. We have not tested C++ intensively yet.


> I'm happy to help further.
>

that would be awesome. What would you like to work on? We're seeking
contributions from the community to help with KLEE integration.
If you're interested in integrating some of the current benchmarks, please
let us know. The klee branch just lays out the foundations, but there's
more to do, as you pointed out.

Something we noticed is KLEE does not support pthread. We're not sure what
the side effects are. Cloud9 does have pthread support, I believe. Not sure
how much work would be needed to port their implementation.
AFAIK, master KLEE does not support floating points either, and there's a
fork somewhere that does (unless it's already been merged). This is also
something worth pursuing?

kren1:
> was not aware of this. Can you explain the meaning of each? We use '-g -O0 -Xclang -disable-O0-optnone' as suggested on the klee tutorial <https://klee.github.io/tutorials/testing-function/>. Will clang perform some builtin LLVM passes if we don't specify '-disable-llvm-pass'? Which fortify functions are not supported? Why do we need to pass -U__OPTIMIZE__? Sorry, lots of questions :)

 The '-g -O0 -Xclang -disable-O0-optnone' set of flags is fine instead of `-g -O1 -Xclang -disable-llvm-passes. https://github.com/klee/klee/issues/902 is a discussion on the difference. The idea is to get non optimized code, without the `optnone` attribute, which prevents optimisation. Both versions  do that, but `-O1` version should in theory prepare the bitcode better for further optimisation.

`  '-D__NO_STRING_INLINES', '-D_FORTIFY_SOURCE=0', '-U__OPTIMIZE__'` These flags try to prevent clang replacing normal libc function with their `_chk` versions (something like `strlen` to `strlen_chk`), I can't tell you what each individual flag does, but this is the set we found to work.

> we're in the process of cleaning up some benchmarks (ie removing dependency) and adding a dependency folder to support these sort of dependencies. We have. some non-committed code that does exactly what you suggest for for libz. Thanks for confirming this is the right way!

Glad to hear that.

> we're aware of this, but does it hurt to include it anyway? The startup may be a little slower. Any other side effects? Currently we always include it because we don't know which programs use C++ vs C. In addition, the driver program used in benchmark is C++ for all targets.

No it doesn't hurt, but don't expect to be able to run C++ programs, just by having this (you also need to compile with the right includes). Note that for purpose of KLEE a C++ program that doesn't use any C++ features is a C program.

> that would be awesome. What would you like to work on? We're seeking contributions from the community to help with KLEE integration. If you're interested in integrating some of the current benchmarks, please let us know. 

What do you mean integrate some current benchmarks? I thought once you do this KLEE integration, you get all the benchmarks for free?

> The klee branch just lays out the foundations, but there's more to do, as you pointed out. Something we noticed is KLEE does not support pthread. We're not sure what the side effects are. Cloud9 does have pthread support, I believe. Not sure how much work would be needed to port their implementation. AFAIK, master KLEE does not support floating points either, and there's a fork somewhere that does (unless it's already been merged). This is also something worth pursuing?

What you seem to be asking for here is a KLEE fork that supports all those features. I wouldn't be interested in maintaining that. 

For `pthreads` there is an old PR https://github.com/klee/klee/pull/240 that would be a good starting point, but I don't think there is much interest in merging it into mainline. The side-effects for `pthreads` are pretty bad. KLEE treats them as an external call, which means it calls `pthread` functions in it's own process, which doesn't really make sense in this context. My guess would be it spawns loads of threads that at best hang. 

For floating point the version with floating point support you are referring to is found here https://github.com/srg-imperial/klee-float . However the constraint solving for floating points is not really good enough yet for this to be usable in any real context. It just takes too long to model floats fully. If floating point is a big problem and you are not really interested in bugs that depend on some weird floating point semantics, that you can just restrict them back to integers. That's not integrated either, but it is a relatively simple patch. Also note that KLEE works with concrete floating points, when a float becomes symbolic KLEE just immediately concretizes it, so it really depends on the benchmark if you need more support.

lmrs2:
On Fri, Apr 24, 2020 at 2:57 AM Timotej Kapus <notifications@github.com>
wrote:

> was not aware of this. Can you explain the meaning of each? We use '-g -O0
> -Xclang -disable-O0-optnone' as suggested on the klee tutorial
> https://klee.github.io/tutorials/testing-function/. Will clang perform
> some builtin LLVM passes if we don't specify '-disable-llvm-pass'? Which
> fortify functions are not supported? Why do we need to pass -U__OPTIMIZE__?
> Sorry, lots of questions :)
>
> The '-g -O0 -Xclang -disable-O0-optnone' set of flags is fine instead of -g
> -O1 -Xclang -disable-llvm-passes. https://github.com/klee/klee/issues/902
> is a discussion on the difference. The idea is to get non optimized code,
> without the optnoneattribute, which prevents optimisation. Both versions
> do that, but-O1` version should in theory prepare the bitcode better for
> further optimisation.
>
> '-D__NO_STRING_INLINES', '-D_FORTIFY_SOURCE=0', '-U__OPTIMIZE__' These
> flags try to prevent clang replacing normal libc function with their _chk
> versions (something like strlen to strlen_chk), I can't tell you what
> each individual flag does, but this is the set we found to work.
>

Thanks for the details, that's very helpful. Looks like our current
config + running KLEE with -optimize should be OK, and only cause a bit of
atency at startup.


> we're in the process of cleaning up some benchmarks (ie removing
> dependency) and adding a dependency folder to support these sort of
> dependencies. We have. some non-committed code that does exactly what you
> suggest for for libz. Thanks for confirming this is the right way!
>
> Glad to hear that.
>
> we're aware of this, but does it hurt to include it anyway? The startup
> may be a little slower. Any other side effects? Currently we always include
> it because we don't know which programs use C++ vs C. In addition, the
> driver program used in benchmark is C++ for all targets.
>
> No it doesn't hurt, but don't expect to be able to run C++ programs, just
> by having this (you also need to compile with the right includes). Note
> that for purpose of KLEE a C++ program that doesn't use any C++ features is
> a C program.
>

thanks.

> that would be awesome. What would you like to work on? We're seeking
> contributions from the community to help with KLEE integration. If you're
> interested in integrating some of the current benchmarks, please let us
> know.
>
> What do you mean integrate some current benchmarks? I thought once you do
> this KLEE integration, you get all the benchmarks for free?
>

Many benchmarks have peculiarities with regards to dependencies and build
process. For example, some benchmarks have inline assembly: we need to
disable this at compilation time. Some benchmarks call functions not
supported by KLEE (we may need to add support in KLEE; Note: I don't recall
which function this was). So integrating benchmarks is the most
time-consuming aspect right now. Help for this is greatly appreciated. If
you have cycles to integrate one or two benchmarks, it would already be of
great help to us.


> The klee branch just lays out the foundations, but there's more to do, as
> you pointed out. Something we noticed is KLEE does not support pthread.
> We're not sure what the side effects are. Cloud9 does have pthread support,
> I believe. Not sure how much work would be needed to port their
> implementation. AFAIK, master KLEE does not support floating points either,
> and there's a fork somewhere that does (unless it's already been merged).
> This is also something worth pursuing?
>
> What you seem to be asking for here is a KLEE fork that supports all those
> features. I wouldn't be interested in maintaining that.
>
> For pthreads there is an old PR klee/klee#240
> <https://github.com/klee/klee/pull/240> that would be a good starting
> point, but I don't think there is much interest in merging it into
> mainline. The side-effects for pthreads are pretty bad. KLEE treats them
> as an external call, which means it calls pthread functions in it's own
> process, which doesn't really make sense in this context. My guess would be
> it spawns loads of threads that at best hang.
>

some benchmarks also rely on this. We need to check if we can disable the
usage through compile-time flags.


> For floating point the version with floating point support you are
> referring to is found here https://github.com/srg-imperial/klee-float .
> However the constraint solving for floating points is not really good
> enough yet for this to be usable in any real context. It just takes too
> long to model floats fully. If floating point is a big problem and you are
> not really interested in bugs that depend on some weird floating point
> semantics, that you can just restrict them back to integers. That's not
> integrated either, but it is a relatively simple patch. Also note that KLEE
> works with concrete floating points, when a float becomes symbolic KLEE
> just immediately concretizes it, so it really depends on the benchmark if
> you need more support.
>

I see. Good to know. Thanks!

kren1:
> Many benchmarks have peculiarities with regards to dependencies and build process. For example, some benchmarks have inline assembly: we need to disable this at compilation time. Some benchmarks call functions not supported by KLEE (we may need to add support in KLEE; Note: I don't recall which function this was). So integrating benchmarks is the most time-consuming aspect right now. Help for this is greatly appreciated. If you have cycles to integrate one or two benchmarks, it would already be of great help to us.

Is there a run of this somewhere online, or do I have to run it locally?

Also I assume you mean modify `fuzzbench/benchmarks` files such that they work with KLEE. Is that desirable? For some benchmarks you need to disable a lot to make KLEE happy and fuzzers could potentially fuzz that too. Or is there a way to compile the benchmarks differently for different fuzzers?

lmrs2:
On Fri, Apr 24, 2020 at 9:23 AM Timotej Kapus <notifications@github.com>
wrote:

> Many benchmarks have peculiarities with regards to dependencies and build
> process. For example, some benchmarks have inline assembly: we need to
> disable this at compilation time. Some benchmarks call functions not
> supported by KLEE (we may need to add support in KLEE; Note: I don't recall
> which function this was). So integrating benchmarks is the most
> time-consuming aspect right now. Help for this is greatly appreciated. If
> you have cycles to integrate one or two benchmarks, it would already be of
> great help to us.
>
> Is there a run of this somewhere online, or do I have to run it locally?
>
not sure what you mean.


> Also I assume you mean modify fuzzbench/benchmarks files such that they
> work with KLEE. Is that desirable?
>
yes. That's option1 to do it.


> For some benchmarks you need to disable a lot to make KLEE happy and
> fuzzers could potentially fuzz that too. Or is there a way to compile the
> benchmarks differently for different fuzzers?
>

that's the option2 to do it. You can compile them differently by modifying
the klee/fuzzer.py file. it's not very clean, but it works for now.

Which option (1 or 2) to choose depends on benchmarks.

If you're interested in picking one or two benchmarks, do let us know, so
we avoid working on the same benchmarks.

kren1:
I'm looking at sqlite3 and libpcap benchmarks. I think I know which compile flags they need, but I'm struggling to even test them.

They are OSS benchmarks. so I tried changing the docker image and then trying to trick fuzzbench into picking it up, but that doesn't seem to be working. The fuzzbench documentation even says it can't be done. Any pointers on this? 

Also when I know the flags for these projects, what should I do? Changing https://github.com/google/oss-fuzz doesn't sound like the best idea for this and I don't see any facitlity to pass `./configure` flags to those `build.sh` scripts.

jonathanmetzman:
> I'm looking at sqlite3 and libpcap benchmarks. I think I know which compile flags they need, but I'm struggling to even test them.
> 
> They are OSS benchmarks. so I tried changing the docker image and then trying to trick fuzzbench into picking it up, but that doesn't seem to be working. The fuzzbench documentation even says it can't be done. Any pointers on this?
> 
> Also when I know the flags for these projects, what should I do? Changing https://github.com/google/oss-fuzz doesn't sound like the best idea for this and I don't see any facitlity to pass `./configure` flags to those `build.sh` scripts.

The way I implemented OSS-Fuzz image suport is a bit hacky but I did so anyway because it was much simpler than the alternative.
Basically OSS-Fuzz benchmarks get built by using fuzzers/$FUZZER/builder.Dockerfile to inherit from a specific build of the OSS-Fuzz image that I did and then pushed to gcr.io/fuzzbench (this is what the has is for).
So one way to change this is to change https://github.com/google/oss-fuzz I don't think this is bad option necessarily, but this doesn't lend itself to quick experimenting locally.
The other way is in the fuzzer.py file for klee, inspect the BENCHMARK environment variable to determine the benchmark you are building, and then set C/CXXFLAGS accordingly.

lmrs2:
Thanks Timotej for looking into these!

Make sure the code returns gracefully if BENCHMARK env variable is not set,
because AFAIK, non-oss-fuzz benchmarks don't have the variable set yet.
I wrote a function called is_benchmark() to check the benchmarks in aflcc
https://github.com/google/fuzzbench/blob/aflcc/fuzzers/aflcc/fuzzer.py#L25
I did not realize BENCHMARK was set, so I used the PWD instead. I will
update it accordingly. If this function is useful to you, please use it.

jonathanmetzman:
>Thanks Timotej for looking into these!

>Make sure the code returns gracefully if BENCHMARK env variable is not set,
because AFAIK, non-oss-fuzz benchmarks don't have the variable set yet.
I wrote a function called is_benchmark() to check the benchmarks in aflcc
https://github.com/google/fuzzbench/blob/aflcc/fuzzers/aflcc/fuzzer.py#L25
I did not realize BENCHMARK was set, so I used the PWD instead. I will
update it accordingly. If this function is useful to you, please use it.

I think it should be set every time `build()` is run thanks to #250

kren1:
Thanks for your pointers! I think I know how to do sqlite3 now, but for libpcap, setting CFLAGS is not enough.

For example for libpcap, I would need to replace

```
cmake ..
make
```

with 

```
../configure --disable-largefile --disable-shared --without-gcc --without-libnl --disable-dbus --without-dag --without-snf
sed -i "s/-fpic//" Makefile
make
```

in `build.sh` (replacing cmake with configure is probably not neccessary, but that's besides the point). 

I don't see this as a useful patch for https://github.com/google/oss-fuzz , because there is no need for fuzzers (in general) to have those things disabled. I also see no way of controlling flags for `configure`/`cmake` from `fuzzer.py`.

kren1:
I've added sqlite3 in #281, but I'm not too happy with it. oss-fuzz also uses `./configure` to build sqlite and there isn't a good way of passing arguments to it. Managed to do it by setting `enable_threadsafe=no` environment, but that doesn't feel right.

jonathanmetzman:
presubmit still failing.

jonathanmetzman:
@lmrs2 What should we do with this PR?

lmrs2:
@rsprabery are there still problems we need to fix before merging?

rsprabery:
@lmrs2 I don't necessarily think so. Were there any code structure things you wanted to change in `fuzzer.py` before we merge?

jonathanmetzman:
I'm pretty sure klee was merged in another PR. making this obsolete.

PR #197, opened 2020-04-01
aflplusplus_fortify

vanhauser-thc:
Can you please add this variant and then start a comparison of aflplusplus, aflplusplus_noalloc and aflplusplus_fortify? (plus what else would be good to have a current evaluation)

Thanks a lot!

### DIFF
--- a/.github/workflows/fuzzers.yml
+++ b/.github/workflows/fuzzers.yml
@@ -18,6 +18,7 @@ jobs:
           - aflplusplus
           - aflplusplus_mopt
           - aflplusplus_noalloc
+          - aflplusplus_fortify
           - aflsmart
           - eclipser
           - entropic
--- a/fuzzers/aflplusplus_fortify/builder.Dockerfile
+++ b/fuzzers/aflplusplus_fortify/builder.Dockerfile
@@ -24,7 +24,7 @@ RUN apt-get update && \
 # Set AFL_NO_X86 to skip flaky tests.
 RUN git clone https://github.com/AFLplusplus/AFLplusplus.git /afl && \
     cd /afl && \
-    git checkout 3c7bc6fd429b597850d5165d0d9f0a306eea9804 && \
+    git checkout 9ca8ff77f3a54c353f55cda621d023514a89ceff && \
     AFL_NO_X86=1 make PYTHON_INCLUDE=/ && \
     cd libdislocator && make && cd .. && \
     cd llvm_mode && CXXFLAGS= make
--- a/fuzzers/aflplusplus_fortify/builder.Dockerfile
+++ b/fuzzers/aflplusplus_fortify/builder.Dockerfile
@@ -1,0 +1,35 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+ARG parent_image=gcr.io/fuzzbench/base-builder
+FROM $parent_image
+
+# Install wget to download afl_driver.cpp. Install libstdc++ to use llvm_mode.
+RUN apt-get update && \
+    apt-get install wget libstdc++-5-dev -y
+
+# Download and compile afl++ (v2.62d).
+# Build without Python support as we don't need it.
+# Set AFL_NO_X86 to skip flaky tests.
+RUN git clone https://github.com/AFLplusplus/AFLplusplus.git /afl && \
+    cd /afl && \
+    git checkout 3c7bc6fd429b597850d5165d0d9f0a306eea9804 && \
+    AFL_NO_X86=1 make PYTHON_INCLUDE=/ && \
+    cd libdislocator && make && cd .. && \
+    cd llvm_mode && CXXFLAGS= make
+
+# Use afl_driver.cpp from LLVM as our fuzzing library.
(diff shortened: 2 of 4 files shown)


jonathanmetzman:
We are already doing a full evaluation. I think tommorow I can start an expeirment with aflplusplus_noalloc and we can combine the results.

jonathanmetzman:
btw, it doesn't look like the seed fixing you did had much of an effect judging from the results I see here: https://www.fuzzbench.com/reports/2020-04-01/index.html

I still see afl++ having different coverage values for min, max, median,25th percentile and 50th percentile.

vanhauser-thc:
> I assume everything is the same except for the hash because the change was made on master and you want to compare it to an older commit on master?

yes exactly. thanks!



> btw, it doesn't look like the seed fixing you did had much of an effect judging from the results I see here: https://www.fuzzbench.com/reports/2020-04-01/index.html
> 
> I still see afl++ having different coverage values for min, max, median,25th percentile and 50th percentile.
> We should probably undo that change I'm thinking.

I saw the same thing. I have now to check if this is a bug in afl++ that the -s is not working as it should - but if it does then fuzzbench results are unreliable. that would be a major issue. I will investigate this and get back to you.

jonathanmetzman:
>but if it does then fuzzbench results are unreliable. that would be a major issue.

If it isn't due to AFL++ it may be a sign of a "major issue" in FuzzBench, but it's possible there isn't anything seriously wrong. The latter would be true if there are small amounts of non-determinism in fuzz targets as I mentioned in a previous thread.

jonathanmetzman:
I tried running a few instances of the sqlite benchmark with libFuzzer and the same seed to see if it was deterministic. I used this command for 4 instances:
```
./ossfuzz seeds/ -seed=1 -max_total_time=120
```
These are the last lines of output I saw on each instance:
```
#24160  DONE   cov: 7148 ft: 31839 corp: 895/129Mb lim: 1048576 exec/s: 199 rss: 198Mb
```

```
#46545  DONE   cov: 7593 ft: 35118 corp: 1231/63Mb lim: 1048576 exec/s: 384 rss: 150Mb
```

```
#54252  DONE   cov: 7942 ft: 36283 corp: 1287/43Mb lim: 1048576 exec/s: 448 rss: 118Mb
```

```
#55621  DONE   cov: 7325 ft: 34116 corp: 1186/55Mb lim: 1048576 exec/s: 459 rss: 144Mb
```

It's possible there is a bug in libfuzzer, or maybe the way I ran them favored the later runs (as the speed (exec/s) seems to be increasing) but my guess is that the targets have enough non-determinism in them to make the seed useless.

vanhauser-thc:
I pushed a PR which fixed an area for fixed seed that was not fixed. now it is deterministic every time (unless the target does random stuff). when this has run through we will know :)

vanhauser-thc:
> ```
> #54252  DONE   cov: 7942 ft: 36283 corp: 1287/43Mb lim: 1048576 exec/s: 448 rss: 118Mb
> ```
> ```
> #55621  DONE   cov: 7325 ft: 34116 corp: 1186/55Mb lim: 1048576 exec/s: 459 rss: 144Mb
> ```

here you can see something is wrong: the run above is slower than the one below - but finds more paths. so yes, either there is an issue in libfuzzer with the seeds, the target does random things (unlikely with sqlite), or it is fuzzbench.

My money is on libfuzzer though. Because I think they made the same mistake I did - although the seed is fixed and all random values generated are the same, the run-time measurements of a new discovered path can be different and that results in different priorizations of queue entries, and therefore queue selections etc.

jonathanmetzman:
I started a fortify experiment this morning btw.

>My money is on libfuzzer though. Because I think they made the same mistake I did - although the seed is fixed and all random values generated are the same, the run-time measurements of a new discovered path can be different and that results in different priorizations of queue entries, and therefore queue selections etc.

Ah do you mean the instrumentation can't tell when two runs have the exact same coverage?
i think AFL uses the timing of testcases, I'd expect this not to behave deterministically.

Did you fix this already (so I can narrow down causes if we see similar problems on the next afl++ run)?

vanhauser-thc:
> Ah do you mean the instrumentation can't tell when two runs have the exact same coverage?
> i think AFL uses the timing of testcases, I'd expect this not to behave deterministically.

no, what I mean:
Expectations are that with the same seed you would get the exact same result (edges found, etc.) with the only difference being the runtime because of CPU scheduling and load.
But this is not the case because if you run the same input several time you will get different run times for that input. and the run time is part of the seed selection algorithm of afl and libfuzzer. so although the random numbers are all the same, a different run of the fuzzer will get different run times for the same input and therefore select different seeds and hence the difference.

> Did you fix this already (asking so I can narrow down causes if we see similar problems on the next afl++ run)?

I fixed this, so the difference we will see in the fortify run will show in what range the available CPU time is different between the runs.

jonathanmetzman:
> > Ah do you mean the instrumentation can't tell when two runs have the exact same coverage?
> > i think AFL uses the timing of testcases, I'd expect this not to behave deterministically.
> 
> no, what I mean:
> Expectations are that with the same seed you would get the exact same result (edges found, etc.) with the only difference being the runtime because of CPU scheduling and load.
> But this is not the case because if you run the same input several time you will get different run times for that input. and the run time is part of the seed selection algorithm of afl and libfuzzer. so although the random numbers are all the same, a different run of the fuzzer will get different run times for the same input and therefore select different seeds and hence the difference.
> 

Got it. By the way, as far as I know, the time a seed takes to execute does not directly impact seed selection in libFuzzer. I think this could happen indirectly though if a testcase times out in one libFuzzer run but doesn't timeout in another.

> > Did you fix this already (asking so I can narrow down causes if we see similar problems on the next afl++ run)?
> 
> I fixed this, so the difference we will see in the fortify run will show in what range the available CPU time is different between the runs.

Maybe, but I still think the targets may have some nondeterminism. In https://github.com/google/fuzzbench/pull/158#issuecomment-605148424 for example, I found that the SQLite target is ~76% stable.

jonathanmetzman:
We can revert this now right?

vanhauser-thc:
already pushed a PR which removes it and adds the next level of tests. after that: different target instrumentations :)

PR #199, opened 2020-04-02
[base-builder] Remove installation of unneeded packages

jonathanmetzman:
Remove installation of unneeded packages.

### DIFF
--- a/docker/base-builder/Dockerfile
+++ b/docker/base-builder/Dockerfile
@@ -22,24 +22,16 @@ RUN apt-get update -y && apt-get install -y \
     automake \
     cmake \
     git \
-    golang \
     libarchive-dev \
-    libbfd-dev \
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
PTAL

docker/base-builder/Dockerfile
```diff
@@ -22,24 +22,16 @@ RUN apt-get update -y && apt-get install -y \
     automake \
     cmake \
     git \
-    golang \
```
inferno-chromium, line 25: Is it because these are in base image ?

jonathanmetzman: That's probably the reason for some of them. TAnother reason is because we removed some benchmarks so we don't need their deps. And a last reason could be that some of these may not have been necessary to begin with.

PR #200, opened 2020-04-02
[service] Initial code for running periodic experiments

jonathanmetzman:
Add code to run weekly fully experiments.

### DIFF
--- a/common/utils.py
+++ b/common/utils.py
@@ -64,3 +64,28 @@ def file_hash(file_path):
             chunk = file_handle.read(chunk_size)
 
     return digest.hexdigest()
+
+
+def get_all_benchmarks():
+    """Returns the list of all benchmarks."""
+    benchmarks_dir = os.path.join(ROOT_DIR, 'benchmarks')
+    all_benchmarks = []
+    for benchmark in os.listdir(benchmarks_dir):
+        benchmark_path = os.path.join(benchmarks_dir, benchmark)
+        if os.path.isfile(os.path.join(benchmark_path, 'oss-fuzz.yaml')):
+            # Benchmark is an OSS-Fuzz benchmark.
+            all_benchmarks.append(benchmark)
+        elif os.path.isfile(os.path.join(benchmark_path, 'build.sh')):
+            # Benchmark is a standard benchmark.
+            all_benchmarks.append(benchmark)
+    return all_benchmarks
+
+
+def get_all_fuzzers():
+    """Returns the list of all fuzzers."""
+    fuzzers_dir = os.path.join(ROOT_DIR, 'fuzzers')
+    return [
+        fuzzer for fuzzer in os.listdir(fuzzers_dir)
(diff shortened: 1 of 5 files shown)


common/utils.py
```diff
@@ -64,3 +64,28 @@ def file_hash(file_path):
             chunk = file_handle.read(chunk_size)
 
     return digest.hexdigest()
+
+
+def get_all_benchmarks():
```
inferno-chromium, line 69: should be fuzzer_utils and benchmark_utils and not the super generic utils.py

service/experiment-config.yaml
```diff
@@ -0,0 +1,7 @@
+trials: 20
+max_total_time: 86400
+cloud_project: fuzzbench
+cloud_compute_zone: us-central1-a
+cloud_experiment_bucket: gs://fuzzbench-data
```
inferno-chromium, line 5: gs://www.fuzzbench.com/reports
since real-time reports work well now.

service/run_experiment.py
```diff
@@ -0,0 +1,69 @@
...
+    raise NotImplementedError('Diff experiments not implemented yet.')
+
+
+def run_full_experiment():
+    """Run a full experiment."""
+    experiment_name = get_experiment_name()
+    fuzzers = utils.get_all_fuzzers()
+    benchmarks = utils.get_all_benchmarks()
+    run_experiment.start_experiment(experiment_name, EXPERIMENT_CONFIG_FILE,
+                                    benchmarks, fuzzers, [])
```
inferno-chromium, line 51: fuzzer_configs=[]

service/run_experiment.py
```diff
@@ -0,0 +1,69 @@
...
+    """Returns the name of the experiment to run."""
+    timezone = pytz.timezone('America/Los_Angeles')
+    time_now = datetime.datetime.now(timezone)
+    return time_now.strftime('%Y-%m-%d')
+
+
+def run_diff_experiment():
+    """Run a diff expeirment. This is an experiment that runs only on
+    fuzzers that have changed since the last experiment."""
+    # TODO(metzman): Finish this.
```
inferno-chromium, line 41: s/Finish/Implement.

common/utils.py
jonathanmetzman: I think fuzzer utils is a bad module. Half the code is meant to run on the runner and half the code is meant to run on the dispatcher. get_all_fuzzers won't work on the runner even though the module fuzzer_utils is used. I need to break this up.

inferno-chromium: How about moving to experiment_utils. i dont feel these function should get mixed with general utils stuff in this file like string and file hashes.

jonathanmetzman: I put them in fuzzer_utils and benchmark_utils for now.

jonathanmetzman:
Also can we think of another name besides run_experiment? Don't like having two modules with this name.

PR #201, opened 2020-04-02
remove noalloc and fix fixed_seed setup

vanhauser-thc:
This removes the noalloc setup as it is not needed anymore.

It also fixed the fixed_seed setup to check how repeatable the tests are when all randomness is removed. After the next run the -s options can be removed from the aflplusplus setups.

Also the results of the aflplusplus_fortify are interesting and just need to run in the next run.


fuzzers/aflplusplus_mopt/fuzzer.py
```diff
@@ -40,7 +40,7 @@ def fuzz(input_corpus, output_corpus, target_binary):
 
     flags = ['-L0']  # afl++ MOpt activation at once.
     flags += ['-prare']  # rare branch scheduling.
-    flags += ['-s666']  # fixed random seed.
+    flags += ['-s123']  # fixed random seed.
```
jonathanmetzman, line 43: I'm not sure I understand why the seed is changing but I have no objections to doing so.

jonathanmetzman:
I think you need to remove noalloc from CI.

fuzzers/aflplusplus_mopt/fuzzer.py
vanhauser-thc: just because then it is the same as in /aflplusplus/ - no other reason and not really helping or hurting :)

vanhauser-thc:
> I think you need to remove noalloc from CI.

oh you are right! commited.

PR #202, opened 2020-04-02
Add fuzzer variants to the repository and run them by default.

mbarbella-chromium:
This is not quite ready to land since we'll likely want to finalize the design. Uploading it as-is for now though.

Some open questions (which I'll prepare a short doc to address):
1) Will the default configuration always just be "fuzzer: <fuzzer_name>"? If so, we can remove the default.yaml files entirely and auto-generate them. This is less work for fuzzer authors but less flexible.
2) Abhishek suggested possibly combining the variants into a single yaml file. I think this would be useful if we want to do things that involve multiple configurations at the same time (e.g. have some way to say take the cross product of these features), but think it might be overkill in simpler cases.

### DIFF
--- a/experiment/run_experiment.py
+++ b/experiment/run_experiment.py
@@ -200,8 +200,8 @@ def start_experiment(experiment_name: str, config_filename: str,
         with open(fuzzer_config, 'rb') as handle:
             hasher = hashlib.sha256(handle.read())
             basename = hasher.hexdigest() + '.yaml'
-            shutil.copy(
-                fuzzer_config, os.path.join(fuzzer_config_dir, basename))
+            shutil.copy(fuzzer_config, os.path.join(fuzzer_config_dir,
+                                                    basename))
     for fuzzer in fuzzers:
         if fuzzers.count(fuzzer) > 1:
             raise Exception('Fuzzer "%s" provided more than once.' % fuzzer)
--- a/experiment/run_experiment.py
+++ b/experiment/run_experiment.py
@@ -16,6 +16,7 @@
 it needs to begin an experiment."""
 
 import argparse
+import hashlib
 import multiprocessing
 import os
 import re
@@ -194,7 +195,13 @@ def start_experiment(experiment_name: str, config_filename: str,
                             fuzzer_config)
         # Validate the fuzzer yaml attributes e.g. fuzzer, env, etc.
         validate_fuzzer_config(fuzzer_config)
-        shutil.copy(fuzzer_config, fuzzer_config_dir)
+
+        # Use the hash of the config data as a filename to avoid collisions.
+        with open(fuzzer_config, 'rb') as handle:
+            hasher = hashlib.sha256(handle.read())
+            basename = hasher.hexdigest() + '.yaml'
+            shutil.copy(
+                fuzzer_config, os.path.join(fuzzer_config_dir, basename))
     for fuzzer in fuzzers:
         if fuzzers.count(fuzzer) > 1:
             raise Exception('Fuzzer "%s" provided more than once.' % fuzzer)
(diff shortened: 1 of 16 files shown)


inferno-chromium:
Probably need a disabled attribute that can be used to exclude fuzzer config from consideration.

inferno-chromium:
Also, need to add tests.

jonathanmetzman:
> This is not quite ready to land since we'll likely want to finalize the design. Uploading it as-is for now though.
> 
> Some open questions (which I'll prepare a short doc to address):
> 
> 1. Will the default configuration always just be "fuzzer: <fuzzer_name>"? If so, we can remove the default.yaml files entirely and auto-generate them. This is less work for fuzzer authors but less flexible.

I think this will be the default. I think I'd rather this config exist implicitly, rather than auto or manually generated.


> 2. Abhishek suggested possibly combining the variants into a single yaml file. I think this would be useful if we want to do things that involve multiple configurations at the same time (e.g. have some way to say take the cross product of these features), but think it might be overkill in simpler cases.

I was thinking 1 file might be cleaner. What do others think? I'm not sure I understand what is meant by "have some way to say take the cross product of these features".

experiment/run_experiment.py
```diff
@@ -194,7 +195,13 @@ def start_experiment(experiment_name: str, config_filename: str,
                             fuzzer_config)
         # Validate the fuzzer yaml attributes e.g. fuzzer, env, etc.
         validate_fuzzer_config(fuzzer_config)
-        shutil.copy(fuzzer_config, fuzzer_config_dir)
+
+        # Use the hash of the config data as a filename to avoid collisions.
```
jonathanmetzman, line 199: I think this will break if someone copies a file right?
I think we should include the name of the variant and directory in "fuzzers/" as well, this will make debugging easier and is collision free.

jonathanmetzman: Also, can you put this in it's own function?

experiment/run_experiment.py
```diff
@@ -321,14 +328,25 @@ def get_all_benchmarks():
...
     """Returns the list of all fuzzers."""
+    configs = []
     fuzzers_dir = os.path.join(utils.ROOT_DIR, 'fuzzers')
-    return [
-        fuzzer for fuzzer in os.listdir(fuzzers_dir)
-        if (os.path.isfile(os.path.join(fuzzers_dir, fuzzer, 'fuzzer.py')) and
-            fuzzer != 'coverage')
-    ]
+    for fuzzer in os.listdir(fuzzers_dir):
+        # Don't create a configuration for the coverage build.
```
jonathanmetzman, line 336: Probably cleaner not to special case this.

experiment/run_experiment.py
```diff
@@ -321,14 +328,25 @@ def get_all_benchmarks():
...
+        if fuzzer == 'coverage':
+            continue
+
+        # Ensure that this is a fuzzer directory.
+        if not os.path.isfile(os.path.join(fuzzers_dir, fuzzer, 'fuzzer.py')):
+            continue
+
+        # Load the config of each variant of this fuzzer.
+        variants_dir = os.path.join(fuzzers_dir, fuzzer, 'variants')
```
jonathanmetzman, line 345: Do you think it's weird to call it "variants" if even the default has one?

experiment/run_experiment.py
```diff
@@ -371,12 +389,12 @@ def main():
     args = parser.parse_args()
 
     if not args.fuzzers and not args.fuzzer_configs:
-        fuzzers = get_all_fuzzers()
+        fuzzer_configs = get_all_fuzzer_configs()
     else:
-        fuzzers = args.fuzzers
+        fuzzer_configs = args.fuzzer_configs
```
jonathanmetzman, line 394: I guess it doesn't need to come right now but do you think we should even have a way to specify configs (aka variants) from outside the repo if we have a way to support it within repo? We should probably get rid of this feature/command line option.

jonathanmetzman:
Please see why CI is failing too.

lszekeres:
> I was thinking 1 file might be cleaner. What do others think? I'm not sure I understand what is meant by "have some way to say take the cross product of these features".

(Mixing and matching yaml/json configs could be done nicely with config languages like https://jsonnet.org.)

However, i'm wondering if we need/want any yaml config files at all. What will the yaml file contain? List of env vars and flags for building and running? What if a config needs an additional command to run, etc? 

Would it make sense to extend the existing `fuzzer.py`-s to support configs? E.g., by adding a `config='default'` parameter to the `build()` and `fuzz()` functions? The supported configs could be listed in each `fuzzer.py`, eg, something like `CONFIGS = ['with-dict', ...]`. (That'd allow stuff like: `if config = 'dataflow': ... build additional dataflow binary ...`, and similar things.)

mbarbella-chromium:
Abandoning this in favor of a newer implementation based on discussion.

PR #207, opened 2020-04-06
[wpantund] Remove benchmark

### LINKED ISSUE #171: wpantund benchmark can initiate network connections
The wpantund benchmark can initiate networking connections to remote networks under some conditions. Attaching a testcase for that:

[1788a8b166e1fe4e854bdbb1572e2bb6.00000080.honggfuzz.cov.txt](https://github.com/google/fuzzbench/files/4391761/1788a8b166e1fe4e854bdbb1572e2bb6.00000080.honggfuzz.cov.txt)

```shell
$ strace -f ~/fuzz/wpantund/wpantund-fuzz ~/fuzz/wpantund/corpus/1788a8b166e1fe4e854bdbb1572e2bb6.00000080.honggfuzz.cov
...
openat(AT_FDCWD, "/tmp", O_RDWR|O_EXCL|O_TMPFILE, 0600) = 6
fcntl(6, F_GETFL)                       = 0x418002 (flags O_RDWR|O_LARGEFILE|O_TMPFILE)
fstat(6, {st_mode=S_IFREG|0600, st_size=0, ...}) = 0
write(6, "\235mfProxy:EnabiG:NNN\305SocketPath \""..., 127) = 127
lseek(6, 0, SEEK_SET)                   = 0
read(6, "\235mfProxy:EnabiG:NNN\305SocketPath \""..., 4096) = 127
read(6, "", 4096)                       = 0
close(6)                                = 0
brk(0xb518000)                          = 0xb518000
socket(AF_INET6, SOCK_STREAM, IPPROTO_IP) = 6
connect(6, {sa_family=AF_INET6, sin6_port=htons(4951), inet_pton(AF_INET6, "::ffff:0.0.3.122", &sin6_addr), sin6_flowinfo=htonl(0), sin6_scope_id=0}, 28^C) = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
..
```

I know nothing about wpantund internals, so maybe this can be fixed (if you want to fix it) in the fuzzer harness code (maybe some API option needs to be set).

If not on API level, a quick hack could be to add something like

```C
unshare(CLONE_NEWUSER|CLONE_NEWNET);
```

at the beginning of the file.

I'm not sure if we want to leave current behavior as-is. On one hand it might be a useful case to test fuzzing timeouts, on the other, benchmarks initiating connections over non-loopback sound somehow wrong (and they also add unnecessary timeouts).

jonathanmetzman:
Remove benchmark as it can make network connections, causing timeouts.

Fixes #171.


jonathanmetzman:
CC @robertswiecki

robertswiecki:
Sounds good, I mean, with irssi which presumably accumulates state (I'll try to confirm it soon) we can test how well timeouts are handled, but a benchmarks like this one, which tries to establish random TCP connections seems a bit shady, so lgtm.

for the record, the benchmark author was asked for a fix here: https://github.com/google/oss-fuzz/issues/3555 - but no response so far.

jonathanmetzman:
> Sounds good, I mean, with irssi which presumably accumulates state (I'll try to confirm it soon) we can test how well timeouts/performance-degradation are handled, but a benchmark like this one, which tries to establish random TCP connections, seems a bit shady, so lgtm.

+1 to point about testing on *some* misbehaving benchmarks.
I also agree that this is a different category of misbehavior

PR #208, opened 2020-04-06
use latest AFL code for lafintel

lmrs2:


fuzzers/lafintel/builder.Dockerfile
```diff
@@ -28,6 +28,9 @@ RUN wget https://lcamtuf.coredump.cx/afl/releases/afl-2.26b.tgz -O /afl-2.26b.tg
     tar xvzf /afl-2.26b.tgz -C / && \
     mv /afl-2.26b /afl && \
     cd /afl && \
+    git clone https://github.com/google/AFL.git /afl/recent_afl && \
```
inferno-chromium, line 31: We need to use a specific revision, even for latest afl, e.g. see
https://github.com/google/fuzzbench/blob/master/fuzzers/afl/builder.Dockerfile#L22
That makes it easy to know what was benchmarked against.

fuzzers/lafintel/builder.Dockerfile
```diff
@@ -28,6 +28,12 @@ RUN wget https://lcamtuf.coredump.cx/afl/releases/afl-2.26b.tgz -O /afl-2.26b.tg
...
+    cd /afl/recent_afl && \ 
+    git checkout 8da80951dd7eeeb3e3b5a3bcd36c485045f40274 && \
+    cd /afl/ && \
+    cp /afl/recent_afl/*.c /afl/ && \
```
inferno-chromium, line 35: Forgot to ask this: Can we just use the new afl code or what is the need to copy afl files into old afl ?

lmrs2:
we need the llvm_mode/ folder from lafintel. The lafintel passes were
written from old llvm_mode/ passes. For the rest, we can use the most
recent afl code (seed scheduling, etc).

lmrs2:
so nothing I can do on my side, correct? Do you know why this is happening
now and not in previous tests?

inferno-chromium:
> so nothing I can do on my side, correct? Do you know why this is happening now and not in previous tests?

CI flakes probably, lets see if it reproduces again. if it does, might need some fixing on CI code - https://github.com/google/fuzzbench/blob/master/.github/workflows/build.py (maybe add retry) ?

PR #209, opened 2020-04-07
[honggfuzz]: update version

robertswiecki:
The previous version can panic() under rare circumstances.


robertswiecki:
hm.. 'no space left on the device' when building the new proxygen benchmark.

jonathanmetzman:
Disk space on GitHub actions is too small. We may need to switch to something else.
I'll land this with my fingers crossed.

PR #210, opened 2020-04-08
[AFL++] CmpLog mode + Power schedules args fix

andreafioraldi:
honggfuzz increased a lot its coverage using tokens extraction from *cmp functions.
https://groups.google.com/forum/m/?utm_medium=email&utm_source=footer#!topic/fuzzing-discuss/9bzAZs_BMUA

I want to add this configuration (never tested here before) that we use to extract tokens from cmp instructions and routines at runtime to see if it can do the same job.

Sorry for the tons of afl++ directories for many different confs, this conf is already in aflplusplus/fuzzer.py and can be enabled changing the BUILD_MODES var if you want to maintain a single file (as the others can be enabled changing the ADDITIONAL_ARGS env var).


vanhauser-thc:
sorry for my last PR with the bugs that andrea is fixing ...

andreafioraldi:
Some build benchmarks are failing but are related to other fuzzers that I didn't touched

robertswiecki:
I think it's the new proxygen benchmark, which is failing for me with OOMs even in a 4GB VM :)

jonathanmetzman:
> I think it's the new proxygen benchmark, which is failing for me with OOMs even in a 4GB VM :)

Does it OOM? I've seen it causing failures when building benchmarks because we run out of disk space. We may need to remove it.

robertswiecki:
It OOMs my personal 8GB VM, but I think parallelism (by default JOBS=8 in proxygen's build.sh) can be lowered by using --jobs <n> in the proxygen's build.sh script. Not sure about the presubmits checks here, it was also not enough space in some cases I've seen.

```
vm-ubuntu login: Out of memory: Killed process 3515 (clang-10) total-vm:1783160kB, anon-rss:1660208kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:3520kB oom_score_adj:0
Out of memory: Killed process 3423 (clang-10) total-vm:1408580kB, anon-rss:1291328kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:2780kB oom_score_adj:0
```

inferno-chromium:
> It OOMs my personal 8GB VM, but I think parallelism (by default JOBS=8 in proxygen's build.sh) can be lowered by using --jobs in the proxygen's build.sh script. Not sure about the presubmits checks here, it was also not enough space in some cases I've seen.
> 
> ```
> vm-ubuntu login: Out of memory: Killed process 3515 (clang-10) total-vm:1783160kB, anon-rss:1660208kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:3520kB oom_score_adj:0
> Out of memory: Killed process 3423 (clang-10) total-vm:1408580kB, anon-rss:1291328kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:2780kB oom_score_adj:0
> ```

Yes that benchmark is removed for now, so CI should finish properly.

jonathanmetzman:
I started an(other) experiment with these fuzzers. 
I'm going to be offline until Sunday so you probably won't be able to see results until then or Monday.

jonathanmetzman:
[Here](https://www.fuzzbench.com/reports/2020-04-08-aflplusplus-and-2020-04-01/index.html) are the results.

I couldn't add all of the fuzzers from 2020-04-01 for some reason (tracked in #219) but I did compare all of the afl++ based fuzzers to one another.

PR #213, opened 2020-04-08
Add short run tests to make and CI and add unittests for fuzzers

### LINKED ISSUE #211: Run fuzzers in CI
Recently Marc opened a PR to test the different power schedules https://github.com/google/fuzzbench/pull/206.

I tested it locally while adding cmplog and he added the flags for the schedules in a wrong way.
`flags = ['-p fast']` instead of `flags = ['-p', 'fast']`

This prevented afl++ to start using make run-afplusplus-target.

In the PR is see that all Travis checks were passed but this is clearly a bug IMO.

jonathanmetzman:
Add a make target that runs fuzzers for 20 seconds and fails if they error out.
Add this to CI (fixes #211).

Add some unittests for fuzzer modules
Test that they don't have TypeErrors and that they exception when fuzzing  or building fails
(this verifies that the run tests will work).


jonathanmetzman:
PTAL @inferno-chromium

common/experiment_utils.py
```diff
@@ -19,7 +19,7 @@
 from common import environment
 
 # Time interval for collecting experiment data (e.g. corpus, crashes).
-SNAPSHOT_PERIOD = 15 * 60  # Seconds.
+SNAPSHOT_PERIOD = environment.get('SNAPSHOT_PERIOD', 15 * 60)  # Seconds.
```
inferno-chromium, line 22: Can we keep this a constant, and move the caller to use environment.get. Not good to set dynamic vars through global.

experiment/runner.py
```diff
@@ -466,6 +469,8 @@ def main():
             'trial_id': str(environment.get('TRIAL_ID')),
         })
     experiment_main()
+    if fuzzer_errored_out:
```
inferno-chromium, line 472: nit: return int(fuzzer_errored_out)

fuzzers/test_fuzzers.py
```diff
@@ -0,0 +1,76 @@
...
+    ]
+
+
+@pytest.mark.parametrize('fuzzer_module', _get_all_fuzzer_modules())
+def test_build_function_errors(fuzzer_module, fs):
+    """Test that calling the build function will cause an error when
+    the benchmark can't be built. Test also ensures that there are no
+    type errors which can be made by having the wrong signature."""
+    with pytest.raises(Exception) as error:
+        fuzzer_module.build()
```
inferno-chromium, line 54: I dont understand these tests. Does this mean it should always error out when calling the build. Shouldn't we have the test the other way that it always passes when you call the fuzz and build functions.

jonathanmetzman:
Looks like eclisper fails to build because of disk space issues :-(

experiment/runner.py
jonathanmetzman: Disagree on this. The current version seems clearer and more explicit.
It seems weird to convert `True` to a returncode.

fuzzers/test_fuzzers.py
jonathanmetzman: >I dont understand these tests.

The tests intentionally call build/fuzz under circumstances in which they should fail and then asserts that they do in fact fail.
The reason I added this test is because when I was adding the run feature to CI, I found that most fuzzers as they were written did not exception when the command they executed returned nonzero. 

>Does this mean it should always error out when calling the build. 

Exactly.

>Shouldn't we have the test the other way that it always passes when you call the fuzz and build functions.

Ideally yes, but it isn't practical to test that everything works, it's easier to test that they fail properly.
We have this in CI, not locally though because it would take too long to build and run every fuzzer.

common/experiment_utils.py
jonathanmetzman: I'll probably replace with a function `get_snapshot_seconds()`

fuzzers/test_fuzzers.py
jonathanmetzman: The `build` test is probably unnecessary but it probably doesn't hurt to have (except hurting test time since we are running each fuzzer as a different test (e.g. one for HF, one for AFL etc.)).

inferno-chromium: What does the test failure look like, is it just timing out ? Should we test that exception is timeout one, as otherwise catching any test can be some more important problem.

alifahmed:
> Looks like eclisper fails to build because of disk space issues :-(
> fastcgs_br seems to run into an issue with bloaty CC @alifahmed

Looking into it. Thanks!

alifahmed:
> Looks like eclisper fails to build because of disk space issues :-(
> fastcgs_br seems to run into an issue with bloaty CC @alifahmed

@jonathanmetzman Submitted a fixed version (#246). I was trying to ensure unique keys for each branch. For bloaty_fuzz_target it did not work because number of branches were more than map size.

fuzzers/test_fuzzers.py
jonathanmetzman: No it isn't just timing out.
I don't know if we should specify which exception is thrown because I don't think it's important to enforce this. I just want to ensure that when building is impossible an exception is thrown.
In most cases it will probably be subprocess.CalledProcessError when utils.build_benchmark is called, but again i don't consider it important that this happens. I only consider it important that they throw exceptions of some kind.

common/experiment_utils.py
jonathanmetzman: Did this.

jonathanmetzman:
I don't know why eclipser is failing :-(

PR #216, opened 2020-04-10
Only use Docker's cache-from when on a CI service.

rsprabery:
This speeds up local caching when developing new fuzzers.

@lszekeres 

This change dynamically sets the `--cache-from` option only if the `GITHUB_RUN_ID` is present in the environment variables - which is for by [default](https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables) for GitHub Workflows (the CI we use).

To understand when this is useful, consider a failing `./configure` script, which we can simulate with a bad command.

## Simulating the Problem

Add a failing command to AFL's `builder.Dockerfile`:

`RUN ln asdf` - insert after AFL has been built (after line 24) so that AFL's `builder.Dockerfile` looks like:

```dockerfile
FROM $parent_image

# Download and compile AFL v2.56b.
# Set AFL_NO_X86 to skip flaky tests.
RUN git clone https://github.com/google/AFL.git /afl && \
    cd /afl && \
    git checkout 8da80951dd7eeeb3e3b5a3bcd36c485045f40274 && \
    AFL_NO_X86=1 make

RUN ln asdf

# Use afl_driver.cpp from LLVM as our fuzzing library.
RUN apt-get update && \
    apt-get install wget -y && \
    wget https://raw.githubusercontent.com/llvm/llvm-project/5feb80e748924606531ba28c97fe65145c65372e/compiler-rt/lib/fuzzer/afl/afl_driver.cpp -O /afl/afl_driver.cpp && \
    clang -Wno-pointer-sign -c /afl/llvm_mode/afl-llvm-rt.o.c -I/afl && \
    clang++ -stdlib=libc++ -std=c++11 -O2 -c /afl/afl_driver.cpp && \
    ar r /libAFL.a *.o
```

Then run `make build-afl-libpng-1.2.56` twice in the `master` branch and noticed how AFL is cloned down and built each time.

## Testing the Fix

In this branch, intermediate layers are cached before the final tagging takes place and running `make build-afl-libpng-1.2.56` twice will only run commands before the failing command once. This is particularly useful when adding fuzzers with complex dependencies (KLEE for example).

### DIFF
--- a/.github/workflows/build.py
+++ b/.github/workflows/build.py
@@ -72,7 +72,7 @@ def make_builds(benchmarks, fuzzer):
 
         # Then build.
         print('Building', build_target)
-        build_command = ['make', '-j', build_target]
+        build_command = ['make', 'RUNNING_ON_CI=yes', '-j', build_target]
         result = subprocess.run(build_command, check=False)
         if not result.returncode == 0:
             return False
--- a/docker/build.mk
+++ b/docker/build.mk
@@ -22,11 +22,18 @@ BASE_TAG ?= gcr.io/fuzzbench
 build-all: $(addsuffix -all, $(addprefix build-,$(FUZZERS)))
 pull-all: $(addsuffix -all, $(addprefix pull-,$(FUZZERS)))
 
+# If we're running on a CI service, cache-from a remote image. Otherwise just
+# use the local cache.
+cache_from = $(if ${RUNNING_ON_CI},--cache-from $(1),)
+
+# For base-* images (and those that depend on it), use a remote cache by
+# default, unless the developer sets DISABLE_REMOTE_CACHE_FOR_BASE.
+cache_from_base = $(if ${DISABLE_REMOTE_CACHE_FOR_BASE},,--cache-from $(1))
 
 base-image:
 	docker build \
     --tag $(BASE_TAG)/base-image \
-    --cache-from $(BASE_TAG)/base-image \
+    $(call cache_from_base,${BASE_TAG}/base-image) \
     docker/base-image
 
 pull-base-image:
@@ -35,8 +42,8 @@ pull-base-image:
 base-builder: base-image
 	docker build \
     --tag $(BASE_TAG)/base-builder \
-    --cache-from $(BASE_TAG)/base-builder \
-    --cache-from gcr.io/oss-fuzz-base/base-clang \
+    $(call cache_from_base,${BASE_TAG}/base-builder) \
+    $(call cache_from_base,gcr.io/oss-fuzz-base/base-clang) \
     docker/base-builder
 
 pull-base-clang:
@@ -48,7 +55,7 @@ pull-base-builder: pull-base-image pull-base-clang
 base-runner: base-image
 	docker build \
     --tag $(BASE_TAG)/base-runner \
-    --cache-from $(BASE_TAG)/base-runner \
+    $(call cache_from_base,${BASE_TAG}/base-runner) \
     docker/base-runner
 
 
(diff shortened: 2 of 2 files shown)


docker/build.mk
```diff
@@ -22,11 +22,13 @@ BASE_TAG ?= gcr.io/fuzzbench
...
 
+# If we're running using GitHub Workflows, cache-from a remote image. Otherwise
+# just use the local cache.
+cache_from = $(if ${GITHUB_RUN_ID},--cache-from $(1),)
 
 base-image:
 	docker build \
-    --tag $(BASE_TAG)/base-image \
-    --cache-from $(BASE_TAG)/base-image \
+    --tag $(BASE_TAG)/base-image $(call cache_from,${BASE_TAG}/base-image) \
```
inferno-chromium, line 31: better to split into new line for easier reading.

docker/build.mk
```diff
@@ -22,11 +22,13 @@ BASE_TAG ?= gcr.io/fuzzbench
 build-all: $(addsuffix -all, $(addprefix build-,$(FUZZERS)))
 pull-all: $(addsuffix -all, $(addprefix pull-,$(FUZZERS)))
 
+# If we're running using GitHub Workflows, cache-from a remote image. Otherwise
+# just use the local cache.
+cache_from = $(if ${GITHUB_RUN_ID},--cache-from $(1),)
```
inferno-chromium, line 27: please add travis as well

docker/build.mk
rsprabery: Fixed!

docker/build.mk
rsprabery: I've changed it to where we set the env var in `build.py` so as long as we use that for Travis or other CI's, it should be fine.

I'm watching the test run below to ensure the env var is properly propagated though.

.github/workflows/build.py
```diff
@@ -100,6 +101,11 @@ def main():
         return 1
     build_type = sys.argv[1]
     fuzzer = sys.argv[2]
+
+    # Set environment variable indicating to our Makefiles that this
+    # is being run on a CI service.
+    os.environ["RUNNING_ON_CI"] = "yes"
```
lszekeres, line 107: Consider adding this to the `make` command instead in `make_builds()`. Ie. `make RUNNING_ON_CI=yes -j ...`

inferno-chromium:
Lets wait on @jonathanmetzman review too, since i think this will bad for building base-* images since that will be too slow locally.

docker/build.mk
```diff
@@ -22,11 +22,14 @@ BASE_TAG ?= gcr.io/fuzzbench
 build-all: $(addsuffix -all, $(addprefix build-,$(FUZZERS)))
 pull-all: $(addsuffix -all, $(addprefix pull-,$(FUZZERS)))
 
+# If we're running using GitHub Workflows, cache-from a remote image. Otherwise
```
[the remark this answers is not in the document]
rsprabery: Done!

.github/workflows/build.py
rsprabery: Done!

rsprabery:
@inferno-chromium - sounds good. I've changed the behavior so that base-* builds still uses `--cache-from` unless a developer specifies `DISABLE_REMOTE_CACHE_FOR_BASE=yes` with the `make` command.

Let's see what @jonathanmetzman thinks. Will be good to get a second set of eyes on things considered "base" incase I've overlooked one of the base targets.

.github/workflows/build.py
```diff
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Script for building fuzzer,benchmark pairs in CI."""
+import os
```
jonathanmetzman, line 16: There's a linting failure because of this line, could you please remove it?

rsprabery:
One minor change to get past linting error. (Removed `import os` in the workflow script)

PR #217, opened 2020-04-13
Update documentation for fuzzer configuration

mboehme:
In the the fuzzer configuration, set cloud_compute_zone to the zone (not the region).


mboehme:
@googlebot I signed it!

PR #220, opened 2020-04-13
[afl++] Delete one-time variants

jonathanmetzman:
These were benchmarked in
https://www.fuzzbench.com/reports/2020-04-08-aflplusplus-and-2020-04-01/index.html
We don't need to keep benchmarking them.


jonathanmetzman:
CC @vanhauser-thc @andreafioraldi 
Is it OK to delete these?

andreafioraldi:
super fine for me

PR #231, opened 2020-04-15
Add support for fuzzer variants specified by variants.yaml files.

mbarbella-chromium:
PTAL. This is a little bit different from what's currently proposed in the one-pager, but Jonathan and I discussed this yesterday and settled on this approach. I'll update the doc to reflect the current implementation soon.

I still plan to add a test to verify that all checked-in variants.yaml files are properly formatted. In the meantime, do either of you have ideas for places where we'd want to use this now? I'd like to check in at least one variants.yaml file with this change if I can so that there's an example handy. I'm testing it locally using value profile, but I don't think that's actually a configuration we want to maintain as a variant.

### DIFF
--- a/common/fuzzer_utils.py
+++ b/common/fuzzer_utils.py
@@ -20,6 +20,7 @@ from typing import Optional
 
 from common import logs
 from common import utils
+from common import yaml_utils
 
 DEFAULT_FUZZ_TARGET_NAME = 'fuzz-target'
 FUZZ_TARGET_SEARCH_STRING = b'LLVMFuzzerTestOneInput'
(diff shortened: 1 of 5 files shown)


common/fuzzer_utils.py
```diff
@@ -77,11 +78,32 @@ def validate(fuzzer):
...
+            # Auto-generate the default configuration for each base fuzzer.
+            fuzzer_configs.append({'fuzzer': fuzzer})
+
+        variant_config_path = os.path.join(fuzzers_dir, fuzzer, 'variants.yaml')
+        if not os.path.isfile(variant_config_path):
+            continue
+
+        variant_config = yaml_utils.read(variant_config_path)
+        for variant in variant_config['variants']:
+            if not fuzzers or variant['name'] in fuzzers:
```
inferno-chromium, line 101: Probably add
assert 'name' in variant, 'Missing name attribute for fuzzer variant.'

common/fuzzer_utils.py
```diff
@@ -77,11 +78,32 @@ def validate(fuzzer):
...
+        if not fuzzers or fuzzer in fuzzers:
+            # Auto-generate the default configuration for each base fuzzer.
+            fuzzer_configs.append({'fuzzer': fuzzer})
+
+        variant_config_path = os.path.join(fuzzers_dir, fuzzer, 'variants.yaml')
+        if not os.path.isfile(variant_config_path):
+            continue
+
+        variant_config = yaml_utils.read(variant_config_path)
+        for variant in variant_config['variants']:
```
inferno-chromium, line 100: Probably add
assert 'variants' in variant_config, 'Missing variants attribute in variants.yaml.'

mbarbella-chromium: Done.

common/fuzzer_utils.py
mbarbella-chromium: Done.

common/fuzzer_utils.py
```diff
@@ -77,11 +78,37 @@ def validate(fuzzer):
         return False
 
 
-def get_all_fuzzers():
+def get_fuzzer_configs(fuzzers=None):
     """Returns the list of all fuzzers."""
```
jonathanmetzman, line 82: Comment should probably be updated.

common/yaml_utils.py
```diff
@@ -23,3 +23,9 @@ def read(yaml_filename):
 
     with open(yaml_filename) as file_handle:
         return yaml.load(file_handle, yaml.SafeLoader)
+
+
+def write(yaml_filename, data):
+    """Writes data to a new yaml file at |yaml_filename|."""
```
jonathanmetzman, line 29: |data|

common/fuzzer_utils.py
```diff
@@ -77,11 +78,37 @@ def validate(fuzzer):
...
+        assert 'variants' in variant_config, (
+            'Missing "variants" section of {}'.format(variant_config_path))
+        for variant in variant_config['variants']:
+            if not fuzzers or variant['name'] in fuzzers:
+                # Modify the config from the variants.yaml format to the
+                # format expected by a fuzzer config.
+                assert 'name' in variant, (
+                    'Missing name attribute for fuzzer variant in {}'.format(
+                        variant_config_path))
+                variant['variant_name'] = variant['name']
```
jonathanmetzman, line 109: Don't really see a lot of value in renaming it "variant_name" I would just leave as is, makes it easier to reason about.

jonathanmetzman:
I just realized that this doesn't add make support. We need make support before anyone can start using this.

PR #247, opened 2020-04-21
Enable huge pages on runner instances.

lszekeres:
cc @alifahmed

### DIFF
--- a/experiment/resources/runner-startup-script-template.sh
+++ b/experiment/resources/runner-startup-script-template.sh
@@ -15,6 +15,9 @@
 
 ## Configure the host.
 
+# Enable 8 2MB huge pages.
+echo 8 > /proc/sys/vm/nr_hugepages
+
 # Make everything ptrace-able.
 echo 0 > /proc/sys/kernel/yama/ptrace_scope
 


jonathanmetzman:
@alifahmed is this what you meant?
@lszekeres Please see presubmit failure.
I'll approve this once alifahmed approves, but I'd like to merge it myself just for the purpose of 1 experiment so I can revert right after, or do you think we should keep this long term (maybe depends on results).

jonathanmetzman:
@alifahmed is this what you meant?
@lszekeres Please see presubmit failure.
I'll approve this once alifahmed approves, but I'd like to merge it myself just for the purpose of 1 experiment so I can revert right after, or do you think we should keep this long term (maybe depends on results).

alifahmed:
@lszekeres @jonathanmetzman : Thanks, this is exactly what I need! However, I think the number of pages could be too low. Each fastcgs instance will use 3 hugepages. It will work fine with 8 pages available only upto two instances running at the same time.

alifahmed:
I think the presumbit failure can be fixed by updating experiment/test_scheduler.py

https://github.com/google/fuzzbench/blob/fa230b554c8ad2030806d423f4e9f0ee06acc12b/experiment/test_scheduler.py#L90-L93

https://github.com/google/fuzzbench/blob/fa230b554c8ad2030806d423f4e9f0ee06acc12b/experiment/test_scheduler.py#L147-L150

lszekeres:
> @lszekeres @jonathanmetzman : Thanks, this is exactly what I need! However, I think the number of pages could be too low. Each fastcgs instance will use 3 hugepages. It will work fine with 8 pages available only upto two instances running at the same time.

OK that's good, because one VM runs a single trial, so for each fastcgs instance there will be 8 pages available, so no changes are needed.

lszekeres:
> @lszekeres Please see presubmit failure.

Fixed in https://github.com/google/fuzzbench/pull/257, so this PR only includes the relevant lines.

> I'll approve this once alifahmed approves, but I'd like to merge it myself just for the purpose of 1 experiment so I can revert right after, or do you think we should keep this long term (maybe depends on results).

If it will have positive result, then sure. But in any case it's fine to keep: in the worst case we're wasting 16 MB of memory :)

inferno-chromium:
I would defer to @jonathanmetzman review on this.

jonathanmetzman:
Is there still a reason to merge this if we land: https://github.com/google/fuzzbench/pull/258 ?

jonathanmetzman:
> Is there still a reason to merge this if we land: #258 ?
> Maybe we should land that, confirm if it leads to a positive result and then enable everywhere if that's the case?

Nevermind, I see that PR depends on this.

experiment/resources/runner-startup-script-template.sh
```diff
@@ -15,6 +15,10 @@
 
 ## Configure the host.
 
+# Enable 8 2MB huge pages.
+echo 8 > /proc/sys/vm/nr_hugepages
+grep Huge /proc/meminfo
```
jonathanmetzman, line 20: Is this grep needed?

jonathanmetzman:
LGTM other than the grep.

experiment/resources/runner-startup-script-template.sh
lszekeres: done

PR #250, opened 2020-04-22
Expose BENCHMARK variable during fuzzer build() function.

### LINKED ISSUE #191: expose BENCHMARK env variable to builder container
It would be useful to expose a BENCHMARK env variable to the runner. This way fuzzer.py can use it adjust the building code (fuzzer.py). For example: currently lafintel uses the PWD env variable to detect the php benchmark. Cleaner to have a BENCHMARK variable

inferno-chromium:
Fixes #191

### DIFF
--- a/docker/benchmark-builder/Dockerfile
+++ b/docker/benchmark-builder/Dockerfile
@@ -34,6 +34,9 @@ ARG benchmark
 # Set up source and output directories.
 ENV SRC /src
 ENV OUT /out
+
+ENV FUZZER $fuzzer
+ENV BENCHMARK $benchmark
 ENV FUZZ_TARGET $OUT/fuzz-target
 
 WORKDIR $SRC
--- a/docker/build.mk
+++ b/docker/build.mk
@@ -216,6 +216,7 @@ define fuzzer_oss_fuzz_benchmark_template
     --file=docker/oss-fuzz-builder/Dockerfile \
     --build-arg parent_image=$(BASE_TAG)/oss-fuzz/builders/$(1)/$($(2)-project-name)-intermediate \
     --build-arg fuzzer=$(1) \
+    --build-arg benchmark=$(2) \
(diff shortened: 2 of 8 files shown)


docker/build.mk
```diff
@@ -216,6 +216,7 @@ define fuzzer_oss_fuzz_benchmark_template
     --file=docker/oss-fuzz-builder/Dockerfile \
     --build-arg parent_image=$(BASE_TAG)/oss-fuzz/builders/$(1)/$($(2)-project-name)-intermediate \
     --build-arg fuzzer=$(1) \
+    --build-arg benchmark=$(2) \
```
jonathanmetzman, line 219: This probably needs to be done in the config files in `experiment/gcs` as well

docker/benchmark-builder/Dockerfile
```diff
@@ -34,6 +34,9 @@ ARG benchmark
 # Set up source and output directories.
 ENV SRC /src
 ENV OUT /out
+
```
jonathanmetzman, line 37: Won't this be inconsistent with ossfuzz?

jonathanmetzman: For the record I'm designing a bit of a refactor of docker images so that ossfuzz images will use this. Should get started later today on this.

docker/build.mk
inferno-chromium: Yes, this is in my CL :)

docker/benchmark-builder/Dockerfile
inferno-chromium: Yes i changed oss-fuzz-builder/Dockerfile as well in this CL.

jonathanmetzman: Didn't see, sorry

docker/build.mk
jonathanmetzman: Ah didn't see.

PR #252, opened 2020-04-22
Don't install benchmark specific dependencies in base-builder

jonathanmetzman:
Install them in the becnhmark's build.sh instead.


jonathanmetzman:
"aflplusplus_ngram2" is caused by failing to connect to sourceforge.

PR #256, opened 2020-04-22
Remove temporary afl++ variants

jonathanmetzman:
The experiment where these were run seems to be working so we won't run these again.
I want to get rid of them because having so many fuzzers is making my CI jobs get queued (since I've done a couple new PRs in the last 30 minutes).


jonathanmetzman:
CC @vanhauser-thc @andreafioraldi

vanhauser-thc:
dont worry, they will be replaced fast ;)

PR #257, opened 2020-04-23
Remove boilerplate from scheduler test.

lszekeres:
Compare only the docker startup part of the script, as the rest doesn't contain
any template variables. Checking the fixed part of the template is not
interesting for the test and just adds unnecessary boilerplate.

### DIFF
--- a/experiment/resources/runner-startup-script-template.sh
+++ b/experiment/resources/runner-startup-script-template.sh
@@ -13,8 +13,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+## Configure the host.
+
+# Make everything ptrace-able.
 echo 0 > /proc/sys/kernel/yama/ptrace_scope
+
+# Do not notify external programs about core dumps.
 echo core >/proc/sys/kernel/core_pattern
+
+## Start docker.
+
 {% if not local_experiment %}
 while ! docker pull {{docker_image_url}}
 do
--- a/experiment/resources/runner-startup-script-template.sh
+++ b/experiment/resources/runner-startup-script-template.sh
@@ -22,7 +22,6 @@ echo 0 > /proc/sys/kernel/yama/ptrace_scope
 echo core >/proc/sys/kernel/core_pattern
 
 ## Start docker.
-
 {% if not local_experiment %}
 while ! docker pull {{docker_image_url}}
 do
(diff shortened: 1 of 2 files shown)


experiment/test_scheduler.py
```diff
@@ -201,7 +171,9 @@ def _test_create_trial_instance(benchmark, expected_image, expected_target,
         startup_script=expected_startup_script_path)
 
     with open(expected_startup_script_path) as file_handle:
-        assert file_handle.read() == expected_format_string.format(
+        content = file_handle.read()
+        script_for_docker = content[content.find('## Start docker.'):]
```
inferno-chromium, line 175: maybe assert '## Start docker.' in content first.

experiment/resources/runner-startup-script-template.sh
```diff
@@ -13,8 +13,16 @@
...
+## Configure the host.
+
+# Make everything ptrace-able.
 echo 0 > /proc/sys/kernel/yama/ptrace_scope
+
+# Do not notify external programs about core dumps.
 echo core >/proc/sys/kernel/core_pattern
+
+## Start docker.
+
```
inferno-chromium, line 25: can remove newline.

lszekeres:
> Not sure I think this is an improvement.
> I think the boilerplate (ugly as it is) is nicer than having extra logic to avoid it, the code is easier to understand with the boilerplate.

The motivation is https://github.com/google/fuzzbench/pull/247 and the problem is redundancy: we have the same code in three different places, which means that if we want to make a change, we need to update it in three places. The part of the test that checks the template instantiation is useful, but the part that tests if X=X is less. wdty?

jonathanmetzman:
> > Not sure I think this is an improvement.
> > I think the boilerplate (ugly as it is) is nicer than having extra logic to avoid it, the code is easier to understand with the boilerplate.
> 
> The motivation is #247 and the problem is redundancy: we have the same code in three different places, which means that if we want to make a change, we need to update it in three places. The part of the test that checks the template instantiation is useful, but the part that tests if X=X is less. wdty?

Heh, you know how to convince me, I think having to change tests in multiple places because they test the same thing is very bad. Yep overall this LGTM, lemme just review for real now.

experiment/test_scheduler.py
lszekeres: Good idea! done

experiment/resources/runner-startup-script-template.sh
lszekeres: done

PR #258, opened 2020-04-23
[fastcgs] Huge page versions

alifahmed:
Enables huge page on fastcgs. Will keep normal or this version based on the performance difference. Depends on PR #247.


.github/workflows/fuzzers.yml
```diff
@@ -36,7 +36,9 @@ jobs:
           - entropic
           - fairfuzz
           - fastcgs
+          - fastcgs_huge
```
jonathanmetzman, line 39: Do we need to test this feature with both? Or do you think one is enough to tell whether it is useful?

jonathanmetzman:
So is this PR downloading a new version of fastcgs that can use these large pages?

.github/workflows/fuzzers.yml
alifahmed: One is actually enough to test this feature. I wanted to compare huge page versions of fastcgs and fastcgs_lm, that's why added both. Please let me know if I should remove one of them.

alifahmed:
> 
> 
> So is this PR downloading a new version of fastcgs that can use these large pages?

Yes, the current fastcgs/fastcgs_lm does not have large pages enabled. This PR adds fastcgs_huge/fastcgs_lm_huge with large pages enabled, everything else being same.

alifahmed:
@lszekeres @jonathanmetzman I rerun the CI, but shmget() still failing to access huge pages. I am not sure where the problem is. Any idea what went wrong?

jonathanmetzman:
> @lszekeres @jonathanmetzman I rerun the CI, but shmget() still failing to access huge pages. I am not sure where the problem is. Any idea what went wrong?

I personally wouldn't know the answer to this, but maybe the CI environment is just weird and this won't happen in production? What happens if you run it locally? Does it work then?

lszekeres:
> > @lszekeres @jonathanmetzman I rerun the CI, but shmget() still failing to access huge pages. I am not sure where the problem is. Any idea what went wrong?
> 
> I personally wouldn't know the answer to this, but maybe the CI environment is just weird and this won't happen in production? What happens if you run it locally? Does it work then?

We enabled huge pages on our production VMs (on GCE) that the FuzzBench service runs the experiments on. The CI runs on a completely different infrastructure (on GitHub).

.github/workflows/fuzzers.yml
jonathanmetzman: Let's do one for now and if it does better we can try it on the variant.

jonathanmetzman: If you think we're likely to miss something by this approach (I don't know much about the big page tables and know nothing about fastcgs_lm) then we can keep. I'm just trying to cut back on unnecessary runs by multiplying features.

alifahmed:
> > @lszekeres @jonathanmetzman I rerun the CI, but shmget() still failing to access huge pages. I am not sure where the problem is. Any idea what went wrong?
> 
> I personally wouldn't know the answer to this, but maybe the CI environment is just weird and this won't happen in production? What happens if you run it locally? Does it work then?

@jonathanmetzman @lszekeres I tried to run locally and have mixed experience. If I enable huge pages outside docker and then run, it works perfectly:
```
# echo 8 > /proc/sys/vm/nr_hugepages
$ make run-fastcgs_lm-<bench>
```

On the other hand, it does not work in the following case:
```
# echo 0 > /proc/sys/vm/nr_hugepages
$ make run-fastcgs_lm-<bench>
```

I thought huge page is being enabled by the docker instance, and both cases should work.

jonathanmetzman:
I guess if it fails in CI than could you please remove it from CI?
Thanks!

alifahmed:
> 
> 
> I guess if it fails in CI than could you please remove it from CI?
> Thanks!

Done.

.github/workflows/fuzzers.yml
alifahmed: Done. Kept fastcgs_lm_huge and removed fastcgs_huge.

PR #214, opened 2020-04-09
Initial contribution from SVRWB benchmark

roachspray:
- This pull request is related to issue #50  and includes:
   - `audiofile 0.3.6` and specifically libaudiofile. The harness is based on `sfconvert"`which is a utility that is a part of this project. 
   - `jasper 1.701.0` and libjasper. The harness is based on jasper.c which is a utility in that project.
  - `tcpdump 4.9.0`
- I was unable to test in the FuzzBench environment, but able to test by hand ... It needs FuzzBench testing. All feedback is good feedback. :)
- I include bugs in `known_bugs` directory. For audiofile, are 4 CVEs that do not want to crash under this setup, but do with the original sfconvert utility. Similarly, there were a handful of CVEs that were not wanting to crash for jasper (or only in 32 or 64 bit mode). All of the samples not included may be found in the original svrwb benchmark for reference (https://github.com/veracode-research/svrwb-fuzz-benchmark-suite/)
- I do not include any seed files, but a seed should be a WAV file. I have one I used, but it was selected w/o any rhyme or reason other than it was the correct type and AFL liked it's size.
- There are another bunch of crashes for `sfinfo` tool, but I believe it should be a different fuzz target rather than build a harness to handle both sfconvert and sfinfo.
- This work is done with the support of the Veracode Applied Research Group.

When this is OK'd (after whatever iterations), I will work on others. Some others will require common.sh changes -- e.g., using wget(1) to pull a .tgz from a URL and unpack it.

Issue #165, opened 2020-03-26
Use bugs to measure fuzzer performance


jonathanmetzman:
My view of fuzzers is that "better" fuzzers can find more exploitable bugs than worse fuzzers. While coverage is a decent, easy-to-use, proxy for this, we should look into using crashes to determine fuzzer performance.

We already save the crashes found in each cycle/corpus snapshot.

Here's what's left to do that I can think of:
1. Implementing a method for identifying crashes. ClusterFuzz's method (similar to stack hashing) works pretty well in practice.

2. Finding a way to rank fuzzers based on the crashes they find.

lszekeres:
I think the two biggest tasks for this are:
- Having numerous benchmarks with a large number of real bugs. https://github.com/google/fuzzbench/issues/50 will be a massive contribution to this! Thanks a bunch for  @roachspray again! (Additionally, we could also look at the history of oss-fuzz bugs to find very buggy versions of some oss-fuzz targets). Once these benchmarks with known set of bugs are available, it's relatively straightforward to rank fuzzers based on which one can find more of the bugs and how fast, similarly to coverage, but having a _large number_ of bugs will be very important.
- The other thing, i think, is making sure that fuzzers can progress after finding the first bug (ie they don't get blocked by constantly keep crashing on the first bug).

roachspray:
In my initial pull request, I put known CVE bugs under the benchmark's directory in `known_bugs`. My file naming is the CVE label.... but I am not sure this is the correct route. ...Whether there should exist some fuzzbench mapping of bugs?

jonathanmetzman:
>Whether there should exist some fuzzbench mapping of bugs?

You mean "where" right? I think the scheme you currently used is good for now.

roachspray:
> > Whether there should exist some fuzzbench mapping of bugs?
> 
> You mean "where" right? I think the scheme you currently used is good for now.

OK that works. I just did not know if there might be a chance of bugs without a CVE label, but I am not going to worry about it. :)

mboehme:
+1 for using bugs to measure fuzzer performance!

> Additionally, we could also look at the history of oss-fuzz bugs to find very buggy versions of some oss-fuzz targets.
That would be awesome. Shouldn't ClusterFuzz have all the buggy versions, fixed versions, and reproducers?

jonathanmetzman:
>That would be awesome. Shouldn't ClusterFuzz have all the buggy versions, fixed versions, and reproducers?

[...]

Issue #50, opened 2020-03-03
Additional benchmark cases from svrwb-fuzz-benchmark-suite


roachspray:
I started a similar project last year and since it probably will not be as successful as this one, I figure I should donate the cases I have. They are various OSS apps that have multiple vulnerabilities in a single version; they may be found: https://github.com/veracode-research/svrwb-fuzz-benchmark-suite/tree/master/cases

It seems after a quick look that about 8 of them are not being done by you all (and perhaps some of the sqlite may have gaps between ours). They 8 are:

- audiofile
- imageworsener
- jasper
- lame
- libarchive
- perl
- tcpdump
- wavpack
- ytnef

I am willing to do the work to prep them to be added to your project, but I am curious: should I create an issue for each?

There are additional tests to add that I have not included either, including ChakraCore 1.4.1 (there are multiple vulns), and a few others from recent papers. Any guidance on how to go forward is appreciated.

jonathanmetzman:
Thank you, what a generous donation! These definitely look interesting.

There are going to be some challenges here like making libFuzzer harnesses for each of these benchmarks.
I think the fuzzbench team needs to iron out some details of adding new benchmarks, I'll get back to you on this later this week.

roachspray:
Definitely no worries and I understand that, so makes sense. The harnesses could be written by me (under veracode) as I should have time coming up. I mostly figured I should offer since I would rather try to have the targets I pulled together used than not and having a variety of cases with multiple vulns would be useful. Thanks for considering.

jonathanmetzman:
@roachspray We would love to have these benchmarks when you get a chance. I think adding a few at a time would be a good idea though.
Since we are only measuring coverage, we've discarded benchmarks (I think libarchive may have been one for example) that cause issues running for 24 hours or for measuring coverage. But I think we'd be less likely to discard benchmarks with known bugs since these are much more expensive than coverage-only benchmarks which we can get from OSS-Fuzz for free.
Could you start with integrating the top 3 or 4 benchmarks, we can see how they fare and then proceed from there?

The guide for integrating benchmarks is [here](https://google.github.io/fuzzbench/developing-fuzzbench/adding-a-new-benchmark/). Please reach out if you have any trouble.

roachspray:
Sounds like a plan. I will let you know if you (all) know if I have any issues and get (the top) 3 or 4 in shape with respect to the guide you pointed out.

roachspray:
[...]

Issue #249, opened 2020-04-22
Experiment Announcement Thread


jonathanmetzman:
New experiments will be announced on this issue as discussed on #205. 
I think this is a good place to discuss the experiments as well.

jonathanmetzman:
https://www.fuzzbench.com/reports/2020-04-20-aflplusplus/index.html contains afl++ variants.
The experiment is currently running, most of the trials have completed, though some started running earlier today, and coverage measurement still needs to finish.

I just started another experiment https://www.fuzzbench.com/reports/2020-04-21/index.html that includes every fuzzer except for the afl++ variants benchmarked above.

Once both of these experiments complete I will make a combined report.

Note that coverage measuring for the previous experiment https://www.fuzzbench.com/reports/2020-04-14/index.html didn't complete because of a bug I have fixed.

vanhauser-thc:
@jonathanmetzman when do you expect the 2020-04-20-aflplusplus one to be finished with the measurement? and is the title then updated to not mention "incomplete"?

no need to combine this with 2020-04-21 btw, this will just make the overall report graphs unreadable with so many entries :)

jonathanmetzman:
> @jonathanmetzman when do you expect the 2020-04-20-aflplusplus one to be finished with the measurement? and is the title then updated to not mention "incomplete"?

Title gets updated automatically.
Hard to say when it will complete, I expect a day or two from now.
The recent fixes we made and removal of irssi should reduce measurement time considerably.
Unfortuantely, afl++ experiment started before that.

> 
> no need to combine this with 2020-04-21 btw, this will just make the overall report graphs unreadable with so many entries :)

OK sounds good to me. Let me know if you want just the AFL++ fuzzers then (I can do that as well).

vanhauser-thc:
> OK sounds good to me. Let me know if you want just the AFL++ fuzzers then (I can do that as well).

that would be good however the new run does not have irssi target in there ... does that work and not f*ck up the benchmark calculation?

jonathanmetzman:
I can exclude that benchmark.

jonathanmetzman:
> https://www.fuzzbench.com/reports/2020-04-20-aflplusplus/index.html contains afl++ variants.
> The experiment is currently running, most of the trials have completed, though some started running earlier today, and coverage measurement still needs to finish.
> 
> I just started another experiment https://www.fuzzbench.com/reports/2020-04-21/index.html that includes every fuzzer except for the afl++ variants benchmarked above.
[...]

PR #259, opened 2020-04-23
Initial contribution from SVRWB benchmark (take 2)

roachspray:
The original pull request was #214 , but due to some issues with an app was canceled and re-spun as this pr:

- This pull request is related to issue #50 and includes:
  - `jasper-1.701.0`
  - `perl-5.21.7`
      - This case is kind of interesting and we should be careful because it uses it's own `Configure` shell script, instead of `configure`.
  - `tcpdump-4.9.0`
- For each of the `_fuzz.cc` harness files also has a main() routine which is included upon ifdef `_HAS_MAIN`.
- For each of these, a single seed is given that was randomly chosen.
- For each of these, a subdirectory named `known_bugs` exists that contains a number of *known* bugs. 
- Currently, there is no test case for checking inputs in `known_bugs`, but that will be forthcoming in a future pr.
- These have passed `make build-afl-$BENCHMARK_NAME` and `make run-afl-$BENCHMARK_NAME` on my setup.

There is some debate to the benefit of these bugs (see #165 for deeper discussion), but it perhaps can act as a starting place for getting other pieces of comparison/measurement in working order.

If these are accepted, I will plan to look to add some of the others from the svrwb set.

This work is done with the support of the Veracode Applied Research Group.

cc @jonathanmetzman  @lszekeres


roachspray:
I see the conflicts and they are no issue -- I am testing them now with them merged...

roachspray:
Ok, this should be good to go (pending CI) .. but my local build-$FUZZER-$BENCH and run-$FUZZER-$BENCH passed for the 3 I have added. I had to re-work the tcpdump case because of dependencies and the desire to have them isolated to the benchmark; might be an OK example for other's to reference if they need to do something similar. 

...I think that's it.

roachspray:
Anyone with any clue on `eclipser` failures?

lszekeres:
> Anyone with any clue on `eclipser` failures?

AFAIK it is unrelated to your PR, so please don't worry about it.

benchmarks/jasper-1.701.0/build.sh
```diff
@@ -0,0 +1,43 @@
...
+# XXX There is no hash to check. Git repo for this only goes back to 1.9x.
+#
+ZIPURL=https://www.ece.uvic.ca/~frodo/jasper/software/jasper-1.701.0.zip
+rm -rf SRC jasper-1.701.0 jasper-1.701.0.zip
+wget ${ZIPURL} || (printf "wget failed" && exit 1)
+unzip jasper-1.701.0.zip || (printf "unzip failed" && exit 1)
+mv jasper-1.701.0 SRC
+
+build_lib
+
```
jonathanmetzman, line 34: nit: remove this empty line

benchmarks/tcpdump-4.9.0/build.sh
```diff
@@ -0,0 +1,60 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+. $(dirname $0)/../common.sh
+
+build_lib() {
+  rm -rf BUILD
+  cp -rf SRC BUILD
+  # we build and install bc was having issues with autogen.sh of dbus
+  # makes building the rest easier too..
```
jonathanmetzman, line 22: nit one period.
Also capitalize the beginning of the sentence please.

benchmarks/tcpdump-4.9.0/build.sh
```diff
@@ -0,0 +1,60 @@
...
+   make install)
+  (cd BUILD/libdbus && 
+    CFLAGS="${CFLAGS} -Wno-error" ./autogen.sh --disable-xml-docs --disable-doxygen-docs --disable-ducktype-docs &&
+    make && make install)
+  (cd BUILD/libpcap && ./configure && make &&
+   cd ../tcpdump && ./configure && make)
+}
+
+#
+# tcpdump really wants libpcap to share the same parent directory.
```
jonathanmetzman, line 33: Can you please drop the empty comment lines below and above this line?

benchmarks/tcpdump-4.9.0/build.sh
```diff
@@ -0,0 +1,60 @@
...
+# 
+mkdir SRC
+git clone https://github.com/libexpat/libexpat.git SRC/libexpat
+get_git_tag https://gitlab.freedesktop.org/dbus/dbus.git  dbus-1.10 SRC/libdbus
+get_git_tag https://github.com/the-tcpdump-group/libpcap.git  libpcap-1.9.0 SRC/libpcap
+get_git_tag https://github.com/the-tcpdump-group/tcpdump.git  tcpdump-4.9.0 SRC/tcpdump
+
+build_lib
+
+# copy over shared lib deps to /out
```
jonathanmetzman, line 43: nit: capitalize the beginning of this sentence and end it with a period please.

benchmarks/tcpdump-4.9.0/build.sh
```diff
@@ -0,0 +1,60 @@
...
+cp BUILD/libexpat/expat/lib/.libs/libexpat.so.1.6.11 $OUT/
+ln -s $OUT/libexpat.so.1.6.11 $OUT/libexpat.so.1
+ln -s $OUT/libexpat.so.1.6.11 $OUT/libexpat.so
+cp BUILD/libdbus/dbus/.libs/libdbus-1.so.3.14.16 $OUT/
+ln -s $OUT/libdbus-1.so.3.14.16 $OUT/libdbus-1.so.3
+ln -s $OUT/libdbus-1.so.3.14.16 $OUT/libdbus-1.so
+
+#
+# To test with the main() in tcpdump_fuzz.cc, use -D_HAS_MAIN and disable any
```
jonathanmetzman, line 52: Can you please drop the empty comment lines below and above this line?

docker/base-builder/Dockerfile
```diff
@@ -21,9 +21,13 @@ FROM gcr.io/fuzzbench/base-image
 # recommended for some reason.
 RUN apt-get update -y && apt-get install -y --no-install-recommends \
     automake \
+    bison \
```
jonathanmetzman, line 24: I've just stopped installing dependencies in this file unless it is very generally useful.
cmake and unzip might be appropriate here but I'd rather everything gets installed in build.sh.
Sorry to change things on you in the middle.
Here's an example of how I do this: https://github.com/google/fuzzbench/blob/master/benchmarks/freetype2-2017/build.sh#L18
(we don't have sudo in the images so the script assumes running as root, Do you need these to run on your host?)

benchmarks/tcpdump-4.9.0/tcpdump_fuzz.cc
```diff
@@ -0,0 +1,180 @@
+// Copyright 2020 Google LLC
```
jonathanmetzman, line 1: We don't have an official style for C/C++ but I may clang-format this in the future.
No need to change anything now though.

benchmarks/perl-5.21.7/build.sh
```diff
@@ -0,0 +1,49 @@
...
+#     -Alddlflags="-shared"
+#
+  (cd BUILD && ./Configure -des -Dusedevel && make)
+}
+
+get_git_tag https://github.com/Perl/perl5.git v5.21.7 SRC
+build_lib
+
+#
+# To test with the main() in perl_fuzz.cc, use -D_HAS_MAIN and disable any
```
jonathanmetzman, line 40: Can you drop the empty comment lines please?

benchmarks/jasper-1.701.0/build.sh
```diff
@@ -0,0 +1,43 @@
...
+
+#
+# To test with the main() in jasper_fuzz.cc, use -D_HAS_MAIN and disable any
```
jonathanmetzman, line 37: nit: can you remove the empty comment lines.

benchmarks/jasper-1.701.0/build.sh
```diff
@@ -0,0 +1,43 @@
...
+. $(dirname $0)/../common.sh
+
+build_lib() {
+  rm -rf BUILD
+  cp -rf SRC BUILD
+  (cd BUILD && ./configure --enable-static --disable-shared && make)
+}
+
+#
+# XXX There is no hash to check. Git repo for this only goes back to 1.9x.
```
jonathanmetzman, line 25: nit: can you remove the empty comment lines.

roachspray:
@jonathanmetzman Thank you! I think I have addressed the issues you have mentioned.

docker/base-builder/Dockerfile
roachspray: Fixed.

benchmarks/jasper-1.701.0/build.sh
roachspray: Fixed

benchmarks/jasper-1.701.0/build.sh
roachspray: Fixed

benchmarks/tcpdump-4.9.0/build.sh
roachspray: Fixed

benchmarks/tcpdump-4.9.0/build.sh
roachspray: Fixed.

benchmarks/tcpdump-4.9.0/build.sh
roachspray: Fixed

benchmarks/tcpdump-4.9.0/build.sh
roachspray: Fixed

benchmarks/jasper-1.701.0/build.sh
roachspray: Fixed

roachspray:
I am uncertain why some are checks are failing. Looking at logs, it was not obvious to me. In one case I saw earlier, there were issues with git clone in one of the other benchmarks, but looked random occurrence.

jonathanmetzman:
I'll land this on Monday so if there are any breakages we somehow missed, they can be fixed on a workday.
Thank you for this contribution!

roachspray:
> I'll land this on Monday so if there are any breakages we somehow missed, they can be fixed on a workday.

Thank you.

In going back through the documentation, one thing that is off with my harnesses is not always returning 0 from `LLVMFuzzerTestOneInput` and, in fact, calling abort() and exit() from there (in some cases). Should I change now or after this lands?

roachspray:
Prior to merge, I have another concern regarding exit():

Deeper in perl and tcpdump, `exit()` might still be called and `abort()` deeper in jasper. This is not ideal/proper, for obvious reasons, but modification to make these OK may be non-trivial.

Should this be considered a best-effort or that we must root all these out?

Is it true all the existing cases (not these new ones) all do not invoke exit() (or another bad undesirable function) somewhere in the library code?

jonathanmetzman:
> > I'll land this on Monday so if there are any breakages we somehow missed, they can be fixed on a workday.
> 
> Thank you.
> 
> In going back through the documentation, one thing that is off with my harnesses is not always returning 0 from `LLVMFuzzerTestOneInput` and, in fact, calling abort() and exit() from there (in some cases). I have pushed these changes and they work locally (`build-afl-*` and `run-afl-*`).
> 
> Another note: the jasper seed, a JPEG-2000 file, was crashing for me on the initial run ... which it was not prior to adding to git and CI was not seeing that, I don't think. I have replaced it with another. Oddly, my first attempt to replace this crashing seed by generating one from Preview.app also crashed ... :-/ Anyway, it is replaced.



> > I'll land this on Monday so if there are any breakages we somehow missed, they can be fixed on a workday.
> 
> Thank you.
> 
> In going back through the documentation, one thing that is off with my harnesses is not always returning 0 from `LLVMFuzzerTestOneInput` and, in fact, calling abort() and exit() from there (in some cases). I have pushed these changes and they work locally (`build-afl-*` and `run-afl-*`).

Oh this is not good. Maybe the nonzero paths weren't being reached? Can we make it so that all paths through program return zero?
Which benchmarks do this, I didn't notice this when I reviewed (except perl benchmark I see now can return nonzero).

> 
> Another note: the jasper seed, a JPEG-2000 file, was crashing for me on the initial run ... which it was not prior to adding to git and CI was not seeing that, I don't think. I have replaced it with another. Oddly, my first attempt to replace this crashing seed by generating one from Preview.app also crashed ... :-/ Anyway, it is replaced.

Also weird, can you share the seed? I can try it out.

roachspray:
> > In going back through the documentation, one thing that is off with my harnesses is not always returning 0 from `LLVMFuzzerTestOneInput` and, in fact, calling abort() and exit() from there (in some cases). I have pushed these changes and they work locally (`build-afl-*` and `run-afl-*`).
> 
> Oh this is not good. Maybe the nonzero paths weren't being reached? Can we make it so that all paths through program return zero?
> Which benchmarks do this, I didn't notice this when I reviewed (except perl benchmark I see now can return nonzero).
> 

I think so. 

 Question: what if we caused a crash at those points? The crashes could be removed as known and would lessen changes on target code base.

jonathanmetzman:
> Prior to merge, I have another concern regarding exit():
> 
> Deeper in perl and tcpdump, `exit()` might still be called and `abort()` deeper in jasper. This is not ideal/proper, for obvious reasons, but modification to make these OK may be non-trivial.
> 
> Should this be considered a best-effort or that we must root all these out?

Hmm...I don't really know the answer. libFuzzer normally is incompatible with these. But because of the weird way we run libFuzzer (-fork=1 -ignore_crashes=1) it may actually be OK if not ideal. This is probably why CI didn't catch this issue. I may need to think about this more to decide if it is OK, but it would probably be easier for me if they were removed. One issue that may come up is we currently measure the coverage of the entire AFL output directory including crashes. Since our coverage measurement is done using libFuzzer, it may be slowed down by having a lot of crashes.

Locally I added a somewhat shallow exit/abort into https://raw.githubusercontent.com/llvm-mirror/compiler-rt/master/test/fuzzer/MultipleConstraintsOnSmallInputTest.cpp. It slows down libFuzzer and AFL but both seem to be able to find new coverage. 
CC @kcc and @morehouse to see if they think it makes sense to benchmark libFuzzer under these conditions.

> 
> Is it true all the existing cases (not these new ones) all do not invoke exit() (or another bad undesirable function) somewhere in the code? Obviously, most stand alone libraries likely should not be invoking exit, as is, but...right.. real world.

I'm pretty sure none of them invoke exit.

jonathanmetzman:
> > > In going back through the documentation, one thing that is off with my harnesses is not always returning 0 from `LLVMFuzzerTestOneInput` and, in fact, calling abort() and exit() from there (in some cases). I have pushed these changes and they work locally (`build-afl-*` and `run-afl-*`).
> > 
> > 
> > Oh this is not good. Maybe the nonzero paths weren't being reached? Can we make it so that all paths through program return zero?
> > Which benchmarks do this, I didn't notice this when I reviewed (except perl benchmark I see now can return nonzero).
> 
> I think so.
> 
> Question: what if we caused a crash at those points in the target library that do use exit() or another function that is undesirable? The crashes could be removed as known and would lessen changes on target code base.

Sorry I didn't see this before my last comment. I think exit and abort are equivalent to crashes (returning nonzero should be removed though). My hunch is libFuzzer in our setup can work despite crashes although it is hurt by them. One issue is that this might not be fair to libFuzzer (though some would argue that it is). But a bigger issue is that this could down measuring. We may just need to try it to find out what happens :-)

roachspray:
> > > > In going back through the documentation, one thing that is off with my harnesses is not always returning 0 from `LLVMFuzzerTestOneInput` and, in fact, calling abort() and exit() from there (in some cases). I have pushed these changes and they work locally (`build-afl-*` and `run-afl-*`).
> > > 
> > > 
> > > Oh this is not good. Maybe the nonzero paths weren't being reached? Can we make it so that all paths through program return zero?
> > > Which benchmarks do this, I didn't notice this when I reviewed (except perl benchmark I see now can return nonzero).
> > 
> > 
> > I think so.
> > Question: what if we caused a crash at those points in the target library that do use exit() or another function that is undesirable? The crashes could be removed as known and would lessen changes on target code base.
> 
> Sorry I didn't see this before my last comment. I think exit and abort are equivalent to crashes (returning nonzero should be removed though). My hunch is libFuzzer in our setup can work despite crashes although it is hurt by them. One issue is that this might not be fair to libFuzzer (though some would argue that it is). But a bigger issue is that this could down measuring. We may just need to try it to find out what happens :-)

That works for me... :) it can always be turned off if it's clear there are changes that should be made.

jonathanmetzman:
I still need to figure what the next experiment will be, it will either be the new fuzzers or all fuzzers on the new benchmarks, but I hope to run this soon.

jonathanmetzman:
I started the experiment (see https://github.com/google/fuzzbench/issues/249#issuecomment-623802265) for details. Sorry it took a while

roachspray:
No worries whatsoever... thank you for the heads up!

PR #260, opened 2020-04-23
Add support for builder arguments on GCB.

mbarbella-chromium:
Not quite ready for review here. I have a few questions that it would help to be able to point to the code for, so I'll leave a few inline comments soon. It also needs more testing.


experiment/build/gcb_build.py
```diff
@@ -19,11 +19,21 @@
...
 from common import experiment_utils
+from common import filesystem
+from common import fuzzer_config_utils
 from common import logs
 from common import new_process
 from common import utils
+from common import yaml_utils
 from experiment.build import build_utils
 
+BUILDER_STEP_IDS = [
```
mbarbella-chromium, line 30: I was having trouble wrapping my head around how cloud build worked. It looked like the yaml files were the only way that we configure what's happening there, so did we want to append the arguments to those configs? If so, are these the correct places to append them?

If not let me know, this should be easy enough to remove.

jonathanmetzman: I wonder how this works given that we use multiprocessing to build?
Sorry, but I realized yesterday that this model probably requires more dynamism than the config files allow.
I think we probably need to switch to something like this: https://github.com/google/oss-fuzz/blob/dbe2731d93f8de04969543e45338af26a6dc3773/infra/gcb/build_base_images.py

mbarbella-chromium: Something like the oss-fuzz implementation does look much cleaner but I don't see what missing functionality it provides (just want to make sure I'm not missing something there). Other than that, are these the places we'd want to be plugging the arguments from the config files in? Is anything else missing?

jonathanmetzman: 1. I think when write a new config file it needs to be in a temp directory, otherwise doing this in parallel breaks right?
2. I don't think it is theoretically more powerful than this approach, just cleaner and easier to get right.

mbarbella-chromium: 1. I think a temp directory makes sense either way. There's no reason for it to be in the configs. That said, I'm not sure that this would break as it should be a unique filename. I might be missing something here.
2. I'm not too confident in my understanding of cloud build to refactor this file to use the other API throughout this file. Would you mind taking the config file rewrite approach for now with a TODO to refactor that in a follow up once the configs are working?

jonathanmetzman: 1. Sorry, I didn't look at the code closely, enough, I think the approach you took doesn't suffer from parallelism issues (though we may need to ensure that no two variants have the same name, unless we append the name to the beginning can't remember right now.
2. Yes I can accept this for now.

mbarbella-chromium: 1. Switched to a temp filename anyway. From what I can tell this happens after the config directory sync so there's no reason to handle it the older way.
2. Added a TODO.

experiment/run_experiment.py
```diff
@@ -127,26 +127,34 @@ def validate_fuzzer(fuzzer: str):
 
 def validate_fuzzer_config(fuzzer_config):
     """Validate |fuzzer_config|."""
-    allowed_fields = ['variant_name', 'env', 'fuzzer']
+    allowed_fields = [
+        'name', 'fuzzer_environment', 'builder_arguments', 'fuzzer'
```
inferno-chromium, line 131: build_arguments sound better ?

experiment/build/gcb_build.py
```diff
@@ -59,6 +68,25 @@ def _build_benchmark_coverage(benchmark: str) -> Tuple[int, str]:
...
+
+    # TODO(mbarbella): Rather than rewrite yaml files, use the GCB API.
+    args = fuzzer_config['builder_arguments']
+    config = yaml_utils.read(base)
+    for step in config['steps']:
+        if 'id' in step and step['id'] in BUILDER_STEP_IDS:
+            # Append additional flags before the final argument.
+            step['args'] = step['args'][:-1] + args + [step['args'][-1]]
+
+    new_config_path = tempfile.NamedTemporaryFile().name
```
inferno-chromium, line 85: Where do we clean this temp file.

mbarbella-chromium: Hmm. We don't, and there isn't a particularly clean way to handle cleaning it up with this implementation (since sometimes the yaml file is the default one that we didn't modify). Though writing to the configs directory wasn't too clean either, I don't think there are actually any conflicts so I'll return to that.

mbarbella-chromium: Done

experiment/run_experiment.py
mbarbella-chromium: Done.

mbarbella-chromium:
This seems to be working properly. Just waiting on a short test experiment to be sure then I'll land it.

docker/gcb/fuzzer.yaml
```diff
@@ -36,12 +36,12 @@ steps:
     '--tag',
```
mbarbella-chromium, line 36: Please review the changes here if you get a chance. I think these are correct but it would be easy to miss an occurrence.

docker/gcb/oss-fuzz-fuzzer.yaml
```diff
@@ -41,7 +41,7 @@ steps:
     '--build-arg',
```
mbarbella-chromium, line 41: Ditto for this file.

docker/gcb/fuzzer.yaml
jonathanmetzman: Will do

experiment/run_experiment.py
```diff
@@ -127,26 +127,32 @@ def validate_fuzzer(fuzzer: str):
 
 def validate_fuzzer_config(fuzzer_config):
```
jonathanmetzman, line 128: We should probably validate that no 2 variants have the same name (either variants of same fuzzer or different ones).

docker/gcb/fuzzer.yaml
```diff
@@ -36,12 +36,12 @@ steps:
...
     '${_REPO}/builders/${_FUZZER}',
 
-    '--file=fuzzers/${_FUZZER}/builder.Dockerfile',
+    '--file=fuzzers/${_BASE_FUZZER}/builder.Dockerfile',
```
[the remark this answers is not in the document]
mbarbella-chromium: We do still use underlying in one place. I'll swap this out here.

common/fuzzer_utils.py
```diff
@@ -104,13 +104,9 @@ def get_fuzzer_configs(fuzzers=None):
             'Missing "variants" section of {}'.format(variant_config_path))
         for variant in variant_config['variants']:
             if not fuzzers or variant['name'] in fuzzers:
-                # Modify the config from the variants.yaml format to the
-                # format expected by a fuzzer config.
                 assert 'name' in variant, (
                     'Missing name attribute for fuzzer variant in {}'.format(
                         variant_config_path))
-                variant['variant_name'] = variant['name']
```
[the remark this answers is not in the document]
mbarbella-chromium: No strong preference. I think we could still call this variant_name and avoid the deletion if we really want to, but I also don't think "name" is too confusing. It gets translated into the fuzzer config as name, and if we don't explicitly specify a name I don't think defaulting to the fuzzer is too bad.

experiment/run_experiment.py
mbarbella-chromium: Added a check while we're parsing the variant configs.

docker/gcb/fuzzer.yaml
```diff
@@ -133,7 +133,7 @@ steps:
     '${_REPO}/runners/${_FUZZER}/${_BENCHMARK}',
 
     '--build-arg',
-    'fuzzer=${_FUZZER}',
+    'fuzzer=${_BASE_FUZZER}',
```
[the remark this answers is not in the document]
mbarbella-chromium: Done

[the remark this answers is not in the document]
jonathanmetzman: Oh sorry this was a real question, I have no idea which needs to be used. Do you know?

mbarbella-chromium:
Also corrected the issue on line 34 of oss-fuzz-fuzzer.yaml. I'm even more surprised that it doesn't let me reply to that comment.

Running one last quick experiment to validate that these fixes didn't break anything else. Will land after that.

docker/gcb/fuzzer.yaml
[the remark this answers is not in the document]
mbarbella-chromium: I'm not sure which is correct. I'll keep an eye on the test run though to see if this works.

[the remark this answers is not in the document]
mbarbella-chromium: Looks like this ends up getting used as the fuzzer directory name under fuzzers/ so _UNDERLYING_FUZZER was required. I don't think this caused any other issues when doing my initial testing before changing it.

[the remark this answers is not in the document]
jonathanmetzman: Got it!

mbarbella-chromium:
I'm guessing these are spurious CI failures since most look like variations of:
fatal: unable to access 'https://github.com/mboehme/aflfast.git/': Failed to connect to github.com port 443: Connection timed out

I'm comfortable with landing this whenever, but want to confirm that these aren't likely real issues. Feel free to submit if so.

docker/gcb/oss-fuzz-fuzzer.yaml
```diff
@@ -31,7 +31,7 @@ steps:
     '--tag',
     '${_REPO}/oss-fuzz/builders/${_FUZZER}/${_BENCHMARK}-intermediate',
 
-    '--file=fuzzers/${_FUZZER}/builder.Dockerfile',
+    '--file=fuzzers/${_BASE_FUZZER}/builder.Dockerfile',
```
inferno-chromium, line 34: Fix this please, should be _UNDERLYING_FUZZER

inferno-chromium:
CI failures are spurious, did you verify with experiment run that it works, probably test with oss-fuzz too since this should be caught by build failure.

mbarbella-chromium:
Good catch, my last run was only using a non-oss-fuzz benchmark. I'll make that fix and run another with both.

jonathanmetzman:
Thanks for this!

PR #266, opened 2020-04-24
Move experiment/gcb/ to docker/gcb

jonathanmetzman:


lszekeres:
Why not into docker/gcb, where the rest of the docker build related stuff are?

lszekeres:
> Why not into docker/gcb, where the rest of the docker build related stuff are?

(I think eventually we could auto-generate these .yaml files, together with docker/build.mk, from the same dependency graph we could define in python)

jonathanmetzman:
> LGTM
> 
> (Not sure if GitHub can track file moves to keep their history, probably not)

I think it does, but I think I messed this up by forgetting to delete the initial files I moved to experiment/gcb.

PR #269, opened 2020-04-24
new afl++ variants

vanhauser-thc:
new temporary variants:
```
aflplusplus_instrim_coe
aflplusplus_instrim_mopt2
aflplusplus_ngram3
aflplusplus_ngram4_17
aflplusplus_ngram4_18
aflplusplus_ngram4_coe
aflplusplus_ngram5
```
if you could afterwards add them to the `2020-04-20-aflplusplus` report (and remove irsii) that would be amazing!

this tests 3 different things:
 * after ngram-4 showed to be the best instrumentation in  `2020-04-20-aflplusplus` (over - among other ngram-2), so lets test ngram-3 and ngram-5
 * after ngram-4 showed to be the best instrumentation in  `2020-04-20-aflplusplus`  maybe it improves with a larger map size (testing map sizes 2^17 and 2^18, default is 2^16)
 * the afl++ variant tests before 2020-04-20 showed in the after analysis that the COE power schedule is actually really good, so test this for the two best instrumentation variants instrim and ngram-4 as well.


vanhauser-thc:
@googlebot I consent.

jonathanmetzman:
The current experiment is still measuring, are you sure it makes sense to do a new one?
Also, please see presubmit failure.

vanhauser-thc:
> The current experiment is still measuring, are you sure it makes sense to do a new one?

start when you have the resources. it is not urgent. thanks!

> Also, please see presubmit failure.

I only find warnings concerning the code.
The only error I see in there is this:
```
ERROR: google-api-core 1.17.0 has requirement google-auth<2.0dev,>=1.14.0, but you'll have google-auth 1.11.0 which is incompatible.
```
and that is nothing I can influence :)

vanhauser-thc:
Also after carefully checking the results if the `2020-04-20-aflplusplus` run the CTX instrumentation is a bit better than ngram4 (when comparing the results in % of edges found vs the one that found most). so I added three CTX variations.

jonathanmetzman:
Sorry, it isn't so clear from the logs (because tests continue after the first fail) but you need to run `make format`

vanhauser-thc:
ah ok. It said "Warnings" and not Error so ...

"make format" fails compiling on my system :-( so I tried to fix it by hand, guessing what it want to have fixed. fingers crossed.

jonathanmetzman:
>ah ok. It said "Warnings" and not Error so ...
Good point, I'll make sure to add something that says "Error" so it can be found more easily.

Another issue was pointed out by linting. Can you add `# pylint: disable=too-many-branches,too-many-statements` to line 30? (i.e.`def build(*args):  # pylint: disable=too-many-branches,too-many-statements`)

vanhauser-thc:
done :)

vanhauser-thc:
of course it needed two spaces. obvious. hope its now fine.

vanhauser-thc:
dont know where else to put my question - `2020-04-20-aflplusplus` is still not finished, and also `2020-04-21` which does not contain the irssi test is not finished either ... is there a bug?

jonathanmetzman:
> dont know where else to put my question - `2020-04-20-aflplusplus` is still not finished, and also `2020-04-21` which does not contain the irssi test is not finished either ... is there a bug?

2020-04-21 seems to still be going but looks close to finishing up. It doesn't look like there's bug other than being atrociously slow. Of the 604800 snapshots that need to be measured (6300 trials * 24*60/15) 604314 have been measured already (we probably wont even get to 604800 because it's basically inevitable that some trials out of 6300 will die early when fuzzing because of OOMS rare bugs etc).

There was a bug that I fixed in 2020-04-20-aflplusplus but the fix will prevent it from finishing. Basically, the disk was filling up so it couldn't rsync the corpora for a day or two (which also has delayed results). I was able to resolve this by hiding the woff corpora on GCS since they took up almost half the space for some reason.

vanhauser-thc:
@jonathanmetzman thanks for the clarification :)

can this PR be merged or are there showstoppers?

PR #270, opened 2020-04-24
Add some small optimizations to the measurer

jonathanmetzman:
1. Keep the names of measured files in a list on disk instead of keeping the actual files on disk. This should modestly reduce disk space usage.
2. Copy unchanged-cycles to disk and only copy it again when it may contain new information (e.g. don't copy it again if the current cycle is in the file already or if some other greater cycle is in the file already).

Also, log the amount of time it takes to measure each snapshot.

### DIFF
--- a/experiment/measurer.py
+++ b/experiment/measurer.py
@@ -317,14 +317,6 @@ class SnapshotMeasurer:  # pylint: disable=too-many-instance-attributes
                                        self.benchmark_fuzzer_trial_dir)
         self.corpus_dir = os.path.join(measurement_dir, 'corpus')
 
-        # Keep a directory containing all the corpus units we've already seen.
-        # This is an easy to implement way of storing this info such that
-        # the measurer can restart and continue where it left off.
-        # A better solution could involve using a file to store this info
-        # instead. Another problem with it is it assumes the measurer is running
-        # on one machine.
-        self.prev_corpus_dir = os.path.join(measurement_dir, 'prev-corpus')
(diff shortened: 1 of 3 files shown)


experiment/measurer.py
```diff
@@ -335,18 +327,21 @@ def __init__(self, fuzzer: str, benchmark: str, trial_num: int,
...
+                                                  'unchanged-cycles')
 
     def initialize_measurement_dirs(self):
         """Initialize directories that will be needed for measuring
         coverage."""
         for directory in [self.corpus_dir, self.sancov_dir, self.crashes_dir]:
             filesystem.recreate_directory(directory)
-        for directory in [self.report_dir, self.prev_corpus_dir]:
-            pathlib.Path(directory).mkdir(exist_ok=True)
+        pathlib.Path(self.report_dir).mkdir(exist_ok=True)
```
lszekeres, line 344: consider using common.filesystem.create_directory()

jonathanmetzman:
@inferno-chromium I currently log how long it takes to measure each snapshot.
Do you think we should use the metrics library like we use in ClusterFuzz instead so we can get something like a bar graph?

inferno-chromium:
> @inferno-chromium I currently log how long it takes to measure each snapshot.
> Do you think we should use the metrics library like we use in ClusterFuzz instead so we can get something like a bar graph?

Depends on if you want to do this one time optimization or want to invest in long term monitoring.

PR #277, opened 2020-04-27
[generate_report] Add --end-time arg to limit reports to time periods

### LINKED ISSUE #276: Important: Add time feature in report generation
This will make it easy to generate report at 1h,6h,12h,24h

jonathanmetzman:
Add end_time argument so that one can specify a time limit on snapshots
they want in the report. For example, this allows one to get a report
from the first 6 hours of an experiment.
Fixes #276

### DIFF
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -62,6 +62,12 @@ def label_fuzzers_by_experiment(experiment_df):
     return experiment_df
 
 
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
Demo: http://commondatastorage.googleapis.com/fuzzbench-reports/2020-04-14-end-time-6-hours/index.html

PR #278, opened 2020-04-27
Add src_analysis module

jonathanmetzman:
Add src_analysis module for doing analysis on the fuzzbench source
code. This will be useful for determining what should be tested in
CI and which fuzzers and benchmarks should be run in the next
experiment.


pytest.ini
```diff
@@ -1,2 +1,2 @@
 [pytest]
-norecursedirs=third_party/*
\ No newline at end of file
+norecursedirs = third_party/* .venv/* docs/vendor/* docs/_site/*
```
jonathanmetzman, line 2: This change is unrelated to the main one but speeds up running tests by 3-8 seconds.

src_analysis/fuzzer_dependencies.py
```diff
@@ -0,0 +1,142 @@
...
+from typing import List
+import sys
+
+from common import fuzzer_utils
+
+MAX_DEPTH = 5
+PY_DEPS_CACHE = {}
+
+
+def _get_fuzzer_module(fuzzer: str) -> str:
```
jonathanmetzman, line 30: Should this be part of the FuzzerDirectory class?

src_analysis/fuzzer_dependencies.py
```diff
@@ -0,0 +1,142 @@
...
+    """Returns the module fuzzer.py module of |fuzzer|."""
+    return 'fuzzers.{}.fuzzer'.format(fuzzer)
+
+
+def is_builtin_module(module) -> bool:
+    """Returns the module for |fuzzer|'s fuzzer.py."""
+    return module.__name__ in sys.builtin_module_names
```
jonathanmetzman, line 37: Maybe this is too fragile since it will incorrectly return true if I have a module blah.time?

.github/workflows/fuzzers.yml
```diff
@@ -46,9 +46,13 @@ jobs:
     env:
       FUZZER: ${{ matrix.fuzzer }}
       BENCHMARK_TYPE: ${{ matrix.benchmark_type }}
+      ALWAYS_BUILD_FUZZER: afl
```
jonathanmetzman, line 49: One problem with using src_analyzer to strictly limit what gets built in CI is testing changes to the build system as a whole rather than individual fuzzers or benchmarks.
For example, what should get built if we change docker/build.mk. In theory this affects the build of every fuzzer.
There are two approaches I have taken in this PR, we could drop either one or even combine them.
The first approach (the one I take in this file) is to always build one fuzzer. It's hard for me to imagine a change to build.mk breaking some fuzzers and not others, so maybe building one is sufficient.
This approach saves builds when we do change these files but is otherwise wasteful since we are always building AFL. However, I think the approach might be most likely to catch changes that src_analysis misses and so in some sense it is safest.
The second approach, which I take in change_utils is to build *every* fuzzer when one of these files changes. 
The third approach, which I haven't implemented is to combine the two. Only build one fuzzer when one of these files changes. This is the least conservative approach but saves the most amount of builds.

Any ideas for which I take?

src_analysis/benchmark_dependencies.py
```diff
@@ -0,0 +1,46 @@
...
+    dependent_benchmarks = []
+    benchmarks = benchmark_utils.get_all_benchmarks()
+    for dependency_file in dependency_files:
+        for benchmark in benchmarks:
+            if not is_subpath_of_benchmark(dependency_file, benchmark):
+                continue
+            if not benchmark_utils.is_oss_fuzz(benchmark):
+                dependent_benchmarks.append(benchmark)
+                break
+            # OSS-Fuzz benchmarks only have an oss-fuzz.yaml file as a
```
jonathanmetzman, line 40: I optimize things here by saying that OSS-Fuzz benchmarks can only have oss-fuzz.yaml as a dependency so if there were some other file in the directory like a README it wouldn't be a dependency.
Maybe this should be loose and say if it is in the directory, it is a dependency regardless of OSS-Fuzz or not.
What do people think?

src_analysis/change_utils.py
```diff
@@ -0,0 +1,67 @@
+# Copyright 2020 Google LLC
```
jonathanmetzman, line 1: The purpose of this module was to be aware of things like gcb/* and docker/* so that fuzzer_dependencies and benchmark_dependencies could ignore them. It is unnecessary if we go with the always_build_fuzzer approach. Should I remove and use those modules directly?

src_analysis/benchmark_dependencies.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2020 Google LLC
```
jonathanmetzman, line 1: In my mind it makes sense to call this directory src_analysis instead of src_analyzer because it isn't an independent module used in experiments like "runner" or "measurer". What do people think?

src_analysis/fuzzer_dependencies.py
```diff
@@ -0,0 +1,185 @@
...
+    fuzzer_directory = fuzzer_utils.FuzzerDirectory(fuzzer)
+
+    if initial_fuzzer == fuzzer:
+        dependencies = set()
+    else:
+        # If fuzzer's base fuzzer is different, fuzzer is a variant, which
+        # means changes to variants.yaml can affect it.
+        dependencies = {fuzzer_directory.variants_yaml}
+
+    # Don't use modulefinder for python dependencies since it doesn't work
```
jonathanmetzman, line 107: Modulefinder did not work here because the modules do not contain __init__.py files. We have three ways of dealing with this:
1. The route I took, use something other than modulefinder. It probably doesn't handle edge cases as well (such as importing functions, rather than modules, which the style guide doesn't allow anyway).
2. Forcing every module in fuzzers to have an __init__.py.
3. Hacking modulefinder to work anyway. I did this in a few lines of code by using fakefs and adding a fake __init__.py. But this seems too. hacky

jonathanmetzman: I personally like 1, especially if we can get it to handle `import fuzzers.afl.fuzzer` since that is really the only use case that isn't against the style guide that actually works but src_analysis module doesn't support (I don't think it's that hard to special case these kinds of imports). But it might not be worth the complexity so 2 is also a reasonable option (at the cost of minor integration inconvenience).

jonathanmetzman:
I think this code is ready for a first pass. PTAL.
There are probably better ways to implement the python dependency analysis (as I note in the many questions I've left on this review, please see those as well), but I think this works well for the way fuzzbench is currently used. It may be worth landing and switching to better techniques later.

pytest.ini
inferno-chromium, line 2: nit: alpha order.

presubmit.py
```diff
@@ -337,7 +304,7 @@ def main() -> int:
 
     args = parser.parse_args()
     os.chdir(_SRC_ROOT)
-    changed_files = get_changed_files()
+    changed_files = [Path(path) for path in diff_utils.get_changed_files()]
```
inferno-chromium, line 307: Path(path).absolute() ?

inferno-chromium: Do we need Path(), if we are later converting to str(path).

common/fuzzer_utils.py
```diff
@@ -77,6 +120,17 @@ def validate(fuzzer):
         return False
 
 
+def get_fuzzer_from_config(fuzzer_config: dict) -> str:
+    """Returns the fuzzer of |fuzzer_config| for a non-variant fuzzer or returns
+    the name for a fuzzer variant."""
+    return fuzzer_config.get('name', fuzzer_config['fuzzer'])
```
inferno-chromium, line 126: Do we enforce fuzzer attribute always exist ?

common/filesystem.py
```diff
@@ -110,3 +102,14 @@ def make_dir_copy(src_dir):
...
     return dst_dir
+
+
+def list_files(directory):
+    """Returns a list of absolute paths to all files in |directory| and its
+    subdirectories."""
+    dir_absolute_path = os.path.abspath(directory)
+    search_path = os.path.join(dir_absolute_path, '**', '*')
+    return [
+        path for path in glob.iglob(search_path, recursive=True)
```
inferno-chromium, line 113: Why glob.iglob and not just os.walk ?

.github/workflows/test_fuzzer_benchmarks.py
```diff
@@ -100,17 +105,30 @@ def do_build(build_type, fuzzer):
...
+    benchmarks = benchmarks.intersection(changed_benchmarks)
     return make_builds(benchmarks, fuzzer)
 
 
 def main():
     """Build OSS-Fuzz or standard benchmarks with a fuzzer."""
-    if len(sys.argv) != 3:
-        print('Usage: %s <build_type> <fuzzer>' % sys.argv[0])
+    if len(sys.argv) != 4:
+        print('Usage: %s <build_type> <fuzzer> <always_build_fuzzer>' %
```
inferno-chromium, line 124: Lets just put always_build_fuzzer as a global in this file, it looks weird to put in this file argument. also, someone can manually run this script as well.

.github/workflows/test_fuzzer_benchmarks.py
```diff
@@ -100,17 +105,30 @@ def do_build(build_type, fuzzer):
...
         raise Exception('Invalid build_type: %s' % build_type)
 
+    if always_build:
+        return make_builds(benchmarks, fuzzer)
+
+    changed_files = diff_utils.get_changed_files()
+    changed_fuzzers = change_utils.get_changed_fuzzers(changed_files)
+    if fuzzer in changed_fuzzers:
+        return make_builds(benchmarks, fuzzer)
+
```
inferno-chromium, line 115: Maybe add comments for the three branches.

# Always build with all benchmarks if always_build is set.
# Build changed fuzzer with all benchmarks.
# Build fuzzer with changed benchmarks, skip rest.

src_analysis/benchmark_dependencies.py
inferno-chromium: this optimization is not needed and more confusing, can remove and simplify.

src_analysis/test_benchmark_dependencies.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def test_is_subpath_of_benchmark():
+    """Tests that is_subpath_of_benchmark returns True for subpaths of a
+    benchmark and returns False for other paths."""
+    assert benchmark_dependencies.is_subpath_of_benchmark(
+        OSS_FUZZ_YAML_PATH, OSS_FUZZ_BENCHMARK)
+    assert not benchmark_dependencies.is_subpath_of_benchmark(
+        STANDARD_BUILD_SH_PATH, OSS_FUZZ_BENCHMARK)
+
```
inferno-chromium, line 35: Can you test this with STANDARD_BENCHMARK

src_analysis/test_benchmark_dependencies.py
```diff
@@ -0,0 +1,49 @@
...
+
+def test_get_files_dependent_benchmarks():
+    """Tests that get_files_dependent_benchmarks returns the benchmarks that are
+    dependent on the files passed to it."""
+    fake_build_sh_path = os.path.join(benchmark_utils.BENCHMARKS_DIR, 'fake',
+                                      'build.sh')
+    changed_files = [
+        STANDARD_BUILD_SH_PATH, OSS_FUZZ_YAML_PATH, fake_build_sh_path
```
inferno-chromium, line 43: Shouldnt fake_build_sh_path be tested on its own without STANDARD_BUILD_SH_PATH

src_analysis/change_utils.py
```diff
@@ -0,0 +1,67 @@
...
+    return changed_fuzzers
+
+
+def get_changed_fuzzers_for_ci(changed_files: List[str] = None) -> List[str]:
+    """Returns a list of fuzzers that have changed functionality based
+    on the files that have changed in |changed_files|.
+    Unlike get_changed_fuzzers this function considers changes that affect
+    building or running fuzzers in CI."""
+    changed_files = get_absolute_paths(changed_files)
+    print(CI_FILES, changed_files)
```
inferno-chromium, line 55: remove print

src_analysis/change_utils.py
```diff
@@ -0,0 +1,67 @@
...
+    if set(changed_files).intersection(CI_FILES):
```
inferno-chromium, line 56: Add comment.

src_analysis/change_utils.py
```diff
@@ -0,0 +1,67 @@
...
+        return fuzzer_utils.get_fuzzer_names()
+    return get_changed_fuzzers(changed_files)
+
+
+def get_changed_benchmarks(changed_files: List[str] = None) -> List[str]:
+    """Returns a list of benchmarks that have changed functionality based
+    on the files that have changed in |changed_files|."""
+    changed_files = get_absolute_paths(changed_files)
+    changed_benchmarks = benchmark_dependencies.get_files_dependent_benchmarks(
```
inferno-chromium, line 65: Why not put get_absolute_paths in get_files_dependent_benchmarks

src_analysis/fuzzer_dependencies.py
```diff
@@ -0,0 +1,182 @@
...
+
+# A cache of Python dependencies for modules. Keys are strings of module paths.
+# Values are sets of module paths.
+PY_DEPENDENCIES_CACHE: Dict[str, Set[str]] = {}
+
+# Cache these so we don't need to do it every time we call
+# get_underlying_fuzzer.
+FUZZER_CONFIGS = fuzzer_utils.get_fuzzer_configs()
+FUZZER_NAMES_TO_UNDERLYING = {
+    fuzzer_utils.get_fuzzer_from_config(config): config['fuzzer']
```
inferno-chromium, line 51: wasnt there a helper function for this part, instead of looking for fuzzer attribute.

src_analysis/fuzzer_dependencies.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def is_builtin_module(module: types.ModuleType) -> bool:
+    """Returns True if |module| is a python builtin module."""
+    return module.__name__ in sys.builtin_module_names
+
+
+def is_fuzzers_subpath(path: str) -> bool:
+    """Returns True if path is a subpath of the fuzzers/ directory."""
+    common_path = os.path.commonpath([path, fuzzer_utils.FUZZERS_DIR])
```
inferno-chromium, line 69: os.path.commonpath use to detect if something is part of that directory, can you make a helper function in common/filesystem or something.

src_analysis/fuzzer_dependencies.py
```diff
@@ -0,0 +1,182 @@
...
+    """Returns the python files that |module| is dependent on if module is a
+    submodule of fuzzers/. Does not return dependencies that are not submodules
+    of fuzzers/, such as std library modules. Has a limit of
+    PY_DEPENDENCIES_MAX_DEPTH so that cyclic imports can easily be detected.
+    Note that this may not work if a fuzzer.py imports modules dynamically or
+    within individual functions. That is ok because we can prevent this during
+    code review."""
+    if depth > PY_DEPENDENCIES_MAX_DEPTH:
+        # Enforce a depth to catch cyclic imports.
+        format_string = ('Depth: {} greater than max: {}. '
```
inferno-chromium, line 125: named args.

src_analysis/fuzzer_dependencies.py
```diff
@@ -0,0 +1,182 @@
...
+
+        # Files in fuzzers/ can only import code from the Python standard
+        # library modules and modules in fuzzers/.
+        if is_fuzzers_submodule(imported_module):
+            imported_module_path = inspect.getfile(imported_module)
+            dependencies.add(imported_module_path)
+            # Now recur to get the dependencies of the dependency.
+            dependencies = dependencies.union(
+                _get_python_dependencies(imported_module, depth + 1))
+
```
inferno-chromium, line 159: global PY_DEPENDENCIES_CACHE ?

src_analysis/fuzzer_dependencies.py
```diff
@@ -0,0 +1,182 @@
...
+    PY_DEPENDENCIES_CACHE[module_path] = dependencies
+    return dependencies
+
+
+def get_underlying_fuzzer(fuzzer_name: str) -> str:
```
inferno-chromium, line 164: i thought there was a helper for this in fuzzer_utils ?

presubmit.py
jonathanmetzman: All of this file uses pathlib.Path. The rest of the code generally does not. I can refactor this file but I don't think it is worth it for now, the PR is pretty large as is.

src_analysis/fuzzer_dependencies.py
jonathanmetzman: I'm not sure that there is now. What function do you think does this?

src_analysis/fuzzer_dependencies.py
jonathanmetzman: I don't think this is needed since we are mutating the `PY_DEPENDENCIES_CACHE`. If we were reassigning where we would need to do this.

src_analysis/fuzzer_dependencies.py
jonathanmetzman: Done

src_analysis/fuzzer_dependencies.py
jonathanmetzman: Done

src_analysis/fuzzer_dependencies.py
jonathanmetzman: Sorry what do you mean by attribute? Not sure I understand.
But if youre wondering why FUZZER_NAMES_TO_UNDERLYING is needed when fuzzer_utils.get_fuzzer_from_config exists it's sort of a cache, but it's one for a really fast operation so I guess we can get rid of it if you think it makes things simpler.

src_analysis/change_utils.py
jonathanmetzman: I think it's more consistent to do it here. I need absolute paths `get_changed_fuzzers_for_ci` when I check if anything is in `CI_FILES`. It's a bit defensive since this module is sort of the main external interface for src_analysis. But I think it isn't actually needed because changed_files is returned by diff_utis.get_changed_files which always uses absolute paths.

src_analysis/change_utils.py
jonathanmetzman: I made this clearer by changing it to `any(changed_file in CI_FILES for changed_file in changed_files)` instead of using sets.

src_analysis/change_utils.py
jonathanmetzman: Done

src_analysis/test_benchmark_dependencies.py
jonathanmetzman: Done

src_analysis/test_benchmark_dependencies.py
jonathanmetzman: It's the same code path I don't really think it adds anything.
The first assertion tests that a subpath is identified as such. The second assert tests that a non-subpath is identified as such

.github/workflows/test_fuzzer_benchmarks.py
jonathanmetzman: Done

.github/workflows/test_fuzzer_benchmarks.py
jonathanmetzman: Done. but if someone were running manually, I think the global is also weird since script will inexplicably behave differently for AFL.

common/filesystem.py
jonathanmetzman: Done

common/fuzzer_utils.py
jonathanmetzman: They should always exist (unlike name).
I think `fuzzer_config.get('fuzzer')` would be bad. I think it's better to exception.
Sorry if this doesn't answer the question, I'm not sure I fully understood.

pytest.ini
jonathanmetzman: Done

jonathanmetzman:
Thanks for the review.
Any thoughts on whether we should have an always build fuzzer or CI_FILES https://github.com/google/fuzzbench/pull/278#discussion_r420297136 ?

And also, any thoughts on whether we should force use of __init__.py so modulefinder can be used https://github.com/google/fuzzbench/pull/278#discussion_r420396452 ?

.github/workflows/fuzzers.yml
inferno-chromium: If dont take second approach, we will silent failures that will show up later when running experiment. We have to do that, otherwise more manual work and followup stuff. it is resource intensive, but what can we do else.

src_analysis/fuzzer_dependencies.py
inferno-chromium: 1 sounds good, no need to force people to add __init__.py, feels crappy.

inferno-chromium:
> Thanks for the review.
> Any thoughts on whether we should have an always build fuzzer or CI_FILES [#278 (comment)](https://github.com/google/fuzzbench/pull/278#discussion_r420297136) ?
> 
> And also, any thoughts on whether we should force use of **init**.py so modulefinder can be used [#278 (comment)](https://github.com/google/fuzzbench/pull/278#discussion_r420396452) ?

responded on the 2 comments.

.github/workflows/fuzzers.yml
jonathanmetzman: I see what you mean but I think 2 is inherently fragile as we must remember to include every file that affects builds in `CI_FILES`. I think the current implementation which always builds AFL but builds everything when CI_FILES is changed is the best approach. Maybe I'll eliminate the AFL part if I get more confident in the CI_FILES approach.

src_analysis/fuzzer_dependencies.py
jonathanmetzman: ACK

PR #282, opened 2020-04-28
Add --merge-with-clobber argument to generate_report.py

jonathanmetzman:
This argument allows you to combine reports that have updated versions
of the same trials. For example, AFL++ from one experiment and then
an updated AFL++ on a later experiment on the same benchmarks.
It combines the reports by clobbering the stale data.

This means that if we do an experiment and then update a fuzzer,
comparing the fuzzer to others from the old experiment can be done
without running all of them in the new experiment, only the changed
fuzzers needs to run.

This can save resources and speed up measuring by doing less unnecessary
trials. Determining what to run in an experiment still must be done manually.

### DIFF
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -69,8 +69,8 @@ def clobber_snapshots(df, experiments):
         covered_pairs = result[['benchmark', 'fuzzer']].drop_duplicates()
         covered_pairs = covered_pairs.apply(tuple, axis=1)
         experiment_data = df[df.experiment == experiment]
-        experiment_pairs = experiment_data[['benchmark', 'fuzzer']].apply(
-            tuple, axis=1)
+        experiment_pairs = experiment_data[['benchmark',
+                                            'fuzzer']].apply(tuple, axis=1)
         to_include = experiment_data[~experiment_pairs.isin(covered_pairs)]
         result = result.append(to_include)
     return result
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -37,7 +37,7 @@ def drop_uninteresting_columns(experiment_df):
     ]]
 
 
-def clobber_snapshots(df, experiments):
+def clobber_experiments_data(df, experiments):
     """Clobber experiment data that is part of lower priority (generally
     earlier) versions of the same trials in |df|. For example in experiment-1 we
     may test fuzzer-a on benchmark-1. In experiment-2 we may again test fuzzer-a
@@ -47,9 +47,7 @@ def clobber_snapshots(df, experiments):
     determined by order of each experiment in |experiments| with the highest
     priority experiment coming last in that list."""
     # We don't call |df| "experiment_df" because it is a misnomer and leads to
-    # confusion in this case where it contains data from multiple experiments
-    # and we actually build DataFrames containing data from individual
-    # experiments.
+    # confusion in this case where it contains data from multiple experiments.
 
     # Include everything from the last experiment.
     experiments = experiments.copy()  # Copy so we dont mutate experiments.
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -43,59 +43,37 @@ FuzzerBenchmark = collections.namedtuple('FuzzerBenchmark',
                                          ['fuzzer', 'benchmark', 'experiment'])
 
 
-def clobber_snapshots(experiment_df, experiments):
+def clobber_snapshots(df, experiments):
     """Clobber snapshots that are part of lower priority (generally earlier)
-    versions of the same trials in |experiment_df|. For example in experiment-1
+    versions of the same trials in |df|. For example in experiment-1
     we may test fuzzer-a on benchmark-1. In experiment-2 we may again test
     fuzzer-a on benchmark-1 because fuzzer-a was updated. This function will
     remove the snapshots from fuzzer-a,benchmark-1,experiment-1 from
-    |experiment_df| because we want the report to only contain the up-to-date
+    |df| because we want the report to only contain the up-to-date
     data. Experiment priority is determined by order of each experiment in
     |experiments| with the highest priority experiment coming last in that
     list."""
-    # Create a set of tuples that contain all the unique
-    # fuzzer-benchmark-experiment combinations in experiment_df. We'll use this
-    # to determine which snapshots to drop.
-    fuzzer_benchmarks_df = experiment_df[['fuzzer', 'benchmark', 'experiment']]
-    fuzzer_benchmarks = {
-        FuzzerBenchmark(*fuzzer_benchmark)
-        for fuzzer_benchmark in fuzzer_benchmarks_df.to_numpy()
-    }
-    experiment_rankings = {
-        experiment: experiments.index(experiment) for experiment in experiments
-    }
+    # We don't call |df| "experiment_df" because it is a misnomer and leads to
+    # confusion in this case where it contains data from multiple experiments
+    # and we actually build DataFrames containing data from individual
+    # experiments.
 
-    highest_rank = max(experiment_rankings.values())
+    # Include everything from the last experiment.
+    experiments = experiments.copy()  # Copy so we dont mutate experiments.
+    experiments.reverse()
+    highest_rank_experiment = experiments[0]
+    result = df[df.experiment == highest_rank_experiment]
 
-    # Get the set of those tuples that we should drop because they are
-    # lower priority/rank than others in experiment_df.
-    fuzzer_benchmarks_to_drop = set()
-    for fuzzer_benchmark in fuzzer_benchmarks:
-        experiment = fuzzer_benchmark.experiment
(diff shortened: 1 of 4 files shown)


jonathanmetzman:
Looks like there is a big backlog of GitHub actions that are being queued.
https://github.com/google/fuzzbench/pull/278 will help prevent this when landed.
It can also help with making the process of running updated fuzzers in a new experiment, combining with old experiment and clobbering automated.

analysis/data_utils.py
```diff
@@ -42,6 +39,65 @@ def drop_uninteresting_columns(experiment_df):
...
+    """Clobber snapshots that are part of lower priority (generally earlier)
+    versions of the same trials in |experiment_df|. For example in experiment-1
+    we may test fuzzer-a on benchmark-1. In experiment-2 we may again test
+    fuzzer-a on benchmark-1 because fuzzer-a was updated. This function will
+    remove the snapshots from fuzzer-a,benchmark-1,experiment-1 from
+    |experiment_df| because we want the report to only contain the up-to-date
+    data. Experiment priority is determined by order of each experiment in
+    |experiments| with the highest priority experiment coming last in that
+    list."""
+    # Create a set of tuples that contain all the unique
```
lszekeres, line 56: Would it be simpler to build up the clobbered table, eg like this?

```python
    # Include everything from the last experiment.
    experiments.reverse()
    last_exp_name = experiments[0]
    result = experiment_df[experiment_df.experiment == last_exp_name]
    for exp_name in experiments[1:]:
        # Include data for not yet covered benchmark/fuzzer pairs.
        covered_pairs = result[['benchmark', 'fuzzer']].drop_duplicates()
        covered_pairs = covered_pairs.apply(tuple, axis=1).tolist()
        exp = experiment_df[experiment_df.experiment == exp_name]
        exp_pairs = exp[['benchmark', 'fuzzer']].apply(tuple, axis=1)
        to_include = exp[~exp_pairs.isin(covered_pairs)]
        result = result.append(to_include)
    return result

```

jonathanmetzman: Yeah much simpler actually. Thanks!

jonathanmetzman: Any opinion on whether I should reverse the order of the `result` so it more resembles the initial order?

analysis/data_utils.py
```diff
@@ -42,6 +39,43 @@ def drop_uninteresting_columns(experiment_df):
     ]]
 
 
+FuzzerBenchmark = collections.namedtuple('FuzzerBenchmark',
```
lszekeres, line 42: Remove?

analysis/data_utils.py
```diff
@@ -12,10 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Utility functions for data (frame) transformations."""
-
-# score.rename('stat wins', inplace=True) in rank_by_stat_test_wins breaks
-# pylint and causes it to stack overflow. Don't lint this file because of it.
-# pylint: skip-file
+import collections
```
lszekeres, line 15: Remove?

analysis/data_utils.py
```diff
@@ -42,6 +39,43 @@ def drop_uninteresting_columns(experiment_df):
     ]]
 
 
+FuzzerBenchmark = collections.namedtuple('FuzzerBenchmark',
+                                         ['fuzzer', 'benchmark', 'experiment'])
+
+
+def clobber_snapshots(df, experiments):
```
lszekeres, line 46: We should call this something else than "snapshots", because in the rest of the code by "snapshot" we refer to a "data frame where all time values are the same": https://github.com/google/fuzzbench/blob/master/analysis/README.md#snapshot.

How about calling it `clobber_experiments_data`, or something like that?

analysis/data_utils.py
```diff
@@ -42,6 +39,43 @@ def drop_uninteresting_columns(experiment_df):
...
+    """Clobber snapshots that are part of lower priority (generally earlier)
```
lszekeres, line 47: Consider: Clobber experiment data that is part of..

analysis/test_data_utils.py
```diff
@@ -69,6 +70,42 @@ def test_drop_uniteresting_columns():
...
+        experiment_df = create_experiment_data(experiment)
+        df = pd.concat([df, experiment_df])
+    not_updated_benchmark = 'libpng'
+    not_updated_fuzzer = 'afl'
+    drop_condition = ((df.experiment != experiments[2]) |
+                      (df.benchmark != not_updated_benchmark) |
+                      (df.fuzzer != not_updated_fuzzer))
+    df = df[drop_condition]
+    df = data_utils.clobber_snapshots(df, experiments)
+    # All of the first experment should be clobbered
```
lszekeres, line 90: This test feels like a bit convoluted way of checking the results. Consider just checking if df[['experiemnt', 'benchmark', 'fuzzer']].drop_duplicates() is the expected 4 triples, eg.:
```
'experiment1', 'libxml', 'afl'
'experiment2', 'libxml', 'libfuzzer'
'experiment2', ...
'experiment2', ...
```

analysis/data_utils.py
jonathanmetzman: I changed it to clobber_experiments_data to be consistent with the graphing code.
But as I understand it, the doc's explanation of the term 'snapshot' is inconsistent with the rest of the code base where snapshot refers to a particular time and a particular trial not a particular time across all trials. See the snapshot model for example: https://github.com/google/fuzzbench/blob/master/database/models.py#L49

lszekeres: Yes, https://github.com/google/fuzzbench/blob/master/analysis/README.md is about the three types of DataFrames we use under analysis/. But yea, naming is hard :) we can think about better names

analysis/test_data_utils.py
```diff
@@ -35,17 +36,17 @@ def create_trial_data(trial_id, benchmark, fuzzer, reached_coverage):
...
     return pd.concat([
-        create_trial_data(0, 'libpng', 'afl', 100),
-        create_trial_data(1, 'libpng', 'afl', 200),
-        create_trial_data(2, 'libpng', 'libfuzzer', 200),
-        create_trial_data(3, 'libpng', 'libfuzzer', 300),
-        create_trial_data(4, 'libxml', 'afl', 1000),
-        create_trial_data(5, 'libxml', 'afl', 1200),
-        create_trial_data(6, 'libxml', 'libfuzzer', 600),
-        create_trial_data(7, 'libxml', 'libfuzzer', 800),
+        create_trial_data(0, 'libpng', 'afl', 100, experiment=experiment),
```
lszekeres, line 42: nit: the `experiments=` is not needed

analysis/test_data_utils.py
```diff
@@ -69,6 +70,37 @@ def test_drop_uniteresting_columns():
...
+def test_experiments_data():
+    """Tests that clobber experiments data clobbers stale snapshots from earlier
+    experiments."""
+    experiments = []
+    df = pd.DataFrame()
+    for experiment_num in range(3):
+        experiment = 'experiment-%d' % experiment_num
+        experiments.append(experiment)
+        experiment_df = create_experiment_data(experiment)
+        df = pd.concat([df, experiment_df])
```
lszekeres, line 82: nit: can this be simplified with
`  df = pd.concat([create_experiment_data('experiment-' + n) for n in range(3)])
`?

analysis/test_data_utils.py
```diff
@@ -69,6 +70,37 @@ def test_drop_uniteresting_columns():
...
+    df = data_utils.clobber_experiments_data(df, experiments)
+    columns = ['experiment', 'benchmark', 'fuzzer']
+    expected_result = pd.DataFrame([
+        ['experiment-2', 'libpng', 'libfuzzer'],
+        ['experiment-2', 'libxml', 'afl'],
+        ['experiment-2', 'libxml', 'libfuzzer'],
+        ['experiment-1', 'libpng', 'afl'],
+    ],
+                                   columns=columns)
+    print(type(expected_result), type(df[columns].drop_duplicates()))
```
lszekeres, line 98: stale debug code?

analysis/test_data_utils.py
```diff
@@ -69,6 +70,37 @@ def test_drop_uniteresting_columns():
...
+    df = pd.DataFrame()
+    for experiment_num in range(3):
+        experiment = 'experiment-%d' % experiment_num
+        experiments.append(experiment)
+        experiment_df = create_experiment_data(experiment)
+        df = pd.concat([df, experiment_df])
+    not_updated_benchmark = 'libpng'
+    not_updated_fuzzer = 'afl'
+    drop_condition = ((df.experiment != experiments[2]) |
+                      (df.benchmark != not_updated_benchmark) |
```
lszekeres, line 86: can this be simplified with
```
to_drop = df[df.experiment=='experiment-2' & df.benchmark=='libpng' & df.fuzzer='afl'].index
df.drop(to_drop, inplace=True)

```
?

PR #284, opened 2020-04-29
Be more explicit about presubmit errors

jonathanmetzman:
Explicitly say that there are errors so that errors are easier to spot when
viewing CI output.

### DIFF
--- a/experiment/test_runner.py
+++ b/experiment/test_runner.py
@@ -57,7 +57,6 @@ BENCHMARK = 'benchmark-1'
 EXPERIMENT = 'experiment-name'
 TRIAL_NUM = 1
 FUZZER = 'fuzzer-name-a'
-FULL_FUZZER_NAME = FUZZER
 
 
 @pytest.yield_fixture
@@ -65,7 +64,6 @@ def trial_runner(fs, environ):
     """Fixture that creates a TrialRunner object."""
     os.environ.update({
         'BENCHMARK': BENCHMARK,
-        'FUZZER_VARIANT_NAME': FULL_FUZZER_NAME,
         'EXPERIMENT': EXPERIMENT,
         'TRIAL_ID': str(TRIAL_NUM),
         'FUZZER': FUZZER,
@@ -240,7 +238,6 @@ class TestIntegrationRunner:
         os.environ['OUTPUT_CORPUS_DIR'] = str(output_corpus_dir)
 
         fuzzer = 'libfuzzer'
-        fuzzer_variant = fuzzer + '_variant'
         fuzzer_parent_path = root_dir / 'fuzzers' / fuzzer
 
         benchmark = 'MultipleConstraintsOnSmallInputTest'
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
I don't think it's practical to put this error message before the failing check's output (I don't think it's a good idea to wait for the tool to return to print its output, things like tests can hang, realtime output is necessary).

jonathanmetzman:
I also upgrade google-auth to 1.14 to silence this error: `ERROR: google-api-core 1.17.0 has requirement google-auth<2.0dev,>=1.14.0, but you'll have google-auth 1.11.0 which is incompatible`
This might be a bit of a risky change given the bad experiences we've had with the auth library. @inferno-chromium Do you think I should test in small experiment first? (the errors we saw probably only happen in a large experiment).

presubmit.py
```diff
@@ -338,24 +339,21 @@ def main() -> int:
...
-        success = pytype(changed_files)
-        return bool_to_returncode(success)
-
-    if args.command == 'licensecheck':
-        success = license_check(changed_files)
-        return bool_to_returncode(success)
-
-    if args.command == 'test_changed_integrations':
-        success = test_changed_integrations(changed_files)
+    command_check_mapping = {
```
inferno-chromium, line 342: Does this code work if we just run
make presubmit

Shouldn't this block be in "if args.command"

presubmit.py
```diff
@@ -338,24 +339,21 @@ def main() -> int:
...
+        'format': yapf,
+        'lint': lint,
+        'typecheck': pytype,
+        'test_changed_integrations': test_changed_integrations
+    }
+    for command, check in command_check_mapping.items():
```
inferno-chromium, line 348: Why iterate and continue, why not

command_check_mapping.get(args.command) and just run on it.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 alembic==1.4.0
-google-auth==1.11.0
+google-auth==1.14.0
```
inferno-chromium, line 2: This is ok, worst case we revert. Our integration test should atleast test this.

laurentsimon:
lgtm.

requirements.txt
jonathanmetzman: Not sure it would catch, i don't think our tests cover any code that use this library.

jonathanmetzman: Actually test_integration_runner should cover this code (but it doesn't run in CI (yet)). Running it locally.

presubmit.py
jonathanmetzman: Done.

presubmit.py
jonathanmetzman: It does work, but I've changed it to avoid unnecessary looping.

requirements.txt
jonathanmetzman: Now that I think about it won't touch the auth code because even though it talks to gcs it uses gsutil. The test pass after I fixed some unrelated issues related to variants.

experiment/test_runner.py
```diff
@@ -57,15 +57,13 @@ def test_run_fuzzer_log_file(mocked_communicate, fs):
 EXPERIMENT = 'experiment-name'
 TRIAL_NUM = 1
 FUZZER = 'fuzzer-name-a'
-FULL_FUZZER_NAME = FUZZER
```
jonathanmetzman, line 60: Marty I had to undo these changes to get `test_integration_runner` to pass.
Do you know if it was passing when you made the change? The test wont run unless you set `TEST_CLOUD_EXPERIMENT_BUCKET` to a bucket.

jonathanmetzman: Sorry, forgot to tag @mbarbella-chromium

mbarbella-chromium: Thanks for the heads up. I didn't set the bucket, so it's very likely that my changes broke it.

PR #294, opened 2020-05-02
Revert "Enable huge pages on runner instances. (#247)"

jonathanmetzman:
This reverts commit 625e0298a7700e9bf849c7d8c7449248bda6a576.

I'm already running the experiment for fastcgs with this commit, we don't need it anymore.


jonathanmetzman:
PTAL. Would like to revert this before doing some other experiments (though kind of a minor difference).

PR #297, opened 2020-05-02
Fix typo in oss-fuzz-coverage.yaml.

inferno-chromium:


jonathanmetzman:
> I don't think this broke the experiments because I don't think the measurer images even changed.

I meant the experiments I was running, not Marcel's.
Measuring is still happening, but it looks like there are some other issues.

PR #299, opened 2020-05-04
Don't use parallel when using gsutil to copy unchanged-cycles

### LINKED ISSUE #296: Dispatcher doesn't generate report.
Running AFL against another AFL-variant. The runners and fuzzers on those work properly. `experiment_folders` exists in the data bucket and on the dispatcher. It contains the corpus and result folders with the expected stdout output. `measurement-folders` exists on the dispatcher and also shows that coverage has been measured. 

However, the `snapshot` table in the SQL database remains empty. The `experiment` table indicates the start of the experiment. The Error Reporting only shows `ValueError: Empty experiment data. Message: Error generating HTML report.` Stackdriver under Global doesn't show anything unusual, except `Error generating HTML report.` The environment variables `CLOUD_SQL_INSTANCE_CONNECTION_NAME` and `POSTGRES_PASSWORD` as well as the config in `config/experiment.yaml` on the dispatcher are set properly - but DB connection should be fine, since the experiment-table is successfully written to. `_DEFAULT_SNAPSHOT_SECONDS` in common/experiment_utils.py is set to `6 * 60  # Seconds` and the experiment is run for 1 hour with two fuzzers on two benchmarks.

Suggestions on how to debug this?

### LINKED ISSUE #298: Cleanly aborting experiments
It would be useful to have a script that kills dispatcher and runner instances, cleans up container registry, and removes authentication keys from the service account.

jonathanmetzman:
Using parallel causes hangs. Use nonparallel (as we did with gsutil cat before we replaced it).
Also switch to non-parallel for copying crashes archive.
Also add more logging.
    
Fixes #296.

### DIFF
--- a/experiment/measurer.py
+++ b/experiment/measurer.py
@@ -115,8 +115,8 @@ def measure_all_trials(experiment: str, max_total_time: int, pool, q) -> bool:  
         return False
 
     measure_trial_coverage_args = [
-        (unmeasured_snapshot, max_cycle, q)
-        for unmeasured_snapshot in unmeasured_snapshots
+        (unmeasured_snapshot, max_cycle, q, i)
+        for i, unmeasured_snapshot in enumerate(unmeasured_snapshots)
     ]
 
     result = pool.starmap_async(measure_trial_coverage,
@@ -148,6 +148,7 @@ def measure_all_trials(experiment: str, max_total_time: int, pool, q) -> bool:  
                 # If "ready" that means pool has finished calling on each
                 # unmeasured_snapshot. Since it is finished and the queue is
                 # empty, we can stop checking the queue for more snapshots.
+                logger.info('ready')
                 break
 
             if len(snapshots) >= SNAPSHOTS_BATCH_SAVE_SIZE * .75:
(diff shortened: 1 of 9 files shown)


experiment/measurer.py
```diff
@@ -392,6 +395,7 @@ def is_cycle_unchanged(self, cycle: int) -> bool:
         def copy_unchanged_cycles_file():
             result = gsutil.cp(exp_path.gcs(self.unchanged_cycles_path),
                                self.unchanged_cycles_path,
+                               parallel=False,
```
inferno-chromium, line 398: Can we please just set to parallel to False by default, this set of bugs is getting annoying.

jonathanmetzman: Definitely, I just thought it would be better to do in another CL.
WDYT?

inferno-chromium: Just 7 places, lets just do it here.

jonathanmetzman: Doing this now but it is more than 7 places if we want to make sure that every place that was previously using parallel=True implicitly still uses it.

inferno-chromium: this is fine, you are breaking measurer anyway soon, so this does not matter much.

jonathanmetzman: Done. Does latest iteration look good to you?

inferno-chromium, line 398: Not needed.

jonathanmetzman: Done.

PR #301, opened 2020-05-04
Remove fastcgs and fastcgs_lm_huge.

jonathanmetzman:
fastcgs_lm_huge was tested in
https://www.fuzzbench.com/reports/2020-05-01-fastcgs/index.html
It also depends on #247 which was reverted in #294.


jonathanmetzman:
CC @alifahmed

alifahmed:
> 
> 
> CC @alifahmed

Thanks for the results! I actually want to remove fastcgs too, since fastcgs_lm performed a little better.

jonathanmetzman:
> Not added in CI right ?

Nope.

> > CC @alifahmed
> 
> Thanks for the results! I actually want to remove fastcgs too, since fastcgs_lm performed a little better.

OK. I'll remove that here as well after I finish something else up.

PR #302, opened 2020-05-04
Add initial support for preemptible trial runners

jonathanmetzman:
This change will use preemptible GCE instances when `preemptible_runners: true` is specified in experiment-config.yaml.
Using preemptible instances is much cheaper but comes with the following cost:
1. Trials can last only up to 24 hours (probably less if we include start up time)
2. Some trials will be preempted and thus not terminate.

This patch doesn't change parts of the scheduler that assume that it shuts down all runners. As a result if a trial is preempted the scheduler (and the measurer) will run forever. Thus the dispatcher will need to be terminated manually. For large experiments, the cost of the dispatcher running for some extra time unnecessarily is unlikely to outweigh the savings from using preemptible instances. Proper support
for preemptibles will come in a later CL.

### DIFF
--- a/common/gcloud.py
+++ b/common/gcloud.py
@@ -103,9 +103,9 @@ def create_instance(instance_name: str,
             '--machine-type=%s' % RUNNER_MACHINE_TYPE,
             '--boot-disk-size=%s' % RUNNER_BOOT_DISK_SIZE,
         ])
-        if config.get('preemptible_runners', False):
+        if config.get('preemptible_runners'):
             # TODO(metzman): Make runners signal to scheduler that they were
-            # preempted.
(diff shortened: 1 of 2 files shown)


common/gcloud.py
```diff
@@ -103,6 +103,11 @@ def create_instance(instance_name: str,
             '--machine-type=%s' % RUNNER_MACHINE_TYPE,
             '--boot-disk-size=%s' % RUNNER_BOOT_DISK_SIZE,
         ])
+        if config.get('preemptible_runners', False):
```
inferno-chromium, line 106: , False is not needed, will be None otherwise.

PR #306, opened 2020-05-06
Start building OSS-Fuzz benchmarks from source commits.

jonathanmetzman:
The current method of using frozen images won't work for bug-based
benchmarking.
This iteration wont work if the source code for the OSS-Fuzz project of the benchmark changed (e.g. oss-fuzz/projects/$PROJECT/Dockerfile, oss-fuzz/projects/$PROJECT/build.sh).

I've modified the zlib benchmark to checkout the same commit used in the frozen image, so while it won't change the benchmark, it is a decent POC.

This iteration still uses the frozen images. Except now the frozen images are not benchmark-specific, they are simply the oss-fuzz/base-builder image that was used to build a particular commit.
This may create an unresolvable problem where some benchmarks need a certain compiler to build but some fuzzers cannot use it. In that case we will probably not use the benchmark.


jonathanmetzman:
I started saving things needed to migrate our other OSS-Fuzz benchmarks.
At least two benchmarks will be problematic to migrate and may need to be dropped:
1. sqlite: It doesn't use git.
2. curl: it has two repos. One called curl_fuzzer for the fuzzers and another, the curl repo. So both may need to be checked out.

These are mainly a problem if I assume git and do the checkout using a script. It's also possible to append commands to checkout the git commit to the Dockerfile we copy over from OSS-Fuzz. If we do that, then the integrator can fix any wrong assumptions we make about git etc.

jonathanmetzman:
Also, there are some things that can be automated that I haven't done so far. Such as.
1. Figuring out the repo_path.
2. Figuring out the date of the commit. This will probably be a lot more involved and may not be worth the effort. The best way I can figure out how to do it is build the docker using current base-builder and then find the date in the commit history. But this complicates benchmark_integration.py a lot. This is what OSS-Fuzz's bisection script does. But I don't think we need that level of automation since unlike with bisection, there is a human involved in integrating the benchmark.

jonathanmetzman:
Another TODO: getting this working with GCB.

jonathanmetzman:
I think once this lands we might be able to get rid of the distinction between OSS-Fuzz benchmarks and non-OSS-Fuzz benchmarks. But that is down the road and won't be done here.

jonathanmetzman:
I switched this PR to merge into transparent-oss-fuzz-benchmarks branch. We can use that branch and work on this feature incrementally.

jonathanmetzman:
Here's a table of the OSS-Fuzz benchmarks so that they can be migrated later:
| benchmark | commit | commit date | repo_path |
| --------- | ------ | ----------- | --------- |
| bloaty_fuzz_target | 08f59b2ae229c75b01c50bfcb52e74984c8019fb | Mon Nov 18 08:52:25 2019 -0800 | /src/bloaty |
| curl_curl_fuzzer_http | 82eb95b410cc84579b8f9d663f2a7e38f3bdb41fo | Mon Nov 18 15:34:56 2019 +0000 | /src/curl_fuzzer (BAD!) |
| jsoncpp_jsoncpp_fuzzer | 3beb37ea14aec1bdce1a6d542dc464d00f4a6cec | Fri Feb 14 05:25:08 2020 +0800 | /src/jsoncpp |
| libpcap_fuzz_both | d615abec7e0237299250c409dca23effb8dd36cc | Tue Feb 11 22:46:17 2020 -0800 | /src/libpcap |
| mbedtls_fuzz_dtlsclient | 4c08dd4e716c0c6d70d78117d3155c2b832f694e | Tue Feb 11 09:17:02 2020 +0100 | /src/mbedtls |
| openssl_x509 | b0593c086dd303af31dc1e30233149978dd613c4 | Fri Feb 7 09:09:50 2020 +0100 | /src/openssl |
| systemd_fuzz-link-parser | 99fdffaa194cbfed659b0c1bfd0ace4bfcd2a245 | Mon Feb 10 17:19:52 2020 +0100 | /src/systemd |

zlib_zlib_uncompress_fuzzer is migrated to the new system in this PR.
sqlite3_ossfuzz does not use git, so our automation will fail there. No big deal, just needs to be done  manually.
curl_curl_fuzzer_http does use git. But our automation will probably fail there as well because it uses two repositories curl_fuzzer and curl. Our automation (which copies the necessary files from OSS-Fuzz and checks out the right commit) will probably fail on benchmarks that depend on multiple source checkouts. 
Also, the automation currently uses a script for checking out the right commit. Maybe this should be changed to automatically appending the commands to the dockerfile for more transparency/configurability (so things can easily be fixed manually in edge cases)

jonathanmetzman:
Note to self: test local changes before pushing.

benchmarks/zlib_zlib_uncompress_fuzzer/checksum_fuzzer.c
```diff
@@ -0,0 +1,67 @@
+#include <stdio.h>
```
jonathanmetzman, line 1: What should we do about licenses? Everything in OSS-Fuzz should have a license header but some don't. For now I added benchmarks/ to the list of ignored directories for the license check.

benchmarks/zlib_zlib_uncompress_fuzzer/compress_fuzzer.c
```diff
@@ -0,0 +1,81 @@
+#include <stdio.h>
```
jonathanmetzman, line 1: A problem with copying oss-fuzz/projects/$PROJECT dir into fuzzbench is a lot of the files will be unnecesary. What should be the policy on this? Just copy all files or just the ones needed for the benchmark (this benchmark isn't using compress_fuzzer to be clear). A problem with removing unneeded files is some build.sh files might depend on them. I think forcing removal might be more complicated than it is worth.

Tanq16:
Running `benchmark_integration.py` requires installation of Google Cloud SDK.

jonathanmetzman:
> Running `benchmark_integration.py` requires installation of Google Cloud SDK.

Tanishq found out later that this isn't actually the case. It was only true for him because he wasn't using FORCE_LOCAL.

Tanq16:
> > Running `benchmark_integration.py` requires installation of Google Cloud SDK.
> 
> Tanishq found out later that this isn't actually the case. It was only true for him because he wasn't using FORCE_LOCAL.

Update - The execution does in fact require Google Cloud SDK and is independent of FORCE_LOCAL.

jonathanmetzman:
This is not going to be merged into master just yet. But merging into the desired branch (transparent-oss-fuzz-benchmarks) will help tanishq land his migrations of benchmarks to the new system.

A big question is what should we do with files copied from OSS-Fuzz. Should we commit all of them? Or only the ones needed for the benchmark?

jonathanmetzman:
Ping.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+OSS_FUZZ_REPO_PATH = os.path.join(utils.ROOT_DIR, 'third_party', 'oss-fuzz',
+                                  'infra')
+sys.path.append(OSS_FUZZ_REPO_PATH)
+
+import bisector
+import build_specified_commit
+import helper
+import repo_manager
+
+from common import logs
```
inferno-chromium, line 43: better to move "from common" imports with the top one "from common import utils"

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+import os
+import sys
+
+from common import utils
+
+# pytype: disable=import-error
+# pylint: disable=import-error,wrong-import-position,ungrouped-imports,too-many-arguments
+
+# TODO(metzman): Don't rely on OSS-Fuzz code. We don't want to depend on it
+# because it can easily break use. Especially becuase:
```
inferno-chromium, line 29: typo becuase

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+    return dir_util.copy_tree(src_dir, dst_dir)
+
+
+def copy_oss_fuzz_files(project, commit_date, benchmark_dir):
+    """Checkout the right files from OSS-Fuzz to build the benchmark based on
+    |project| and |commit_date|. Then copy them to |benchmark_dir|."""
+    cwd = os.getcwd()
+    oss_fuzz_repo_manager = repo_manager.BaseRepoManager(helper.OSS_FUZZ_DIR)
+    try:
+        projects_dir = os.path.join(helper.OSS_FUZZ_DIR, 'projects', project)
```
inferno-chromium, line 58: can these two lines be moved out of try.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+        os.chdir(helper.OSS_FUZZ_DIR)
+        oss_fuzz_commit, _, _ = oss_fuzz_repo_manager.git([
```
inferno-chromium, line 60: Add comment on what this is doing.

docker/oss-fuzz-builder/Dockerfile
```diff
@@ -43,4 +45,11 @@ RUN touch $SRC/__init__.py
 ENV ASAN_OPTIONS="detect_leaks=0"
 
 ENV OSS_FUZZ 1
+
+ENV CHECKOUT_COMMIT $checkout_commit
+ENV CHECKOUT_COMMIT_REPO_PATH $checkout_commit_repo_path
+RUN mkdir /opt/fuzzbench/
```
inferno-chromium, line 51: Why need a dir at all, can copy to / ?

docs/developing-fuzzbench/adding_a_new_benchmark.md
```diff
@@ -35,22 +35,48 @@ powerful but also more work.
 
 ## OSS-Fuzz benchmarks
 
-You can use any existing OSS-Fuzz project as a benchmark. You just need to
-create a single file `oss-fuzz.yaml` in the benchmark directory.
+You can use most existing OSS-Fuzz projects a benchmark. First find out which
+project you want to use as a benchmark, and which commit and fuzz target you
+want to use. Then find out the date and time (UTC) of that commit and the path
```
inferno-chromium, line 40: How to find date and time, why not find using hash, this feels hard.

benchmarks/zlib_zlib_uncompress_fuzzer/zlib_uncompress_fuzzer.cc
```diff
@@ -0,0 +1,21 @@
+// Copyright 2015 The Chromium Authors. All rights reserved.
```
inferno-chromium, line 1: Why do we need to put these files here. Why not have some setup experiment step copy these on-demand. This feels weird to commit to fuzzbench source.

docker/oss-fuzz-builder/Dockerfile
jonathanmetzman: Seems much cleaner to put things in a specific location rather than `/`?

docs/developing-fuzzbench/adding_a_new_benchmark.md
jonathanmetzman: This would require building twice and is too much of a burden on fuzzbench IMO.

inferno-chromium: Why cant be derive this by calling a git command on an oss-fuzz hash.

benchmarks/zlib_zlib_uncompress_fuzzer/project.yaml
```diff
@@ -0,0 +1,19 @@
+homepage: "http://www.zlib.net/"
```
lszekeres, line 1: Is this a copy of https://github.com/google/oss-fuzz/blob/master/projects/zlib/project.yaml ?
Why are we checking it into fuzzbench?

benchmarks/zlib_zlib_uncompress_fuzzer/zlib_uncompress_fuzzer.cc
lszekeres: +1

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+                        '--fuzz-target',
+                        help='Fuzz target benchmark.',
+                        required=True)
+    parser.add_argument('-r',
+                        '--repo-path',
+                        help='Project repo path in OSS-Fuzz image.',
+                        required=True)
+    parser.add_argument('-n',
+                        '--benchmark-name',
+                        help='Benchmark name.',
```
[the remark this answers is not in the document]
jonathanmetzman: The usage string shows it's optional.

docs/developing-fuzzbench/adding_a_new_benchmark.md
jonathanmetzman: Because the commit is from the project repo, not the OSS-Fuzz repo. AFAIK the OSS-Fuzz commit isn't given in revisions by CF: https://oss-fuzz.com/revisions?job=libfuzzer_ubsan_opencv&revision=201912170255

Actually, the date should be pretty easy to get since it is in the regression range.

docs/developing-fuzzbench/adding_a_new_benchmark.md
```diff
@@ -35,22 +35,48 @@ powerful but also more work.
...
-project: <project-name>
-fuzz_target: <fuzz-target-name>
-oss_fuzz_builder_hash: <sha>
+Once you have this information, run
+`benchmarks/oss_fuzz_benchmark_integration.py` to copy the necessary integration
+files, like so:
+
+```shell
+PYTHONPATH=. python3 benchmarks/oss_fuzz_benchmark_integration.py -p $PROJECT
+    -f $FUZZ_TARGET -r $REPO_PATH -c $ COMMIT_HASH -d $COMMIT_DATE
```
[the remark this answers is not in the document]
jonathanmetzman: No, it's the absolute path of the repo in the OSS-Fuzz image.

benchmarks/oss_fuzz_benchmark_integration.py
jonathanmetzman: import order is usually 1. standard libs 2. third-party libs 3. 1st party
Here utils is needed to import third party code which is why it comes earlier, but everything else comes in the right order.

benchmarks/oss_fuzz_benchmark_integration.py
jonathanmetzman: Done.

benchmarks/oss_fuzz_benchmark_integration.py
jonathanmetzman: Done.

benchmarks/zlib_zlib_uncompress_fuzzer/zlib_uncompress_fuzzer.cc
jonathanmetzman: This is a question I want feedback on. The choices here are:
1. Put everything in our repo and have a simple build system with very little differences between OSS-Fuzz and other benchmarks (only difference is OSS-Fuzz benchmarks inherit from a specified version  of OSS-Fuzz base-builder)

2. Have some kind of magic build system where OSS-Fuzz benchmarks are just a yaml file and everything gets done behind the scenes by FuzzBench.

You might be able to tell that I favor option 1. I think FuzzBench's build system is too complex and I think option 2. would make it far worse. But I'd be interested in hearing why people think 2 is better.

docs/developing-fuzzbench/adding_a_new_benchmark.md
jonathanmetzman: Also, the purpose of the date is so we can figure out the right version of the base-builder image to inherit from. If we get rid of date we might limit the number of benchmarks we can use since some benchmarks wont build with new clang versions. But the benefit is there will be 0 distinction between OSS-Fuzz and non-oss-fuzz benchmarks for the build system, everything can inherit from gcr.io/oss-fuzz/base-images/base-builder:latest. What do people think of that option?

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+    if benchmark_name is not None:
+        return benchmark_name
+    return project + '_' + fuzz_target
+
+
+def replace_base_builder(benchmark_dir, commit_date):
+    """Replace the parent image of the Dockerfile in |benchmark_dir|,
+    base-builder (latest), with a version of base-builder that is likely to
+    build the project as it was on |commit_date| without issue."""
+    base_builder_repo = bisector._load_base_builder_repo()  # pylint: disable=protected-access
```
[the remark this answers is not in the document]
jonathanmetzman: Now that I think of it, we probably shouldn't rely on OSS-Fuzz code at all, especially because we checkout different commits. I'm going to investigate how much effort it will take to break this dependency.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+                        repo_path,
+                        benchmark_name=None):
+    """Copies files needed to integrate an OSS-Fuzz benchmark and creates the
+    benchmark's oss-fuzz.yaml file."""
+    benchmark_name = get_benchmark_name(project, fuzz_target, benchmark_name)
+    benchmark_dir = os.path.join(utils.ROOT_DIR, 'benchmarks', benchmark_name)
+    # TODO(metmzna) Replace with dateutil since fromisoformat isn't supposed to
+    # work on arbitrary iso format strings. Also figure out if i this timezone
+    # replace correct.
+    commit_date = datetime.datetime.fromisoformat(commit_date).replace(
```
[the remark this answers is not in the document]
jonathanmetzman: It breaks without this. This is how you add a timezone to a datetime without one.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+                        fuzz_target,
+                        commit,
+                        commit_date,
+                        repo_path,
+                        benchmark_name=None):
+    """Copies files needed to integrate an OSS-Fuzz benchmark and creates the
+    benchmark's oss-fuzz.yaml file."""
+    benchmark_name = get_benchmark_name(project, fuzz_target, benchmark_name)
+    benchmark_dir = os.path.join(utils.ROOT_DIR, 'benchmarks', benchmark_name)
+    # TODO(metmzna) Replace with dateutil since fromisoformat isn't supposed to
```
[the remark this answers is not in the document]
jonathanmetzman: oof :-)

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+def integrate_benchmark(project,
+                        fuzz_target,
+                        commit,
+                        commit_date,
+                        repo_path,
+                        benchmark_name=None):
+    """Copies files needed to integrate an OSS-Fuzz benchmark and creates the
+    benchmark's oss-fuzz.yaml file."""
+    benchmark_name = get_benchmark_name(project, fuzz_target, benchmark_name)
+    benchmark_dir = os.path.join(utils.ROOT_DIR, 'benchmarks', benchmark_name)
```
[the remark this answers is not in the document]
jonathanmetzman: Done.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+    parser.add_argument('-r',
+                        '--repo-path',
+                        help='Project repo path in OSS-Fuzz image.',
+                        required=True)
+    parser.add_argument('-n',
+                        '--benchmark-name',
+                        help='Benchmark name.',
+                        required=False)
+    parser.add_argument('-c', '--commit', help='Project commit.')
+    parser.add_argument('-d', '--date', help='Date.')
```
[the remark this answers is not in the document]
jonathanmetzman: Done.
It's good to have so we can pin the benchmarks to a clang that works. But see the conversation below on date for why we may not want it.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+                        required=True)
+    parser.add_argument('-r',
+                        '--repo-path',
+                        help='Project repo path in OSS-Fuzz image.',
+                        required=True)
+    parser.add_argument('-n',
+                        '--benchmark-name',
+                        help='Benchmark name.',
+                        required=False)
+    parser.add_argument('-c', '--commit', help='Project commit.')
```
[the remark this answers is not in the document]
jonathanmetzman: Done.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+                        '--project',
+                        help='Project of benchmark.',
+                        required=True)
+    parser.add_argument('-f',
+                        '--fuzz-target',
+                        help='Fuzz target benchmark.',
+                        required=True)
+    parser.add_argument('-r',
+                        '--repo-path',
+                        help='Project repo path in OSS-Fuzz image.',
```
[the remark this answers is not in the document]
jonathanmetzman: Done.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+    """Copies files needed to integrate an OSS-Fuzz benchmark and creates the
+    benchmark's oss-fuzz.yaml file."""
+    parser = argparse.ArgumentParser(description='Integrate a new benchmark.')
+    parser.add_argument('-p',
+                        '--project',
+                        help='Project of benchmark.',
+                        required=True)
+    parser.add_argument('-f',
+                        '--fuzz-target',
+                        help='Fuzz target benchmark.',
```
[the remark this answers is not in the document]
jonathanmetzman: Done.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+# because it can easily break use. Especially becuase:
+# 1. We use private methods
+# 2. The OSS-Fuzz code depends on absolute imports.
+# 3. The OSS-Fuzz code assumes it is run from the OSS-Fuzz directory and can
+# accdidentaly break our repo.
+OSS_FUZZ_REPO_PATH = os.path.join(utils.ROOT_DIR, 'third_party', 'oss-fuzz',
+                                  'infra')
+sys.path.append(OSS_FUZZ_REPO_PATH)
+
+import bisector
```
oliverchang, line 39: What's the plan here to get rid of (or at least reduce) the dependency on oss-fuzz code? These were never really intended to be depended on by other projects and will easily break, as you mention in your comment.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+import helper
+import repo_manager
+
+from common import logs
+from common import yaml_utils
+
+
+def copy_dir_contents(src_dir, dst_dir):
+    """Copy the contents of |src_dir| into |dst_dir|."""
+    return dir_util.copy_tree(src_dir, dst_dir)
```
oliverchang, line 51: why this instead of shutil.copytree ?

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def copy_oss_fuzz_files(project, commit_date, benchmark_dir):
+    """Checkout the right files from OSS-Fuzz to build the benchmark based on
+    |project| and |commit_date|. Then copy them to |benchmark_dir|."""
+    cwd = os.getcwd()
```
oliverchang, line 57: why do we need to set cwd at all? I don't think these git calls below need this, as they do not rely on the current cwd being set?

benchmarks/oss_fuzz_benchmark_integration.py
oliverchang, line 94: Just making sure that the image which this is being run on has gcloud installed?

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,164 @@
...
+        if not oss_fuzz_commit:
+            logs.warning('No suitable earlier OSS-Fuzz commit found.')
+            return False
+        oss_fuzz_repo_manager.git(['checkout', oss_fuzz_commit, projects_dir],
+                                  check_result=True)
+        copy_dir_contents(projects_dir, benchmark_dir)
+        return True
+    finally:
+        oss_fuzz_repo_manager.git(['reset', '--hard'])
+        # This must be done in this order or else we reset our fuzzbench repo.
```
oliverchang, line 78: I don't think this comment is true. git() sets the cwd

benchmarks/oss_fuzz_benchmark_integration.py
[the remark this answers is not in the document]
oliverchang: This was moved to build_specified_commit, so this is no longer up to date (an example of this breaking already :) )

Tbh I would prefer that these are re-implemented here as they're straightforward functions and to prevent more issues in the future with oss-fuzz breaking fuzzbench.

inferno-chromium:
+++++1 on @oliverchang comment - "Tbh I would prefer that these are re-implemented here as they're straightforward functions and to prevent more issues in the future with oss-fuzz breaking fuzzbench." - this reduces dependency on OSS-Fuzz, keeps code clean here and removes all the sys.path hacks.

jonathanmetzman:
Blocking: #444

benchmarks/zlib_zlib_uncompress_fuzzer/project.yaml
jonathanmetzman: Yes.
Good question, let's discuss this issue over here: https://github.com/google/fuzzbench/pull/306#discussion_r441220851

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,168 @@
...
+                        '--project',
+                        help='Project for benchmark.',
+                        required=True)
+    parser.add_argument('-f',
+                        '--fuzz-target',
+                        help='Fuzz target for benchmark.',
+                        required=True)
+    parser.add_argument('-r',
+                        '--repo-path',
+                        help=('Absolute path of the project repo in the '
```
lszekeres, line 152: to me `repo_path` suggests path to a repository. consider calling this `project_path`

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,168 @@
...
+                              'OSS-Fuzz image (e.g. /src/systemd).'),
+                        required=True)
+    parser.add_argument('-n',
+                        '--benchmark-name',
+                        help='Benchmark name.',
```
lszekeres, line 157: consider adding an example here too in the help message

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,168 @@
...
+                         benchmark_dir)
+
+
+def main():
+    """Copies files needed to integrate an OSS-Fuzz benchmark and creates the
+    benchmark's oss-fuzz.yaml file."""
+    parser = argparse.ArgumentParser(description='Integrate a new benchmark.')
+    parser.add_argument('-p',
+                        '--project',
+                        help='Project for benchmark.',
```
lszekeres, line 144: consider adding examples to all flag descriptions, so users can understand better what format is expected

benchmarks/zlib_zlib_uncompress_fuzzer/checksum_fuzzer.c
lszekeres: Why not add the licence? They were google code in the oss-fuzz repo with the same licence.

benchmarks/zlib_zlib_uncompress_fuzzer/compress_fuzzer.c
lszekeres: If we're copying the files over, then we i think we should definitely not copy unnecessary files, and should fix the build.sh. By copying these files over, these files will have a separate, independent life in fuzzbench.

benchmarks/zlib_zlib_uncompress_fuzzer/zlib_uncompress_fuzzer.cc
lszekeres: Makes sense, option 1 SGTM, ie importing benchmarks from oss-fuzz, by copying necessary files that will be independent from their origin and can be modified in fuzzbench.

The PR was a bit confusing because it was seemingly mirroring the oss-fuz directory as it is. E.g., it has 9 zlib _fuzzer.cc-s, however we probably don't want to have 9 zlib benchmarks in fuzzbench. It also copies the project.yaml file -- do we need that in fuzzbench? Can we drop the unnecessary files and fix build.sh?

Tanq16:
Removed OSS-Fuzz imports in this commit.

jonathanmetzman:
@Tanq16 is going to take over this PR.

inferno-chromium:
Having a script to import OSS-Fuzz files (excluding project.yaml) is fine. Lets move the files in another CL, to keep this one clean.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -34,18 +37,88 @@
 # accdidentaly break our repo.
 OSS_FUZZ_DIR = os.path.join(utils.ROOT_DIR, 'third_party', 'oss-fuzz')
 OSS_FUZZ_REPO_PATH = os.path.join(OSS_FUZZ_DIR, 'infra')
-
 sys.path.append(OSS_FUZZ_REPO_PATH)
```
jonathanmetzman, line 40: Can you get rid of this sys.path.append and these constants can now be defined after the imports.

jonathanmetzman:
I think script can probably exclude project.yaml.

benchmarks/oss_fuzz_benchmark_integration.py
inferno-chromium: yes right, @Tanq16  also can start in a new PR, first adding the bisector, git, etc function first with tests in one PR, then this PR of using OSS-Fuzz projects, and then another one migrating projects with OSS-Fuzz files.

Tanq16: I removed the zlib extra fuzz targets and modified only the integration python script. Will do PR for all migration separately. Hope that's alright.

benchmarks/oss_fuzz_benchmark_integration.py
Tanq16: Uses shutil.copytree now.

jonathanmetzman: I'm pretty sure copytree didn't work and that is why I used dir_util.copy_tree.
I can't exactly remember why but I think it's because copy tree will copy the directory including the name, whereas dir_util.copytree copies the *contents*

Tanq16: Right, that works for a test entry but not if it exists. I'll change that.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,243 @@
...
+from common import benchmark_utils
+from common import logs
+from common import new_process
+from common import yaml_utils
+
+
+# pytype: disable=import-error
+# pylint: disable=import-error,wrong-import-position,ungrouped-imports,too-many-arguments
+
+# TODO(metzman): Don't rely on OSS-Fuzz code. We don't want to depend on it
```
jonathanmetzman, line 36: Let's remove this TODO since it isn't needed anymore.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,243 @@
...
+import subprocess
+
+from common import utils
+from common import benchmark_utils
+from common import logs
+from common import new_process
+from common import yaml_utils
+
+
+# pytype: disable=import-error
```
jonathanmetzman, line 33: Let's remove these pylint and pytype directives as they aren't needed anymore.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,243 @@
...
+# because it can easily break use. Especially because:
+# 1. We use private methods
+# 2. The OSS-Fuzz code depends on absolute imports.
+# 3. The OSS-Fuzz code assumes it is run from the OSS-Fuzz directory and can
+# accdidentaly break our repo.
+OSS_FUZZ_DIR = os.path.join(utils.ROOT_DIR, 'third_party', 'oss-fuzz')
+OSS_FUZZ_REPO_PATH = os.path.join(OSS_FUZZ_DIR, 'infra')
+
+
+class BaseRepoManager:
```
jonathanmetzman, line 46: Let's just call this `GitRepoManager`.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,243 @@
...
+	      check_result: Should an exception be thrown on failed command.
+
+	    Returns:
+	      stderr, stdout, error code.
+	    """
+	    return new_process.execute(['git'] + cmd,
+	                         cwd=self.repo_dir)
+
+
+class BaseBuilderRepo:
```
jonathanmetzman, line 66: Can you call this `BaseBuilderDockerRepo` I think it's confusing to have two classes with `Repo` in the name where one refers to a docker repo and the other refers to a git repo.

benchmarks/oss_fuzz_benchmark_integration.py
jonathanmetzman: Oliver makes a good point, can you get rid of this cwd/chdir stuff from copy_oss_fuzz_files?

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,243 @@
...
+OSS_FUZZ_REPO_PATH = os.path.join(OSS_FUZZ_DIR, 'infra')
+
+
+class BaseRepoManager:
+	"""Base repo manager."""
+
+	def __init__(self, repo_dir):
+	    self.repo_dir = repo_dir
+
+	def git(self, cmd, check_result=False):
```
jonathanmetzman, line 52: Let's get rid of this `check_result` argument as it isn't used.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,243 @@
...
+	    """
+	    return new_process.execute(['git'] + cmd,
+	                         cwd=self.repo_dir)
+
+
+class BaseBuilderRepo:
+	"""Repo of base-builder images."""
+
+	def __init__(self):
+		self.timestamps = []
```
jonathanmetzman, line 70: Indentation seems off on this class.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,243 @@
...
+
+def get_benchmark_name(project, fuzz_target, benchmark_name=None):
+    """Returns the name of the benchmark. Returns |benchmark_name| if is set.
+    Otherwise returns a name based on |project| and |fuzz_target|."""
+    if benchmark_name is not None:
+        return benchmark_name
+    return project + '_' + fuzz_target
+
+
+def _load_base_builder_repo():
```
jonathanmetzman, line 122: can you add `docker` to this function name? `repo` is unclear on its own.

benchmarks/zlib_zlib_uncompress_fuzzer/zlib_uncompress_fuzzer.cc
jonathanmetzman: I think it will be hard to automate copying of only "necessary" files. I think the scripts should copy everything (except for obviously unnecessary stuff like the project.yaml) and then the burden for removing the unnecessary files can be on the integrator/user.

benchmarks/oss_fuzz_benchmark_integration.py
Tanq16: Done

benchmarks/oss_fuzz_benchmark_integration.py
Tanq16: Required functions from oss-fuzz code now independent in `benchmark_integration.py`.

benchmarks/oss_fuzz_benchmark_integration.py
Tanq16: Done

Tanq16:
Please review changes. @inferno-chromium @jonathanmetzman 
Thanks

benchmarks/zlib_zlib_uncompress_fuzzer/zlib_uncompress_fuzzer.cc
lszekeres: Yep, that's what i meant too. (Removing the unnecessary files from the PR, not automating it)

inferno-chromium: Importing the OSS-Fuzz files for a benchmark should be done in another CL.
As @lszekeres said, unneccessary files can be manually removed, and build.sh improved to only build the necessary fuzzers. (as part of second cl)

Tanq16: So even the already migrated zlib should be removed from this CL ?

inferno-chromium: yes right, none of the projects should be migrated in this CL. Then it is easier to review code changes, and then in next one, you can add it.

docker/build.mk
```diff
@@ -66,9 +65,18 @@ $(1)-project-name := $(shell cat benchmarks/$(1)/oss-fuzz.yaml | \
                              grep project | cut -d ':' -f2 | tr -d ' ')
 $(1)-fuzz-target  := $(shell cat benchmarks/$(1)/oss-fuzz.yaml | \
                              grep fuzz_target | cut -d ':' -f2 | tr -d ' ')
-$(1)-oss-fuzz-builder-hash := $(shell cat benchmarks/$(1)/oss-fuzz.yaml | \
-                                      grep oss_fuzz_builder_hash | \
-                                      cut -d ':' -f2 | tr -d ' ')
+$(1)-commit := $(shell cat benchmarks/$(1)/oss-fuzz.yaml | \
+                           grep commit: | cut -d ':' -f2 | tr -d ' ')
```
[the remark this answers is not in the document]
Tanq16: : is needed to get the correct value, because "commit" is present in another line in oss-fuzz.yaml.

benchmarks/oss_fuzz_benchmark_integration.py
Tanq16: Sorry if I misunderstand the question, but if you mean local test for running the integration, yes it was tested with gcloud installed.

jonathanmetzman: >Just making sure that the image which this is being run on has gcloud installed?

This is run on developers hosts so actually we don't know that gcloud is installed. We could add it as a project wide dependency or check if it is installed in this script and if it isn't point to the installation instructions.
I favor this over adding it as a project-wide dependency since I'm not 100% sure we want to keep this dependency because I'm not sure we want to use the "right" OSS-Fuzz builder image instead of "latest", since at this point it's one of the only things seperating OSS-Fuzz benchmarks from standard ones. If we get rid of this it may cause some benchmarks not to build but will simplify things. I will weight these tradeoffs at a later time.

benchmarks/zlib_zlib_uncompress_fuzzer/zlib_uncompress_fuzzer.cc
jonathanmetzman: I don't agree with this. Let's keep the zlib migration in this PR.
Keeping it in this PR is a nice proof of concept that the new build system actually works. Without it, I suspect something in this system will break while this PR is being revised and won't be noticed until any of the projects are migrated.

docker/oss-fuzz-builder/Dockerfile
```diff
@@ -21,6 +21,7 @@ FROM $parent_image
```
lszekeres, line 21: high level question (not strictly about this PR):  how difficult would it be to integrate imported oss-fuzz benchmarks with the fuzzbench builder/runner docker images, instead of using these oss-fuzz ones?

benchmarks/oss_fuzz_benchmark_integration.py
lszekeres: +1 not just this class, the whole file. i think tabs need to be converted to spaces.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,224 @@
...
+	def git(self, cmd):
+	    """Run a git command.
+
+	    Args:
+	      command: The git command as a list to be run.
+
+	    Returns:
+	      stderr, stdout, error code.
+	    """
+	    return new_process.execute(['git'] + cmd,
```
lszekeres, line 54: super nit: fine to keep as it is, but using `['git', '-C', self_repo_dir] + cmd` might be a tiny bit cleaner than setting cwd

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,224 @@
...
+    finally:
+        oss_fuzz_repo_manager.git(['reset', '--hard'])
+
+
+def get_benchmark_name(project, fuzz_target, benchmark_name=None):
+    """Returns the name of the benchmark. Returns |benchmark_name| if is set.
+    Otherwise returns a name based on |project| and |fuzz_target|."""
+    if benchmark_name is not None:
+        return benchmark_name
+    return project + '_' + fuzz_target
```
lszekeres, line 107: consider:
```python
return benchmark_name if benchmark_name else project + '_' + fuzz_target
```

docker/oss-fuzz-builder/checkout_commit.py
```diff
@@ -0,0 +1,62 @@
...
+    return git(['checkout', '-f', commit], repo_dir)
+
+
+def main():
+    """Check out an OSS-Fuzz project repo."""
+    commit = sys.argv[1]
+    if not commit:
+        print('Not checking out commit.')
+        return 0
+    src_dir = sys.argv[2]
```
lszekeres, line 47: Consider documenting the second argument. The source directory of what? What are these directory entries in it?

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -0,0 +1,224 @@
...
+		'gcr.io/oss-fuzz-base/base-builder',
+		'--format=json',
+		'--sort-by=timestamp',
+	])
+	result = json.loads(result)
+
+	repo = BaseBuilderDockerRepo()
+	for image in result:
+		timestamp = datetime.datetime.fromisoformat(
+	    	image['timestamp']['datetime']).astimezone(datetime.timezone.utc)
```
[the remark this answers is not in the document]
Tanq16: No, the image timestamp is not in utc by default.

docker/oss-fuzz-builder/Dockerfile
jonathanmetzman: I definitely want to minimize or eliminate the differences between OSS-Fuzz and standard benchmarks. This PR eliminates some of them.
After this PR lands there will be three main differences between OSS-Fuzz and standard benchmarks:
1. OSS-Fuzz benchmarks need to run checkout_commit.py (and hence need this image).
2. OSS-Fuzz benchmarks inherit from gcr.io/oss-fuzz/base-images/base-builder instead of gcr.io/fuzzbench/base-builder
3. OSS-Fuzz benchmarks inherit from a particular version of gcr.io/oss-fuzz/base-images/base-builder instead of "latest" like standard benchmarks do.

1. is a relatively minor difference, we could reduce it further by running the script for standard benchmarks and making it a noop, or we could eliminate it by making the integration script edit the docker file to checkout the right commit (or putting this (minor) burden on users).

2. is the biggest of differences/complexity. It can be solved by making all benchmarks inherit from gcr.io/oss-fuzz/base-images/base-builder. Or it can be solved by making all benchmarks inherit from gcr.io/fuzzbench/base-builder. I think gcr.io/oss-fuzz/base-images/base-builder makes more sense.

3. is done to ensure that OSS-Fuzz benchmarks build and can continue to build forever. By using the version of gcr.io/oss-fuzz/base-images/base-builder that has built the desired benchmark commit, we know that the build wont suddenly start failing because the compiler changed (in OSS-Fuzz as a whole compiler rolls frequently break projects). It's worth keeping in mind that non of our standard benchmarks use a pinned clang and have broken because of this. In the long run this will probably reduce maintenance burden (though we probably don't want to have benchmarks using clang 11.0 in 10 years from now when everyone is using clang 20 for example).

For now, I think the PR can land since it doesn't really increase differences, it makes eliminating them easier.

Maybe we could discuss this during the weekly meeting?

lszekeres: Thanks for the details, this is very useful! Sure, let's continue discussing later, I was asking independently of the PR.

benchmarks/zlib_zlib_uncompress_fuzzer/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+#
+################################################################################
+
+FROM gcr.io/oss-fuzz-base/base-builder@sha256:84df55d087d82b65e3a594c01248d59f49d513d4d7362014b417e640317b11f4
+MAINTAINER inferno@chromium.org
+RUN apt-get update && apt-get install -y make autoconf automake libtool
+RUN git clone --depth 1 -b develop https://github.com/madler/zlib.git
+WORKDIR zlib
+COPY build.sh *_fuzzer.c* $SRC/
```
lszekeres, line 22: nit: can  `*_fuzzer.c*` be replaced with `zlib_uncompress_fuzzer.cc`?

docker/oss-fuzz-builder/checkout_commit.py
```diff
@@ -0,0 +1,66 @@
...
+    return git(['checkout', '-f', commit], repo_dir)
+
+
+def main():
+    """Check out an OSS-Fuzz project repo."""
+    commit = sys.argv[1]
+    if not commit:
+        print('Not checking out commit.')
+        return 0
+    # The SRC env var which is the /src directory in the oss-fuzz builder image
```
lszekeres, line 47: Consider adding some documentation, e.g. to the top of this file, describing what this script does. It's still unclear to me because I don't know what is the input is. E.g., is `commit` a commit of github.com/google/oss-fuzz, or some upstream project repo? what is in the "/src directory in the oss-fuzz builder"? What are the directories under src_dir that we are iterating over, and why? What is the "correct project repo directory" and why do we need to search for it?

docker/oss-fuzz-builder/checkout_commit.py
```diff
@@ -0,0 +1,66 @@
...
+        entry_to_check = os.path.join(src_dir, dir_entry)
+        if os.path.isdir(entry_to_check):
+            try:
+                checkout_success = checkout_repo_commit(commit, entry_to_check)
+            except subprocess.CalledProcessError:
+                continue
+            if not checkout_success.returncode:
+                return 0
+    print("Checkout unsuccessful.")
+    return 0
```
jonathanmetzman, line 62: return 1 if unsuccessful.

docs/developing-fuzzbench/adding_a_new_benchmark.md
```diff
@@ -35,22 +35,57 @@ powerful but also more work.
 
 ## OSS-Fuzz benchmarks
 
-You can use any existing OSS-Fuzz project as a benchmark. You just need to
-create a single file `oss-fuzz.yaml` in the benchmark directory.
+You can use most existing OSS-Fuzz projects a benchmark. First find out which
```
jonathanmetzman, line 38: Change the first two sentences to
`You can use most OSS-Fuzz projects as benchmarks. First decide which project and fuzz target you want to use as a benchmark.`

docs/developing-fuzzbench/adding_a_new_benchmark.md
```diff
@@ -35,22 +35,57 @@ powerful but also more work.
...
+project and the fuzz target you want to use as a benchmark. Then find out the commit
```
jonathanmetzman, line 39: Change this third sentence to `Next, find out a project commit you want to use to build the benchmark.`

docs/developing-fuzzbench/adding_a_new_benchmark.md
```diff
@@ -35,22 +35,57 @@ powerful but also more work.
...
+(UTC) of that commit in iso format. This can be done in the project repo as follows:
+```shell
+git --no-pager log -1 $COMMIT_HASH --format=%cd --date=iso-strict
+```
 
-```yaml
-project: <project-name>
-fuzz_target: <fuzz-target-name>
-oss_fuzz_builder_hash: <sha>
+Once you have this information, run
```
jonathanmetzman, line 46: Doesn't need to come now, but we should probably walkthrough the process of getting this info.
For example, we could show how to get this info to make benchmark for the "foo" fuzz target of LLVM using the latest commit.

docs/developing-fuzzbench/adding_a_new_benchmark.md
```diff
@@ -35,22 +35,57 @@ powerful but also more work.
...
+```shell
+PYTHONPATH=. python3 benchmarks/oss_fuzz_benchmark_integration.py -p bloaty
+    -f fuzz_target -c f572d396fae9206628714fb2ce00f72e94f2258f -d 2019-10-19T09:07:25+01:00
+```
+
+The script should create the benchmark directory in
+`benchmarks/$PROJECT_$FUZZ_TARGET` (unless you specify the name manually) with
+all the files needed to build the benchmark. You must then remove unecessary files
+such as fuzz targets which are not used for the benchmark. Further, the `build.sh`
+file must be modified accordingly so as to only build the required fuzz target.
```
jonathanmetzman, line 64: "may need to be modified"
Did you need to modify it in your case?
If this is very common and annoying I may get rid of this requirement.

docs/developing-fuzzbench/adding_a_new_benchmark.md
```diff
@@ -35,22 +35,57 @@ powerful but also more work.
...
+
+Add the files in the directory to git (and then commit them):
```
jonathanmetzman, line 66: "in the benchmark directory"

docs/developing-fuzzbench/adding_a_new_benchmark.md
```diff
@@ -35,22 +35,57 @@ powerful but also more work.
...
+git add benchmarks/$BENCHMARK/*
+```
+
+Test your integration:
+
+```shell
+export FUZZER_NAME=afl
+export BENCHMARK_NAME=zlib_zlib_uncompress_fuzzer
+
+make build-$FUZZER_NAME-$BENCHMARK_NAME
```
jonathanmetzman, line 78: Get rid of "build" command. Not needed if you do "run-"

docs/developing-fuzzbench/adding_a_new_benchmark.md
```diff
@@ -35,22 +35,57 @@ powerful but also more work.
...
+make run-$FUZZER_NAME-$BENCHMARK_NAME
```
jonathanmetzman, line 79: mention that this command will run the fuzzer until Ctrl-C is used.

docs/developing-fuzzbench/adding_a_new_benchmark.md
```diff
@@ -35,22 +35,57 @@ powerful but also more work.
...
+```
+
+Add your benchmark to the list of OSS-Fuzz benchmarks in
+[test_fuzzer_benchmarks.py](https://github.com/google/fuzzbench/blob/master/.github/workflows/test_fuzzer_benchmarks.py)
```
jonathanmetzman, line 83: `[.github/workflows/build_and_test_run_fuzzer_benchmarks.py](https://github.com/google/fuzzbench/blob/master/.github/workflows/build_and_test_run_fuzzer_benchmarks.py)`.
Can you add a TODO to that file to get the list of benchmarks automatically?

docs/developing-fuzzbench/adding_a_new_benchmark.md
```diff
@@ -35,22 +35,57 @@ powerful but also more work.
...
+The script should create the benchmark directory in
+`benchmarks/$PROJECT_$FUZZ_TARGET` (unless you specify the name manually) with
+all the files needed to build the benchmark. You must then remove unecessary files
+such as fuzz targets which are not used for the benchmark. Further, the `build.sh`
+file must be modified accordingly so as to only build the required fuzz target.
+
+Add the files in the directory to git (and then commit them):
+
+```shell
+git add benchmarks/$BENCHMARK/*
```
jonathanmetzman, line 69: `$BENCHMARK_NAME` for consistency.

benchmarks/zlib_zlib_uncompress_fuzzer/Dockerfile
jonathanmetzman: It's OK if we ask for this here, but I'm not going to ask external contributors to make this many changes to benchmarks. Just what it is necessary to avoid clogging up our repo. I'm with benchmarks containing some unneeded lines if it means less work for contributors.

docker/oss-fuzz-builder/checkout_commit.py
Tanq16: `commit` is the commit for the project repo not the oss-fuzz repo. The search is over all the directories to try git checkout using the commit hash, so it's successful only in the project repo (the correct directory) to infer the path to the repo without the integrator having to specify. I'll add the same detail in the documentation.

docs/developing-fuzzbench/adding_a_new_benchmark.md
Tanq16: It is common if we want unnecessary fuzz targets to be deleted after copying from the OSS-Fuzz repo. That also follows editing `build.sh` which commonly loops over all fuzz targets to build.

docker/oss-fuzz-builder/checkout_commit.py
```diff
@@ -0,0 +1,66 @@
...
+def checkout_repo_commit(commit, repo_dir):
+    """Checkout |commit| in |repo_dir|."""
+    fetch_unshallow(repo_dir)
+    # TODO(metzman): Figure out if we need to run clean.
+    return git(['checkout', '-f', commit], repo_dir)
+
+
+def main():
+    """Check out an OSS-Fuzz project repo."""
+    commit = sys.argv[1]
```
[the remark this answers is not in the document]
Tanq16: I used 2 arguments there, one for commit and the other for `/src`, both of which are in `os.environ`. Is argparse required even if only the Dockerfile calls it ?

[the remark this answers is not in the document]
jonathanmetzman: No need for argparse, just add this:
```
if len(sys.argv) != 3:
  print('Usage: %s <commit> <src_dir>')
  return 1
```
note the code in my last version was wrong.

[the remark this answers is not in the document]
Tanq16: Noted, will do.

docker/build.mk
```diff
@@ -66,9 +65,18 @@ $(1)-project-name := $(shell cat benchmarks/$(1)/oss-fuzz.yaml | \
...
+$(1)-repo-path := $(shell cat benchmarks/$(1)/oss-fuzz.yaml | \
+                           grep repo_path | cut -d ':' -f2 | tr -d ' ')
+
+.$(1)-project-builder:
+	docker build \
+    --tag $(BASE_TAG)/builders/project/$(1) \
```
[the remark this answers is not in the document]
Tanq16: Done

docker/oss-fuzz-builder/checkout_commit.py
lszekeres: Thanks for the explanation! This feels pretty convoluted to me. I wonder why don't we checkout the commit right away after cloning the repo in the Dockerfile, ie
``` 
RUN git clone ... https://github.com/madler/zlib.git
RUN git checkout ...
```

docker/gcb/oss-fuzz-fuzzer.yaml
```diff
@@ -1,156 +0,0 @@
-# Copyright 2020 Google LLC
```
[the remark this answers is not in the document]
Tanq16: Okay done.

docker/oss-fuzz-builder/checkout_commit.py
Tanq16: That's a good point. As far as I understand, this can be done without intervention by the integrator by modifying the dockerfile using code if the name of the project repo directory is same as the project. However, that would circle back to the issue because of which we check all directories to checkout the commit (because the name of the directory may not be same as the project). So to get the name from the dockerfile and appending the checkout command would also become convoluted and may not look clean at all.

jonathanmetzman: I think doing the checkout in the dockerfile is a good goal, but I think this is fine for now. Doing it in the dockerfile is harder to get right and would probably require parsing the `git clone ...` line to figure out which is the correct clone line to modify (many oss-fuzz dockerfiles clone multiple repos)?

Tanq16: Okay. I'll add it to `checkout_commit.py`.

PR #307, opened 2020-05-06
Don't start trials for failed builds

### LINKED ISSUE #112: afl++ shows up on openssl_x509 benchmark in 2020-03-11 report even though build is broken
See https://www.fuzzbench.com/reports/2020-03-11/index.html
As i reported in #110 the afl++/openssl_x509 failed.
I assume FuzzBench used old docker image. I need to change things to not run trials for a fuzzer,benchmark pair when the build fails.

jonathanmetzman:
Don't start a trial (or save it to the db) for a fuzzer-benchmark
if the build failed (not counting builds that fail but succeed on a
retry).
Fixes #112.

### DIFF
--- a/experiment/dispatcher.py
+++ b/experiment/dispatcher.py
@@ -16,7 +16,6 @@
 configuration, spawns a runner VM for each benchmark-fuzzer combo, and then
 records coverage data received from the runner VMs."""
 
-import itertools
 import multiprocessing
 import os
 import posixpath
@@ -48,8 +47,7 @@ def create_work_subdirs(subdirs: List[str]):
 
 
 def _initialize_experiment_in_db(experiment: str, git_hash: str,
-                                 benchmarks: List[str], fuzzers: List[str],
-                                 num_trials: int):
+                                 trials: List[models.Trial]):
     """Initializes |experiment| in the database by creating the experiment
     entity and entities for each trial in the experiment."""
     db_utils.add_all([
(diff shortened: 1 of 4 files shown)


experiment/dispatcher.py
```diff
@@ -75,22 +67,44 @@ class Experiment:
...
+    # This call will raise an exception if the images can't be built which will
+    # halt the experiment.
+    builder.build_base_images()
+
+    # Only build fuzzers for benchmarks whose measurers built successfully.
+    benchmarks = builder.build_all_measurers(benchmarks)
+    build_successes = builder.build_all_fuzzer_benchmarks(fuzzers, benchmarks)
+    experiment_name = experiment_utils.get_experiment_name()
+    trials = []
+    for fuzzer, benchmark in build_successes:
```
inferno-chromium, line 98: Do we log_error when things fail ? i think we should remember to check these after an experimetn run.

experiment/dispatcher.py
```diff
@@ -75,22 +67,44 @@ class Experiment:
...
-                                     self.fuzzers, self.config['trials'])
+        self.num_trials = self.config['trials']
+        self.experiment_name = self.config['experiment']
+        self.git_hash = self.config['git_hash']
 
         self.web_bucket = posixpath.join(self.config['cloud_web_bucket'],
                                          experiment_utils.get_experiment_name())
 
 
+def build_for_trials(fuzzers: List[str], benchmarks: List[str],
```
inferno-chromium, line 84: build_images_for_trials ?

experiment/test_dispatcher.py
```diff
@@ -49,8 +50,15 @@ def mock_split_successes_and_failures(inputs, results):
 @mock.patch('multiprocessing.pool.ThreadPool', test_utils.MockPool)
 @mock.patch('experiment.build.builder.split_successes_and_failures',
             mock_split_successes_and_failures)
-def dispatcher_experiment(fs, db, experiment):
+def dispatcher_experiment(fs, db, environ):
     """Creates a dispatcher.Experiment object."""
+    work = '/work'
+    os.environ['WORK'] = work
+    fs.create_dir(work)
+    os.environ['EXPERIMENT'] = 'test-experiment'
```
inferno-chromium, line 58: Why are we not using experiment helper that sets all these vars.

experiment/dispatcher.py
jonathanmetzman: We do log these.

experiment/test_dispatcher.py
jonathanmetzman: Done.

experiment/dispatcher.py
jonathanmetzman: Done.

PR #309, opened 2020-05-06
Dont run test_measure_snapshot_coverage by default

### LINKED ISSUE #308: master cannot pass "make presubmit"
Current master branch has one failed at "``make presubmit``".

To reproduce: ``git clone https://github.com/google/fuzzbench ; cd fuzzbench ; git submodule update --init ; make install-dependencies ; source ./vene/bin/activate ; make presubmit``

https://github.com/google/fuzzbench/blob/41aa9b4267907185bd51514a6fb335ff70f4990d/experiment/test_measurer.py#L347

E            assert 0 == 3798

Reference: https://google.github.io/fuzzbench/getting-started/prerequisites/

jonathanmetzman:
The test relies on a binary that may not run on everyone's machine.
Run it in CI though.
Fixes #308.

Also, avoid using tmp_path with fakefs. Change the experiment
fixture to not use tmp_path. Combining the two causes weird errors.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -75,7 +75,7 @@ def environ():
 @pytest.fixture
 def experiment(environ):  # pylint: disable=redefined-outer-name,unused-argument
     """Mock an experiment."""
-    os.environ['WORK'] = '/directory-not-on-host'
+    os.environ['WORK'] = '/work'
     os.environ['EXPERIMENT'] = 'test-experiment'
     os.environ['CLOUD_EXPERIMENT_BUCKET'] = 'gs://experiment-data'
     os.environ['CLOUD_WEB_BUCKET'] = 'gs://web-bucket'
--- a/conftest.py
+++ b/conftest.py
@@ -73,9 +73,9 @@ def environ():
 
 
 @pytest.fixture
-def experiment(tmp_path, environ):  # pylint: disable=redefined-outer-name,unused-argument
+def experiment(environ):  # pylint: disable=redefined-outer-name,unused-argument
     """Mock an experiment."""
-    os.environ['WORK'] = str(tmp_path)
+    os.environ['WORK'] = '/directory-not-on-host'
     os.environ['EXPERIMENT'] = 'test-experiment'
     os.environ['CLOUD_EXPERIMENT_BUCKET'] = 'gs://experiment-data'
     os.environ['CLOUD_WEB_BUCKET'] = 'gs://web-bucket'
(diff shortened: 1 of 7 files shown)


conftest.py
```diff
@@ -73,9 +73,9 @@ def environ():
 
 
 @pytest.fixture
-def experiment(tmp_path, environ):  # pylint: disable=redefined-outer-name,unused-argument
+def experiment(environ):  # pylint: disable=redefined-outer-name,unused-argument
     """Mock an experiment."""
-    os.environ['WORK'] = str(tmp_path)
+    os.environ['WORK'] = '/directory-not-on-host'
```
inferno-chromium, line 78: Does this work ?

experiment/test_measurer.py
```diff
@@ -298,6 +298,8 @@ def _create_measurer(fuzzer, benchmark, trial_num):
 class TestIntegrationMeasurement:
     """Integration tests for measurement."""
 
+    @pytest.mark.skipif(not os.getenv('FUZZBENCH_TEST_INTEGRATION'),
```
inferno-chromium, line 301: Why disable, why not use some other benchmark, this feels like important test.

conftest.py
jonathanmetzman: Not unless the user of this fixture makes it in fakefs, or reassigns it to something else.

experiment/test_measurer.py
jonathanmetzman: Frankly I'm a worried there will always be some reason a random binary wont run on someone's machine, I could try a different benchmark but I'm worried well see the same issue.

lszekeres: Sorry, missing some context. Are we building and running a coverage binary outside docker? If yes, would it make sense to either:
1) do it in docker, or
2) use a fake coverage binary (could be a python script that outputs what we expect)
?

jonathanmetzman: I think we essentially do 2 elsewhere, just not here because this is the integration test and should actually run the binary.
You're probably right that this should be done in docker.

conftest.py
inferno-chromium: This is just confusing, lets just use /work, it wont exist on host, but a little clear when debugging, since /directory-not-on-host is not clear when you get in output.

experiment/test_measurer.py
inferno-chromium: +1 docker is better, dont know how much work is this ?

jonathanmetzman: I don't think getting it to work elsewhere is a priority right now. I don't expect many people to modify the code that this test covers. I'll leave it as a TODO.

conftest.py
jonathanmetzman: Done

PR #310, opened 2020-05-06
doc update prerequisites.md

zchcai:


docs/getting-started/prerequisites.md
```diff
@@ -35,7 +35,7 @@ Install Docker using the instructions
 Googlers can visit [go/installdocker](https://goto.google.com/installdocker).
 
 If you want to run `docker` without `sudo`, you can
-[create a docker group](https://docs.docker.com/engine/installation/linux/ubuntulinux/#/create-a-docker-group).
+[create a docker group](https://docs.docker.com/engine/install/linux-postinstall/).
```
jonathanmetzman, line 38: Let's link to https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user to be extra specific.

docs/getting-started/prerequisites.md
```diff
@@ -61,6 +61,7 @@ If you already have Python installed, you can verify its version by running
 Install the python dependencies by running the following command:
 
 ```bash
+sudo apt-get install python3-dev python3-venv
```
jonathanmetzman, line 64: You needed this to get things running right?

zchcai: Yes, otherwise the next step will report things missing. In fact, I first install ``python3-venv`` and then ``python3-dev``, according to reported error messages.

docs/getting-started/prerequisites.md
zchcai: Sure. Updated.

PR #317, opened 2020-05-08
Preserve gsutil rsync and cp with file permissions.

### LINKED ISSUE #316: Preserve permissions of files when starting an experiment
Currently the source code of fuzzbench is copied to GCS, before being copied to the dispatcher.
This destroys the permissions on any file in the repo.
Instead we should tar the source code and then copy it.

inferno-chromium:
This is needed when source is copied from local to gcs when running
an experiment.

Fixes #316


inferno-chromium:
This does not work since uids are different on different machines.

PR #321, opened 2020-05-10
Add details on what we want to test privately.

jonathanmetzman:
Ask for tested patch.


docs/faq.md
```diff
@@ -77,11 +77,15 @@ good to them.
 
 ## I'd like to get my fuzzer evaluated on the free FuzzBench service, but I don't want the results to be public yet.
 
-Please reach out to us at fuzzbench@google.com.
+Please reach out to us at fuzzbench@google.com. If we agree to benchmark your
+fuzzer, please send us a patch file for fuzzbench and please test your patch
```
inferno-chromium, line 81: maybe add "that applies cleanly on master"

docs/faq.md
```diff
@@ -77,11 +77,15 @@ good to them.
...
+using `make build-$FUZZER-all`.
 
 ## I'm working on a new tool, but I'm not ready to make my code public yet. Can I use the FuzzBench service?
 
-Yes, please reach out to us at fuzzbench@google.com.
+Yes, please reach out to us at fuzzbench@google.com. If we agree to benchmark
+your tool, please send us a patch file for fuzzbench and please test your
+patch using `make build-$TOOL-all`.
```
jonathanmetzman: Do you think `make build-$TOOL-all` is too burdensome? It will take a long time. It might be better just to ask for `make build test-run-$TOOL-$OSS_FUZZ_BENCHMARK test-run-$TOOL-$STANDARD_BENCHMARK`

docs/faq.md
jonathanmetzman: Done.

docs/faq.md
inferno-chromium: Yes right, i think can ask them to try against one benchmark OR one oss-fuzz,regular benchmark. wonder if example benchmark can go stale, so can ask them to try against 1-2 benchmarks ?

docker/build.mk
```diff
@@ -21,6 +21,7 @@ BASE_TAG ?= gcr.io/fuzzbench
 
 build-all: $(addsuffix -all, $(addprefix build-,$(FUZZERS)))
 pull-all: $(addsuffix -all, $(addprefix pull-,$(FUZZERS)))
+test-run-all: $(addsuffix -all, $(addprefix test-run-,$(FUZZERS)))
```
inferno-chromium, line 24: This need a rebase and move to docker/generate_makefile.py

docs/faq.md
lszekeres: We can also say: try at at least 1-2 benchmark, ideally all.

inferno-chromium: ah right, all will be a nightmare. 1-2 seems better ?

inferno-chromium:
friendly ping, this needs minor rebase and nice to get committed.

docs/faq.md
jonathanmetzman: OK updated the docs to say this.

docker/build.mk
jonathanmetzman: We don't have a test-run-all target now but we do have test-run-$FUZZER-all. I think that's probably enough. WDYT?

jonathanmetzman:
Please take another look at this. I made some more substantial changes to the docs so a rereview would be good.

docs/faq.md
```diff
@@ -75,13 +75,27 @@ to misunderstand configuration details that can have an impact on the results.
...
+
+Probably yes. We run private experiments for this purpose.
+Please reach out to us at fuzzbench@google.com. If we agree to benchmark your
+fuzzer, please follow the guide on
+[adding a new fuzzer]({{ site.baseurl }}/getting-started/adding-a-new-fuzzer/)
+on how to integrate your fuzzer with FuzzBench.
+
+You can ignore the sections on [Requesting an experiment]({{ site.baseurl }}/getting-started/adding-a-new-fuzzer/#requesting-an-experiment) and
+[Submitting your integration]({{ site.baseurl }}/getting-started/adding-a-new-fuzzer/#submitting-your-integration).
+Please test your fuzzer builds and works with our benchmarks, we won't have
```
lszekeres, line 88: nit, consider wording: "Please make sure that your fuzzer works with our benchmarks"

docs/faq.md
```diff
@@ -75,13 +75,27 @@ to misunderstand configuration details that can have an impact on the results.
...
+CI to do this for private experiments.
+Ideally, you should test all benchmarks using `make -j test-run-$FUZZER-all`.
+This takes too long on most machines so you should try a few of them:
```
lszekeres, line 91: do some of the benchmarks tend to be more "problematic" then others? this is probably hard to tell but if you happen to have some in mind consider suggesting to prioritize testing with those.

jonathanmetzman: The ones I picked were probably some of the less problematic ones but are pretty fast. bloaty_fuzz_target might be more problematic but is a bit slower. Maybe I should point out that if your fuzzer doesn't change the build process with AFL/some other fuzzer we already have you probably don't need to test others but if it does change a lot (e.g. KLEE) you should test others like bloaty_fuzz_target?

lszekeres: or just suggest of shortlist of benchmarks that includes a few fast ones and a few potentially problematic ones too

PR #323, opened 2020-05-11
Building libcxx with KLEE.

rsprabery:
More work is needed to get things like re2 to run. The `klee_driver.cpp`
needs to be aware of size limits.

@laurentsimon PTAL.

### DIFF
--- a/fuzzers/klee/builder.Dockerfile
+++ b/fuzzers/klee/builder.Dockerfile
@@ -49,19 +49,36 @@ RUN git clone https://github.com/stp/stp.git /stp && \
     -DCMAKE_INSTALL_PREFIX=/user/local/ -DCMAKE_BUILD_TYPE=Release .. && \
     make -j`nproc` && make install
 
-# Install klee-uclibc. TODO: do we need libcxx?
 RUN git clone https://github.com/lmrs2/klee-uclibc.git /klee-uclibc && \
     cd /klee-uclibc && \
     CC=`which clang-6.0` CXX=`which clang++-6.0` \
     ./configure --make-llvm-lib --with-llvm-config=`which llvm-config-6.0` && \
     make -j`nproc` && make install
 
-# Install KLEE. Use my personal repo containing additional scripts we need for now.
+
+# Install KLEE. Use my personal repo containing additional scripts we need for
+# now.
 RUN git clone https://github.com/lmrs2/klee.git /klee && \
     cd /klee && \
-    git checkout debug && \
+    git checkout debug
+
+# The libcxx build script in the KLEE repo depends on wllvm:
+# We'll upgrade pip first to avoid a build error with the old version of pip.
+RUN pip install --upgrade pip
+RUN pip install wllvm
+
+# Before building KLEE, build libcxx.
+RUN cd /klee && \
+    LLVM_VERSION=6.0 SANITIZER_BUILD= ENABLE_OPTIMIZED=0 ENABLE_DEBUG=1 \
+    DISABLE_ASSERTIONS=1 REQUIRES_RTTI=1 \
+    BASE=/out \
+    ./scripts/build/build.sh libcxx
+
+RUN cd /klee &&  \
     mkdir build && cd build && \
     CXXFLAGS= cmake -DENABLE_SOLVER_STP=ON -DENABLE_POSIX_RUNTIME=ON \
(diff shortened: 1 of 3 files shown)


fuzzers/klee/builder.Dockerfile
```diff
@@ -49,38 +49,53 @@ RUN git clone https://github.com/stp/stp.git /stp && \
...
-# Install KLEE. Use my personal repo containing additional scripts we need for now.
+
+# Install KLEE. Use my personal repo containing additional scripts we need for
+# now.
 RUN git clone https://github.com/lmrs2/klee.git /klee && \
     cd /klee && \
-    git checkout debug && \
+    git checkout debug
+
+# the libcxx build script in the KLEE repo depends on wllvm:
```
jonathanmetzman, line 65: nit: capitalize first letter in comments please.

fuzzers/klee/fuzzer.py
```diff
@@ -38,7 +38,8 @@ def prepare_build_environment():
     os.environ['CC'] = 'gclang'
     os.environ['CXX'] = 'gclang++'
 
-    os.environ['FUZZER_LIB'] = '/libAFL.a -L/ -lKleeMock'
+    os.environ['FUZZER_LIB'] = '/libAFL.a -L/ -lKleeMock -lpthread'
```
jonathanmetzman, line 41: I think `-pthread` is set by default in CXXFLAGS and CFLAGS.
Is it actually needed?

fuzzers/klee/fuzzer.py
```diff
@@ -66,13 +67,23 @@ def emptydir(path):
...
+    """Run the command |command|, optionally, run |ulimit_cmd| first."""
     cmd = ' '.join(command)
     print('[run_cmd] {}'.format(cmd))
 
     output_stream = subprocess.DEVNULL if hide_output else None
-    ret = subprocess.call(command, stdout=output_stream, stderr=output_stream)
+    if ulimit_cmd:
+        ulimit_command = [ulimit_cmd + ';']
+        ulimit_command.extend(command)
+        ret = subprocess.call(' '.join(ulimit_command),
```
jonathanmetzman, line 79: Don't think the join call is needed. `call` should work on a list.

fuzzers/klee/fuzzer.py
```diff
@@ -66,13 +67,23 @@ def emptydir(path):
     os.mkdir(path)
 
 
-def run(command, hide_output=False):
-    """Run a the command |command|"""
+def run(command, hide_output=False, ulimit_cmd=None):
```
jonathanmetzman, line 70: Why is this the ulimit stuff needed?

kren1: STP uses recursion when constraint solving. If you have a stack limit KLEE reaches it quite quickly and dies. However I'm not sure if these work inside docker, usually you have to tell docker, that the container has no stack limit.

fuzzers/klee/fuzzer.py
rsprabery: It was needed to resolve an issue with libpng, though the rest of that fix is missing in this PR.

fuzzers/klee/fuzzer.py
rsprabery: As @kren1 mentioned, this is necessary for STP. 

I've tested this with both libpng and re2 (fixes for those specific benchmarks coming in another PR) and this fixed the issue.

I had set the unlimit in the dockerfile, but it didn't appear to take. This was the only way I was able to clear up the issue with STP. Open to suggestions if anyone has a cleaner way to do this.

laurentsimon:
FYI, AFL also needs this, but does it itself in afl-fuzz.c
<https://github.com/google/AFL/blob/master/afl-fuzz.c#L2023> I don't think
KLEE has a similar option, does it?

>

fuzzers/klee/fuzzer.py
rsprabery: I double checked, in general, call works with lists as you said - but I'm getting an issue where the side effects of ulimit -s aren't visible to klee (increased stack size) unless I do this join.

rsprabery:
Thanks for the feedback! I've addressed the comments. I think the ulimit thing I've got going here is a temporary fix for now. We'll want to do something to make that cleaner. Maybe add an option to KLEE itself?

jonathanmetzman:
Didn't bother checking that presubmit wasn't one of the failures. Could you fix the please? (no rush)

PR #324, opened 2020-05-11
Fix incorrect bool in gcloud.delete_instances.

### LINKED ISSUE #322: Experiment is not ended and dispatcher stays alive even at end of experiment.
Looks like measurer is not finished, scheduler did finish.

inferno-chromium:
Fixes #322

### DIFF
--- a/common/gcloud.py
+++ b/common/gcloud.py
@@ -120,7 +120,8 @@ def create_instance(instance_name: str,
 
 
 def delete_instances(instance_names: List[str], zone: str, **kwargs) -> bool:
-    """Delete gcloud instance |instance_names|."""
+    """Delete gcloud instance |instance_names|. Returns true if the operation
+    succeeded or false otherwise."""
     error_occurred = False
     # Delete instances in batches, otherwise we run into rate limit errors.
     for idx in range(0, len(instance_names), INSTANCE_BATCH_SIZE):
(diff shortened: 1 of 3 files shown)


inferno-chromium:
Verified that this was the bug, aarya-test-9 exited successfully.

common/test_gcloud.py
```diff
@@ -170,9 +170,9 @@ def test_delete_instances_greater_than_batch_size(mocked_execute):
   than batch size."""
     instances = ['instance-%d' % i for i in range(103)]
     mocked_execute.return_value = new_process.ProcessResult(0, '', False)
-    # -q is needed otherwise gcloud will prompt "Y/N?".
```
jonathanmetzman, line 173: Why delete this?

common/gcloud.py
```diff
@@ -131,7 +131,7 @@ def delete_instances(instance_names: List[str], zone: str, **kwargs) -> bool:
         result = new_process.execute(command, expect_zero=False, **kwargs)
         error_occurred = error_occurred or result.retcode != 0
 
-    return error_occurred
+    return not error_occurred
```
jonathanmetzman, line 134: Can you update the docstring to say what return value is?

common/test_gcloud.py
inferno-chromium: It is there in source code in gcloud.delete_instances, not needed in test again.

common/gcloud.py
inferno-chromium: Done!

PR #325, opened 2020-05-12
Add option to drop trials when generating reports

jonathanmetzman:
This is useful when combining experiments that used different numbers of trials.

### DIFF
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -96,10 +96,9 @@ def filter_max_time(experiment_df, max_time):
     return experiment_df[experiment_df['time'] <= max_time]
 
 
-def filter_fuzzer_benchmark_max_trials(experiment_df, max_trials):
+def drop_fuzzer_benchmark_trials_above_max(experiment_df, max_trials):
     """Returns the table without snapshots from trials of a fuzzer-benchmark
     that is greater than |max_trials|."""
-
     # For each fuzzer-benchmark store the set of trials we will include in the
     # result.
     trials = collections.defaultdict(set)
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Utility functions for data (frame) transformations."""
+import collections
+
 from analysis import stat_tests
 from common import environment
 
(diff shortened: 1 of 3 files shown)


lszekeres:
> This is useful when combining experiments that used different numbers of trials.

Is having different numbers of trials an issue? I think having more (or less) trials for one fuzzer than for another doesn't make the experiment unfair. It just gives more precise results for the one that has more trials. Wdyt?

jonathanmetzman:
> > This is useful when combining experiments that used different numbers of trials.
> 
> Is having different numbers of trials an issue? I think having more (or less) trials for one fuzzer than for another doesn't make the experiment unfair. It just gives more precise results for the one that has more trials. Wdyt?

It is generally not an issue and I agree it isn't unfair. I think it is an aesthetic preference people may have. So I think it would be useful for that purpose (even if the results within the greater trial experiment are now less accurate). Does this sound like a reason to keep it for you?

lszekeres:
> > > This is useful when combining experiments that used different numbers of trials.
> > 
> > 
> > Is having different numbers of trials an issue? I think having more (or less) trials for one fuzzer than for another doesn't make the experiment unfair. It just gives more precise results for the one that has more trials. Wdyt?
> 
> It is generally not an issue and I agree it isn't unfair. I think it is an aesthetic preference people may have. So I think it would be useful for that purpose (even if the results within the greater trial experiment are now less accurate). Does this sound like a reason to keep it for you?

Personally to me, better data/precision is far more important than the symmetry of the report. I'd never drop useful data (obtained by burning a lot of cpu hours :). Others might have different opinion, in which case it makes sense to keep it. Wdyt?

mboehme:
Agree with @lszekeres. You probably want to maximaze statistical power. Would only remove pre-empted trials (shorter than specified time). 

If you want an equal number of trials, what about asking the scheduler to repeat trials for preempted instances?

inferno-chromium:
> Agree with @lszekeres. You probably want to maximaze statistical power. Would only remove pre-empted trials (shorter than specified time).

We wouldn't want to drop trials, but the reason is for report clarity. If we have different trials for each benchmark, it would just raise confusion in user's mind on why.

> 
> If you want an equal number of trials, what about asking the scheduler to repeat trials for preempted instances?

We are thinking on this and probably implementing this soon. Only problem is if that preemtible instance dies that, how many times to retry and it just increase total report generation time. We have to go with preemptibles since cost savings are just immense to ignore at this point.

analysis/data_utils.py
```diff
@@ -94,6 +96,39 @@ def filter_max_time(experiment_df, max_time):
...
+
+        if len(fuzzer_benchmark_trials) >= max_trials:
+            drop_trials.add(trial_id)
+            return False
+
+        fuzzer_benchmark_trials.add(trial_id)
+        return True
+
+    return experiment_df[experiment_df.apply(filter_function, axis=1)]
+
```
lszekeres, line 130: Can we do something like: 
```python
return experiment_df.group_by('benchmark', 'fuzzer', 'trial_id').head(max_trials)
```
?

analysis/generate_report.py
```diff
@@ -65,6 +65,15 @@ def get_arg_parser():
...
               'the report.'))
+
+    parser.add_argument(
+        '-x',
+        '--max-num-trials',
+        default=None,
+        type=int,
+        help=('The maximum number of trials per fuzzer-benchmark to include in '
+              'the report. Useful if you want to join experiments that used '
+              'different numbers of trials.'))
```
lszekeres, line 76: Consider adding something discouraging here, e.g.: "The maximum number of trials per fuzzer-benchmark to include in the report. Using this options drops measurement data, lowers the precision of the report, and might lead to less statistically significant results. Only use this feature if you know what you are doing".

analysis/test_data_utils.py
```diff
@@ -56,14 +56,37 @@ def test_validate_data_empty():
...
 
+def test_drop_fuzzer_benchmark_trials_above_max():
+    """Tests that drop_fuzzer_benchmark_trials_above_max drops trials for a
+    fuzzer-benchmark if they are beyond the maximum number of trials per
+    fuzzer-benchmark that we specify."""
+    experiment_df = create_experiment_data()
+    new_experiment_df = data_utils.drop_fuzzer_benchmark_trials_above_max(
+        experiment_df, 1)
+    # Sanity check test.
+    assert len(new_experiment_df) < len(experiment_df)
```
lszekeres, line 67: I'm not sure i understand this test.
Isn't what we want to check is something like:
`new_experiement_df.groupby('benchmark', 'fuzzer')['trial_id'].nunique() == 1` for all rows?

inferno-chromium:
Is this still applicable, otherwise lets close, it is getting hard to review open PR list.

jonathanmetzman:
Not going to land this now. Too controversial :-)

PR #329, opened 2020-05-12
Initial/WIP tag docker images by experiment.

jonathanmetzman:
This should make reproducing results easier.

ACTUALLY: shoudln't we just use the SHAs instead? this approach still
seems susceptible to races.


jonathanmetzman:
This doesn't need a PR yet.

Issue #40, opened 2020-03-02
Logscale X-axis on the time scales on coverage/time graphs


gamozolabs:
Without having a log scale X axis the first part of the fuzz run (typically the densest), it's pretty hard to read meaningful data from the charts.

-B

inferno-chromium:
Two solutions:
* Either use log scale
* Or switch to interactive graphs just for coverage growth plot [that helps to select a part of time interval, allow selecting only particular fuzzers, etc].

We might start with 1) and in a few weeks, switch to 2)

PR #330, opened 2020-05-12
Add option for log scale coverage growth plot.

lszekeres:
Example:

Linear:
image

Log scale:
image

### DIFF
--- a/analysis/plotting.py
+++ b/analysis/plotting.py
@@ -20,6 +20,7 @@ import scikit_posthocs as sp
 import seaborn as sns
 
 from analysis import data_utils
+from common import experiment_utils
 
 _DEFAULT_SPINE_OFFSET = 10
 _DEFAULT_TICKS_COUNT = 12
@@ -36,11 +37,11 @@ def _formatted_hour_min(seconds):
     hours = int(seconds / 60 / 60)
     minutes = int(seconds / 60) % 60
     if hours:
-        time_string += '%dhr' % hours
+        time_string += '%dh' % hours
     if minutes:
         if hours:
             time_string += ':'
-        time_string += '%dmin' % minutes
+        time_string += '%dm' % minutes
     return time_string
 
 
@@ -69,7 +70,7 @@ class Plotter:
         '#17becf', '#9edae5'
     ]
 
-    def __init__(self, fuzzers, quick):
+    def __init__(self, fuzzers, quick=False, logscale=False):
         """Instantiates plotter with list of |fuzzers|. If |quick| is True,
         creates plots faster but, with less detail.
         """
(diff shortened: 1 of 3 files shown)


analysis/plotting.py
```diff
@@ -141,10 +142,18 @@ def coverage_growth_plot(self, benchmark_df, axes=None):
...
         axes.set(xlabel='Time (hour:minute)')
 
-        ticks = np.arange(
-            0,
-            snapshot_time + 1,  # Include tick at end time.
-            snapshot_time / _DEFAULT_TICKS_COUNT)
+        if self._logscale:
+            axes.set_xscale('log')
+            ticks = np.logspace(
+                np.log10(5 * 60),  # Start from minute 5.
```
inferno-chromium, line 148: Should use constant from https://github.com/google/fuzzbench/blob/master/common/experiment_utils.py#L21 instead of 5.
our snapshot start from 15 min now.

analysis/plotting.py
```diff
@@ -141,10 +142,18 @@ def coverage_growth_plot(self, benchmark_df, axes=None):
...
+                np.log10(snapshot_time + 1),  # Include tick at end time.
+                _DEFAULT_TICKS_COUNT)
+        else:
+            ticks = np.arange(
+                0,
```
inferno-chromium, line 153: Does this need to match the logscale plot ? Should this start from snapshot time as well, will graph look weird [it currently does when we look at results in realtime, e.g. after 30 minutes, it shows empty ticks, values at like 5 min, 10 min, any tick < 15 min(first snapshot)..

jonathanmetzman:
This addresses (fixes?) #40.

analysis/plotting.py
```diff
@@ -125,7 +126,7 @@ def coverage_growth_plot(self, benchmark_df, axes=None):
             palette=self._fuzzer_colors,
             ax=axes)
 
-        sns.despine(offset=_DEFAULT_SPINE_OFFSET, trim=True)
+        sns.despine(ax=axes, offset=_DEFAULT_SPINE_OFFSET)
```
jonathanmetzman, line 129: Is this change here interesting? I don't really know what it's doing.

analysis/plotting.py
lszekeres: Done

analysis/plotting.py
lszekeres: Yes, it makes sense to match them.

analysis/plotting.py
lszekeres: Yes, this is a bug fix, but i should be doing this in a separate commit. PTAL at https://github.com/google/fuzzbench/pull/331

PR #331, opened 2020-05-12
Fix plot "despining".

lszekeres:
Despining the plot depends on the ticks, so it should happen after the ticks
were updated. Also set `ax` to make the code more OO.

https://seaborn.pydata.org/generated/seaborn.despine.html

### DIFF
--- a/analysis/plotting.py
+++ b/analysis/plotting.py
@@ -125,8 +125,6 @@ class Plotter:
             palette=self._fuzzer_colors,
             ax=axes)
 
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Something weird is going on with GitHub actions. Seeing warnings like: "The remote provider has asked us to wait until 5/13/2020 12:01:26 AM before attempting to provision a resource. This will cause delays in the running of this request." on https://github.com/google/fuzzbench/actions/runs/102941128 
So I don't know if presubmit will run for a while.

PR #334, opened 2020-05-13
Move from InsTrim to TracePCGuard

vanhauser-thc:
trace PC guard is better than instrim and everything else (except LTO I hope).
So this switches all instrim tests to trace pc guard (but leaves the fuzzer name as before - otherwise the PR is a big mess. and the results are just for me anyway :).
(I also removed the ngram3 variants as it is pointless for this experiment)

You can either keep the current state for the next bench or accept this one.
I would prefer this one as it should have better results.

I also fixed the long line in the builder Dockerfile.

Please run all aflplusplus_* fuzzers (you can skip or add the "aflplusplus" one)


fuzzers/aflplusplus_lto_mopt/builder.Dockerfile
```diff
@@ -25,7 +25,11 @@ RUN apt-get update && \
     wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
     apt-key adv --recv-keys --keyserver keyserver.ubuntu.com 1E9377A2BA9EF27F && \
     apt-get update && \
-    apt-get install -y clang-11 clang-11-doc clang-11-examples clangd-11 clang-format-11 clang-tidy-11 clang-tools-11 libc++1-11 libc++-11-dev libc++abi1-11 libc++abi-11-dev libclang1-11 libclang-11-dev libclang-common-11-dev libclang-cpp11 libclang-cpp11-dev libfuzzer-11-dev liblld-11 liblld-11-dev liblldb-11 liblldb-11-dev libllvm11 libomp-11-dev libomp-11-doc libomp5-11 lld-11 lldb-11 llvm-11 llvm-11-dev llvm-11-runtime llvm-11-tools && \
+    apt-get install -y clang-11 clangd-11 clang-tools-11 libc++1-11 libc++-11-dev \
+      libc++abi1-11 libc++abi-11-dev libclang1-11 libclang-11-dev libclang-common-11-dev \
+      libclang-cpp11 libclang-cpp11-dev libfuzzer-11-dev liblld-11 liblld-11-dev \
```
jonathanmetzman, line 30: Sorry if we discussed this already but is the clang-11 installation needed? We provide a pretty recent clang+LLVM in our standard images that gets updated pretty frequently.
(OSS-Fuzz images are another story, the clang provided there will be tied to the benchmark, this is intentional so that we know the clang is capable of compiling the benchmarks).
I don't mind if fuzzers use their own compilers, but I imagine using the provided ones will be less likely to break and may make the comparison slightly fairer (e.g. a binary produced by a different clang might have different edges).

>libfuzzer-11-dev

slightly amusing

jonathanmetzman:
Gah, I wish github reviews were smart like gerrit and selected "approve" when first word in review is "LGTM".

vanhauser-thc:
Without llvm 11 the LTO instrumentation does not work. And it’s only in there because I was persistent enough to ask(or pester) for it ... Wasn’t easy and took three attempts ;-)

jonathanmetzman:
> And it’s only in there because I was persistent enough to ask(or pester) for it ... Wasn’t easy and took three attempts ;-)

Right, which is why I'm surprised you need to install your own.
I'm fine with whatever you want to do here. I'm not going to make you remove it, I'm just curious.

vanhauser-thc:
> > And it’s only in there because I was persistent enough to ask(or pester) for it ... Wasn’t easy and took three attempts ;-)
> 
> Right, which is why I'm surprised you need to install your own.
> If it doesn't work with the LLVM we provide then I definitely understand.
> Either way, I'm fine with whatever you want to do here. I'm not going to make you remove it or spend time figuring out if it can be removed, I'm just curious.

the llvm that comes with the fuzzbench image is llvm 10. The feature I need was added in version 11 (which is still in development). So unless you have an image that has a current llvm 11 I have to install my own :)

jonathanmetzman:
> > > And it’s only in there because I was persistent enough to ask(or pester) for it ... Wasn’t easy and took three attempts ;-)
> > 
> > 
> > Right, which is why I'm surprised you need to install your own.
> > If it doesn't work with the LLVM we provide then I definitely understand.
> > Either way, I'm fine with whatever you want to do here. I'm not going to make you remove it or spend time figuring out if it can be removed, I'm just curious.
> 
> the llvm that comes with the fuzzbench image is llvm 10. The feature I need was added in version 11 (which is still in development). So unless you have an image that has a current llvm 11 I have to install my own :)

Ah, I didn't know this. I assumed it was closer to HEAD than it actually is. Sorry to bug you.

jonathanmetzman:
> Ah, I didn't know this. I assumed it was closer to HEAD than it actually is. Sorry to bug you.

Plot twist: OSS-Fuzz (and thus FuzzBench) is supposed to be using clang-11. However because of a bug that I fix [here](https://github.com/google/oss-fuzz/pull/3806) it was using clang 10. At least for me this discussion was useful,  since it revealed this bug. Once that fix lands, standard benchmark images should have a newer clang (bringing your own is still a good idea because we wont necessarily provide it with OSS-Fuzz images/benchmarks).

vanhauser-thc:
haha :)
when can you start the aflplusplus benchmark? I am eager for the results :)

vanhauser-thc:
ah they are running already! great, thanks :)
so excited for the results ... we should have place 1 or 2 with both LTO variants compared to the other fuzzers.
can you combined the results with the batch of the other fuzzers then? thanks!

jonathanmetzman:
https://www.fuzzbench.com/reports/2020-05-13-aflplusplus/index.html should be done in ~10 hours

jonathanmetzman:
Ah, I messed up and didn't pull this change. I'll start a new one with that change.

vanhauser-thc:
no keep it

vanhauser-thc:
put you pulled it?
 -> jonathanmetzman merged commit eb91087 into google:master 

if its somehow not in it, keep it running, it is still useful data for me

jonathanmetzman:
I already stopped it unfortunately sorry. 
Currently we start experiments by someone manually running an experiment from a checkout. Mine wasn't updated when I started this experiment. I'll try to be more careful but I filed #337 to avoid doing this in the future.

vanhauser-thc:
I am pretty sure you used this commit, because it deleted a variant and added a variant and the tests show that the one deleted is not present and the new one is. :)

vanhauser-thc:
@jonathanmetzman before you start the next run - 3-4 LTO fuzzing targets were not build.

http://commondatastorage.googleapis.com/fuzzbench-data/index.html?prefix=2020-05-13-aflplusplus/build-logs/

the ones that are not build (e.g. aflplusplus_lto_mopt for curl or php) show no clue why. there is no error, it is just that the file is incomplete.
My guess is that there is a timeout watcher, that terminates maybe if the there is no output for 5 minutes or the build is not completed in 30 minutes.
the LTO targets need quite some time to build, for these the values would need to be increased.

jonathanmetzman:
> I am pretty sure you used this commit, because it deleted a variant and added a variant and the tests show that the one deleted is not present and the new one is. :)

You're correct. Sorry about this, I was confused and should have double checked before terminating the experiment.

I'll check out the build failures.

jonathanmetzman:
Yep our build lengths are limited to two hours. Do you have an idea what timeout I should use for this?
It's weird how all builds (i.e. curl+php and then some) finished in CI in 3.5 hours https://github.com/google/fuzzbench/runs/670248723 but take longer than 2 on GCB.

Actually this might've made my mistake of terminating the experiment less of a waste since the experiment must have started well after I thought it did since we retry failed builds 3 times.

vanhauser-thc:
How long they take will depend on the CPU power. Maybe try 4 hours?
Why is it a good idea to timeout anyway? I mean either a build fails or it succeeds. a deadlock should be very, very rare.

jonathanmetzman:
> How long they take will depend on the CPU power. Maybe try 4 hours?
> Why is it a good idea to timeout anyway? I mean either a build fails or it succeeds. a deadlock should be very, very rare.

I have somewhat similar thoughts on timeouts generally, but I think some kind of cutoff is a good idea so we are prepared for crazy edge cases that would use infinite resources.

jonathanmetzman:
I started the build again. I see one lto build so far: aflplusplus_lto_instrim_mopt failed to build once for curl_curl_fuzzer_http:

```
Step #3 - "build-fuzzer-benchmark-builder": libtool: link: /afl/afl-clang-lto++ -g -I/src/curl_install/include -I/src/curl_install/utfuzzer -DFUZZ_PROTOCOLS_IMAP -stdlib=libc++ -pthread -Wl,--no-as-needed -Wl,-ldl -Wl,-lm -Wno-unused-command-line-argument -O3 -o curl_fuzzer_imap curl_fuzzer_imap-curl_fuzzer.o curl_fuzzer_imap-curl_fuzzer_tlv.o curl_fuzzer_imap-curl_fuzzer_callback.o  /src/curl_install/lib/libcurl.a -L/src/curl_install/lib /src/curl_install/lib/libnghttp2.a -lssl -lcrypto -lz /libAFLDriver.a -lpthread -lm -pthread
Step #3 - "build-fuzzer-benchmark-builder": [91mclang: error: unable to execute command: Killed
Step #3 - "build-fuzzer-benchmark-builder": [0m[91mclang: error: linker command failed due to signal (use -v to see invocation)
Step #3 - "build-fuzzer-benchmark-builder": [0mMakefile:724: recipe for target 'curl_fuzzer_https' failed
Step #3 - "build-fuzzer-benchmark-builder": [91mmake: *** [curl_fuzzer_https] Error 254
Step #3 - "build-fuzzer-benchmark-builder": [0m[91mmake: *** Waiting for unfinished jobs....
```

Hopefully it will pass on the second and third attempt (I'm skeptical) but if not it's fine the experiment will just run without that pair.

vanhauser-thc:
the previous run it didnt came to that point so its an improvement :)
Maybe its the available memory that is not enough? how much GB RAM is per docker instance approx?

jonathanmetzman:
> the previous run it didnt came to that point so its an improvement :)
> Maybe its the available memory that is not enough? how much GB RAM is per docker instance approx?

7.5 GB. Odd that it doesn't fail on GitHub actions which have 7 GB of RAM: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources

vanhauser-thc:
On the plus side, the PHP target is building now.
Either it needs more than 4 hours or more RAM :)
But lets keep this running. eager for the results ...

jonathanmetzman:
ACK

Issue #336, opened 2020-05-14
Use one source of truth for building images


jonathanmetzman:
Currently, making changes to building images is extremely bug prone.
Steps for building images are defined in docker/build.mk and gcb/*.yaml.
Instead we should have one source of truth for building images and generate these files from those steps.

jonathanmetzman:
The system should define the relationships between docker images using something like Python classes.

inferno-chromium:
This is tracked in build simplification issue now.

PR #342, opened 2020-05-15
Add basic support for variants to local builds.

mbarbella-chromium:
This change adds very basic support for variants to make by appending the output from a new script to the docker arguments for the builder. The script parses the appropriate fuzzer's variants.yaml file for one with a name matching a VARIANT_NAME environment variable. This will be documented in a later change.

Let me know if you have any concerns with the general approach. If it looks good it should be easy to add support for this to local experiments as well.

### DIFF
--- a/docker/build.mk
+++ b/docker/build.mk
@@ -18,7 +18,6 @@ OSS_FUZZ_BENCHMARKS := $(notdir $(shell find benchmarks -type f -name oss-fuzz.y
 
 BASE_TAG ?= gcr.io/fuzzbench
 
-
 build-all: $(addsuffix -all, $(addprefix build-,$(FUZZERS)))
 pull-all: $(addsuffix -all, $(addprefix pull-,$(FUZZERS)))
 
@@ -97,6 +96,7 @@ define fuzzer_benchmark_template
     --build-arg fuzzer=$(1) \
     --build-arg benchmark=$(2) \
     $(call cache_from,${BASE_TAG}/builders/$(1)/$(2)) \
+    $(shell python3 docker/get_docker_args.py $1) \
     --file docker/benchmark-builder/Dockerfile \
     .
 
--- a/docker/build.mk
+++ b/docker/build.mk
@@ -18,6 +18,7 @@ OSS_FUZZ_BENCHMARKS := $(notdir $(shell find benchmarks -type f -name oss-fuzz.y
 
 BASE_TAG ?= gcr.io/fuzzbench
 
+
 build-all: $(addsuffix -all, $(addprefix build-,$(FUZZERS)))
 pull-all: $(addsuffix -all, $(addprefix pull-,$(FUZZERS)))
 
--- a/docker/get_docker_args.py
+++ b/docker/get_docker_args.py
@@ -1,0 +1,48 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Helper script to get docker arguments for local building."""
+
+import os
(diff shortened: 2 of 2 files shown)


docker/build.mk
```diff
@@ -97,6 +97,7 @@ define fuzzer_benchmark_template
     --build-arg fuzzer=$(1) \
     --build-arg benchmark=$(2) \
     $(call cache_from,${BASE_TAG}/builders/$(1)/$(2)) \
+    $(shell python3 docker/get_docker_args.py $1) \
```
jonathanmetzman, line 100: Does this build target depend on the virtualenv target Make has? If not then it needs to.
Alternatively https://stackoverflow.com/questions/5014632/how-can-i-parse-a-yaml-file-from-a-linux-shell-script

@lszekeres how do you feel about having Python dependencies in our build process?  I figure for fuzzer integrators they are supposed to run presubmit anyway (though I'm sure many don't and let CI do it). Maybe we should have a different requirements.txt for building that is depended on.

off topic:
@lszekeres since the dam is bursting here by incorporating our Python code into the build process what do you think about each make target a thin wrapper that calls into Python to actually execute the build? I think this would make https://github.com/google/fuzzbench/issues/336 easier. I started work on generating the makefile using jinja and I found it hard to reason about, particularly since in make we are really generating make templates that are used to generate targets.

lszekeres: Would this build different images based on an implicit environment variable containing the variant name? If yes, i think that'd be difficult to use. Can we put the variant name in the make target just like the benchmark and fuzzer, ie.:
build-openssl-libfuzzer-default or build-openssl-libfuzzer-dictionary, and pass the variant name as build argument just like we do it for benchmark and fuzzer, eg:
```
--build-arg fuzzer=$(1) \
--build-arg benchmark=$(2) \
--build-arg variant=$(3) \
```

> I started work on generating the makefile using jinja and I found it hard to reason about, particularly since in make we are really generating make templates that are used to generate targets.

One of the benefits of generating the Makefile would be that we wouldn't need to use make templates any more, but have an explicit make target for every single thing, ie instead of build-$1-$1, we'd have `build-openssl-libfuzzer-default`, ...

Wdyt?

jonathanmetzman: >Would this build different images based on an implicit environment variable containing the variant name? 

I'm not sure what this means exactly, could you please give an example?

>One of the benefits of generating the Makefile would be that we wouldn't need to use make templates any more, but have an explicit make target for every single thing, ie instead of build-$1-$1, we'd have build-openssl-libfuzzer-default

This makes generating the makefile much easier, so I'd like to do this. The problem with doing it is then when someone adds a new variant, the won't be able to build it until they run the script to generate the makefile. Seems a bit convoluted.

lszekeres: > > Would this build different images based on an implicit environment variable containing the variant name?
> 
> I'm not sure what this means exactly, could you please give an example?

I was just asking if i understood the goal of this PR. If i understand correctly, with this PR, one would build say the "dataflow" variant of libfuzzer by setting the VARIANT_NAME env var to "dataflow" and run `make libfuzzer-libxml`. It'd be nicer if we could do this by running say "make libfuzzer-dataflow-libxml" instead. One practical benefit of this would be that we could leverage the auto-completion in bash/emacs/etc for variants too, eg. by typing "make libfuzzer-", one would automatically get all the variant options. Does that make sense?

> This makes generating the makefile much easier, so I'd like to do this. The problem with doing it is then when someone adds a new variant, they won't be able to build it until they run the script to generate the makefile. Seems a bit convoluted.

Users wouldn't need to run the script to generate the makefile, `make` itself would do that. Make supports doing this seamlessly, by allowing include files to be targets (this is a commonly used feature). We'd have something like this above the current `include docker/build.mk` line:

```
docker/build.mk: docker/generate_makefile.py $(wildcard fuzzers/*/variants.yaml)
  $< > $@

include docker/build.mk
```

When one runs `make whatever`, build.mk gets automatically regenerated - before its inclusion and evaluation - if necessary (i.e., if the generator script, or a variants.yaml, or some other dependency would change). (If it's fast enough we could even force its generation every time.)

mbarbella-chromium: Yes, you understood the intention correctly. Your proposal sounds a lot cleaner. I'll update this soon.

inferno-chromium: friendly ping @mbarbella-chromium ! Laszlo approach seems straightforward, lets implement this and document as part of CL please.

mbarbella-chromium:
Abandoning in favor of #388

Issue #286, opened 2020-05-01
afl++ benchmark batches


vanhauser-thc:
@jonathanmetzman 
I dont want to spam the announcement thread so I think communicating in a dedicated issue is better?

you wanted to run the next batch today. Will that happen? And if so - when do you roughly expect the measurements to be finished? With the changes I expect the total time to be less than the previous runs.

We have so many developments in our repo we want/need to measure we get into a backlog and need to plan the order etc.

so also - what is the minimum and maximum number of entries that make sense to run in a batch?

in about 2-3 iterations I hope we can put a "best setup" for default fuzzing in fuzzbench. until then lots of testing :)

jonathanmetzman:
>@jonathanmetzman
>I dont want to spam the announcement thread so I think communicating in a dedicated issue is better?

Probably, discussing everything experiment related on one issue might become to noisy.

>you wanted to run the next batch today. Will that happen? 

Yep. Will start it shortly

>And if so - when do you roughly expect the measurements to be finished? With the changes I expect the total time to be less than the previous runs.

Still not sure. 
The main issue with measuring is that we use a 96 core machine to measure the output of many more cores, this issue still exists.
We're still discussing how to fix this (it's more difficult when we want FuzzBench to work outside of Google Cloud). It's probably the highest priority issue we have now, but fixing it is a bit of an invasive change.

I might break up the afl++ experiment into two so we get 192 measurement cores.
I'd guess measuring finishes 2-4 days after the trials depending on whether I break it up or not.


>so also - what is the minimum and maximum number of entries that make sense to run in a batch?

I think 10 is a maximum for now. 2-3 is probably a good minimum.

jonathanmetzman:
This is blocked on #291 landing btw.

Also, we've made some changes that could possibly affect all fuzzers like 252 (https://github.com/google/fuzzbench/pull/252) since 2020-04-21 and 2020-04-20-afl++. I personally think it's fine to compare these two even if it isn't 1000% scientific to do so. Are you fine with this? I personally don't think there will be any real difference by building the targets in docker images with less unnecessary dependencies.

jonathanmetzman:
Sorry also blocked on #293.

vanhauser-thc:
if there are block then there are blocks.
If you want you can merge the new run with the results of the old ones if you want. I get the resultsdata myself and import it to Excel and run my analysis, so for me it is not necessary. I only need the standard report table with the median edges found for each target :)

[...]

PR #348, opened 2020-05-16
new alf++ variants

vanhauser-thc:
sadly the last attempt failed twice.
As eager I want to see how good the LTO variant works out (the 2x 12h attempts already showed it should be the best), I rather want compete runs to test other features.

this this compares two things:
cmplog vs laf-intel vs laf-intel with a 2^17 map (because of many more map collisions with laf-intel)

and normal vs mopt vs mopt2 (mopt2 is normal + mopt)

So basically: please test all aflplusplus variants incl. the main "aflplusplus" one. thanks! (it is a different one than from the 2020-05-11 run)


jonathanmetzman:
> sadly the last attempt failed twice.

See my comments on https://github.com/google/fuzzbench/issues/286#issuecomment-630367637 about how the last attempt was't actually a failure.

jonathanmetzman:
I started the experiment with all 7 AFL++ based fuzzers. But I don't really see what changed in aflplusplus_cmplog or aflplusplus.
Note that the compiler was updated since the last experiment if you need to compare them to the old results, let me know and I can try to figure out something to make this work.

jonathanmetzman:
Sorry, I had to kill this experiment because I started the experiment using the wrong checkout (one which only contained AFL++ and _cmplog). I'll start a fresh one shortly.

>I can try to figure out something to make this work.

To clarify what I mean here, we can pin the clang version we take from OSS-Fuzz so there's much less that can majorly change between experiments.

jonathanmetzman:
Restarted the experiment.

vanhauser-thc:
> But I don't really see what changed in aflplusplus_cmplog or aflplusplus.

before they were using the classic map[prev >> 1 ^ cur] instrumentation, now they are using trace pc guard sanitizer. so faster and less collision.
basically "aflplusplus" is now the same as "afl" in the settings (which wasnt the case before)

jonathanmetzman:
Ah sorry, didn't notice those changes came here: https://github.com/google/fuzzbench/commit/eb910871dc14d823dffa6aef024415eae9a2e01c

PR #352, opened 2020-05-18
Add support for retrying preemptible trials.

jonathanmetzman:


jonathanmetzman:
Some TODOs:
1. Update the measurer so it actually knows when to shutdown now.
2. (probably in a later CL) Don't restart trials. Instead start a new replacement trial.

database/models.py
```diff
@@ -38,6 +38,8 @@ class Trial(Base):
     benchmark = Column(String, nullable=False)
     time_started = Column(DateTime(), nullable=True)
     time_ended = Column(DateTime(), nullable=True)
+    preemptible = Column(Boolean, default=False)
+    preempted = Column(Boolean, default=False)
```
lszekeres, line 42: Can we avoid this platform specific thing leaking into our db model?

jonathanmetzman: An earlier implementation avoided this but it was much harder to get right and had uglier code. I think adding it to the db model makes for cleaner code.

jonathanmetzman: I could also have a foreign key to a GoogleCloudInstanceMetadata entity that can keep this info if you think it would be cleaner.

jonathanmetzman: I don't think this should make it harder to support local experiments by the way. Non-preemptible experiments should work without really using these fields.

lszekeres: I wasn't worried about local experiments, but about the experiment abstraction being "leaky". It'd be ideal to keep this model about the experiment itself, not including information about how it was run on GCE. Having this metadata in a separate table would def be better, but perhaps we could store this data outside the DB too. Maybe we can have a chat about it.

jonathanmetzman:
PTAL.

database/models.py
jonathanmetzman: Changing to the seperate model is a bit complicated right now because there's an experiment running that uses the current (in this PR) schema and I want to run another one later today. I'll look into changing it when the experiment is done.

common/experiment_utils.py
```diff
@@ -82,3 +82,19 @@ def get_base_docker_tag(cloud_project=None):
...
+
+def get_trial_dir(fuzzer, benchmark, trial_id):
+    """Returns the unique directory for |fuzzer|, |benchmark|, and
+    |trial_id|."""
+    benchmark_fuzzer_directory = '%s-%s' % (benchmark, fuzzer)
+    trial_subdir = 'trial-%d' % trial_id
+    return posixpath.join(benchmark_fuzzer_directory, trial_subdir)
+
+
+def get_trial_gcs_dir(fuzzer, benchmark, trial_id):
```
inferno-chromium, line 95: nit: s/get_trial_gcs_dir/get_cloud_trial_dir to match get_cloud* functions and avoid putting gcs in function name.

common/gce.py
```diff
@@ -0,0 +1,76 @@
...
+# limitations under the License.
+"""Module for using the Google Compute Engine (GCE) API."""
+import threading
+
+import dateutil.parser
+
+from googleapiclient import discovery
+from oauth2client.client import GoogleCredentials
+
+thread_local = threading.local()  # pylint: disable=invalid-name
```
inferno-chromium, line 22: Why not just
_THREAD_LOCAL
why we disable invalid name instead of naming them to avoid lint issue.

common/gce.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def initialize():
+    """Initialize the thread-local configuration with things we need to use the
+    GCE API."""
+    thread_local.credentials = GoogleCredentials.get_application_default()
```
inferno-chromium, line 28: This does not need to be in thread local ? just credentials = ?

common/gce.py
```diff
@@ -0,0 +1,76 @@
...
+def get_operations(project, zone):
+    """Generator that yields GCE operations for compute engine |project| and
+    |zone| in descendending order by time."""
+    zone_operations = thread_local.service.zoneOperations()  # pylint: disable=no-member
+    request = zone_operations.list(project=project,
+                                   zone=zone,
+                                   orderBy='creationTimestamp desc')
+    while request is not None:
+        response = request.execute()
+        for operation in response['items']:
```
inferno-chromium, line 43: Do we need to check for bad response or does it always have items.

common/gce.py
```diff
@@ -0,0 +1,76 @@
...
+            yield operation
+
+        request = zone_operations.list_next(previous_request=request,
+                                            previous_response=response)
+
+
+def get_preemption_operations(operations):
```
inferno-chromium, line 50: s/preemption/preempted ?

database/models.py
inferno-chromium: We will always run preemptible experiments from now on, so does it makes sense to even have bool there for it.

requirements.txt
```diff
@@ -9,12 +10,14 @@ pandas==1.0.1
...
 pyfakefs==3.7.1
 pytest==5.3.5
+python-dateutil==2.8.1
 PyYAML==5.3
 scikit-posthocs==0.6.2
 scipy==1.4.1
 seaborn==0.10.0
 sqlalchemy==1.3.13
 pytz==2019.3
+oauth2client==4.1.3
```
inferno-chromium, line 20: nit: alpha order move oauth2client up.

common/gce.py
inferno-chromium, line 50: Where are these functions used ? Is some code missing from this CL ?

common/gce.py
jonathanmetzman: Because the linter is wrong. This variable is not a constant and therefore should not be capitalized.

common/gce.py
jonathanmetzman: They are in scheduler.py. One of the many great features of github reviews is hiding files that have the most changes (aka the most important files in a PR). :-(

database/models.py
jonathanmetzman: 1. I'm not sure that is true, especially considering non-service users.
2. It is needed even when we use preemptibles because even then we can launch a limited number of nonpreemptibles.

common/gce.py
jonathanmetzman: I'm not sure how I should handle things if 'items' isn't in `response`, i'm fine letting it exception now.

common/gce.py
jonathanmetzman: A lot of Google libraries break when used in a multithreaded environment. Since we use threads in FuzzBench, I'm making it thread local. This is similar to how we use locks in database/utils.py but here we don't need to worry about connection limits (unlike in db_utils).

analysis/queries.py
```diff
@@ -30,6 +30,7 @@ def get_experiment_data(experiment_names):
         .select_from(Experiment)\
         .join(Trial)\
         .join(Snapshot)\
-        .filter(Experiment.name.in_(experiment_names))
+        .filter(Experiment.name.in_(experiment_names))\
+        .filter(~Trial.preempted.is_(True))
```
inferno-chromium, line 34: Why not
Trial.preempted.is_(False)

common/gce.py
inferno-chromium: I meant, you only need to do thread_local.service, not thread_local.credentials

common/gce.py
```diff
@@ -0,0 +1,76 @@
...
+    """Generator that yields GCE operations in |operations| that finished before
+    |min_end_time|. |operations| must be an iterable that is ordered by time."""
+    for operation in operations:
+        end_time = operation.get('endTime')
+        if not end_time:
+            # Try to handle cases where the operation hasn't finished.
+            yield operation
+            continue
+        operation_end_time = dateutil.parser.isoparse(end_time)
+
```
inferno-chromium, line 73: nit: can remove newline since connected to next statement.

experiment/scheduler.py
```diff
@@ -121,6 +140,436 @@ def end_expired_trials(experiment_config: dict):
...
+            trial.preempted = True
+            trial.time_ended = time_ended
+
+            # We try to start each replacement trial as a preemptible before
+            # trying nonpreemptible to minimize cost.
+            if self.can_start_preemptible(preemptible_starts):
+                # See if we can replace with a preemptible.
+                preemptible_starts += 1
+                num_to_replace -= 1
+                replacements.append(replace_trial(trial, True))
```
inferno-chromium, line 384: Use named argument for True, False.

analysis/queries.py
jonathanmetzman: I could make this work, but I would have to change the migration to add preempted=False to each trial. Should I do this? Otherwise this query fails when combining new data with old because all of the old trials have null in this column.

inferno-chromium: nah, not needed then.

experiment/scheduler.py
```diff
@@ -89,6 +106,14 @@ def all_trials_ended(experiment: str) -> bool:
         models.Trial.time_ended.is_(None)).all()
 
 
+def delete_instances(instances, experiment_config):
+    """Deletes |instances|."""
+    running_instances = gcloud.list_instances()
+    instances_to_delete = [i for i in instances if i in running_instances]
+    return gcloud.delete_instances(instances_to_delete,
```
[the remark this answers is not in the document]
jonathanmetzman: You mean check if it is None?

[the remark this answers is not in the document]
inferno-chromium: ah ok

experiment/scheduler.py
jonathanmetzman: Done.

common/gce.py
jonathanmetzman: Done.

experiment/scheduler.py
```diff
@@ -121,6 +140,436 @@ def end_expired_trials(experiment_config: dict):
...
+        self.preemptible_window = (experiment_config['max_total_time'] *
+                                   self.PREEMPTIBLE_WINDOW_MULTIPLIER)
+        self._initial_trials = list(
+            get_experiment_trials(experiment_config['experiment']))
+        self._max_time_started = None
+
+        self.preempted_trials = {}
+        self.preemptible_starts_futile = False
+
+        self.base_resource_url = (
```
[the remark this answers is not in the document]
jonathanmetzman: I renamed this base_target_link and I still keep it as an attribute to avoid having to format the same string every time I call get_instance_from_preempted_operation. But I moved the formatting to a function in gce.py and moved get_instance_from_preempted_operation to there as well.

experiment/scheduler.py
```diff
@@ -121,6 +140,436 @@ def end_expired_trials(experiment_config: dict):
...
+        self.max_nonpreemptibles = min(
+            math.ceil(self.num_trials * self.NONPREEMPTIBLES_FRACTION),
+            self.MAX_NONPREEMPTIBLES)
+        logger.info('Max nonpreemptibles: %d.', self.max_nonpreemptibles)
+
+        # Bound for the number of preemptibles we can start if the experiment
+        # specified preemptible_runners.
+        self.max_preemptibles = (self.num_trials *
+                                 self.MAX_PREEMPTIBLES_MULTIPLIER)
+        logger.info('Max nonpreemptibles: %d.', self.max_preemptibles)
```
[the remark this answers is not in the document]
jonathanmetzman: Done.

experiment/scheduler.py
```diff
@@ -121,6 +140,436 @@ def end_expired_trials(experiment_config: dict):
...
+    This class object should be created at the start of scheduling and the
+    handle_preempted_trials method should be called in the scheduling loop.
+    See the docstring for handle_preempted_trials for how it works.
+    """
+    # Hard limit on the number of nonpreemptibles we will use. This bounds
+    # costs.
+    MAX_NONPREEMPTIBLES = 500
+
+    # The maximum fraction of total trials in the experiment that can be done
+    # using preemptibles. This helps bound the cost in unexpected situations.
```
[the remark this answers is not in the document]
jonathanmetzman: Done.

experiment/scheduler.py
```diff
@@ -121,6 +140,436 @@ def end_expired_trials(experiment_config: dict):
...
+                return None
+
+            if max_time_started is None:
+                max_time_started = time_started
+                continue
+
+            max_time_started = max(time_started, max_time_started)
+
+        assert max_time_started is not None
+        max_time_started = max_time_started.replace(
```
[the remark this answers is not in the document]
jonathanmetzman: We do store them as utc, but we don't keep the timezone info in there because all times are utc in postgres. It's not clear to me how/if we can store the times as timezone aware: https://stackoverflow.com/questions/414952/sqlalchemy-datetime-timezone

analysis/queries.py
jonathanmetzman: I'll try, I'd rather not have to contort our code forever just because the migration wasn't thorough enough.

database/models.py
jonathanmetzman: I started changing this and I'm not sure the complexity is worth it (especially since I already have some data using the current schema). Changing this seems to add an annoying layer of indirection for not much benefit.
I'm going to land as is, since this PR is sort of blocking my scalable measurer work

PR #353, opened 2020-05-18
Add a size argument to klee_driver.cpp

rsprabery:
This allows klee to target things like re2.

Also, using this PR to clean up failing presubmit tests once I can see how the CI system handles this branch.

@laurentsimon

### DIFF
--- a/fuzzers/klee/fuzzer.py
+++ b/fuzzers/klee/fuzzer.py
@@ -16,10 +16,14 @@
 import shutil
 import os
 import glob
+import pathlib
 import struct
 import subprocess
 
 from fuzzers import utils
+LIB_BC_DIR = "lib-bc"
+SYMBOLIC_BUFFER = "KleeInputBuf"
+MODEL_VERSION = "model_version"
 
 
 def is_benchmark(name):
@@ -62,7 +66,7 @@ def get_bcs_for_shared_libs(fuzz_target):
     ldd_cmd = ["/usr/bin/ldd", "{target}".format(target=fuzz_target)]
     output = ""
     try:
-        output = subprocess.check_output(ldd_cmd, text=True)
+        output = subprocess.check_output(ldd_cmd, universal_newlines=True)
     except subprocess.CalledProcessError:
         raise ValueError("ldd failed")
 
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
> This allows klee to target things like re2.
> 
> Also, using this PR to clean up failing presubmit tests once I can see how the CI system handles this branch.
> 
> @laurentsimon

The hardcoded values are a bit worrying. What needs to be "64" for re2?

laurentsimon:
see this file
https://github.com/google/fuzzbench/blob/master/benchmarks/re2-2014-12-09/target.cc#L19
We've decided to set the size of the input to the maximum allowed (64). For
other benchmarks we've arbitrarily selected 4096.

rsprabery:
Let's wait on the merge until i can clean up the presubmit issues btw.

kren1:
Side note. 4096 as a default is huge and should be kept as small as possible. I don't think KLEE has any hope of using inputs of this size and having big symbolic arrays makes KLEE slower. I would be careful with increasing the input size above 100 bytes. I guess larger inputs are needed for some formats and the input needs to fit some seeds.

laurentsimon:
thanks for the input. Do you think 100 bytes would be enough for other
formats such as pcap, x509 certs? We could change on a per-benchmark
basis... Something we thought about is to use -sym-args <N>
<https://klee.github.io/docs/options/>. Do you think this would be a viable
alternative? I don't know if this adds complexity to KLEE.

kren1:
Yes I agree, 100 bytes is not enough for all benchmarks. Ideally it should be adjusted automatically depending on benchmarks, maybe by taking the size of the largest seed. But those might be too large.

What `-sym-arg <N>` does is replace that argument with `N` symbolic bytes. so for example `klee ... ls -sym-arg 2` would explore paths where `ls` is run with `ls -l`, `ls -h` etc... . So I don't think it is particularly useful here. What you could do is make `klee_driver.cpp` read from a file (or stdin) and then run `klee ... benchmark.bc --sym-stdin <N>`. This would be a viable alternative and it doesn't add any meaningful complexity to klee.

With this second option there is also a nicer interface to klee ( at klee/klee#1222 ), where you can run the above `cat seed | klee-zesti ... benchmark.bc`, which is equivalent to the above, but stdin is seeded with `seed`. But I'm not sure if that is that useful with many seeds. Unless you run one instance of klee for each seed.

laurentsimon:
On Tue, May 19, 2020 at 8:22 AM Timotej Kapus <notifications@github.com>
wrote:

> Yes I agree, 100 bytes is not enough for all benchmarks. Ideally it should
> be adjusted automatically depending on benchmarks, maybe by taking the size
> of the largest seed. But those might be too large.
>
some seeds may be many MB sometimes :)

> What -sym-arg <N> does is replace that argument with N symbolic bytes. so
> for example klee ... ls -sym-arg 2 would explore paths where ls is run
> with ls -l, ls -h etc... . So I don't think it is particularly useful
> here. What you could do is make klee_driver.cpp read from a file (or
> stdin) and then run klee ... benchmark.bc --sym-stdin <N>. This would be
> a viable alternative and it doesn't add any meaningful complexity to klee.
>

this is essentially the same as what the PR does, but using a different
argument to pass the length. Is that correct?
If this is cleaner, we could use this instead. rsprabery@ what do you think?

With this second option there is also a nicer interface to klee ( at

kren1:
You still have to read that file, so it's a bit more code, I think the current way is fine.

PR #359, opened 2020-05-20
[run_experiment] Only permit selecting fuzzers in the repo.

jonathanmetzman:


jonathanmetzman:
@mbarbella-chromium should fuzzer_utils.get_fuzzer_configs(fuzzers) exception if there is a fuzzer without an available config?

Issue #198, opened 2020-04-01
Ankou: Request for new fuzzer integration


inferno-chromium:
https://github.com/SoftSec-KAIST/Ankou

@Jiliac, @sangkilc - can you please consider integration Ankou with our fuzzer benchmarking platform. We would love to have the community benefit from your research work.

Jiliac:
Hello. Thanks for pinging us :-) We are preparing for it.

zchcai:
Hi @Jiliac, how about the progress of integration in your side? Maybe I can help. :)

Zhicheng
FuzzBench

Jiliac:
Ah sorry I totally missed the activity on this thread. Thanks a ton for handling it @zchcai !

jonathanmetzman:
We have results https://www.fuzzbench.com/reports/2020-05-28/index.html, but I think they aren't accurate in Ankou's case because the integration has some bugs: https://github.com/google/fuzzbench/issues/392

PR #364, opened 2020-05-21
[Ankou] initial fuzzer integration

zchcai:
- This PR relates to [this issue](https://github.com/google/fuzzbench/issues/198).
- The current approach reuses/bases on [afl](https://github.com/zchcai/fuzzbench/blob/master/fuzzers/afl) integration.
- For more usage options of Ankou, please refer its repo to help integrate in the future. For example, "-dict", "-dur", etc.

cc: @lszekeres


zchcai:
A full list of usages of Ankou:
```
$ ./Ankou --help
Usage of ./Ankou:
  -app string
        Fuzz target: AFL instrumented binary path
  -args string
        App arguments. Must be quoted. @@ for file fuzzing.
        If not specified, fuzzed in standard input.
  -debug
        Print debug information
  -dict string
        Fuzzer dictionary (see AFL doc for format)
  -dur string
        Time to fuzz (default is infinite)
  -i string
        Input directory with seeds, or seed files.
        If multiple, comma separated, no space.
  -o string
        Output directory (crashes and discovered seeds).
  -select string
        Seed selection algorithm: WMO or random  (default "rand")
  -threads int
        Number of threads (i.e. number of fork servers) (default 1)
```

fuzzers/ankou/builder.Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+ARG parent_image=gcr.io/fuzzbench/base-builder
+FROM $parent_image
+
+# Go
```
jonathanmetzman, line 18: "Install Go."
Also, end comments with a period please.

fuzzers/ankou/fuzzer.py
```diff
@@ -0,0 +1,93 @@
...
+import shutil
+import subprocess
+import os
+
+from fuzzers import utils
+
+# OUT environment variable is the location of build directory (default is /out).
+
+
+def prepare_build_environment():
```
jonathanmetzman, line 25: You don't really need to define this function, we do it in the AFL fuzzer.py because it is useful for people who import it.

jonathanmetzman, line 25: Actually, if this is the same as AFL, why not just copy it?

fuzzers/ankou/fuzzer.py
```diff
@@ -0,0 +1,93 @@
...
+    prepare_build_environment()
+
+    utils.build_benchmark()
+
+    print('[post_build] Copying Ankou to $OUT directory')
+    # Copy out the Ankou binary as a build artifact.
+    shutil.copy('/Ankou', os.environ['OUT'])
+
+
+def prepare_fuzz_environment(input_corpus):
```
jonathanmetzman, line 48: Same comments here that I wrote on `prepare_build_environment`

fuzzers/ankou/fuzzer.py
```diff
@@ -0,0 +1,93 @@
...
+    os.environ['AFL_I_DONT_CARE_ABOUT_MISSING_CRASHES'] = '1'
+
+    # AFL needs at least one non-empty seed to start.
+    if len(os.listdir(input_corpus)) == 0:
+        with open(os.path.join(input_corpus, 'default_seed'),
+                  'w') as file_handle:
+            file_handle.write('hi')
+
+
+def run_ankou_fuzz(input_corpus,
```
jonathanmetzman, line 67: I would just do this in fuzz() since it's unlikely anyone will want to import this.

fuzzers/ankou/fuzzer.py
```diff
@@ -0,0 +1,93 @@
...
+                   hide_output=False):
+    """Run Ankou."""
+    print('[run_fuzzer] Running target with Ankou')
+    command = [
+        './Ankou', '-app', target_binary, '-i', input_corpus, '-o',
+        output_corpus
+    ]
+    if additional_flags:
+        command.extend(additional_flags)
+    #dictionary_path = utils.get_dictionary_path(target_binary)
```
jonathanmetzman, line 80: Just delete this code and leave a comment about how ankou doesn't support dicts.

fuzzers/ankou/fuzzer.py
```diff
@@ -0,0 +1,93 @@
...
+    #if dictionary_path:
+    #    command.extend(['-dict', dictionary_path])
+
+    print('[run_fuzzer] Running command: ' + ' '.join(command))
+    output_stream = subprocess.DEVNULL if hide_output else None
```
jonathanmetzman, line 85: The hide output stuff is only really needed for AFLFast which was filling up disks with logs.

jonathanmetzman:
Fixes #198.

fuzzers/ankou/fuzzer.py
zchcai: I uncommented out this part and did a local test. It also builds up code coverage, and thus maybe the previous error is due to a potential dictionary format mismatch.

I suggest we put this option in the code at this stage.

jonathanmetzman: If it doesn't work everywhere I would recommend not using it, or at least explicitly not doing it in places where it doesn't work.

fuzzers/ankou/builder.Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+# Install Go.
+RUN mkdir -p /application
+RUN wget https://dl.google.com/go/go1.14.3.linux-amd64.tar.gz
+RUN tar -C /application -xzf go1.14.3.linux-amd64.tar.gz
+
+# Clone Ankou and its dependencies.
+RUN /application/go/bin/go get github.com/SoftSec-KAIST/Ankou
+# Compile Ankou.
+RUN /application/go/bin/go build github.com/SoftSec-KAIST/Ankou
+# /Ankou exists now.
```
lszekeres, line 27: nit: i think we can remove this comment

fuzzers/ankou/fuzzer.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def build():
+    """Build benchmark."""
+    afl_fuzzer.prepare_build_environment()
+
+    utils.build_benchmark()
+
+    print('[post_build] Copying Ankou to $OUT directory')
+    # Copy out the Ankou binary as a build artifact.
```
lszekeres, line 33: nit: can remove this comment (redundant with print() above)

fuzzers/ankou/fuzzer.py
zchcai: All right, let me change this.

zchcai:
updated `docs/_data/fuzzers.yaml` and `.github/workflows/ci.yml`

PR #365, opened 2020-05-21
add clear-cache command

laurentsimon:


jonathanmetzman:
I think this solution is somewhat of a blunt hammer. I think we could acheive the same things if we tell people to pull and we pull the right images. I assume the issue laurent saw before was that the base-builder uses oss-fuzz/base-clang but doesn't pull it first so a stale one was used.

inferno-chromium:
> I think this solution is somewhat of a blunt instrument. I think we could acheive the same things if we tell people to pull and we pull the right images. I assume the issue laurent saw before was that the base-builder uses oss-fuzz/base-clang but doesn't pull it first so a stale one was used.

We can add a pull_images command to update base-*. But this command does not hurt to have, like if we have bad intermediate project images.

zchcai:
According to https://github.com/google/fuzzbench/pull/480#discussion_r446370910, this PR may need to revert.

PR #319, opened 2020-05-09
Next LTO attempt

vanhauser-thc:
Don't merge yet, I first want to be sure it compiles and see which test cases work and don't and if I can fix that. thanks!

PR #367, opened 2020-05-21
adding a 2nd default variant for aflplusplus

vanhauser-thc:
This adds an "optimal" aflplusplus to be added permanently to the benchmarks.

In afl++ we try to keep everything as close as possible as vanilla afl is doing it.
Hence LTO, Instrim, CTX, ngram, mopt, laf-intel, etc. are not enabled by default.
But being able to customize perfectly to a target is what makes afl++ so much better than afl.
afl is a great workhorse. however when you repeatedly fuzz the same target, it is nonsense not to configure it to fit best to the target, e.g. deterministic mutation on/off, dictionaries, etc.

this variant should never be run alone, but rather together with the default "aflplusplus" variant, to show how much better it can be if it is adapted to a target. and how good it is even when run in standard mode.

Note that in previous benchmark runs afl++ looked pretty bad. that is because we were using the classic instrumentation (map[prev_loc >> 1 ^ cur_loc]++) however afl was configured to use the much better pcguard instrumentation. so we switched that in the "aflplusplus" variant as well. we should now be a very little better than afl in the benchmarks.


vanhauser-thc:
I dont get it.
when I do `make run-aflplusplus_optimal-jsoncpp_jsoncpp_fuzzer` and `run-aflplusplus_optimal-openthread-2019-12-23` the run perfectly fine.
Here they fail. and they were not failing when testing LTO.

jonathanmetzman:
I'll take a look

jonathanmetzman:
Only difference I see from LTO (https://github.com/google/fuzzbench/pull/319/files#diff-342474d7284c7f8f1af7feccc4b6b533) is the sed call to increase the map size. I'm skeptical this is breaking things.
Can you link to logs from GitHub actions that you think should pass but aren't. I see failures for systemd (here for example) but not the becnhmarks you've mentioned.

vanhauser-thc:
> Only difference I see from LTO (https://github.com/google/fuzzbench/pull/319/files#diff-342474d7284c7f8f1af7feccc4b6b533) is the sed call to increase the map size. I'm skeptical this is breaking things.
> Can you link to logs from GitHub actions that you think should pass but aren't. I see failures for systemd (here for example) but not the becnhmarks you've mentioned.

no you misunderstand me :)

on my system: `make run-aflplusplus_optimal-jsoncpp_jsoncpp_fuzzer`  works.

in the build system which also does exactly the same `make run-aflplusplus_optimal-jsoncpp_jsoncpp_fuzzer` it fails (ok it does `...-all` but my point is that the sed etc calls are all the same.)

so the exact same setup (I ensured my fuzzbench is current) succeeds for me and fails on travis.
the build failure is in the docker container so ... how can there be a difference?

vanhauser-thc:
wonderful, it runs for over an hour and then fails because of github issues ;)
```
 --2020-05-22 22:47:46--  https://raw.githubusercontent.com/llvm/llvm-project/5feb80e748924606531ba28c97fe65145c65372e/compiler-rt/lib/fuzzer/afl/afl_driver.cpp
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 151.101.248.133
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|151.101.248.133|:443... connected.
HTTP request sent, awaiting response... 500 Internal Server Error
2020-05-22 22:47:58 ERROR 500: Internal Server Error.
```
can you please restart the aflplusplus_optimal failed ones? thank you

jonathanmetzman:
> wonderful, it runs for over an hour and then fails because of github issues ;)
> 
> ```
>  --2020-05-22 22:47:46--  https://raw.githubusercontent.com/llvm/llvm-project/5feb80e748924606531ba28c97fe65145c65372e/compiler-rt/lib/fuzzer/afl/afl_driver.cpp
> Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 151.101.248.133
> Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|151.101.248.133|:443... connected.
> HTTP request sent, awaiting response... 500 Internal Server Error
> 2020-05-22 22:47:58 ERROR 500: Internal Server Error.
> ```
> 
> can you please restart the aflplusplus_optimal failed ones? thank you

I cancelled the currently running ones and restarted them all because that is the only way we can restart them. In the future, it's probably better for you to just push again to trigger a build. That way the currently running builds you probably didn't want cancelled (I assume I shouldn't have cancelled AFL++ build) can finish.

vanhauser-thc:
yes, it built!!! :)
there is still the risk that the LTO ones will take to long and be killed when fuzzing though, but the I am hopeful.
No need to run this now, just do it together in the next run with aflsmart, honggfuzz - and Ankou, really curious how that one will perform ...

PR #369, opened 2020-05-22
Start using task queue for measuring.

jonathanmetzman:
Start using task queue for measuring.
    
This commit uses a task queue to distribute measurement tasks to
workers. Although these workers still run on the dispatcher.
This uses rq and redis.
    
This is the first commit needed for horizontally scalable measuring.
Still TODO is the following:
1. Break up the measurer module into manager and worker submodules.
2. Make workers able to run on different machines. This means unpacking
their own builds and saving state to GCS.
3. Have the scheduler start running measurer workers on different machines.
4. Optimize so we don't need as many trips to the task queue or can
measure even more horizontally (e.g. different snapshots in trial in
parallel.
5. Document setting up redis.


jonathanmetzman:
I'm going to do these 5 steps incrementally to make code review easier. So I'll merge this branch into "queue" and then merge queue into master once it is ready.

experiment/measurer/measure_manager.py
```diff
@@ -0,0 +1,327 @@
...
+    logs.initialize(default_extras={
+        'component': 'dispatcher',
+        'subcomponent': 'measurer',
+    })
+    with multiprocessing.Pool() as pool:
+        set_up_coverage_binaries(pool, experiment)
+    # Using Multiprocessing.Queue will fail with a complaint about
+    # inheriting queue.
+    redis_connection = redis.Redis(host=redis_host)
+    q = rq.Queue(connection=redis_connection)
```
lszekeres, line 68: readability nit: q -> queue

experiment/measurer/measure_worker.py
```diff
@@ -11,23 +10,16 @@
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
-# limitations under the License.
+# limitations under the License. for now
```
lszekeres, line 13: for now?

docker/dispatcher-image/startup-dispatcher.sh
```diff
@@ -34,9 +34,22 @@ iam_account=`gcloud config get-value account`
 gcloud iam service-accounts keys create ${credentials_file} \
     --iam-account=${iam_account}
 
+for i in {1..$(nproc)}
+do
+  PYTHONPATH=${WORK}/src GOOGLE_APPLICATION_CREDENTIALS=${credentials_file}
+
+done
```
lszekeres, line 41: stale?

experiment/measurer/measure_manager.py
jonathanmetzman: Done.

docker/dispatcher-image/startup-dispatcher.sh
jonathanmetzman: Done.

experiment/measurer/measure_worker.py
jonathanmetzman: Done.

experiment/measurer/measure_manager.py
```diff
@@ -0,0 +1,327 @@
...
+
+    if not unmeasured_snapshots:
+        return False
+
+    job_timeout = get_job_timeout()
+    results = [
+        queue.enqueue(measure_worker.measure_trial_coverage,
+                      unmeasured_snapshot,
+                      job_timeout=job_timeout)
+        for unmeasured_snapshot in unmeasured_snapshots
```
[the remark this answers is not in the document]
jonathanmetzman: Currently we won't ever create a task for measuring snapshot. I'll look into using `depends_on` because it seems more elegant than my current approach. But I need to make sure it works well with with preemptibles where somtimes we won't want to measure snapshot N+1 just because N was measured.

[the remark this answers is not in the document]
lszekeres: > Currently we won't ever create a task for measuring snapshot. 

Oh, i just realized that we enqueue `measure_worker.measure_trial_coverage` at this point! 

>I'll look into using `depends_on` because it seems more elegant than my current approach. 

SG, thanks!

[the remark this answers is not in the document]
jonathanmetzman: Sorry, I left out an important part of this sentence, we won't create a task for measuring snapshot N+1 if N hasn't been measured.

jonathanmetzman:
For the record, I'm merging this into a branch called queue, that is different than master. I'm going to test this a few times before merging it into master.

PR #371, opened 2020-05-23
Add pull-clang as depdendency of base-builder.

jonathanmetzman:
Otherwise, a stale clang might be used that is different from the one
used in CI and in experiments.

### DIFF
--- a/docker/build.mk
+++ b/docker/build.mk
@@ -39,15 +39,15 @@ base-image:
 pull-base-image:
 	docker pull $(BASE_TAG)/base-image
 
-base-builder: base-image
+pull-base-clang:
+	docker pull gcr.io/oss-fuzz-base/base-clang
+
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
I haven't verified this, but I think there will be issues if we don't do this (based on what I know about docker).

jonathanmetzman:
I think we should get rid of the clear-cache build target as it doesn't only affect fuzzbench. THoughts?

jonathanmetzman:
Btw, I suspect this would have prevented the issues laurent saw.

inferno-chromium:
> I think we should get rid of the clear-cache build target as it doesn't only affect fuzzbench. THoughts?

Can we match images name in there ? It does not hurt, it is a nice helper command for clearing mess and we can ask folks to do when they ask us questions.

PR #372, opened 2020-05-23
Create an image for the measurer's worker and add steps to build it.

jonathanmetzman:
Also:
Add .dockerignore
Add some more files to .gitignore


jonathanmetzman:
This PR depends on https://github.com/google/fuzzbench/pull/369

.dockerignore
```diff
@@ -0,0 +1,40 @@
...
+*.py[cod]
+*$py.class
+
+.pytype/
+
+# Virtual environments.
+.env
+.venv
+env/
+venv/
```
lszekeres, line 28: Don't we only have .venv ?

docker/measure-worker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# limitations under the License.
+
+FROM gcr.io/fuzzbench/base-runner
+
+ENV WORK /work
+WORKDIR $WORK
+
+ENV VIRTUALENV_DIR $WORK/.venv
+
+RUN mkdir -p /work && virtualenv --python=$(which python3) $VIRTUALENV_DIR
```
lszekeres, line 22: should we use the python3 venv instead? (https://docs.python.org/3/library/venv.html)

docker/build.mk
```diff
@@ -68,6 +68,15 @@ dispatcher-image: base-image
     $(call cache_from,${BASE_TAG}/dispatcher-image) \
     docker/dispatcher-image
 
+measure-worker: base-runner
+	docker build \
+    --tag $(BASE_TAG)/measure-worker \
```
lszekeres, line 73: super nit: docker build lines is not aligned

jonathanmetzman: I think I'm missing something because I can't tell what isn't aligned :-)
The first line uses a tab and each subsequent line uses four spaces, this is what we do in other targets also.
Could you maybe edit it using the online interface or specify how it should be alighned.?

docker/measure-worker/Dockerfile
jonathanmetzman: Sure, I'd rather keep this here to be consistent with the other Docker images and then switch them all at once in a later PR.  It seems like a low risk change so it would probably be fine to change them all here, but I don't see a problem with waiting to do this.

.dockerignore
jonathanmetzman: Yep. I deleted the other ones from the gitignore and here.

docker/measure-worker/Dockerfile
lszekeres: sgtm, thanks!

docker/build.mk
lszekeres: never mind then, looks like on the github code review view a tab is 6 spaces... :)

jonathanmetzman:
Meant to merge this into queue not master branch

Issue #267, opened 2020-04-24
Add an "experimental" report template


lszekeres:
This report would contain any analysis result that currently does not fit in the default report that we try to keep as concise as possible. These results can include alternative ranking techniques, statistical tests, and various other plots. The analysis library already contains alternative rankings, statistical significance tests, and plots that are not included in the default report template. Users will be able to generate this experimental report using the `--report-type` flag for `generate_report.py`. 
If a particular analysis/ranking/plot grains enough interest, it might even graduate into the default report.

Issue #366, opened 2020-05-21
default benchmark ranking is bad. seriously.


vanhauser-thc:
In https://www.fuzzbench.com/reports/2020-05-20-aflplusplus-2/index.html different afl++ variants were run.
The overall ranking is: default (simple aflplusplus), cmplog, laf, mopt

However when we measure the overall total blocks found:

|fuzzer|total bbs|
|----|----|
|aflplusplus_instrim|69899.0|
|aflplusplus|70521.5|
|aflplusplus_mopt2|70357.0|
|aflplusplus_mopt|69426.5|
|aflplusplus_cmplog|70817.5|
|aflplusplus_laf|71393.0|
|aflplusplus_laf_17|71147.0|

the ranking is: laf, laf17, cmplog, default, ...

When we look at % of maximum coverage achieved per target (so the best always receives 1.0 and the others less) and add them up:

|fuzzer|total % of bb per target|
|----|----|
|aflplusplus_instrim|19.095|
|aflplusplus|19.357|
|aflplusplus_mopt2|19.241|
|aflplusplus_mopt|18.918|
|aflplusplus_cmplog|20.248|
|aflplusplus_laf|20.339|
|aflplusplus_laf_17|20.124|

the ranking is: laf, cmplog, laf17, default, ...

IMHO the second ranking method is better than the first, however in this benchmark run they come to  similar ranking order.

If I compare that with the report ranks:
 * default (simple aflplusplus) is displayed the best, in both alternative rankings it is 4th place
 * mopt is in both alternative rankings the last place, in the report it is in the middle (4th place)
 * the best in the alternative rankings merely makes rank 3 in the report

Because of this  I always do my assessment in an excel sheet that I auto-generate from a copy-paste from the median coverages table in the report.

TLDR - what I tried to show is that the ranking method in the reports give a false impression what is good and what is bad.
IMHO that needs to be fixed.

wideglide:
I should probably understand @mboehme 's paper a bit better before commenting, because he would argue that we should also rank fuzzers based on their time to reach coverage X (#327) instead of coverage improvement in a fixed time budget.

But that still doesn't account for how to rank fuzzers based on performance across N number of benchmarks.  Should each benchmark receive equal weighting in the ranking?  For instance, how do you compare fairly when cmplog achieves better coverage on several benchmarks, but significantly worse coverage on sqlite3, vorbis and woff2?

[...]

PR #373, opened 2020-05-23
Add experimental report type.

lszekeres:
This contains multiple different top level ranking results.

Fixes https://github.com/google/fuzzbench/issues/267, and can be extended later with more analysis results.

### DIFF
--- a/analysis/generate_report.py
+++ b/analysis/generate_report.py
@@ -39,7 +39,7 @@ def get_arg_parser():
     parser.add_argument(
         '-t',
         '--report-type',
-        choices=['default'],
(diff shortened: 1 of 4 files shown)


jonathanmetzman:
This PR contributes to #366

lszekeres:
> LGTM. Thanks!
> Could you maybe change https://github.com/google/fuzzbench/blob/master/experiment/reporter.py#L33 so that the experimental report is also included in a fuzzbench report? I think this PR is fine for now, but it would be better if the service generated it instead of it only being an option to do manually.

Agreed! Will add it soon.

PR #374, opened 2020-05-24
Add new fuzzers to the list.

lszekeres:


docs/_data/fuzzers.yaml
```diff
@@ -13,6 +13,24 @@
 
 Fuzzers:
 
+- name: WEIZZ
```
jonathanmetzman, line 16: CC @andreafioraldi 
Are there plans to release the source code for this tool at some point?

vanhauser-thc: AFAIK andrea is still very busy so I quickly answer: https://github.com/andreafioraldi/weizz-fuzzer "Will be released after ISSTA 2020", so I guess end of July. Havent seen it so I am curious too :)

jonathanmetzman: Thanks!
(I assumed at least the same features would be upstreamed into AFL++ at some point).

jonathanmetzman:
Could you also add the qemu mode versions of AFL/AFL++?

docs/_data/fuzzers.yaml
vanhauser-thc: I think that is the plan, at least he wrote that in a chat. dunno when and which features though. he also has to write his thesis (and code and testing etc.) :)

lszekeres: Thanks @vanhauser-thc for the heads up!

andreafioraldi: weizz is hard to evaluate in fuzzbench, and its scope are only binary file formats. In produce invalid inputs in the queue (with patched checksums) and eventually fix them, but is not guaranteed so like T-Fuzz a tool that measure coverage fails to measure the coverage of weizz. In the paper I used the same patched qemu. However, I can consider to upstream parts of weizz in afl++ without checksums patchinng, just the mutator.
In that case we can add an afl++ conf in the future, but IMO not the weizz implementation used in the paper.

andreafioraldi:
> Could you also add the qemu mode versions of AFL/AFL++?

Would be super!

PR #376, opened 2020-05-24
Add a basic colab notebook

lszekeres:
Adds a basic example showing how to use analysis functions that are not included in the default report.

The notebook can be viewed here: https://colab.research.google.com/github/google/fuzzbench/blob/lszekeres/colab/analysis/notebooks/example.ipynb


analysis/notebooks/README.md
```diff
@@ -1,9 +1,6 @@
-# Example Jupyter / Google Colab Notebooks
+# Example Google Colab / Jupyter Notebooks
 
+- [![Open example.ipynb In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/google/fuzzbench/blob/master/analysis/notebooks/example.ipynb) A basic example of using analysis functions whose results are not included in the default report.
 
-## Ranking
+- [![Open ranking.ipynb In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/google/fuzzbench/blob/master/analysis/notebooks/ranking.ipynb) A simple example that compares the existing ranking methods available in the [data_utils](../data_utils.py) module.
```
jonathanmetzman, line 5: Is there any way to ensure that the notebooks don't get out of sync with the code they depend on.

jonathanmetzman, line 5: Maybe we should go further and include a link to a notebook for the experiment in the reports?

analysis/notebooks/README.md
```diff
@@ -1,9 +1,6 @@
-# Example Jupyter / Google Colab Notebooks
+# Example Google Colab / Jupyter Notebooks
```
jonathanmetzman, line 1: Maybe we should just put this document in the real docs? under https://google.github.io/fuzzbench/reference/report/ maybe?

analysis/notebooks/README.md
lszekeres: Yes! Let me do that in a following pr.

lszekeres: Yes, a typical way to validate notebooks is to execute them with the nbconvert command line tool, eg:

`jupyter nbconvert --ExecutePreprocessor.timeout=600 --to notebook --execute example.ipynb`

I tested this, it works, and I can add it in another PR. The only thing is that `pip install jupyter nbconvert` adds a large number of deps: https://github.com/google/fuzzbench/commit/9814a03db72c11d62f1a8c638bb61bd2b76a5617, so we might want to do this in a separated workflow. 

Let me know what you think!

There's also pytest plugin: "Testing with Jupyter notebooks: NoteBook VALidation (nbval) plug-in for pytest": https://arxiv.org/abs/2001.04808, https://github.com/computationalmodelling/nbval
however, I think this is is better for more unit test like tests. our notebook is more like an end-to-end test downloading the repo and sample data, etc.

analysis/notebooks/README.md
lszekeres: Added a subsection in custom_analysis_and_reports.md

PR #377, opened 2020-05-25
Fix target compilation, we had 10%+ speed loss ...

vanhauser-thc:


vanhauser-thc:
there is still a 5% performance loss ...

vanhauser-thc:
now it is ... ok. could be better.
I compiled afl++ without any optimization and the targets with over-optimization and hence the outcome was 20% lower than it should have been.
For some things I am still clueless. a binary that is generated with clang is better compiled than via afl-clang-fast, although the only difference are some -D__AFL defines and -Warnings stuff. stumps me.

vanhauser-thc:
OK this is done now.

I added a very special variant - shared memory fuzzing. This should result in ~ 150-200% speed in all targets.

If you could run aflplusplus, aflpluspus_optimal and aflplusplus_shmem and then combine it with 2020-05-24?
that would be amazing. That would finally be a real comparison of afl++ with the other fuzzers. After I fucked up the afl++ build for so long ... .... (I feel stupid)

PR #378, opened 2020-05-25
Add freeze-dependencies to Makefile.

lszekeres:
For updating requirements.txt after installing new dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,9 @@ ${VENV_ACTIVATE}: requirements.txt
 
 install-dependencies: ${VENV_ACTIVATE}
 
+freeze-dependencies:
+	source $(VENV_ACTIVATE) && python3 -m pip freeze > requirements.txt
+
 presubmit: install-dependencies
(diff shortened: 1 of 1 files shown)


lszekeres:
> nit: Will this be useful for other users ?

I believe so. (I personally never remember the pip command to do this)

jonathanmetzman:
I don't really think this is useful. Typically when I add a dependency (like rq for example) I'll only add that dependency to requirements.txt (i.e. `pip freeze | grep rq`) This would add all the dependencies of that dependency to requirements.txt

inferno-chromium:
> I don't really think this is useful. Typically when I add a dependency (like rq for example) I'll only add that dependency to requirements.txt (i.e. `pip freeze | grep rq`) This would add all the dependencies of that dependency to requirements.txt

ah right, forgot about that. @lszekeres - thoughts, feel free to revert.

lszekeres:
makes sense! let me revert it

PR #380, opened 2020-05-26
Show two types of aggregate rankings in default report.

lszekeres:
- Show ranking based on average normalized score.
- Show ranking based on average rank (previous default).
- Show critical difference diagram in a collapsible.
- Clean up pre-conditions for showing aggregate ranking.
- Experimental report shows all 6 types of aggregate ranking.

Fixes https://github.com/google/fuzzbench/issues/366.

### DIFF
--- a/analysis/report_templates/default.html
+++ b/analysis/report_templates/default.html
@@ -65,31 +65,62 @@
         <div id="summary" class="section scrollspy">
             <h2 class="green-text text-darken-3">experiment summary</h2>
 
+            {% if experiment.rank_by_median_and_average_rank.size < 2 %}
+
+            No aggregate ranking as the data contains a single fuzzer.
+
+            {% elif experiment.benchmarks|length < 2 %}
+
+            No aggregate ranking as the data contains a single benchmark.
+
+            {% else %}
+
             {% block top_level_ranking %}
 
-            {% if experiment.rank_by_mean_and_average_rank.size < 2 %}
-            No ranking as the available data is only for a single fuzzer.
+            We show two different aggreage (cross-benchmark) ranking of fuzzers.
+            The first is based on the average of per-benchmarks scores, where
+            the score represents the percentage of the highest reached median
+            coverage on a given benchmark (higher value is better).
 
-            {% elif experiment.rank_by_mean_and_average_rank.size > 20 %}
-            Showing cross-benchmark ranking as a table, because the critical
-            difference plot currently only supports up to 20 fuzzers. The table
-            shows the average rank (lower better) of fuzzers after ranking them
-            on each benchmark according to the median reached coverages.
+            The second ranking shows the average rank of fuzzers, after we rank
+            them on each benchmark according to their median reached covereges
+            (lower value is better).
+
             <div class="row">
-                <div class="col s7 offset-s3">
-                    {{ experiment.rank_by_mean_and_average_rank.to_frame().to_html() }}
+                <div class="col s5 offset-s1">
+                    <h5 class="center-align">By avg. score</h5>
+                    {{ experiment.rank_by_median_and_average_normalized_score.to_frame().to_html() }}
+                </div>
+                <div class="col s5">
+                    <h5 class="center-align">By avg. rank</h5>
+                    {{ experiment.rank_by_median_and_average_rank.to_frame().to_html() }}
                 </div>
             </div>
 
-            {% elif experiment.benchmarks|length < 2 %}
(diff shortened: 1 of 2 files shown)


lszekeres:
Looks like this:
image

analysis/report_templates/default.html
```diff
@@ -65,46 +65,84 @@ <h5 class="center-align brown-text text-darken-4">
...
+            currently only supports up to 20 fuzzers.
 
             {% else %}
-            Aggregate critical difference diagram showing average ranks when
-            ranking fuzzers on each benchmark according to their median reached
-            coverages. Critical difference is based on Friedman/Nemenyi post-hoc
-            test. See more in the <a href="https://google.github.io/fuzzbench/reference/report/">
-            documentation</a>.<br>
+
+            The diagram visualizes the second ranking while showing the
```
inferno-chromium, line 115: Can be more explicit on what this second ranking is, in case, people want to copy this text with image ?

jonathanmetzman:
> Looks like this:
> image

I think we should round those numbers to 2-3 decimal places? Or does it become hard to distinguish some cases if we do that?

analysis/report_templates/default.html
lszekeres: Good idea. Done

lszekeres:
> I think we should round those numbers to 2-3 decimal places? Or does it become hard to distinguish some cases if we do that?

Done.

jonathanmetzman:
Does it break if we only have 1 fuzzer/benchmark? Maybe don't special case these if it doesn't

analysis/report_templates/experimental.html
```diff
@@ -18,20 +18,31 @@
...
-<h5>Rank by median on benchmarks, then by avereage normalized score</h5>
+<h5>Rank by mean on benchmarks, then by average rank</h5>
 <div class="row">
     <div class="col s7 offset-s3">
-        {{ experiment.rank_by_median_and_average_normalized_score.to_frame().to_html() }}
+        {{ experiment.rank_by_mean_and_average_rank.to_frame().to_html() }}
+    </div>
+</div>
+
+<h5>Rank by pair-wise statistical test wins on benchmarks, then by average rank</h5>
```
jonathanmetzman, line 42: nit: "pairwise"

analysis/report_templates/experimental.html
```diff
@@ -18,20 +18,31 @@
...
-{{ super() }}
+<h5>Rank by median on benchmarks, then by avereage normalized score</h5>
+<div class="row">
+    <div class="col s7 offset-s3">
+        {{ experiment.rank_by_median_and_average_normalized_score.to_frame().to_html() }}
+    </div>
+</div>
 
-<h5>Rank by pair-wise statistical test wins on benchmarks, then by average rank</h5>
+<h5>Rank by median on benchmarks, then by average rank</h5>
```
jonathanmetzman, line 28: Do we need both mean and median? I vaguely remember we concluded mean was better because the distributions are nonnormal.

analysis/report_templates/experimental.html
```diff
@@ -18,20 +18,31 @@
 
 {% block top_level_ranking %}
 
-<h5>Rank by median on benchmarks, then by average rank (Critical difference)</h5>
-{{ super() }}
+<h5>Rank by median on benchmarks, then by avereage normalized score</h5>
```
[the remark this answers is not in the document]
lszekeres: Done

analysis/report_templates/experimental.html
lszekeres: done

analysis/report_templates/experimental.html
lszekeres: The idea was to put everything here just in case. But it might be too redundant, let me remove it.

lszekeres:
> Does it break if we only have 1 fuzzer/benchmark? Maybe don't special case these if it doesn't

I don't remember, i think the crit diff plot does. But in general this is only useful/meaningful if we have more fuzzers and benchmarks.

vanhauser-thc:
looks very good IMHO

lszekeres:
> looks very good IMHO

Thanks for suggesting it!

jonathanmetzman:
> > Does it break if we only have 1 fuzzer/benchmark? Maybe don't special case these if it doesn't
> 
> I don't remember, i think the crit diff plot does. But in general this is only useful/meaningful if we have more fuzzers and benchmarks.

Hopefully no one asks for this so I didn't create more work unnecessarily

PR #383, opened 2020-05-26
Create an image for the measurer's worker and add steps to build it. 

jonathanmetzman:
Also:
Add .dockerignore
Add some more files to and remove unnecessary patterns for virtualenvs from .gitignore.


jonathanmetzman:
Reland of https://github.com/google/fuzzbench/pull/372 into the right branch.

docker/measure-worker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM gcr.io/fuzzbench/base-runner
+
+ENV WORK /work
+WORKDIR $WORK
+
+ENV VIRTUALENV_DIR $WORK/.venv
```
inferno-chromium, line 20: nit: good to keep all ENVs at top.

docker/measure-worker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+RUN mkdir -p /work && virtualenv --python=$(which python3) $VIRTUALENV_DIR
```
inferno-chromium, line 22: /work -> $WORK

inferno-chromium: python3 -m venv ?

docker/measure-worker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+ENV SRC $WORK/src
+
+# Copy requirements.txt and install dependencies before copying all of the
+# source code so the cache hits more often.
+COPY requirements.txt $SRC/
+ENV PYTHONPATH=$SRC
+RUN /bin/bash -c "source $VIRTUALENV_DIR/bin/activate && \
+    pip3 install -r $SRC/requirements.txt"
```
inferno-chromium, line 31: just pip is ok, after you are in virtualenv ?

docker/measure-worker/Dockerfile
jonathanmetzman: I don't agree. I think defining variables close to where they are used is most readable.

docker/measure-worker/Dockerfile
jonathanmetzman: You mean why use a virtualenv at all in our docker images?

inferno-chromium: No i mean, once you active virtualenv, you dont need to do pip3, pip should point to pip3.

docker/measure-worker/Dockerfile
jonathanmetzman: >/work -> $WORK

Done.

>python3 -m venv ?

Spoke to laszlo about this in the original PR. I'll change all uses to venv at once rather than have this one be inconsistent.

docker/measure-worker/Dockerfile
jonathanmetzman: Ah got it.
Done.

jonathanmetzman:
Not sure why the AFL++ fuzzers are actually doing CI builds. I bet our diffing assumes merging into master and not a branch.

PR #386, opened 2020-05-27
Provide Python 3.7 in OSS-Fuzz builder image.

inferno-chromium:
This is at least needed for KLEE integration.


inferno-chromium:
@laurentsimon @rsprabery - fyi

PR #388, opened 2020-05-28
Remove make templates for fuzzers/benchmarks and support variants.

mbarbella-chromium:
PTAL when you have time. The variants part is fairly simple, but could you please confirm that these are the only places the build arguments are needed (search the templates for {build_arguments}). I would still like to test this a bit more locally before I'm comfortable landing it since it's a fairly invasive change to local builds.


mbarbella-chromium:
Any idea what we'd have to change to ensure that the yaml module is installed in CI, or to ensure that we're using the correct python environment for this? I'm not too familiar with the setup we're using here and CI is failing to generate the new makefile.

inferno-chromium:
Please fix travis failures.

Makefile
```diff
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+docker/generated.mk: docker/generate_makefile.py $(wildcard fuzzers/*/variants.yaml)
+	/usr/bin/env python3 $< > $@
```
lszekeres, line 16: If you'd like to make sure the yaml module to be installed when this runs, you can make this rule depend on `install-dependencies`, and do `source ${VENV_ACTIVATE} && python3 $< > $@` instead.

docker/generate_makefile.py
```diff
@@ -0,0 +1,327 @@
...
+
+FUZZER_BENCHMARK_TEMPLATE = """
+.{fuzzer}-{benchmark}-builder: .{fuzzer}-builder
+	docker build \\
+    --tag {base_tag}/builders/{fuzzer}/{benchmark} \\
+    --build-arg fuzzer={fuzzer} \\
+    --build-arg benchmark={benchmark} \\
+    $(call cache_from,{base_tag}/builders/{fuzzer}/{benchmark}) \\
+    --file docker/benchmark-builder/Dockerfile \\
+    {build_arguments} \\
```
inferno-chromium, line 48: I dont understand how this will just work, are these arguments to fuzzer.py or to docker build ?

lszekeres: +1

Would it be possible to do https://github.com/google/fuzzbench/pull/342#discussion_r426094426, i.e.:

have rules like:

{fuzzer}-{variant}-{benchmark}

and just pass:

--build-arg fuzzer=$(1) \
--build-arg variant=$(2) \
--build-arg benchmark=$(3) \

docker/generate_makefile.py
```diff
@@ -0,0 +1,327 @@
...
+    .
+
+.pull-{fuzzer}-{benchmark}-builder: .pull-{fuzzer}-builder
+	docker pull {base_tag}/builders/{fuzzer}/{benchmark}
+
+ifneq ({fuzzer}, coverage)
```
lszekeres, line 54: Consider doing logic like this in the generation script itself. I believe that'd be easier to read.

Makefile
```diff
@@ -12,7 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+docker/generated.mk: docker/generate_makefile.py $(wildcard fuzzers/*/variants.yaml) install-dependencies
```
lszekeres, line 15: to fix the infinite loop issue, replace `install-depenedencies` with `${VENV_ACTIVATE}`

docker/generate_makefile.py
mbarbella-chromium: Sorry for the late reply here, Abhishek and I discussed this in a VC last week. It shouldn't be necessary to separate out the variant from the fuzzer name. The variant names must already be unique based on how other pieces of this are implemented. In other parts of the code we handle things in a similar way.

As for the first question (just to have it documented) this will work because when a variant is defined we have both a fuzzer_environment section for environment variables set before calling into fuzzer.py (already implemented and not relevant to this change) and a separate build_arguments section, which is just a set of docker arguments passed to the builder. This was the most generic way we could handle things, and doesn't limit us to environment variables only. If we decide that we want to use the environment in both cases it would be a fairly small change and I'm open to it, but Jonathan suggested handling it this way a while back.

Makefile
mbarbella-chromium: Thanks! This works.

docker/generate_makefile.py
mbarbella-chromium: I'm not opposed to this but I'm honestly not sure that it's cleaner. If this is removed we need several subtemplates which also have to be formatted ahead of time and combined together, which feels complicated.

Any thoughts?

inferno-chromium:
Would let @lszekeres @jonathanmetzman review this!

docker/generate_makefile.py
lszekeres: You might consider having one template per docker image type for this, but of course if it just further complicates things then never mind.

docker/generate_makefile.py
```diff
@@ -0,0 +1,327 @@
...
+
+        with open(variants_path) as file_handle:
+            config = yaml.load(file_handle, yaml.SafeLoader)
+            assert 'variants' in config
+
+        for variant in config['variants']:
+            assert 'name' in variant
+            variant_name = variant['name']
+            # Build arguments are optional for variants.
+            build_arguments = variant.get('build_arguments', [])
```
lszekeres, line 309: Consider showing an example build argument in the comment to help readers understand what they are.

docker/generate_makefile.py
lszekeres: Thanks!

(Left another comment on the script where `build_argument` is created to help readers understand how this works)

docker/generate_makefile.py
mbarbella-chromium: It would complicate updates to the portions outside of the if/else as they'd need to be updated twice. I think I'll leave it as-is, but if there would ever be cases where more things are handled conditionally I think it would make sense to refactor this to actually include the logic in python rather than make.

docker/generate_makefile.py
mbarbella-chromium: Good idea. Done!

docker/generate_makefile.py
lszekeres: Sounds good thanks

jonathanmetzman: I prefer breaking up the logic so that we could have one source of truth for GCB yaml files and make. But I'm with landing this for now. Probably better to figure out what that logic looks like when we generate the yaml files as well, rather than ahead of time.

inferno-chromium: +1, landing this.

mbarbella-chromium:
Thanks for the reviews!

PR #391, opened 2020-05-30
afl++ optimal shmem variant

vanhauser-thc:
The shmem variants performed good in some variants but showed no progress in many.
This attempt tries to combines LTO and shmem to see where the issue for that could be.
In theory this new variant should be the best variant. Where it will not perform I can then debug where the issue might be.

### DIFF
--- a/.github/workflows/fuzzers.yml
+++ b/.github/workflows/fuzzers.yml
@@ -19,6 +19,7 @@ jobs:
           - aflplusplus
           - aflplusplus_optimal
           - aflplusplus_shmem
+          - aflplusplus_optimal_shmem
           - aflsmart
           - ankou
           - eclipser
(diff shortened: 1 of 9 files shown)


jonathanmetzman:
I assume the build (run) isn't supposed to fail right?

vanhauser-thc:
nah isn't :)

I found a way to debug the shmem issue on my system, no need for this lto shmem variant at the moment. once I have fixed the shmem, there will be an update here plus the other shmem variant.

vanhauser-thc:
yes all builds work and shmem is fixed :)
And basically all aflpluplus instances have small to big enhancements.

if somewhere during in the week you can start a run with all aflplusplus* that would be great, and if you can add afl + honggfuzz to have a direct comparison, that would be amazing :)

PR #393, opened 2020-06-01
Support running measure_worker on different machines than the manager

jonathanmetzman:
To do this, make each worker:
1. Set up the coverage binary for the benchmark it is measuring.
2. Save state such as measured-files and covered-pcs to GCS and associating it with the cycle.
Having a statefile per trial/cycle instead of per trial lets us save measurement in a non-atomic way.
So if we save the covered-pcs but get preempted and don't get to save the snapshot to the db,
measurement of the trial can continue from the same cycle.


jonathanmetzman:
Instead of having a measured-files-0001.json, measured-files-0002.json... file for each cycle I suppose we could have one json file where cycles are keys and the value is the measured-files for that cycle. Does this seem like a better design?

jonathanmetzman:
Ping

common/experiment_utils.py
```diff
@@ -59,14 +59,23 @@ def get_trial_instance_name(experiment: str, trial_id: int) -> str:
...
+    filename = basename + '-' + ('%04d' % cycle)
+    if suffix is None:
+        return filename
+    return filename + suffix
+
+
 def get_corpus_archive_name(cycle: int) -> str:
     """Returns a corpus archive name given a cycle."""
-    return 'corpus-archive-%04d.tar.gz' % cycle
+    return get_cycle_file_name('corpus-archive', cycle, '.tar.gz')
```
mbarbella-chromium, line 73: Would these files ever be anything other than .tar.gz archives? I think it's probably fine to leave it as-is, but it might be simpler to always append the extension and drop the suffix argument if so.

experiment/measurer/measure_worker.py
```diff
@@ -78,6 +80,49 @@ def extract_corpus(corpus_archive: str, sha_blacklist: Set[str],
...
+        self.cycle = cycle
+        self._prev_state = None
+
+    def _get_gcs_cycle_state_file_path(self, cycle: int) -> str:
+        state_file_name = experiment_utils.get_cycle_file_name(
+            self.name, cycle, '.json')
+        state_file_path = os.path.join(self.state_dir, state_file_name)
+        return exp_path.gcs(pathlib.Path(state_file_path))
+
+    def _get_previous_cycle_state(self, first_cycle_default) -> str:
```
mbarbella-chromium, line 100: Optional nit: It also seems unlikely that we'd ever want anything besides an empty list as the value for our default in the first cycle case. Should that also be omitted?

experiment/measurer/measure_worker.py
```diff
@@ -231,26 +273,40 @@ def archive_crashes(self, cycle):
...
 
-    def get_measured_files(self):
-        """Returns a the set of files that have been measured for this
-        snapshot's trials."""
-        if not os.path.exists(self.measured_files_path):
-            return set()
-        return set(filesystem.read(self.measured_files_path).splitlines())
+    def update_state_for_unchanged_cycle(self, cycle):
+        """Update the  covered-pcs and  measured-files state  files so  that the
+        states for |cycle| are the same as |cycle -1|."""
```
mbarbella-chromium, line 290: Nit: "- 1"

experiment/measurer/measure_worker.py
inferno-chromium: +1. this first_cycle_Default param is confusing.

experiment/measurer/measure_worker.py
```diff
@@ -134,37 +170,22 @@ def run_cov_new_units(self):
...
+        prev_pcs = self.get_prev_covered_pcs(cycle)
+        covered_pcs_state = self.get_covered_pcs_state(cycle)
+        sancov_files = glob.glob(os.path.join(self.sancov_dir, '*.sancov'))
+        if not sancov_files:
+            self.logger.error('No sancov files.')
+            return list(prev_pcs)
+
+        self.logger.info('Sancov files: %s.', str(sancov_files))
+        new_pcs = set(sancov.GetPCs(sancov_files))
+        all_pcs = list(sorted(prev_pcs.union(new_pcs)))
```
inferno-chromium, line 184: i think list should come first, then sorted ? Also, this can be put inside set_current to avoid caller forgetting the order.

experiment/measurer/measure_worker.py
```diff
@@ -134,37 +170,22 @@ def run_cov_new_units(self):
...
-            file_handle.seek(0)
-            file_handle.write('\n'.join(all_pcs))
+    def merge_new_pcs(self, cycle: int) -> List[str]:
+        """Merge new pcs into and return the list of all covered pcs."""
+        prev_pcs = self.get_prev_covered_pcs(cycle)
+        covered_pcs_state = self.get_covered_pcs_state(cycle)
+        sancov_files = glob.glob(os.path.join(self.sancov_dir, '*.sancov'))
+        if not sancov_files:
+            self.logger.error('No sancov files.')
+            return list(prev_pcs)
```
inferno-chromium, line 180: Does this need be sorted ?

experiment/measurer/measure_worker.py
```diff
@@ -231,26 +273,40 @@ def archive_crashes(self, cycle):
...
+    def update_state_for_unchanged_cycle(self, cycle):
+        """Update the  covered-pcs and  measured-files state  files so  that the
+        states for |cycle| are the same as |cycle -1|."""
+        state_files = [
+            self.get_covered_pcs_state(cycle),
+            StateFile('measured-files', self.state_dir, cycle)
+        ]
+        for state_file in state_files:
+            prev_state = state_file.get_previous(first_cycle_default=[])
+            state_file.set_current(prev_state)
```
inferno-chromium, line 297: Why we do need to set state for unchanged cycle, why we can't use a bool ?

experiment/measurer/measure_worker.py
```diff
@@ -78,6 +80,49 @@ def extract_corpus(corpus_archive: str, sha_blacklist: Set[str],
...
+    particular cycle. Objects of this class are backed by files stored in the
+    bucket."""
+
+    def __init__(self, name: str, state_dir: str, cycle: int):
+        self.name = name
+        self.state_dir = state_dir
+        self.cycle = cycle
+        self._prev_state = None
+
+    def _get_gcs_cycle_state_file_path(self, cycle: int) -> str:
```
lszekeres, line 94: nit: does "gcs" needs to be in the function name, or could we start making these function name generic, so we won't have to rename them when we have local support?

experiment/measurer/measure_worker.py
```diff
@@ -78,6 +80,49 @@ def extract_corpus(corpus_archive: str, sha_blacklist: Set[str],
...
+    def get_previous(self, first_cycle_default=None):
+        """Returns the previous state."""
+        if self._prev_state is None:
+            self._prev_state = self._get_previous_cycle_state(
+                first_cycle_default=first_cycle_default)
+
+        return self._prev_state
+
+    def set_current(self, state):
+        """Sets the state for this cycle on GCS."""
```
lszekeres, line 118: consider referring to the directory instead, ie: "on GCS" -> "in the experiment data directory" (or something like that)

jonathanmetzman: Changed to "bucket"

experiment/measurer/measure_worker.py
jonathanmetzman: I changed name to bucket.

experiment/measurer/measure_worker.py
jonathanmetzman: Because then we would have to retrieve the previous file's PCs. And then what if that's also unchanged? We could end up doing 20 network requests instead of one. A bool is more space efficient but I think it is not as good.
I think a bool would work well if it was all one file though as I mentioned it could be here:
>Instead of having a measured-files-0001.json, measured-files-0002.json... file for each cycle I suppose we could have one json file where cycles are keys and the value is the measured-files for that cycle. Does this seem like a better design?

WDYT of that solution vs multiple files as I have now?

experiment/measurer/measure_worker.py
jonathanmetzman: No. Nothing actually needs sorting, I just do it to make analysis afterwards easier since the files will be the same if the PCs are unchanged if they are sorted.

experiment/measurer/measure_worker.py
jonathanmetzman: I got rid of `list` an inner list isn't needed because sorting a set returns a list.
Why put it in `set_current` if we are returning it?

experiment/measurer/test_measure_worker.py
```diff
@@ -57,39 +58,47 @@ def db_experiment(experiment_config, db):
...
-def test_merge_new_pcs(new_pcs, fs, experiment):
+@pytest.mark.parametrize('new_pcs', [{'0x1', '0x2'}, set()])
+@mock.patch('common.gsutil.cp')
+@mock.patch(
+    'experiment.measurer.measure_worker.SnapshotMeasurer.get_prev_covered_pcs')
+def test_merge_new_pcs(mocked_get_prev_covered_pcs, mocked_cp, new_pcs, fs,
+                       experiment):
     """Tests that merge_new_pcs merges new PCs, and updates the covered-pcs
     file."""
+    prev_pcs = {"0x425221"}
```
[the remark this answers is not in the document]
jonathanmetzman: Done.

experiment/measurer/test_measure_worker.py
```diff
@@ -57,39 +58,47 @@ def db_experiment(experiment_config, db):
...
 
-    covered_pcs_filename = get_test_data_path('covered-pcs.txt')
-    fs.add_real_file(covered_pcs_filename, read_only=False)
-    snapshot_measurer.sancov_dir = os.path.dirname(covered_pcs_filename)
-    snapshot_measurer.covered_pcs_filename = covered_pcs_filename
+    mocked_cp.side_effect = mock_cp
 
-    with open(covered_pcs_filename) as file_handle:
-        initial_contents = file_handle.read()
+    cycle = 2
```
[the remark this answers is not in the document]
jonathanmetzman: Done.

experiment/measurer/measure_worker.py
```diff
@@ -78,6 +80,49 @@ def extract_corpus(corpus_archive: str, sha_blacklist: Set[str],
...
+        return exp_path.gcs(pathlib.Path(state_file_path))
+
+    def _get_previous_cycle_state(self, first_cycle_default) -> str:
+        if self.cycle == 1:
+            return first_cycle_default
+
+        previous_state_file_gcs_path = self._get_gcs_cycle_state_file_path(
+            self.cycle - 1)
+        return json.loads(
+            gsutil.cat(previous_state_file_gcs_path, expect_zero=False))
```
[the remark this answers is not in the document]
jonathanmetzman: It exceptions, no need.

experiment/measurer/measure_worker.py
```diff
@@ -78,6 +80,49 @@ def extract_corpus(corpus_archive: str, sha_blacklist: Set[str],
...
+        self._prev_state = None
+
+    def _get_gcs_cycle_state_file_path(self, cycle: int) -> str:
+        state_file_name = experiment_utils.get_cycle_file_name(
+            self.name, cycle, '.json')
+        state_file_path = os.path.join(self.state_dir, state_file_name)
+        return exp_path.gcs(pathlib.Path(state_file_path))
+
+    def _get_previous_cycle_state(self, first_cycle_default) -> str:
+        if self.cycle == 1:
```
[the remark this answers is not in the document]
jonathanmetzman: Done.

experiment/measurer/measure_worker.py
```diff
@@ -78,6 +80,49 @@ def extract_corpus(corpus_archive: str, sha_blacklist: Set[str],
...
+    bucket."""
+
+    def __init__(self, name: str, state_dir: str, cycle: int):
+        self.name = name
+        self.state_dir = state_dir
+        self.cycle = cycle
+        self._prev_state = None
+
+    def _get_gcs_cycle_state_file_path(self, cycle: int) -> str:
+        state_file_name = experiment_utils.get_cycle_file_name(
```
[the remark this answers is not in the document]
jonathanmetzman: Done.

common/experiment_utils.py
```diff
@@ -59,14 +59,23 @@ def get_trial_instance_name(experiment: str, trial_id: int) -> str:
     return 'r-%s-%d' % (experiment, trial_id)
 
 
+def get_cycle_file_name(basename, cycle: int, suffix: str = None) -> str:
```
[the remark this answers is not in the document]
jonathanmetzman: Done.

experiment/measurer/measure_worker.py
jonathanmetzman: Done.

experiment/measurer/measure_worker.py
jonathanmetzman: I now return an empty list on first cycle. I think the first_cycle_default thing would be necessary if these methods ever returned anything besides a list, but since they don't in practice the empty list is fine.

common/experiment_utils.py
jonathanmetzman: Done.

PR #395, opened 2020-06-03
Don't delete entire venv when requirements.txt is updated.

lszekeres:
Install new and upgrade existing packages instead.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,9 +18,8 @@ SHELL := /bin/bash
 VENV_ACTIVATE := .venv/bin/activate
 
 ${VENV_ACTIVATE}: requirements.txt
-	rm -rf .venv
 	python3 -m venv .venv
-	source ${VENV_ACTIVATE} && python3 -m pip install -r requirements.txt
+	source ${VENV_ACTIVATE} && python3 -m pip install --upgrade -r requirements.txt
 
 install-dependencies: ${VENV_ACTIVATE}
 


Makefile
```diff
@@ -18,9 +18,8 @@ SHELL := /bin/bash
 VENV_ACTIVATE := .venv/bin/activate
 
 ${VENV_ACTIVATE}: requirements.txt
-	rm -rf .venv
 	python3 -m venv .venv
```
inferno-chromium, line 21: Will this recreate .venv or use existing one.

lszekeres: It reuses the existing directories/files, but also recreates or updates the timestamp on `.venv/bin/activate`, so it seems to be doing exactly what we need.

Makefile
```diff
@@ -18,9 +18,8 @@ SHELL := /bin/bash
...
-	source ${VENV_ACTIVATE} && python3 -m pip install -r requirements.txt
+	source ${VENV_ACTIVATE} && python3 -m pip install --upgrade -r requirements.txt
```
jonathanmetzman, line 22: Why did we add this?

jonathanmetzman:
This caused me an issue which I was able to resolve by deleting .venv:
```
Remove .venv before install requirements again. Not doing this
gave me the following error:
```
python3 -m venv .venv
The virtual environment was not created successfully because
ensurepip is not available.  On Debian/Ubuntu systems, you need
to install the python3-venv package using the following command.

    apt-get install python3-venv

You may need to use sudo with that command.
After installing the python3-venv
package, recreate your virtual environment.

Failing command: ['/home/user/fuzzbench/.venv/bin/python3', '-Im',
'ensurepip', '--upgrade', '--default-pip']

Makefile:16: docker/generated.mk: No such file or directory
make: *** [Makefile:22: .venv/bin/activate] Error 1
```

But running `apt-get install python3-venv` did not solve the problem.
Instead, rm-ing .venv fixed it. Weirdly, it doesn't seem to happen again.
I might undo this to be safe if we can't figure out what happened.

jonathanmetzman:
I wonder if this has anything to do with me having activate .venv already before running `make presubmit`?

Makefile
lszekeres: Ah, sorry the --upgrade flag may not be necessary.

lszekeres:
I was able to repro this after syncing to master and run make presubmit, but then it couldn't reproduce it again by trying to add new packages, etc. Let me know if you can reproduce it again!

lszekeres:
Leaving this here for reference if this comes up again 
https://stackoverflow.com/questions/39539110/pyvenv-not-working-because-ensurepip-is-not-available

jonathanmetzman:
Thanks, wow I never would have solved this without googling.

PR #396, opened 2020-06-03
AFL++ Qemu mode

andreafioraldi:
let's try QEMU mode in fuzzbench, the chosen configuration is floating-point CompCov + persistent mode

Issue #385, opened 2020-05-26
Ping clang version used in standard benchmarks.


jonathanmetzman:
Having clang update automatically makes it necessary to rerun experiments. The fact that it is updated silently is especially bad. 
Instead we should use a specific version of the gcr.io/oss-fuzz/base-clang image so that we have to explicitly roll clang.

vanhauser-thc:
also it is just a limited package. e.g. llvmgold is not there (at least by default)

inferno-chromium:
@vanhauser-thc - that can be added with a PR here - https://github.com/google/fuzzbench/blob/master/docker/base-builder/Dockerfile#L35 ?

jonathanmetzman:
> @vanhauser-thc - that can be added with a PR here - https://github.com/google/fuzzbench/blob/master/docker/base-builder/Dockerfile#L35 ?

Maybe not, it depends whether it is built [here](https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-clang/checkout_build_install_llvm.sh)

I've avoided building clang directly during our build process because I think it would be bad UX for a first time user to have to build clang before building their fuzzers. I'll probably start with pinning the version we copy and then switch to building it directly in our build process. Make will probably pull clang instead of building it by default.

vanhauser-thc:
for my needs also lld is missing etc. - hence I install the nightly llvm 11-dev build in the aflplusplus_lto dockerfile (and the lld needs to have a specific option compiled in. seems to be default though). 
But yes, the dockerfile could be expanded.

but why fixing the version? a newer version might make the fuzzer better or worse. A new feature might be available that a fuzzer can then use.

similar topic - gcc-10 would also be a good idea in the image. the old version in the image is not good for the performance of the fuzzers.

jonathanmetzman:
>but why fixing the version? a newer version might make the fuzzer better or worse. A new feature might be available that a fuzzer can then use.

Basically, I'm worried (perhaps I shouldn't be) that it is not fair to compare fuzzers that used one version to compile targets to fuzzers that use a different version. 
To account for this, we rerun the same experiment pretty frequently, but this is wasteful IMO. 
For example, suppose I run a full experiment today. Then the next day clang rolls and you give me a new fuzzer to compare to the ones I ran the day before. Now I need to rerun everything.

jonathanmetzman:
>a newer version might make the fuzzer better or worse. A new feature might be available that a fuzzer can then use.

[...]

Issue #304, opened 2020-05-05
2020-05-04-preempt-new-bench discussion


jonathanmetzman:
I think there will be a fair amount to say about 2020-05-04-preempt-new-bench because it has new benchmarks and uses preemptible instances so I'm doing it here to not spam the announcement thread.

jonathanmetzman:
tcpdump benchmark seems to have no results. It looks like that's because the runners are running out of diskspace. The jasper benchmark also is running into this error (unfortunate since I think the results were pretty interesting from that benchmark).

roachspray:
Do you think this is something environmental or specific to the benchmark(s)? The tcpdump should be, essentially, mimicking `-n -r @@` arguments to tcpdump, should not be writing to disk. I would need to double check with jasper.

vanhauser-thc:
Link to the report being generated:
https://www.fuzzbench.com/reports/2020-05-04-preempt-new-bench/index.html

jonathanmetzman:
> Do you think this is something environmental or specific to the benchmark(s)? The tcpdump should be, essentially, mimicking `-n -r @@` arguments to tcpdump, so should not be writing to disk. I would need to double check with jasper.

Not really sure I understand what you mean by environmental. But other benchmarks generally don't have this issue so I think it must be something about the benchmarks.

jonathanmetzman:
Running the perl benchmark locally it looks like coverage stays pretty low. Benchmark probably needs some seeds and/or dictionary to demonstrate differences in fuzzer ability.

roachspray:
> > Do you think this is something environmental or specific to the benchmark(s)? The tcpdump should be, essentially, mimicking `-n -r @@` arguments to tcpdump, so should not be writing to disk. I would need to double check with jasper.
> 
> Not really sure I understand what you mean by environmental. But other benchmarks generally don't have this issue so I think it must be something about the benchmarks.

For both tcpdump and jasper, my response ignored the `buf_to_file()` write to disk. This takes mutated input bytes and writes them to a write that would then be read by the target library. It should be cleaning that up... re-reading the code now.

jonathanmetzman:
> > > Do you think this is something environmental or specific to the benchmark(s)? The tcpdump should be, essentially, mimicking `-n -r @@` arguments to tcpdump, so should not be writing to disk. I would need to double check with jasper.
> > 
> > 
[...]

PR #400, opened 2020-06-04
Remove tcpdump benchmark

jonathanmetzman:
@roachspray This is causing failures in https://github.com/google/fuzzbench/pull/396 so I'm removing it at least temporarily. I may preemptively (temporarily) remove the other benchmarks as well since they don't work in production as far as I can remember.

I'm not sure why this started failing, but I suspect it may be because the compiler changed. I'm thinking of pinning the compiler so this won't happen: https://github.com/google/fuzzbench/issues/385.


jonathanmetzman:
See #304 for where I mention that this benchmark is broken in production.

roachspray:
Thank you. I will still plan to look into it, just been short on time here.

PR #399, opened 2020-06-04
[NO MERGE] experimental pure local running support

zchcai:
Hi, this branch is an initial working plan for a pure running support in Linux hosts. (No GCS needed)

However, this support involves a lot of files to change, and we are trying to refactor and split this PR into several PRs to merge into `master` branch.

This PR is not for merging, but a reminder for current active PRs for local support. They are:
- refactor: `filetore_utils` interface for `gsutil`. In future, we can add `local_utils` or other `<file_storage>_utils`. #394
- refactor, minor change: support new configuration options, and rename most `cloud_*` (#402)
- `*gcs*` related path conversion: it will not affect functional behavior and thus no change now.
- new feature: add `local_utils` (for linux) and related unit tests. (#405)
- add `rsync` support #406 
- add some error handling logic: this is related to SQL distinct, fixed in #444.
- local dispatcher logic change
- final integration switching on: #422 (filestore_utils), #440 (local rsync fix), #441 (set env for local running), #444 (fix SQL error and enable end-to-end run).

PR #402, opened 2020-06-05
[refactor] rename cloud*; modified config requirements

zchcai:
This PR is the second PR for #399: rename most `cloud_*_bucket`.

Though, for non-local running, these two options, `cloud_experiment_bucket` and `cloud_web_bucket`, still work, we suggest to use `experiment_filestore` and `web_filestore` instead.

- refactor: rename `cloud_experiment_bucket` to `experiment_filestore`.
- refactor: rename `cloud_web_bucket` to `web_filestore`.
- new: `experiment_filestore` and `web_filestore` are only required when `local_experiment` is set as `True`.
- new: path format checking.
- modify: `cloud_experiment_bucket` and `cloud_web_bucket` are only required when
  - (1) `local_experiment` is not set or `False`.
  - (2) `experiment_filestore` or `web_filestore` is not set respectively.
- new & modify: related unit tests and test data files.

- Note: after this PR, partial local support #8 will not work (because of unmatched path format), until a future pure local support is finished.
- Note: after this PR, prefer to use `experiment_filestore` and `web_filestore`, instead of `cloud_experiment_bucket` and `cloud_web_bucket`.
- Note: I suggest to change related documentation after a whole local support finishes. Currently, no need to change for documentation because `cloud_experiment_bucket` and `cloud_web_bucket` still work.


experiment/run_experiment.py
```diff
@@ -66,12 +66,27 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
     # it isn't like anything expensive happens before this validation is done so
     # rerunning it is cheap.
     config = yaml_utils.read(config_filename)
-    bucket_params = {'cloud_experiment_bucket', 'cloud_web_bucket'}
-    string_params = {
-        'cloud_compute_zone', 'cloud_experiment_bucket', 'cloud_web_bucket'
-    }
+    cloud_bucket_params = {'cloud_experiment_bucket', 'cloud_web_bucket'}
```
jonathanmetzman, line 69: Why are these still called bucket? I think we shouldn't support `cloud_{experiment,web}_bucket` anymore, users will specify a `{experiment,cloud}_filestore` that will be used for local and cloud experiments.

experiment/run_experiment.py
```diff
@@ -95,11 +110,24 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
...
-            logs.error(
-                'Config parameter "%s" is "%s". It must start with gs://.',
-                param, value)
+        if param in filestore_params:
+            # Checks file storage paths format respectively.
+            if not config.get('local_experiment', False):
+                if not value.startswith('gs://'):
+                    valid = False
+                    logs.error(
+                        'Config parameter "%s" is "%s".' \
```
jonathanmetzman, line 119: I generally avoid backslashes.
They aren't needed in this case, Python will join the strings.

experiment/run_experiment.py
```diff
@@ -95,11 +110,24 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
...
             continue
 
-        if param in bucket_params and not value.startswith('gs://'):
-            valid = False
-            logs.error(
-                'Config parameter "%s" is "%s". It must start with gs://.',
-                param, value)
+        if param in filestore_params:
+            # Checks file storage paths format respectively.
+            if not config.get('local_experiment', False):
```
jonathanmetzman, line 115: I generally find very nested code hard to read. Let's restructure this:
```
if config.get('local_experiment', False):
  if not value.startswith('/'):
     valid = False
     logs.error(...)
   continue
if not value.startswith('gs://'):
  valid = False
  logs.error(...)
  continue
```

experiment/run_experiment.py
```diff
@@ -66,12 +66,27 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
...
     int_params = {'trials', 'max_total_time'}
-    required_params = int_params.union(string_params)
+    required_params = int_params.union(filestore_params)
+
+    if not config.get('local_experiment', False):
+        required_params = required_params.union(cloud_config)
+
+    # Searches alternatives for file storages settings.
+    if 'experiment_filestore' not in config:
+        assert 'cloud_experiment_bucket' in config, \
```
jonathanmetzman, line 82: I purposely don't use `assert` here and instead do `valid=False` so that we can report more than one error at a time. Can you change please?

experiment/test_run_experiment.py
```diff
@@ -61,7 +61,21 @@ def setUp(self):
     def test_missing_required(self, mocked_error):
         """Tests that an error is logged when the config file is missing a
         required config parameter."""
-        # All but cloud_compute_zone.
+        # All but trials
+        del self.config['trials']
+        with mock.patch('common.yaml_utils.read') as mocked_read_yaml:
+            mocked_read_yaml.side_effect = lambda config_filename: self.config
```
jonathanmetzman, line 67: nit: You can just do `mocked_read_yaml.return_value = self.config`.

experiment/test_run_experiment.py
```diff
@@ -71,46 +85,75 @@ def test_missing_required(self, mocked_error):
             mocked_error.assert_called_with('Config does not contain "%s".',
                                             'cloud_compute_zone')
 
+    @mock.patch('common.logs.error')
+    def test_cloud_alternative_take_effect_but_wrong_format(self, mocked_error):
+        """Tests that an error is logged when no alternative settings."""
+        del self.config['experiment_filestore']
+        self.config['cloud_experiment_bucket'] = '/local'
+        with mock.patch('common.yaml_utils.read') as mocked_read_yaml:
+            mocked_read_yaml.side_effect = lambda config_filename: self.config
```
jonathanmetzman, line 94: Same comment here about `return_value`

experiment/test_run_experiment.py
```diff
@@ -71,46 +85,75 @@ def test_missing_required(self, mocked_error):
...
+            with pytest.raises(run_experiment.ValidationError):
+                run_experiment.read_and_validate_experiment_config(
+                    'config_file')
+                mocked_error.assert_called_with('Config parameter "%s" ' \
```
jonathanmetzman, line 98: I think something like this is nicer formatting:
```
mocked_error.assert_called_with(
  'Config parameter "%s"  is "%s".It must start with gs:// when running '
  'localy....')
```
Less lines

common/filestore_utils.py
```diff
@@ -22,7 +22,8 @@
 def _using_gsutil():
     """Returns True if using Google Cloud Storage for filestore."""
     try:
-        experiment_path_format = experiment_utils.get_cloud_experiment_path()
+        experiment_path_format = experiment_utils.get_experiment_filestore_path(
```
jonathanmetzman, line 25:
```
experiment_path-format = (
  experiment_utils.get_experiment_filestore_path())
```

experiment/measurer.py
```diff
@@ -441,7 +441,7 @@ def extract_corpus(self, corpus_archive_path) -> bool:
         return True
 
     def archive_crashes(self, cycle):
-        """Archive this cycle's crashes into cloud bucket."""
+        """Archive this cycle's crashes into file storage."""
```
jonathanmetzman, line 444: `filestore` instead of `file storage` please.

experiment/run_experiment.py
```diff
@@ -95,11 +110,24 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
...
-                'Config parameter "%s" is "%s". It must start with gs://.',
-                param, value)
+        if param in filestore_params:
+            # Checks file storage paths format respectively.
+            if not config.get('local_experiment', False):
+                if not value.startswith('gs://'):
+                    valid = False
+                    logs.error(
+                        'Config parameter "%s" is "%s".' \
+                        'It must start with gs:// when running not locally.',
```
jonathanmetzman, line 120: `not locally` -> `on google cloud`

experiment/test_run_experiment.py
```diff
@@ -71,46 +85,75 @@ def test_missing_required(self, mocked_error):
...
 
-    def test_invalid_bucket(self):
+    def test_invalid_local_filestore(self):
+        """Tests that an error is logged when the config file has a config
+        parameter that should be a local bucket but is not."""
+        self.config['local_experiment'] = True
+        self.config['experiment_filestore'] = '/user/test/folder'
+        self._test_invalid(
+            'web_filestore', 'gs://wrong-here',
+            'Config parameter "%s" is "%s".' \
```
jonathanmetzman, line 127: No backslash please, not needed.

experiment/test_run_experiment.py
```diff
@@ -71,46 +85,75 @@ def test_missing_required(self, mocked_error):
...
+        )
+
+    def test_invalid_cloud_bucket(self):
         """Tests that an error is logged when the config file has a config
         parameter that should be a GCS bucket but is not."""
         self._test_invalid(
-            'cloud_experiment_bucket', 'invalid',
-            'Config parameter "%s" is "%s". It must start with gs://.')
+            'experiment_filestore', 'invalid',
+            'Config parameter "%s" is "%s".' \
```
jonathanmetzman, line 136: No backslash please, not needed.

zchcai:
> Thanks!
> 
> > Note: I suggest to change related documentation after a whole local support finishes. Currently, no need to change for documentation because cloud_experiment_bucket and cloud_web_bucket still work.
> 
> Let's migrate to `filestore` all at once please. Don't support both please.

I mentioned this feature in the internal design plan. Please next time give me feedback on design plan first. Then we can have a discussion before coding.

To reserve these options, keep other users not to need to update their configurations. It also gives other users time for a transition.
Also, for refactoring, it should not change the outer behavior. For example, the current workable configuration file format.

Please do consider this feature again. It really took for a while to support this. Actually there is no harm for other things.

zchcai:
> Thanks!
> 
> > Note: I suggest to change related documentation after a whole local support finishes. Currently, no need to change for documentation because cloud_experiment_bucket and cloud_web_bucket still work.
> 
> Let's migrate to `filestore` all at once please. Don't support both please.

To reserve these options, keep other users not to need to update their configurations. It also gives other users time for a transition.
Also, for refactoring, it should not change the outer behavior. For example, the current workable configuration file format.

Please do consider this feature again.

lszekeres:
> * refactor: rename `cloud_experiment_bucket` to `experiment_filestore`.
> * refactor: rename `cloud_web_bucket` to `web_filestore`.

`web_filestore` is the directory where the reports will be generated, right? To make it less confusing in the local case, should we call it `reports_filestore` instead?

zchcai:
> Why are these still called bucket? I think we shouldn't support `cloud_{experiment,web}_bucket` anymore, users will specify a `{experiment,cloud}_filestore` that will be used for local and cloud experiments.

From this PR, users can use `experiment_filestore` and `web_filestore` instead of `cloud_experiment_bucket` and `cloud_web_bucket`, to configure the file storage places for storing experiment running data and final report. Here, `web_filestore` is for report generation usages.

Since this PR is a partial effort towards pure local running, the storage places will not only be cloud buckets, but also local folders. Thus `bucket` is renamed as `filestore`.

Finally, the `cloud_{experiment,web}_bucket` configuration options will be removed. This PR chooses to keep them because it does not plan to affect current users' configuration files.

Do we still want to completely remove these configuration options from this PR? Or leave it to a later time and do it with documentation changes together?

experiment/test_run_experiment.py
zchcai: `return_value` looks clearer in semantics than `side_effect`.
Just to confirm there will be no functionality loss for changing from `side_effect` to `return_value`. Is there any reason behind choosing `side_effect` at other similar test functions in https://github.com/google/fuzzbench/blob/master/experiment/test_run_experiment.py?

experiment/run_experiment.py
zchcai: I see. So I guess the following `todo` comment in this function actually means: (1) it wants to log errors in this function and (2) it should except early after this `read_and_validate_experiment_config` from the whole experiment view.

https://github.com/google/fuzzbench/blob/e52d53714c7bf4f371628af68ce12926252538b8/experiment/run_experiment.py#L62-L67

Thus, we choose `valid=False` here and no assertions, in order to log more errors. Please comment here if this reply hasn't got the point.

experiment/test_run_experiment.py
jonathanmetzman: In some cases, it makes sense to use side_effect, but I don't think there are any such cases in `test_run_experiment` I'll change the others now.

jonathanmetzman: https://github.com/google/fuzzbench/pull/409

jonathanmetzman:
> > * refactor: rename `cloud_experiment_bucket` to `experiment_filestore`.
> > * refactor: rename `cloud_web_bucket` to `web_filestore`.
> 
> `web_filestore` is the directory where the reports will be generated, right? To make it less confusing in the local case, should we call it `reports_filestore` instead? @jonathanmetzman wdyt?

That sounds good to me. Can come here or in another PR.

jonathanmetzman:
> > Why are these still called bucket? I think we shouldn't support `cloud_{experiment,web}_bucket` anymore, users will specify a `{experiment,cloud}_filestore` that will be used for local and cloud experiments.
> 
> From this PR, users can use `experiment_filestore` and `web_filestore` instead of `cloud_experiment_bucket` and `cloud_web_bucket`, to configure the file storage places for storing experiment running data and final report. Here, `web_filestore` is for report generation usages.
> 
> Since this PR is a partial effort towards pure local running, the storage places will not only be cloud buckets, but also local folders. Thus `bucket` is renamed as `filestore`.
> 
> Finally, the `cloud_{experiment,web}_bucket` configuration options will be removed. This PR chooses to keep them because it does not plan to affect current users' configuration files.
> 
> Do we still want to completely remove these configuration options from this PR? Or leave it to a later time and do it with documentation changes together?

Can come at a later time.

experiment/run_experiment.py
jonathanmetzman: The TODO should probably be removed because it's definitely more of a maybe. The TODO is saying that that this is complicated by the fact that tries to find as many errors as possible instead of quitting after finding the first. I'd like to keep this behavior for now. Could you please remove the TODO and do the same instead of asserting? Also could you please mention in the docstring for this fucnction that it tries to find as many errors as possible?

Thanks!

I think your understanding of what the function does based on the TODO is correct.

experiment/run_experiment.py
```diff
@@ -88,12 +69,13 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
...
 
-    if not config.get('local_experiment', False):
+    local_experiment = config.get('local_experiment', False)
+    if not local_experiment:
         required_params = required_params.union(cloud_config)
 
     valid = True
     if 'cloud_experiment_bucket' in config or 'cloud_web_bucket' in config:
-        logs.warning('"cloud_experiment_bucket" and "cloud_wen_bucket" are now '
+        logs.warning('"cloud_experiment_bucket" and "cloud_web_bucket" are now '
```
jonathanmetzman, line 78: `error` not `warning` please.

docs/running-your-own-experiment/running_an_experiment.md
```diff
@@ -63,10 +63,10 @@ cloud_project: $PROJECT_NAME
 cloud_compute_zone: $PROJECT_ZONE
 
 # The Google Cloud Storage bucket that will store most of the experiment data.
-cloud_experiment_bucket: gs://$DATA_BUCKET_NAME
+experiment_filestore: gs://$DATA_BUCKET_NAME
 
-# The bucket where HTML reports and summary data will be stored.
-cloud_web_bucket: gs://$REPORT_BUCKET_NAME
+# The filestore where HTML reports and summary data will be stored.
```
jonathanmetzman, line 68: Don't like this change, please change it to `Google Cloud Storage bucket`. It makes sense to talk about GCS here since this document is about running experiments in cloud.

conftest.py
```diff
@@ -77,6 +77,6 @@ def experiment(environ):  # pylint: disable=redefined-outer-name,unused-argument
     """Mock an experiment."""
     os.environ['WORK'] = '/work'
     os.environ['EXPERIMENT'] = 'test-experiment'
-    os.environ['CLOUD_EXPERIMENT_BUCKET'] = 'gs://experiment-data'
-    os.environ['CLOUD_WEB_BUCKET'] = 'gs://web-bucket'
+    os.environ['EXPERIMENT_FILESTORE'] = 'gs://experiment-data'
+    os.environ['REPORTS_FILESTORE'] = 'gs://web-bucket'
```
jonathanmetzman, line 81: @lszekeres @inferno-chromium Since we're changing this API do you think it makes sense to call it "reports_filestore" or "report_filestore" given that the "experiment_filestore" is singular?

experiment/run_experiment.py
```diff
@@ -95,11 +99,22 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
...
+        if param not in filestore_params:
+            continue
+
+        if local_experiment and not value.startswith('/'):
             valid = False
             logs.error(
-                'Config parameter "%s" is "%s". It must start with gs://.',
-                param, value)
+                'Config parameter "%s" is "%s".'
+                'Local running only supports unix file system.', param, value)
```
jonathanmetzman, line 109: nit: `Local experiments only support using Posix file systems as filestores.`

experiment/run_experiment.py
```diff
@@ -95,11 +99,22 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
...
+            continue
+
+        if not local_experiment and not value.startswith('gs://'):
+            valid = False
+            logs.error(
+                'Config parameter "%s" is "%s".'
+                'It must start with gs:// when running on google cloud.', param,
```
jonathanmetzman, line 116: nit: capitalize "Google Cloud" please.

experiment/run_experiment.py
```diff
@@ -262,17 +277,17 @@ def start_dispatcher(config: Dict, config_dir: str):
...
     def filter_file(tar_info):
         """Filter out unnecessary directories."""
         if FILTER_SOURCE_REGEX.match(tar_info.name):
             return None
         return tar_info
 
-    cloud_experiment_path = os.path.join(config['cloud_experiment_bucket'],
-                                         config['experiment'])
-    base_destination = os.path.join(cloud_experiment_path, 'input')
+    experiment_path = os.path.join(config['experiment_filestore'],
```
jonathanmetzman, line 288: nit: `experiment_filestore_path`

experiment/test_run_experiment.py
```diff
@@ -61,7 +61,21 @@ def setUp(self):
     def test_missing_required(self, mocked_error):
         """Tests that an error is logged when the config file is missing a
         required config parameter."""
-        # All but cloud_compute_zone.
+        # All but trials
```
jonathanmetzman, line 64: nit end with period.

experiment/test_run_experiment.py
```diff
@@ -61,7 +61,21 @@ def setUp(self):
...
+                run_experiment.read_and_validate_experiment_config(
+                    'config_file')
+            mocked_error.assert_called_with('Config does not contain "%s".',
+                                            'trials')
+
+    @mock.patch('common.logs.error')
+    def test_missing_required_cloud(self, mocked_error):
+        """Tests that an error is logged when the config file is missing a
+        required cloudconfig parameter."""
+        # All but cloud_compute_zone
```
jonathanmetzman, line 78: nit end with period.

experiment/test_run_experiment.py
```diff
@@ -76,41 +90,50 @@ def test_invalid_upper(self):
...
         parameter that should be a string but is not."""
         self._test_invalid(
-            'cloud_experiment_bucket', 1,
+            'experiment_filestore', 1,
             'Config parameter "%s" is "%s". It must be a lowercase string.')
 
-    def test_invalid_bucket(self):
+    def test_invalid_local_filestore(self):
+        """Tests that an error is logged when the config file has a config
+        parameter that should be a local bucket but is not."""
```
jonathanmetzman, line 105: nit: "filestore" not bucket.

conftest.py
inferno-chromium: report_filestore seems better since this is in the context of this one report generation.

zchcai: I can do this change in this PR.

zchcai: Done.

lszekeres: Good point! We store the data of multiple experiments in the former and multiple reports in the latter, right? Singular might suggest otherwise, but having consistent names would be good indeed. So either experiments/reports_filestore, or experiment/report_filestore.

lszekeres: (sorry sent this from on old tab which didn't show the discussion above, but LGTM)

jonathanmetzman: I Sorta agree about both plural being ideal, but I think it doesn't matter as long as they are consistent.

PR #403, opened 2020-06-05
[scheduler] Scale measurers horizontally using queue-based arch

jonathanmetzman:
Make measure worker scheduling its own module: experiment/schedule_measure_worker.py
Also:
Use an n1-standard-32 for the Dispatcher.
Fix some errors building measure-worker on GCB.
Move queue initialization code to new module, common/queue_utils.py

### DIFF
--- a/common/gce.py
+++ b/common/gce.py
@@ -100,13 +100,14 @@ def get_instance_group_size(instance_group: str, project: str,
     return request.execute()['targetSize']
 
 
-def resize_instance_group(instance_group, size, project, zone):
+def resize_instance_group(size, instance_group, project, zone):
     """Changes the number of instances running in |instance_group| to |size|."""
     assert size >= 1
     managers = thread_local.service.instanceGroupManagers()
-    request = managers.get(instanceGroupManager=instance_group,
-                           project=project,
-                           zone=zone)
+    request = managers.resize(instanceGroupManager=instance_group,
+                              size=size,
+                              project=project,
+                              zone=zone)
     return request.execute()
 
 
@@ -116,4 +117,22 @@ def delete_instance_group(instance_group, project, zone):
     request = managers.delete(instanceGroupManager=instance_group,
                               zone=zone,
                               project=project)
+    return request.execute()
+
+
+def create_instance_group(name: str, instance_template_url: str,
+                          experiment: str, project: str, zone: str):
+    """Creates an instance group named |name| from the template specified by
+    |instance_template_url|."""
+    managers = thread_local.service.instanceGroupManagers()
+    target_size = 1  # !!!
+    base_instance_name = 'w-' + experiment
+
+    body = {
+        'baseInstanceName': base_instance_name,
+        'targetSize': target_size,
+        'name': name,
+        'instanceTemplate': instance_template_url
+    }
+    request = managers.insert(body=body, project=project, zone=zone)
     return request.execute()
--- a/common/gce.py
+++ b/common/gce.py
@@ -90,10 +90,15 @@ def get_instance_from_preempted_operation(operation, base_target_link) -> str:
     return operation['targetLink'][len(base_target_link):]
 
 
+def get_instance_group_managers():
+    """Returns the instance group managers resource."""
+    return thread_local.service.instanceGroupManagers()
+
+
 def get_instance_group_size(instance_group: str, project: str,
                             zone: str) -> int:
     """Returns the number of instances running in |instance_group|."""
-    managers = thread_local.service.instanceGroupManagers()
+    managers = get_instance_group_managers()
     request = managers.get(instanceGroupManager=instance_group,
                            project=project,
                            zone=zone)
@@ -103,7 +108,7 @@ def get_instance_group_size(instance_group: str, project: str,
 def resize_instance_group(size, instance_group, project, zone):
     """Changes the number of instances running in |instance_group| to |size|."""
     assert size >= 1
-    managers = thread_local.service.instanceGroupManagers()
+    managers = get_instance_group_managers()
     request = managers.resize(instanceGroupManager=instance_group,
                               size=size,
                               project=project,
@@ -113,7 +118,7 @@ def resize_instance_group(size, instance_group, project, zone):
 
 def delete_instance_group(instance_group, project, zone):
     """Deletes |instance_group|."""
-    managers = thread_local.service.instanceGroupManagers()
+    managers = get_instance_group_managers()
     request = managers.delete(instanceGroupManager=instance_group,
                               zone=zone,
                               project=project)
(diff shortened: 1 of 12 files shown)


jonathanmetzman:
I think we can merge the queue branch into master after this PR.
Some things left to do:
Ensure a minimum number of measurers and a maximum so that experiments that run out of CPU quota aren't hurt by this.

jonathanmetzman:
I think our src_analysis module assumes we are merging into master so because this branch has changes to docker it's rebuilding everything even though this PR doesn't.

jonathanmetzman:
Ping

common/gce.py
```diff
@@ -88,3 +88,56 @@ def get_base_target_link(experiment_config):
...
     """Returns the instance name from a preempted |operation|."""
     return operation['targetLink'][len(base_target_link):]
+
+
+def get_instance_group_managers():
+    """Returns the instance group managers resource."""
+    return thread_local.service.instanceGroupManagers()
+
+
+def get_instance_group_size(instance_group: str, project: str,
```
inferno-chromium, line 98: Can you create an InstanceGroupManager class and keep project, zone, instance_group in __init__, it is repeated across all functions. functions become self.size, self.resize, self.delete, self.create

common/gce.py
```diff
@@ -88,3 +88,56 @@ def get_base_target_link(experiment_config):
...
+    return request.execute()
+
+
+def create_instance_group(name: str, instance_template_url: str,
+                          experiment: str, project: str, zone: str):
+    """Creates an instance group named |name| from the template specified by
+    |instance_template_url|."""
+    managers = get_instance_group_managers()
+    target_size = 1
+    base_instance_name = 'w-' + experiment
```
inferno-chromium, line 134: what does w stand for ?
probably add '-' after experiment

common/gcloud.py
```diff
@@ -22,10 +22,10 @@
 from common import logs
 from common import new_process
 
-# Constants for dispatcher specs.
-DISPATCHER_MACHINE_TYPE = 'n1-standard-96'
-DISPATCHER_BOOT_DISK_SIZE = '4TB'
-DISPATCHER_BOOT_DISK_TYPE = 'pd-ssd'
+# TODO(metzman): Lower this when we figure out how else we can take advantage
+# of the 30 concurrent GCB builds.
```
inferno-chromium, line 26: Dont understand this comment, can you explain more why we cannot go lower.

common/gcloud.py
```diff
@@ -147,3 +146,33 @@ def run_local_instance(startup_script: str = None) -> bool:
...
+        'gcloud', 'compute', '--project', project, 'instance-templates',
+        'create-with-container', template_name, '--no-address',
+        '--image-family=cos-stable', '--image-project=cos-cloud',
+        '--region=%s' % zone, '--scopes=cloud-platform',
+        '--machine-type=n1-standard-1', '--boot-disk-size=50GB',
+        '--preemptible', '--container-image', docker_image
+    ]
+    for item in env.items():
+        command.extend(['--container-env', '%s=%s' % item])
+    new_process.execute(command)
```
inferno-chromium, line 167: Should you check if this operation succeeded and return None/assert if it failed.

experiment/schedule_measure_workers.py
```diff
@@ -0,0 +1,91 @@
...
+def get_instance_group_name(experiment: str):
+    """Returns the name of the instance group of measure workers for
+    |experiment|."""
+    return experiment + '-measure-worker'
+
+
+def get_measure_worker_instance_template_name(experiment: str):
+    """Returns an instance template name for measurer workers running in
+    |experiment|."""
+    return experiment + '-measure-worker'
```
inferno-chromium, line 35: this seems to create same name as get_instance_group_name ? something to differentiate.

experiment/schedule_measure_workers.py
```diff
@@ -0,0 +1,91 @@
...
+
+    num_instances_needed = counts['queued'] + counts['started']
+    instance_group_name = get_instance_group_name(
+        experiment_config['experiment'])
+    project = experiment_config['cloud_project']
+    zone = experiment_config['cloud_compute_zone']
+    num_instances = gce.get_instance_group_size(instance_group_name, project,
+                                                zone)
+
+    if num_instances_needed < num_instances and num_instances == 1:
```
inferno-chromium, line 83: why not just
if not num_instances_needed:

jonathanmetzman:
> I dont see docker/measure-worker in this CL.

I'm merging this into the `queue` branch: https://github.com/google/fuzzbench/tree/queue/docker/measure-worker

common/gcloud.py
jonathanmetzman: Currently we use multiprocessing.Pool() which creates `N` processes for `N` cores.
GCB lets us do 30 concurrent builds.
I could always do 30 processes regardless of cores. I think this probably won't break because most processes will be waiting on GCB not doing expensive computation. I'm being cautious and not doing this for now since we've been bitten by parallelism in the past.

inferno-chromium: Ok makes sense.
What will happen to other folks who will see default gcb limit to 10 ?

inferno-chromium:
> > I dont see docker/measure-worker in this CL.
> 
> I'm merging this into the `queue` branch: https://github.com/google/fuzzbench/tree/queue/docker/measure-worker

Wish i had a better memory of remembering my own reviews! LGTM

common/gcloud.py
jonathanmetzman: I think it happens to work because GCB can queue up to 30.

Not a great answer but I haven't had time to make concurrent builds configurable. I thinkn I'll actually probably use the GCP API to find out instead.

common/gce.py
jonathanmetzman: w stands for worker. I'm going to change it to "m" for measurer and leave a comment about this.
GCE adds the hyphen at the end. No need.

common/gcloud.py
jonathanmetzman: I don't think this is needed. This will exception if it fails.

experiment/schedule_measure_workers.py
jonathanmetzman: I think it's fine if they are named the same because they are for different things. I could add "template" to the template name and "group" to the group name but I'm not sure what we gain by doing this. And I don't think merging them is useful either. I like it as is.

experiment/schedule_measure_workers.py
jonathanmetzman: Done.

common/gce.py
jonathanmetzman: Need to think about how I want to do this. I think I will land as is for now.

PR #341, opened 2020-05-14
Add tooling for running experiments only on changed fuzzers.

jonathanmetzman:
Add function to change_utils that determines the changed fuzzers and add option to run_experiment
to use it to determine which fuzzers to run.

PR #404, opened 2020-06-06
make stdin work for afl++ shmem

vanhauser-thc:
Strange phenomenon - reading from stdin didn't work in fuzzbench binaries but worked fine on local builds with aflpp_driver. Maybe some optimization differences. Now I made basically no change that has an impact however now my fuzzbench generated binaries work with stdin.
I hope this is the same when benchmarking ...


vanhauser-thc:
@jonathanmetzman if you can please verify with one benchmark of aflplusplus_optimal_shmem that the fuzz-target is working for stdin? it works for my local fuzzbench, but ... ... thanks!

jonathanmetzman:
>Strange phenomenon - reading from stdin didn't work in fuzzbench binaries but worked fine on local builds with aflpp_driver.

Sorry is this the same thing as `afl_driver.cpp`? I don't really know how shared memory works with afl++ but it's very possible afl_driver.cpp breaks assumptions it makes. Maybe worth checking and/or using something else: 

https://github.com/llvm/llvm-project/blob/master/compiler-rt/lib/fuzzer/afl/afl_driver.cpp

jonathanmetzman:
Btw I tried running an experiment on the last PR you sent me, but it failed because of https://github.com/google/fuzzbench/pull/407

jonathanmetzman:
>if you can please verify with one benchmark of aflplusplus_optimal_shmem that the fuzz-target is working for stdin? it works for my local fuzzbench, but ... ... thanks!

How can I do this? Just observe the output and see if it looks normalish? Or will it fail loudly if something is wrong?

vanhauser-thc:
> > if you can please verify with one benchmark of aflplusplus_optimal_shmem that the fuzz-target is working for stdin? it works for my local fuzzbench, but ... ... thanks!
> 
> How can I do this? Just observe the output and see if it looks normalish? Or will it fail loudly if something is wrong?

if it finds one other path by mutation its fine.
I tested every single fuzzing target created (make build-aflplusplus_optimal_shmem-all) and they work for me. but as you notice yourself, it can be different in other circumstances although it is all docker ...

> Note that this is with the old binaries not ones from this PR.
> I wonder why this didn't happen in CI?

see above ;)

jonathanmetzman:
OK starting to figure out what's going on.
If I run the target I can get this error message:
```
./fuzz_pcap: error while loading shared libraries: libc++.so.1: cannot open shared object file: No such file or directory
```
(side note: it would probably be nice for AFL/AFL++ to print the output of the target on failure but it may hurt performance to keep track of it).

But if I do the command executed by CI and `make test-run-...`, this error doesn't happen.
I can also get it to stop happening using `patchelf`
I'm guessing the target will fail to run if executed from the wrong directory (weird since I'm executing it from /out`). 

I'm not sure though that there's anything to fix in FuzzBench here. I don't think we've seen evidence that this fails in production but not under testing right? The fix here is probably making the targets built for AFL++ (in FuzzBench) behave more consistently by linking against origin.

jonathanmetzman:
Well, it looks like this line is responsible for why the fuzzer works when called by fuzzer.py:
https://github.com/google/fuzzbench/blob/master/fuzzers/aflplusplus/fuzzer.py#L176

I assume this was the issue you encountered. Probably worth doing something more robust like linking against `$ORIGIN` or setting `LD_LIBRARY_PATH` in runner.Dockerfile (though this might make copy-pasting those files more annoying).

vanhauser-thc:
> Well, it looks like this line is responsible for why the fuzzer works when called by fuzzer.py:
> https://github.com/google/fuzzbench/blob/master/fuzzers/aflplusplus/fuzzer.py#L176
> 
> I assume this was the issue you encountered. Probably worth doing something more robust like linking against `$ORIGIN` or setting `LD_LIBRARY_PATH` in runner.Dockerfile (though this might make copy-pasting those files more annoying).

I am well aware of this and this has nothing to do with the issues I encountered :)

Setting that in the runner is a good idea, I should definitely do that in the next PR.

jonathanmetzman:
> > Well, it looks like this line is responsible for why the fuzzer works when called by fuzzer.py:
> > https://github.com/google/fuzzbench/blob/master/fuzzers/aflplusplus/fuzzer.py#L176
> > I assume this was the issue you encountered. Probably worth doing something more robust like linking against `$ORIGIN` or setting `LD_LIBRARY_PATH` in runner.Dockerfile (though this might make copy-pasting those files more annoying).
> 
> I am well aware of this and this has nothing to do with the issues I encountered :)

My bad, then I guess I wasn't able to reproduce your issue and found a different (non)issue instead. But in any case it looks like the targets built during the last (failed) experiment worked.

jonathanmetzman:
I started an experiment for this. Hopefully the fuzzbench code (both AFL++ related and not) is working now. I'll check the binaries tomorrow morning.

I've been thinking of ways to take manual running out of this process since it would make everyone's life easier I think. I'd like to hear your thoughts as the main fuzzbench user.
My initial idea was to do diffs on the fuzzbench repo to find changed fuzzers (done in #341). I'm not sure it will work for a few reasons (I can elaborate if wanted). But the next best idea is to have a yaml file that external people (e.g. you) specify the fuzzers they want in the experiment. So when you make changes like this, you would need to add something like this:
```
experiment: 2020-05-08
fuzzers: ['aflplusplus', 'aflplusplus_optimal', 'aflplusplus_optimal_shmem']
```
to the file. But the experiment could start immediately (or within a day, not sure) of landing the PR. Does this sound like a big improvement over the current system?

jonathanmetzman:
I started an experiment for this. Hopefully the fuzzbench code (both AFL++ related and not) is working now. I'll check the binaries tomorrow morning.

I've been thinking of ways to take manual running out of this process since it would make everyone's life easier I think. I'd like to hear your thoughts on this as the most frequent fuzzbench user.
My initial idea was to do diffs on the fuzzbench repo to find changed fuzzers (done in #341). I'm not sure it will work for a few reasons (I can elaborate if wanted). But the next best idea is to have a yaml file that external people (e.g. you) specify the fuzzers they want in the experiment. So when you make changes like this, you would need to add something like this:
```
experiment: 2020-05-08
fuzzers: ['aflplusplus', 'aflplusplus_optimal', 'aflplusplus_optimal_shmem']
```
to the file. But the experiment could start immediately (or within a day, not sure) of landing the PR. Does this sound like a big improvement over the current system?

vanhauser-thc:
the auto-running is difficult I think because of multiple includes, e.g. change in afl/fuzzer.py would result in changes for all afl based fuzzers. also a change in the a docker base would change everything.

I like the idea of the yaml file that is included in a PR.
do you still have to fix the reports once the a run i competed?

jonathanmetzman:
>the auto-running is difficult I think because of multiple includes, e.g. change in afl/fuzzer.py would result in changes for all afl based fuzzers. 

Yeah, in some cases I think this is a plus (like if we removed `-d` in afl/fuzzer.py, it would affect every AFL based fuzzer), but I think it wouldn't work as well with the way aflplusplus/fuzzer.py works where it seems like there is a small amount of code for each variant in aflplusplus/fuzzer.py.

>I like the idea of the yaml file that is included in a PR.

Great!

>do you still have to fix the reports once the a run i competed?

Not anymore luckily. I suppose I could just run things more frequently too :-)

vanhauser-thc:
so fingers crossed the this run, that it runs through smoothly and the report is automatically generated :)

jonathanmetzman:
Combining experiments will actually still be manual. I'll come up with a solution for that later.

vanhauser-thc:
@jonathanmetzman I think you have to manually handle `2020-06-09` :)

jonathanmetzman:
That report is still being generated because there are about ~20 trials that still haven't finished.
I generated this one though from the data we already have from 2020-06-09 and 05-24 https://www.fuzzbench.com/reports/2020-06-09-and-05-24/index.html

PR #414, opened 2020-06-09
Some last cleanup to the queue/scalable measurer branch.

jonathanmetzman:


jonathanmetzman:
PTAL.
TBRing this.

.dockerignore
```diff
@@ -34,3 +34,6 @@ report/
 # TODO(metzman): Put main code in subdirectories of src/ so that we don't need
 # to enumerate useless directories.
 docs/
+
+# Auto-generated build files.
+docker/generated.mk
```
inferno-chromium, line 39: Add new line?

docker/dispatcher-image/startup-dispatcher.sh
```diff
@@ -34,13 +34,6 @@ iam_account=`gcloud config get-value account`
 gcloud iam service-accounts keys create ${credentials_file} \
     --iam-account=${iam_account}
 
-# Start up rq workers.
```
inferno-chromium, line 37: Do you not start redis instance anymore ? Is this a manual step for everyone now ? (does it need documentation)

.dockerignore
jonathanmetzman: Will try to do this in next PR.

docker/dispatcher-image/startup-dispatcher.sh
jonathanmetzman: This didn't start up redis instances. It started workers on the dispatcher which used the redis instance. Since they now run on other machines, this isn't necessary or useful anymore.

Redis set up is manual. I documented it here: https://github.com/google/fuzzbench/blob/queue/docs/running-your-own-experiment/setting_up_a_google_cloud_project.md#set-up-redis

It might be better to do automatic set up and tear down of redis (would be more in line with Marcel's request that we provide a way to clean up eveything after experiment to avoid charges, and unlike DB the data isn't useful after an experiment except maybe for debugging). But I think it's fine to worry about this at a later time as this feature is complex enough for now :-)
I also don't know if we want to add more Google Cloud Specific set up.

PR #415, opened 2020-06-09
Use filestore_utils in queue code

jonathanmetzman:
The filestore_utils module was recently added as a module that can be used locally.

### DIFF
--- a/experiment/measurer/measure_worker.py
+++ b/experiment/measurer/measure_worker.py
@@ -109,7 +109,7 @@ class StateFile:
 
         return json.loads(
             filestore_utils.cat(previous_state_file_bucket_path,
-                                must_exist=False))
+                                must_exist=False).output)
 
     def get_previous(self):
         """Returns the previous state."""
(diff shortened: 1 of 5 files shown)


jonathanmetzman:
I think I jinxed it last time by saying last. I'll probably write some more tests for this before merging queue into master.
I'm also going to start an experiment using queue for #404 so there will probably be some fixes there as well.
TBRing this as well.

experiment/measurer/measure_worker.py
```diff
@@ -108,7 +108,8 @@ def _get_previous_cycle_state(self) -> list:
             self._get_bucket_cycle_state_file_path(self.cycle - 1))
 
         return json.loads(
-            gsutil.cat(previous_state_file_bucket_path, expect_zero=False))
+            filestore_utils.cat(previous_state_file_bucket_path,
+                                must_exist=False).output)
```
inferno-chromium, line 112: We can just return output from cat function itself.

jonathanmetzman: I want it to have the same interface (e.g. returns a ProcessResult) as the rest of the functions in the module.

PR #416, opened 2020-06-09
Make automatic_run_experiment run experiments based on requests

jonathanmetzman:
Users can request experiments to run in service/experiment-requests.yaml.
automatic_run_experiment.py is configured to run each requested experiment in that file at least once.

NOTE THIS FUNCTIONALITY IS NOT MEANT TO BE USED BY USERS YET!

TODO:
1. Set up a cron job somewhere to run service/run.bash.
2. Document this feature.


service/experiment-requests.yaml
```diff
@@ -0,0 +1,4 @@
+- experiment: 2020-06-08
```
jonathanmetzman, line 1: There was no need to run this experiment. I just did it to test my code and since it provides a decent example (I think examples are as important as docs since many will copy paste).

service/automatic_run_experiment.py
```diff
@@ -12,27 +12,29 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Determines if an experiment should be run and runs one if necessary.
-Note that this code uses a config file for experiments that is not generic.
-Thus, it only works on the official fuzzbench service. This script can be
-run manually but is intended to be run by a cronjob."""
+"""Reads experiment-requests.yaml and determines if an experiment should run and
```
lszekeres, line 15: It's a bit fuzzy to me what an experiment request is. Are we going to run experiments defined in this yaml file, or there will be a default periodic experiment that re-measures everything that has changed, and these "experiment requests" are extra measurements on top of that that otherwise wouldn't run?

jonathanmetzman: We won't use change detection for the service, I implemented that but I don't really think it is viable anymore.
Basically, the current plan is for me to manually run full experiments when big changes happen and the service to run experiments when users ask for them (e.g. when they modify their fuzzer).
Do you think this needs to be clarified in the docstring, or do you think it is fuzzy because I gave you ideas about how this will work that are no longer true?

jonathanmetzman:
This is ready for review.
I think all of the proposed changes can wait.
1. The requirement for specifying the experiment name is good because it's easier to map the request to the actual report generated.
2. Combining with previous experiments can't be done so easily and I'm still figuring out how I should do it.

jonathanmetzman:
I'm undoing the documentation change. I'm not going to ask users to add requests until this system is improved.

service/automatic_run_experiment.py
```diff
@@ -62,26 +73,128 @@
...
+            valid = False
+
+    if not valid:
+        # Don't try the next validation step if the previous failed, we might
+        # exception.
+        return valid
+
+    # Make sure experiment requests have a unique name, we can't run the same
+    # experiment twice.
+    counts = collections.Counter(
```
inferno-chromium, line 144: Why not just keep a list of experiment name inside the loop above and do validation there instead of another loop.

service/automatic_run_experiment.py
```diff
@@ -62,26 +73,128 @@
...
+
+    if PAUSE_SERVICE_KEYWORD in requested_experiments:
+        # Check if automated experiment service is paused.
+        logs.warning('Pause service requested, not running experiment.')
+        return None
+
+    requested_experiment = None
+    for experiment_config in reversed(requested_experiments):
+        experiment_name = _get_experiment_name(experiment_config)
+        experiment_request_is_new = db_utils.query(models.Experiment).filter(
```
inferno-chromium, line 172: s/experiment_request_is_new/is_new_experiment to shorten ?

service/automatic_run_experiment.py
```diff
@@ -62,26 +73,128 @@
...
+            models.Experiment.name == experiment_name).first() is None
+        if experiment_request_is_new:
+            requested_experiment = experiment_config
+            break
+
+    if requested_experiment is None:
+        logs.info('Not running new experiment.')
```
inferno-chromium, line 179: No new experiment available, exiting ?

service/automatic_run_experiment.py
```diff
@@ -62,26 +73,128 @@
...
+            continue
+
+        experiment_fuzzers = request['fuzzers']
+        for fuzzer in experiment_fuzzers:
+            if fuzzer in all_fuzzers:
+                continue
+            # Fuzzer isn't valid.
+            logger.error('Fuzzer: %s in experiment %s is not valid.', fuzzer,
+                         experiment)
+            valid = False
```
lszekeres, line 135: Consider factoring this out to a function

service/test_automatic_run_experiment.py
```diff
@@ -82,3 +112,84 @@ def sort_key(dictionary):
...
+            'experiment': EXPERIMENT,
+            'fuzzers': []
+        }, False),  # No fuzzers.
+        ({
+            'experiment': EXPERIMENT
+        }, False),  # No fuzzers.
+        ({
+            'fuzzers': ['afl']
+        }, False),  # No experiment.
+        # Invalid experiment.
```
inferno-chromium, line 148: either move comments to end to match others, or line before block is better.

service/automatic_run_experiment.py
```diff
@@ -62,26 +73,128 @@
...
+            logger.error('Experiment: "%s" appears %d times.',
+                         str(experiment_name), count)
+            valid = False
+
+    return valid
+
+
+def run_requested_experiment(dry_run):
+    """Run the oldest requested experiment that hasn't been run yet in
+    experiment-requests.yaml."""
```
lszekeres, line 158: This means we will run one request a day? If multiple requests are made one day, would it make sense to merge them?

service/experiment-requests.yaml
```diff
@@ -0,0 +1,22 @@
+# Experiment requests have the following format:
+#
+# - experiment: 2020-06-08  # The name of the experiment
+#   fuzzers:  # The fuzzers to run in the experiment.
+#     - aflplusplus
+#     - libfuzzer
+#
+# The name of the experiment must begin with a date using this format:
+# YYYY-MM-DD (year-month-day). It's not important what timezone is used in
+# deciding the date or if this is a day or two off. The most important thing is
```
lszekeres, line 10: We do we require names to start with dates? Asking because if the date doesn't really mean anything (eg. we won't necessary run the exp on the given date) than this might be confusing for people

service/automatic_run_experiment.py
```diff
@@ -12,27 +12,38 @@
...
+logger = logs.Logger('automatic_run_experiment')  # pylint: disable=invalid-name
 
 EXPERIMENT_CONFIG_FILE = os.path.join(utils.ROOT_DIR, 'service',
                                       'experiment-config.yaml')
 
+REQUESTED_EXPERIMENTS_PATH = os.path.join(utils.ROOT_DIR, 'service',
+                                          'experiment-requests.yaml')
+
+# Don't run an experiment if we have a "request" just containing this keyword.
+PAUSE_SERVICE_KEYWORD = 'PAUSE_SERVICE'
```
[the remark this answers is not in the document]
jonathanmetzman: Let this as a TODO.

service/automatic_run_experiment.py
jonathanmetzman: It probably would. There are definitely some problems with the current approach that should be fixed (I assume many pull requests won't even be merged on the day they were worked on, so I'd expect with this system well run experiments named 2020-01-01 on 2020-01-02 etc.)

service/experiment-requests.yaml
jonathanmetzman: Right. The reason why I want dates is because that is the format we've been using: https://www.fuzzbench.com/reports/

jonathanmetzman: But yes there would be a problem with running experiments on the right date.

presubmit.py
```diff
@@ -242,6 +246,30 @@ def yapf(paths: List[Path], validate: bool = True) -> bool:
...
 
 
+def validate_experiment_requests(paths: List[Path]):
+    """Returns False if service/experiment-requests.yaml it is in |paths| and is
+    not valid."""
+    if Path(automatic_run_experiment.REQUESTED_EXPERIMENTS_PATH) not in paths:
+        return True
+
+    try:
+        experiment_requests = yaml_utils.read(
```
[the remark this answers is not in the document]
jonathanmetzman: I'd rather have a self contained validation function and not couple this functionality.

service/automatic_run_experiment.py
jonathanmetzman: Done.

service/test_automatic_run_experiment.py
jonathanmetzman: Done.

service/automatic_run_experiment.py
jonathanmetzman: Done.

service/automatic_run_experiment.py
jonathanmetzman: Done.

service/automatic_run_experiment.py
jonathanmetzman: I think that has n^2 complexity right? The list will probably be small enough that it doesn't matter, but that is the reason I didn't do this.

presubmit.py
```diff
@@ -242,6 +246,30 @@ def yapf(paths: List[Path], validate: bool = True) -> bool:
...
+            automatic_run_experiment.REQUESTED_EXPERIMENTS_PATH)
+    except yaml.parser.ParserError:
+        print('Error validating %s.' %
```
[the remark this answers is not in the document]
jonathanmetzman: Good point. Done.

service/experiment-requests.yaml
lszekeres: Some alternatives to consider:
- Ask for just some short tag/identifier, e.g., 'libfuzzer-update' or 'my-fuzzer-v6.8' and we would auto-generate the experiment name from the `<data-of-run>-<tag>`, e.g., `2020-06-22-my-fuzzer-v6.8`
- Don't ask for any name, but auto-generate experiment name from the provided fuzzer list: `<DATE-OF-RUN>-<list-of-updated-fuzzers>`, e.g., `2020-06-22-afl-libfuzzer-honggfuzz`. (I imagine typically the list of updated fuzzers would be small in individual requests)

PR #417, opened 2020-06-09
Clean up filestore_utils.py API

jonathanmetzman:
Make the following changes to the API:
1. Don't use `*cp_args` for `cp`, *ls_args` for `ls`, `*rm_args` for `rm`. This is adds complexity and can cause errors because it exposes details of the `cp` and `ls` implementations to callers.
2. Make `parallel` an explicit parameter in the API instead of implicitly passing it to `command/gsutil_command` by passing it `**kwargs`
3. Don't pass `**kwargs` from API users or API functions to `new_process.execute`. This also adds complexity.
4. Don't pass `*args` from API functions to `new_process.execute`. This is very bug prone.
5. Don't rely on `expect_zero=False` in functions other than `ls`. Because we no longer pass `**kwargs` to `execute` (or `gsutil_command`/`command`) we must explicitly support `expect_zero=False` in functions where they were needed. Rather than support it, make callers catch exceptions from `execute`.

### DIFF
--- a/common/filestore_utils.py
+++ b/common/filestore_utils.py
@@ -36,24 +36,22 @@ else:
 
 # TODO(zhichengcai): Change all implementations of cp, ls, and rm to stop using
 # special handling of *args. This is error prone now that there are wrappers.
-def cp(*cp_arguments, **kwargs):  # pylint: disable=invalid-name
+def cp(*cp_arguments, parallel=False):  # pylint: disable=invalid-name
     """Copy source to destination."""
-    return filestore_utils_impl.cp(*cp_arguments, **kwargs)
+    return filestore_utils_impl.cp(*cp_arguments, parallel=parallel)
 
 
-def ls(*ls_arguments, must_exist=True, **kwargs):  # pylint: disable=invalid-name
+def ls(*ls_arguments, must_exist=True, parallel=False):  # pylint: disable=invalid-name
     """List files or folders."""
     return filestore_utils_impl.ls(*ls_arguments,
-                                   must_exist=must_exist,
-                                   **kwargs)
+                                   must_exist=must_exist, parallel=parallel)
 
 
-def rm(*rm_arguments, recursive=True, force=False, **kwargs):  # pylint: disable=invalid-name
+def rm(*rm_arguments, recursive=True, force=False, parallel=False):  # pylint: disable=invalid-name
     """Remove files or folders."""
     return filestore_utils_impl.rm(*rm_arguments,
                                    recursive=recursive,
-                                   force=force,
-                                   **kwargs)
+                                   force=force, parallel=parallel)
 
 
 def rsync(  # pylint: disable=too-many-arguments
@@ -63,7 +61,7 @@ def rsync(  # pylint: disable=too-many-arguments
         recursive=True,
         gsutil_options=None,
         options=None,
-        **kwargs):
+        parallel=False):
     """Synchronize source and destination folders."""
     return filestore_utils_impl.rsync(source, destination, delete, recursive,
-                                      gsutil_options, options, **kwargs)
+                                      gsutil_options, options, parallel=parallel)
--- a/common/filestore_utils.py
+++ b/common/filestore_utils.py
@@ -36,22 +36,22 @@ else:
 
 # TODO(zhichengcai): Change all implementations of cp, ls, and rm to stop using
 # special handling of *args. This is error prone now that there are wrappers.
-def cp(*cp_arguments, parallel=False):  # pylint: disable=invalid-name
+def cp(*cp_arguments, parallel=False, expect_zero=True):  # pylint: disable=invalid-name
     """Copy source to destination."""
-    return filestore_utils_impl.cp(*cp_arguments, parallel=parallel)
+    return filestore_utils_impl.cp(*cp_arguments, parallel=parallel, expect_zero=expect_zero)
 
 
-def ls(*ls_arguments, must_exist=True, parallel=False):  # pylint: disable=invalid-name
+def ls(*ls_arguments, expect_zero=True):  # pylint: disable=invalid-name
     """List files or folders."""
-    return filestore_utils_impl.ls(*ls_arguments,
-                                   must_exist=must_exist, parallel=parallel)
+    return filestore_utils_impl.ls(*ls_arguments, expect_zero=expect_zero)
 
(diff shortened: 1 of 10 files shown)


jonathanmetzman:
I still need to fix `rm`.

Some questions:
1. When we have parameters with default values, I set the default value everywhere. So for example if you call `filestore_utils.rm('blah')`, `filestore_utils.rm` has a default value for `parallel` it calls `gsutil/local_filestore.rm` which also have default values for parallel which call `gsutil.gsutil_command/local_filestore.command` which also have defaults for `parallel` since `gsutil_command` is really something for `gsutil_command`. I don't like this "viralness" I think it's a lot of typing and error prone to set a default value in 5 places. I think only the public APIs should have default values? The nonpublic APIs don't need them IMO. WDYT?

2. `ls` is tricky to support because of differences between gsutil's implementation and the real ls.  But the only place it is used for now is `exists_in_experiment_filestore` which actually only needs the returncode and not the results. Should we get rid of it and replace it with `exists`? I think we are pretty likely to need it in the future. So i'm not really in favor of this (better to have a working `ls` now than when we really need it later).

experiment/measurer.py
```diff
@@ -61,9 +62,9 @@ def get_experiment_folders_dir():
     return exp_path.path('experiment-folders')
 
 
-def remote_dir_exists(directory: pathlib.Path) -> bool:
-    """Does |directory| exist in the CLOUD_EXPERIMENT_BUCKET."""
-    return filestore_utils.ls(exp_path.gcs(directory), must_exist=False)[0] == 0
+def exists_in_experiment_filestore(path: pathlib.Path) -> bool:
+    """Returns True if |path| exists in the experiment_filestore."""
+    return bool(filestore_utils.ls(exp_path.gcs(path), must_exist=False))
```
jonathanmetzman, line 67: I may need to change this. It returns `False` for a directory that is empty. That is impossible in GCS, but is possible in a unix filesystem.

jonathanmetzman:
> I still need to fix `rm`.
> 

Done.

> Some questions:
> 
> 1. When we have parameters with default values, I set the default value everywhere. So for example if you call `filestore_utils.rm('blah')`, `filestore_utils.rm` has a default value for `parallel` it calls `gsutil/local_filestore.rm` which also have default values for parallel which call `gsutil.gsutil_command/local_filestore.command` which also have defaults for `parallel` since `gsutil_command` is really something for `gsutil_command`. I don't like this "viralness" I think it's a lot of typing and error prone to set a default value in 5 places. I think only the public APIs should have default values? The nonpublic APIs don't need them IMO. WDYT?

I started doing this and I'm not sure I want to change it. If we change this then parallel has to come at the end of the nondefault parameters which I don't really like the idea of. I don't like having a cp function with this signature `cp(source, dest, parallel, recursive=False)` instead of `cp(source, dest, recursive=False, parallel=False)` even if this function is nonpublic :-\

> 2. `ls` is tricky to support because of differences between gsutil's implementation and the real ls.  But the only place it is used for now is `exists_in_experiment_filestore` which actually only needs the returncode and not the results. Should we get rid of it and replace it with `exists`? I think we are pretty likely to need it in the future. So i'm not really in favor of this (better to have a working `ls` now than when we really need it later).

We do use it in some tests. I think `ls` is too basic to get rid of.

experiment/measurer.py
jonathanmetzman: Done.

common/test_gsutil.py
```diff
@@ -29,6 +29,15 @@ def test_gsutil_command():
     assert mocked_popen.commands == [['gsutil'] + arguments]
 
 
+@mock.patch('common.new_process.execute')
+def test_ls_not_must_exist(mocked_gsutil_command):
+    """Tests that ls makes a correct call to new_process.execute when
+    must_exist=False."""
+    gsutil.ls('gs://hello', must_exist=False)
```
inferno-chromium, line 36: Shouldn't this be tested with both true and False ?

jonathanmetzman: Done.

common/gsutil.py
```diff
@@ -19,43 +19,46 @@
...
     """Executes a gsutil command with |arguments| and returns the result."""
     command = ['gsutil']
     if parallel:
         command.append('-m')
-    return new_process.execute(command + arguments, *args, **kwargs)
+    return new_process.execute(command + arguments, expect_zero=expect_zero)
 
 
-def cp(*cp_arguments, **kwargs):  # pylint: disable=invalid-name
+def cp(source, destination, parallel, recursive=False):  # pylint: disable=invalid-name
```
zchcai, line 30: What about putting a default value for `parallel=False` as `rm` and `rsync`?

common/filestore_utils.py
```diff
@@ -36,24 +36,28 @@ def _using_gsutil():
...
 
 
-def ls(*ls_arguments, must_exist=True, **kwargs):  # pylint: disable=invalid-name
-    """List files or folders."""
-    return filestore_utils_impl.ls(*ls_arguments,
-                                   must_exist=must_exist,
-                                   **kwargs)
+def ls(path, must_exist=True):  # pylint: disable=invalid-name
+    """Lists files or folders in |path|. Doesn't except on failure if must_exist
+    is False otherwise can raise subprocess.CalledProcessError.
```
lszekeres, line 49: nit: consider simplify this sentence, eg:  if `must_exist` is `True` raises subprocess.CalledProcessError when `path` doesn't exists.

common/filestore_utils.py
```diff
@@ -36,24 +36,28 @@ def _using_gsutil():
...
+    Returns the output of ls split line-by-line (for example ls('directory/')
+    returns every file in the directory."""
```
lszekeres, line 51: unclear from the sentence whether a list of strings is returned or a string with one file per line

common/gsutil.py
lszekeres: +1

common/gsutil.py
```diff
@@ -19,43 +19,46 @@
 logger = logs.Logger('gsutil')
 
 
-def gsutil_command(arguments, *args, parallel=False, **kwargs):
+def gsutil_command(arguments, expect_zero=True, parallel=False):
```
lszekeres, line 22: consider describing `expect_zero` in the docstring

PR #418, opened 2020-06-09
[make][runner] Use FORCE_LOCAL in debug-, test-run-, run- make targets

### LINKED ISSUE #412: make debug/test-run/run don't respect FORCE_LOCAL and FORCE_NOT_LOCAL
This breaks when someone needs to use FORCE_LOCAL (see github.com/google/fuzzbench/issues/82).
I think the quick-and-dirty solution here is to always pass FORCE_LOCAL to these docker runs since they are never run in production. The real fix is making this unnecessary.

jonathanmetzman:
They targets are only used locally, and FORCE_LOCAL won't be propagated
from the user's environment to the docker container unless we do this.
So best to assume it is needed.
Fixes #412.


jonathanmetzman:
CC @tanq16

PR #421, opened 2020-06-11
Alternative coverage measurement

vanhauser-thc:
@wideglide made a comment in a question issue at afl++: `I also wonder if there's a bit of bias in the fuzzbench results since they also measure coverage using SanitizerCoverage?`

So I tried this for a benchmark with the following result:

| BENCHMARK | FUZZER | LINES | PC |
| :-: |:-: | :-: | :-: |
| freetype2-2017 | aflplusplus_classic | 965 | 1758 |
| freetype2-2017 | aflplusplus_nosingleblock | 965 | 1756 |
| freetype2-2017 | aflplusplus_nozero | 943 | 1756 |
| freetype2-2017 | aflplusplus_nozerosingleblock | 805 | 1735 |

LINES = lines covered from llvm coverage , PC = program counters from sancov.

We can see that there are differences where PC values are the same there is a difference in LINES and vice versa.

So for an experiment I propose the following:
Let's create a new branch in fuzzbench and then change the base of this PR to that new branch and apply it, then building the "coverage2" containers.

From any previous fuzzbench run then apply the queues to the fuzz-target in the coverage2 container.
(basically a "cd /src; /out/fuzz-target /import/queue/id*", then followed with an analysis, "llvm-cov-11 gcov -a -b -c -f -u *.gcda" and summing up the "Lines executed")

And afterwards comparing the PC and LINES coverage and analyzing if the sancov coverage is really a good approach.


jonathanmetzman:
I really appreciate this patch.
How many benchmarks does this compile? Just the one mentioned in https://github.com/AFLplusplus/AFLplusplus/issues/397?
I think there would be quite a bit more required to get this working, probably the easiest route would be to just change the coverage image to use lcov and even then there would still be more work.

I mentioned that @shadoom7 will work on using Clang's source based coverage (which isn't used by libFuzzer at all) for measuring coverage in FuzzBench. If we think those results are independent enough I think waiting until that work is done would be easiest. What do you think?

vanhauser-thc:
> How many benchmarks does this compile? Just the one mentioned in [AFLplusplus/AFLplusplus#397](https://github.com/AFLplusplus/AFLplusplus/issues/397)?

this does compile all benchmarks, however I have not checked if they also are all fine for coverage.
(in a previous different way I did it not all libraries were instrumented)

> I think there would be quite a bit more required to get this working, probably the easiest route would be to just change the coverage image to use lcov and even then there would still be more work.

Yes. but it is actually not much.
1. the GCNO file of the builders/ docker image have to be copied to the /src of the target container
2. run the coverage2 target with the queue input from a fuzzer
3. llvm-cov-11 gcov -a -b -c -f -u THE_GCDA_FILE 

step 3 produces output like 
```
Function main:
Lines executed: 86.7% of 249
```
and these can just be easily calculated with a script and voila there are the line numbers.

I am not a python guy and my version is a shell script so I did not bother to push that ;)
here it is:
```
          llvm-cov-11 gcov -a -b -c -f -u "$GCDA" 2>&1 | grep '% of ' > cov.out
          FUNCTIONS=`grep 'Lines executed:' cov.out | wc -l`
          LINES=0
          for k in `grep 'Lines executed:' cov.out | sed 's/.*://' | tr -d % | sed 's/ of /*/'`; do
            LINES=`expr $LINES + $(echo "$k/100"|bc)`
          done
          BRANCHES=0
          for k in `grep 'Branches executed:' cov.out | sed 's/.*://' | tr -d % | sed 's/ of /*/'`; do
            BRANCHES=`expr $BRANCHES + $(echo "$k/100"|bc)`
          done
          test "$LINES" -gt 0 && echo "$BENCHMARK,$FUZZER,$FUNCTIONS,$LINES,$BRANCHES" | tee -a $i/RESULTS2
```
> I mentioned that @Shadoom7 will work on using Clang's source based coverage (which isn't used by libFuzzer at all) for measuring coverage in FuzzBench. If we think those results are independent enough I think waiting until that work is done would be easiest. What do you think?

@Shadoom7 is using the profiling version of the coverage, this one is using the gcc like version which uses debuginfo. I have no clue what the differences in the results are :)

He is also doing all the work to get it working in fuzzbench where this is just a lower effort proposal to run things by hand and see if they make sense to add.

funny that these two approaches come at the same time :)

vanhauser-thc:
I am closing this for now.
I tested 10 binaries generated and only a few generate useful coverage information. the others show minimal coverage and all being the same which means that just the harness seems to be covered and none of the libraries for most targets, which is weird.
I hope @Shadoom7 's implementation performs better. Will try once he is done.

PR #406, opened 2020-06-08
local dispatcher: from gsutil to rsync

zchcai:
## Description

After PR #405, it will run `rsync` in dispatcher and runner docker containers. Thus, add install instruction for these two docker containers running scripts.

Other changes:
- In local dispatcher creation, it also needs to add a `/` at the first parameter in `rsync` command.
- To comply with PR #402, change the name from `CLOUD_EXPERIMENT_BUCKET` to `EXPERIMENT_FILESTORE` to avoid potential merging conflict later after #402 merged.

## Motivation and Context

FuzzBench will support a complete local running in the future, and this PR is a preparation.
`rsync` is not a system command and it needs manual installation.

It relates to #399, #402, #405 and #8.

## How Has This Been Tested?

No test for this. At last, it will be a required component for local running.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Issue #398, opened 2020-06-03
filestore_utils: _using_gsutil() implementation


zchcai:
After this PR, it will have an initial version for deciding whether using gsutil.
However, we may want to raise an error and not providing a default behavior. Then we need to modify 7 related unit tests.
Details to see discussion in this PR https://github.com/google/fuzzbench/pull/394.

zchcai:
Hi @lszekeres @jonathanmetzman, in fact, if we remove the try-catch-exception in deciding `_using_gsutil`, then the errors from unit tests are due to `from experiment import builder` and builder has `from common import filestore_utils`. Changing test function body will not help.

I prefer to keep the current pattern, but also consider other ways to decide whether to use gsutil.
For example, if we can support `source` and `destination` in the `filestore_utils` functions' argument, then we can use these two to decide which command to use. The checking logic can be: if `source` or `destination` starts with `gs://`, then we switch to `gsutil`.

zchcai:
We need a better plan for deciding local or not.

Initially, last week, when I did my [initial experimental plan](https://github.com/zchcai/fuzzbench/compare/5d8feaa...1918184) (https://github.com/google/fuzzbench/pull/399), I didn't put this logic inside `filestore_utils` and `experiment_path`. Now I know the reason behind it:

If this logic locates inside `filestore_utils` and `experiment_path`, then it should be better independent to other implementation.

However, whether or not to use local or cloud, is specified by users, which is not a pre-known things.
That is to say, whether to use local or cloud should let the outside callers tell `filestore_utils` and `experiment_path` what to do, other than `filestore_utils` and `experiment_path` decide it by themselves.

Right now, in my mind, is to specify the file or folder path arguments in function signature. Then we decide by checking whether one of `source` or `destination` have a special file prefix. However, in this way, if two file storages share a same prefix, then no way to decide.

Or we provide a default choice for `filestore_utils` like right now I choose it as `gsutil`. When some user-specified environments have been set, we switch to that. For example, like https://github.com/google/fuzzbench/pull/394 does.

@lszekeres @jonathanmetzman wdyt?

zchcai:
For the next design plan, we use an environment variable (like the current `CLOUD_EXPERIMENT_BUKCET`) to indicate which platform/file storage to use during running. Also, `_using_gsutil()` will move to the `experiment_utils.py`.

PR #422, opened 2020-06-11
filestore: enable local_filestore; add unit test

zchcai:
## Description

This PR enables `local_filestore` in `filestore_utils` and adds some initial integration unit tests for local behaviors.
This PR fixes the bug for importing correct filestore. Previously, since `import` happens at the very beginning, the whole running will only use `gsutil`, even if we change the environment variables for `EXPERIMENT_FILESTORE` and `EXPERIMENT` later. This is not the desired behavior.
Now, when `filestore_utils.*` is called, the decorator will re-define the `filestore_utils_impl`, according to environment at the calling time.
The test cases in `common/test_filestore_utils.py` will be helpful.

## Motivation and Context

FuzzBench will support a complete local running in the future, and this PR is an initial switch-on test.
Currently, for filestore usages, there is only one gsutil implementation, which supports google cloud buckets. After this PR, the other option will turn on when `EXPERIMENT_FILESTORE` is not a Google Cloud pattern path (e.g., starting with `gs://`) and `EXPERIMENT` is not NULL.

After this PR is merged, the last local support PR will be submitted. And from that future PR, when users set `local_experiment` as `true`, then the input validation will automatically check, whether `EXPERIMENT_FILESTORE` is a POSIX style path. Then this value will pass to every layer of execution environment.

This PR is a follow-up for PR #405. Please review #405 and let it merged, and then review this one.
This PR is also a continued PR for local series PRs: #8, #394, #399, #402, #406 and #405.

## How Has This Been Tested?

To enable local or cloud environments, this PR adds two helpful `pytest.fixture` for testing: `switch_on_local` and `switch_on_gsutil`. Then it adds some specific desired behaviors for local or cloud respectively.

The testing environment: SMP Debian 5.2.17-1rodete3 (2019-10-21 > 2018) x86_64 GNU/Linux

In fact, almost all other unit tests should add `switch_on_gsutil` now. Since we define the default behavior is in Google Cloud mode, thus there is no need to rewrite them yet.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My code follows the code style of this project.


common/filestore_utils.py
```diff
@@ -13,8 +13,12 @@
 # limitations under the License.
 """Helper functions for interacting with the file storage."""
 
+import functools
+
 from common import experiment_utils
 
+#pylint: disable=invalid-name,global-at-module-level,redefined-outer-name,import-outside-toplevel
```
jonathanmetzman, line 20: Which of these directives are actually needed?
Is it all of them? To me this looks copied from some of my unittests.
Also, disabling at the toplevel is generally bad. I do it in unittests some times because pylint constantly complains about some things with fixtures. But the better way to do it is to put the pylint directive on the line pylint complains about.

common/filestore_utils.py
```diff
@@ -27,16 +31,35 @@ def _using_gsutil():
     return experiment_filestore_path.startswith('gs://')
 
 
-if _using_gsutil():
-    from common import gsutil as filestore_utils_impl
-else:
-    # When gsutil is not used in the context, here it should use local_utils.
-    # TODO(zhichengcai): Implement local_filestore.py and import it here.
-    from common import gsutil as filestore_utils_impl
+global filestore_utils_impl
```
jonathanmetzman, line 34: is this needed? Why not just do it on line 48?

common/filestore_utils.py
```diff
@@ -27,16 +31,35 @@ def _using_gsutil():
...
+filestore_utils_impl = None
+
+
+def get_filestore_utils_impl():
+    """Imports filestore_utils_impl dynamically."""
```
jonathanmetzman, line 39: This approach feels clean and even though I use a similar one in `database/utils.py` I think the more explicit `get_impl().cp` approach I showed you yesterday is better.

common/filestore_utils.py
```diff
@@ -27,16 +31,35 @@ def _using_gsutil():
...
+
```
jonathanmetzman, line 40: Btw I don't think you made this decorator quite right. You should probably only have one layer of nested functions not two. That's why your decorators need to call get_filestore_utils_impl() which is very weird for a decorator.

common/test_filestore_utils.py
```diff
@@ -21,27 +21,76 @@
...
+    FILESTORE = '/experiment_data'
+    DIR1 = '/dir1'
+    DIR2 = '/dir2'
+
+    def test_local_filestore_on(self, switch_on_local_filestore):
+        """Tests that local_filestore switches on correctly."""
+        with mock.patch('common.new_process.execute') as mocked_execute:
+            filestore_utils.cp(self.DIR1, self.DIR2, recursive=True)
+            mocked_execute.assert_called_with(
+                ['cp', '-r', self.DIR1, self.DIR2], expect_zero=True)
```
jonathanmetzman, line 38: I would just do `assert 'gsutil' not in mocked_execute.call_args_list

common/test_filestore_utils.py
```diff
@@ -21,27 +21,76 @@
...
+
+    def test_local_filestore_parallel_off(self, fs, switch_on_local_filestore):
+        """Tests that `parallel` is False for local execution."""
+        fs.create_dir(self.DIR1)
+
+        with mock.patch('common.local_filestore.local_filestore_command'
+                       ) as mocked_local_filestore_command:
+            filestore_utils.rsync(self.DIR1, self.DIR2, parallel=True)
+            test_args_list = mocked_local_filestore_command.call_args_list
```
jonathanmetzman, line 47: nit: Just call this `call_args_list`

common/local_filestore.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def ls(path, must_exist=True):  # pylint: disable=invalid-name
+    """Executes "ls" command for |path|. If |must_exist| is True then it can
+    raise subprocess.CalledProcessError."""
+    # Add '-l' to behave like `gsutil.ls`.
+    command = ['ls', '-l', path]
+    process_result = local_filestore_command(command, expect_zero=must_exist)
+    return process_result
+
```
jonathanmetzman, line 52: Seems like there is code added here that was already added in #405.

common/filestore_utils.py
zchcai: They are all needed and not copied from other places.
I can move some one-time use to that line. Thanks.

common/filestore_utils.py
zchcai: I tried two layers, but it doesn't work. This is the only workable version. :)
If only two layers, it reports for arguments amount conflict.

common/test_filestore_utils.py
zchcai: Good suggestion! Let me change to it. That is more related to the testing goal.

common/local_filestore.py
zchcai: yes, so we need to first merge #405, and I rebase it.
This version is for working since #405 not merged yet.

common/filestore_utils.py
zchcai: Let me check this later.

common/filestore_utils.py
zchcai: I first tried that one (since that is easier), but failed yesterday. I guess that was because the assignment issues.
Let me do another check later.

common/test_filestore_utils.py
zchcai: This unit test is not needed.

common/filestore_utils.py
zchcai: We can also use `return common.gsutil` in a function `get_impl()`.
Which one do you prefer? The current decorator one, or I rewrite to use `get_impl()`?
@lszekeres @jonathanmetzman

jonathanmetzman: Let's do `get_impl()` please.

lszekeres: +1

zchcai: Done.

conftest.py
```diff
@@ -80,3 +80,17 @@ def experiment(environ):  # pylint: disable=redefined-outer-name,unused-argument
     os.environ['EXPERIMENT_FILESTORE'] = 'gs://experiment-data'
     os.environ['REPORT_FILESTORE'] = 'gs://web-bucket'
     os.environ['CLOUD_PROJECT'] = 'fuzzbench'
+
+
+@pytest.fixture
+def switch_on_local_filestore(environ):  # pylint: disable=redefined-outer-name,unused-argument
```
jonathanmetzman, line 86: name this `use_local_filestore` and `use_gsutil`

conftest.py
```diff
@@ -80,3 +80,17 @@ def experiment(environ):  # pylint: disable=redefined-outer-name,unused-argument
...
+    """Mock a local filestore usage experiment."""
+    os.environ['EXPERIMENT_FILESTORE'] = '/experiment-data'
+    os.environ['EXPERIMENT'] = 'test-experiment'
+
+
+@pytest.fixture
+def switch_on_gsutil(environ):  # pylint: disable=redefined-outer-name,unused-argument
```
jonathanmetzman, line 93: Change this to:
```
@pytest.fixture
def use_gsutil(experiment):
"""Mock a Google Cloud Storage bucket usage experiment."""
```

conftest.py
jonathanmetzman, line 86: Change this to 
```
def use_local_filestore(experiment):
os.environ['EXPERIMENT_FILESTORE'] = '/experiment-data'
```

jonathanmetzman:
Please add a test for this: https://github.com/google/fuzzbench/pull/405#discussion_r439649089

common/test_filestore_utils.py
```diff
@@ -21,27 +21,66 @@
 # TODO(zhichengcai): Figure out how we can test filestore_utils when using
 # the local_filestore implementation.
```
zchcai, line 22: Shall we delete this TODO?

zchcai:
> Please add a test for this: [#405 (comment)](https://github.com/google/fuzzbench/pull/405#discussion_r439649089)

Thanks! Done.

common/filestore_utils.py
```diff
@@ -27,34 +29,34 @@ def _using_gsutil():
...
-    from common import gsutil as filestore_utils_impl
-else:
+def get_impl():
+    """Returns the implementation for filestore_utils."""
+    if _using_gsutil():
+        return gsutil
     # Use local_filestore when not using gsutil.
-    # TODO(zhichengcai): Implement local_filestore.py and import it here.
-    from common import gsutil as filestore_utils_impl
+    return local_filestore
```
lszekeres, line 37: Should we do the imports too here? (eg only import gsutil when actually needed)

common/test_filestore_utils.py
```diff
@@ -22,26 +22,63 @@
...
-                                          expect_zero=True)
-
-    filestore_path2 = filestore_path + '2'
-
-    with mock.patch('common.new_process.execute') as mocked_execute:
-        filestore_utils.cp(filestore_path, filestore_path2, parallel=True)
-        mocked_execute.assert_called_with(
-            ['gsutil', '-m', 'cp', filestore_path, filestore_path2],
-            expect_zero=True)
+class TestLocalFilestore:
```
lszekeres, line 25: Is the class needed, given that we have one test?

common/test_filestore_utils.py
```diff
@@ -22,26 +22,63 @@
...
+    """Tests for switching on local_filestore."""
+    FILESTORE = '/experiment_data'
```
lszekeres, line 27: Where is this used?

common/test_filestore_utils.py
```diff
@@ -22,26 +22,63 @@
...
+    DIR1 = '/dir1'
+    DIR2 = '/dir2'
```
lszekeres, line 29: Do these need to be globals?

common/test_filestore_utils.py
```diff
@@ -22,26 +22,63 @@
...
+        with mock.patch('common.new_process.execute') as mocked_execute:
+            filestore_utils.cp(self.DIR1, self.DIR2, recursive=True)
+            assert 'gsutil' not in mocked_execute.call_args_list[0][0][0]
+
+
+class TestGsutil():
+    """Tests for switching on gsutil."""
+    FILESTORE = 'gs://experiment-data'
+    LOCAL_DIR = '/dir'
+    GSUTIL_DIR = 'gs://fake_dir'
```
lszekeres, line 42: consider using GOOGLE_STORAGE_DIR or GCS_DIR

common/test_filestore_utils.py
```diff
@@ -22,26 +22,63 @@
...
-            ['gsutil', '-m', 'cp', filestore_path, filestore_path2],
-            expect_zero=True)
+class TestLocalFilestore:
+    """Tests for switching on local_filestore."""
+    FILESTORE = '/experiment_data'
+    DIR1 = '/dir1'
+    DIR2 = '/dir2'
+
+    def test_local_filestore_on(self, use_local_filestore):  # pylint: disable=unused-argument
+        """Tests that local_filestore switches on correctly."""
```
lszekeres, line 32: Consider saying: 'Tests if local_filestore is used when experiment_filestore is a local path'

common/test_filestore_utils.py
```diff
@@ -22,26 +22,63 @@
...
+
+
+class TestGsutil():
+    """Tests for switching on gsutil."""
+    FILESTORE = 'gs://experiment-data'
+    LOCAL_DIR = '/dir'
+    GSUTIL_DIR = 'gs://fake_dir'
+
+    def test_gsutil_on(self, fs, use_gsutil):  # pylint: disable=invalid-name,unused-argument
+        """Tests that gsutil switches on correctly."""
```
lszekeres, line 45: Consider saying: 'Tests if gsutil is used when experiment_filestore is a google storage path'

common/test_filestore_utils.py
jonathanmetzman: +1

common/test_local_filestore.py
```diff
@@ -47,6 +47,16 @@ def test_ls_non_must_exist(tmp_path):
...
 
 
+def test_ls_one_file_per_line(tmp_path):
+    """Tests ls will list files as one per line."""
+    dir_path = tmp_path
+    file1 = dir_path / 'file1'
+    file2 = dir_path / 'file2'
+    open(file1, "w+").close()
+    open(file2, "w+").close()
+    assert local_filestore.ls(str(dir_path))[1].count('\n') == 2
```
jonathanmetzman, line 57: Use `.output` instead of `[1]` please.

common/test_filestore_utils.py
```diff
@@ -22,26 +22,63 @@
...
+
+        with mock.patch('common.new_process.execute') as mocked_execute:
+            fs.create_dir(self.LOCAL_DIR)
+            filestore_utils.cp(self.LOCAL_DIR, self.GSUTIL_DIR, recursive=True)
+            mocked_execute.assert_called_with(
+                ['gsutil', 'cp', '-r', self.LOCAL_DIR, self.GSUTIL_DIR],
+                expect_zero=True)
+
+    def test_keyword_args(self, use_gsutil):  # pylint: disable=unused-argument
```
jonathanmetzman, line 54: I think we need some tests that ensure parallel doesn't affect local_filestore

common/test_filestore_utils.py
```diff
@@ -22,26 +22,63 @@
...
-        mocked_execute.assert_called_with(
-            ['gsutil', '-m', 'cp', filestore_path, filestore_path2],
-            expect_zero=True)
+class TestLocalFilestore:
+    """Tests for switching on local_filestore."""
+    FILESTORE = '/experiment_data'
+    DIR1 = '/dir1'
+    DIR2 = '/dir2'
+
+    def test_local_filestore_on(self, use_local_filestore):  # pylint: disable=unused-argument
```
jonathanmetzman, line 31: I think the name `test_using_local_filestore` is more consistent with `use_local_filestore`

common/test_filestore_utils.py
jonathanmetzman: +1 please don't use so many classes in tests. I think `test_gsutil` is one of the rare files where I do this.

common/test_filestore_utils.py
zchcai: @jonathanmetzman wdyt? same for the next one.

common/test_filestore_utils.py
zchcai: I delete that, since right now, there are 4 related `# pylint: disable=unused-arguments` for `parallel` in their arguments.
I don't think we need to write a unit test for it since the argument is not used at all.

Previously, if we have `local_filestore_command` then I can check whether `parallel` has passed to that function. However, since we deleted that function, it is also awkward to write unit tests.

common/test_filestore_utils.py
zchcai: Alright.
Previously I used `switch_on` so I used `on` here.
Nvm

common/test_filestore_utils.py
zchcai: The suggested comment is more consistent with current implementation, where `_using_gsutil()` function checks the format of experiment_filestore path.
However, this implementation is depended on two things:
(1) first, user set `local_experiment` as true
(2) during the read_and_valid_input function, it will check `experiment_filestore` format as Posix format if `local_experiment` is true.
(3) Then, when we use the filestore_utils, it checks the path.

In fact, it is not because of the path setting that introduces the  local running, but the `local_experiment` setting. Choosing the current implementation is a tradeoff and way to utilize current structure, where `experiment_filestore` is passed to all layers of running environments.

jonathanmetzman: We can switch filestore_utils to use the local impl if `LOCAL_EXPERIMENT` env var is set

common/filestore_utils.py
jonathanmetzman: Don't think so, it's considered better to do imports at the toplevel (though our usage of the modules is probably not "ideal" either).

common/test_filestore_utils.py
zchcai: Right. Do we switch it in this PR?

zchcai: If we switch this on, maybe we can use `experiment_utils.is_local_experiment()` to replace the `_using_gsutil()`? And also pass environment variable `LOCAL_EXPERIMENT` to docker containers.

zchcai: This is also related to this issue: https://github.com/google/fuzzbench/issues/398

lszekeres: Sorry, my comment was about making the docstring more clear: i.e., consider replacing " switches on correctly" with something more explicit, so it's more clear what it is testing exactly.

common/test_local_filestore.py
```diff
@@ -47,6 +47,16 @@ def test_ls_non_must_exist(tmp_path):
...
 
 
+def test_ls_one_file_per_line(tmp_path):
+    """Tests ls will list files as one per line."""
+    dir_path = tmp_path
+    file1 = dir_path / 'file1'
+    file2 = dir_path / 'file2'
+    open(file1, "w+").close()
+    open(file2, "w+").close()
+    assert local_filestore.ls(str(dir_path)).output.count('\n') == 2
```
lszekeres, line 57: why not compare to expected output, ie something like "file1\nfile2\n" ?

jonathanmetzman: +1

common/test_filestore_utils.py
jonathanmetzman: Just check that parallel doesn't affect the command passed to `execute` this is the more important thing to test anyway.

common/test_local_filestore.py
zchcai: May I know the reason why a complete output comparison is better?

jonathanmetzman: It's much more obvious what is being tested. The current version is a bit odd and implicit.

zchcai: (1) others will know exactly what the execution results should be.
(2) only use `\n` cannot distinguish the running results between `-l` and `-1`.

common/test_filestore_utils.py
```diff
@@ -18,30 +18,110 @@
...
 
-def test_keyword_args():
+def test_using_local_filestore(fs, use_local_filestore):  # pylint: disable=invalid-name,unused-argument
+    """Tests that local_filestore is used in local running settings."""
+    fs.create_dir(LOCAL_DIR)
+    fs.create_dir(LOCAL_DIR_2)
+
+    with mock.patch('common.new_process.execute') as mocked_execute:
+        filestore_utils.cp(LOCAL_DIR, LOCAL_DIR_2, recursive=True)
+        assert 'gsutil' not in mocked_execute.call_args_list[0][0][0]
```
jonathanmetzman, line 34: Can you use pdb and figure out what `call_args_list[0][0]` is? If it's just the command than we don't need [0][0][0].

common/test_local_filestore.py
```diff
@@ -47,6 +47,16 @@ def test_ls_non_must_exist(tmp_path):
     local_filestore.ls(str(file_path), must_exist=False)
 
 
+def test_ls_one_file_per_line(tmp_path):
+    """Tests ls will list files as one per line."""
+    dir_path = tmp_path
+    file1 = dir_path / 'file1'
+    file2 = dir_path / 'file2'
+    open(file1, "w+").close()
```
jonathanmetzman, line 55: nit: use `with`

common/test_filestore_utils.py
zchcai: It seems we do need [0][0][0], otherwise it is not the actually argument list: 
```
-> assert 'gsutil' not in mocked_execute.call_args_list[0][0][0]
(Pdb) p mocked_execute.call_args_list
[call(['cp', '-r', '/dir', '/dir2'], expect_zero=True)]
(Pdb) p mocked_execute.call_args_list[0]
call(['cp', '-r', '/dir', '/dir2'], expect_zero=True)
(Pdb) p mocked_execute.call_args_list[0][0]
(['cp', '-r', '/dir', '/dir2'],)
(Pdb) p mocked_execute.call_args_list[0][0][0]
['cp', '-r', '/dir', '/dir2']
(Pdb)
```

jonathanmetzman: ACK

PR #423, opened 2020-06-11
Aflcc integration

laurentsimon:
This aims at testing the integration for https://github.com/Samsung/afl_cc


fuzzers/aflcc/fuzzer.py
```diff
@@ -0,0 +1,344 @@
...
+
+def is_benchmark(name):
+    """ Check if the benchmark contains the string 'name' """
+
+    # TODO: return name in os.environ['BENCHMARK]
+    # return benchmark is not None and name in benchmark
+
+    # OSS-fuzz benchmarks do not have a consistent folder content
+    # but typically have the name of the benchmark in the PWD
+    pwd_dir = os.getenv("PWD", None)
```
jonathanmetzman, line 33: `os.getcwd()`

fuzzers/aflcc/fuzzer.py
```diff
@@ -0,0 +1,344 @@
...
+    command += [
+        '--',
+        target_binary,
+        # Pass INT_MAX to afl the maximize the number of persistent loops it
+        # performs.
+        '2147483647'
+    ]
+    print('[run_fuzzer] Running command: ' + ' '.join(command))
+    output_stream = subprocess.DEVNULL if hide_output else None
+    subprocess.call(command, stdout=output_stream, stderr=output_stream)
```
jonathanmetzman, line 309: Use `check_call` `call` doesn't exception on failure.

fuzzers/aflcc/fuzzer.py
```diff
@@ -0,0 +1,344 @@
...
+    # but typically have the name of the benchmark in the PWD
+    pwd_dir = os.getenv("PWD", None)
+    if pwd_dir is None:
+        raise ValueError('PWD not defined')
+
+    if name in pwd_dir:
+        return True
+
+    # non-OSS-fuzz benchmarks
+    src_dir = os.getenv("SRC", None)
```
jonathanmetzman, line 41: nit single quotes for strings (other than docstrings please).

fuzzers/aflcc/fuzzer.py
```diff
@@ -0,0 +1,344 @@
...
+    # OSS-fuzz benchmarks do not have a consistent folder content
+    # but typically have the name of the benchmark in the PWD
+    pwd_dir = os.getenv("PWD", None)
+    if pwd_dir is None:
+        raise ValueError('PWD not defined')
+
+    if name in pwd_dir:
+        return True
+
+    # non-OSS-fuzz benchmarks
```
jonathanmetzman, line 40: nit: being comments with capital letters and end with periods please.

fuzzers/aflcc/fuzzer.py
```diff
@@ -0,0 +1,344 @@
...
+        if is_benchmark(target):
+            return [os.path.join(os.environ['OUT'], fuzzname)]
+
+    # For the reamining oss-projects, use some heuristics.
+    # We look for binaries in the OUT directory and take it as our targets.
+    # Note that we may return multiple binaries: this is necessary because
+    # sometimes multiple binaries are generated and we don't know which will
+    # be used for fuzzing (e.g., zlib benchmark)
+    # TODO(laurentsimon): hardcode targets for oss-fuzz projects in the 'targets' dictionary above
+    out_dir = os.getenv('OUT', None)
```
jonathanmetzman, line 193: I think instead of using getenv and checking `if None` and excepting if is, you could just use os.environ. It will do the same thing.

fuzzers/aflcc/fuzzer.py
```diff
@@ -0,0 +1,344 @@
...
+    os.environ['LLVM_CONFIG'] = 'llvm-config-3.8'
+    os.environ['CC'] = '/afl/aflc-gclang'
+    os.environ['CXX'] = '/afl/aflc-gclang++'
+    #  -L/ -lAflccMock -lpthread
+    os.environ['FUZZER_LIB'] = '/libAFL.a'
+
+    # Fix FUZZER_LIB for various benchmarks
+    fix_fuzzer_lib()
+
+def get_fuzz_targets():
```
jonathanmetzman, line 168: Why didn't you copy `fuzzer_utils.get_fuzz_target_binary`?

laurentsimon: I actually don't know what the name of the fuzz target is, because non-oss-fuzz projects don't set FUZZ_TARGET. So I just look for binaries in /src. For a few benchmarks, I did take the time to run the fuzzer and see which binary was called, so I hardcoded their names. Is FUZZ_TARGET set for all projects now?

fuzzers/aflcc/fuzzer.py
laurentsimon: done

fuzzers/aflcc/fuzzer.py
laurentsimon: done

fuzzers/aflcc/fuzzer.py
laurentsimon: done

fuzzers/aflcc/fuzzer.py
laurentsimon: done

fuzzers/aflcc/fuzzer.py
jonathanmetzman: Oh right. Yeah I don't think so. Need to land https://github.com/google/fuzzbench/pull/263

laurentsimon: ok, so I'll fix this once it's landed. Thx!

.github/workflows/fuzzers.yml
```diff
@@ -31,6 +31,7 @@ jobs:
           - lafintel
           - libfuzzer
           - mopt
+          - aflcc
```
jonathanmetzman, line 34: nit: alpha-order please.

laurentsimon:
> I think you need to add this fuzzer to `.github/workflows/fuzzers.yml`.
> Also please look into why presubmit fails.

done. The presubmit was failing because the license was missing fr some files. I added this. Note that when I run `make presubmit` on my machine, I get `Fatal Python error: Cannot recover from stack overflow.` ... and the following calls are stuck in an infinite loop:

```
File "/usr/local/google/home/laurentsimon/fuzzing/platform/fuzzbench/common/retry.py", line 88 in _wrapper
  File "/usr/local/google/home/laurentsimon/fuzzing/platform/fuzzbench/common/logs.py", line 211 in info
  File "/usr/local/google/home/laurentsimon/fuzzing/platform/fuzzbench/common/retry.py", line 64 in handle_retry
```

jonathanmetzman:
Can you set `FORCE_LOCAL=True` please?

jonathanmetzman:
Is your machine a cloudtop by any chance?

laurentsimon:
> Is your machine a cloudtop by any chance?

yes it is! `FORCE_LOCAL=True` solved the stack overflow pb, thx!

jonathanmetzman:
This failed on perl. Because we fail early (probably bad thing), I don't know if it works with the next benchmarks:
`proj4-2017-08-14
`re2-2014-12-09`
`vorbis-2017-12-11`
 `woff2-2016-05-06`
I'll take your word for it that it does. I think it actually is fine even if it doesn't FuzzBench can handle this gracefully.

laurentsimon:
I did not know about the perl benchmark. Has it been added recently?
proj4-2017-08-14 re2-2014-12-09 vorbis-2017-12-11 woff2-2016-05-06 should
compile as I tested them before. I will run a local build to confirm.

jonathanmetzman:
It's newer than our other benchmarks but we don't use it because of issues discussed in https://github.com/google/fuzzbench/issues/304
Since it's not used, I might remove it from CI.

PR #426, opened 2020-06-11
Ensure FORCE_LOCAL=1 for presubmit.

jonathanmetzman:
Many googlers (using cloudtop) need this and it will probably never hurt
to have.

### DIFF
--- a/common/utils.py
+++ b/common/utils.py
@@ -20,12 +20,17 @@ import urllib.error
 
 ROOT_DIR = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
 
+assert not (os.getenv('FORCE_NOT_LOCAL') and os.getenv('FORCE_LOCAL')), (
+    'You can\'t set FORCE_LOCAL and FORCE_NOT_LOCAL environment variables to '
+    'True at the same time. If you haven\'t set either of these and/or don\'t '
+    'understand why this is happening please file a bug.')
+
 # pylint: disable=invalid-name
 _is_local = None
 
 if os.getenv('FORCE_NOT_LOCAL'):
     # Allow local users to force is_local to return False. This allows things
-    # like stackdriver logging to happen when running code locally.
+    # like logging to happen when running code locally.
     _is_local = False
 
 if os.getenv('FORCE_LOCAL'):
--- a/presubmit.py
+++ b/presubmit.py
@@ -16,6 +16,12 @@
 # pylint: disable=wrong-import-position
 import os
 
+# Many users need this if they are using a Google Cloud instance for development
+# or if their system has a weird setup that makes FuzzBench think it is running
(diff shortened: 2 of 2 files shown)


presubmit.py
```diff
@@ -13,10 +13,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Presubmit script for fuzzbench."""
+# pylint: disable=wrong-import-position
+import os
+
+os.environ['FORCE_LOCAL'] = '1'
+if 'FORCE_NOT_LOCAL' in os.environ:
```
laurentsimon: should we throw an error instead of silently deleting the env variable? If a user passed that argument, they expect it to take effect, right?

jonathanmetzman: Yeah I just deleted that part.

presubmit.py
```diff
@@ -13,10 +13,19 @@
...
+# pylint: disable=wrong-import-position
+import os
+
+# Many users need this if they are using a Google Cloud instance for development
+# or if their system has a weird setup that makes FuzzBench think it is running
+# on Google Cloud. It's unlikely that setting this will mess anything up so set
+# it.
+# TODO(metzman): Make local the default setting and propagate 'NOT_LOCAL' to all
+# production environments so we don't need to worry about this any more.
+os.environ['FORCE_LOCAL'] = '1'
```
laurentsimon, line 25: what if the user passed FORCE_NOT_LOCAL? Shall we detect that we're using both FORCE_NOT_LOCAL and FORCE_LOCAL and throw an error (since they are mutually exclusive)? Maybe this case never happens in practice?

jonathanmetzman: I deleted this, good point.

jonathanmetzman: Sorry. I assumed I was replying to your previous comment.

jonathanmetzman: I added an assertion to ensure this isn't happening.

common/utils.py
```diff
@@ -20,12 +20,17 @@
 
 ROOT_DIR = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
 
+assert not (os.getenv('FORCE_NOT_LOCAL') and os.getenv('FORCE_LOCAL')), (
+    'You can\'t set FORCE_LOCAL and FORCE_NOT_LOCAL environment variables to '
+    'True at the same time. If you haven\'t set either of these and/or don\'t '
+    'understand why this is happening please file a bug.')
+
```
lszekeres, line 27: In the future consider dropping these flags and have a single flag like 'RUNNING_ON_GCE' or 'GOOGLE_CLOUD_LOGGING'. Local logging should be the default not just for presubmit but in general for any tool or unit test started manually, so developers never have to specify this flag. (with this PR we still need to)

Having to set this flag for the GCE service on the other hand is fine, because it'd be set in scripts anyway.

Issue #427, opened 2020-06-11
Benchmark libFuzzer's comparison instrumentation feature


jonathanmetzman:
@kcc + @morehouse would like some hard evidence on the utility of this feature (it's similar to lafintel).
This feature has a runtime part, and a compile-time instrumentation part.
In our discussion, we decided there are two ways this can be tested:
1. Set `-use_cmp=0` at runtime. If this does as well as normal libFuzzer, then the feature is useless.
2. Turn it off at compile time. This is the right way to do it. If 1 is worse we need this method to verify that it isn't the slowdown without the benefit causing problems.

@kcc/@morehouse could you make a copy of `fuzzers/libfuzzer` that sets the right flags here https://github.com/google/fuzzbench/blob/master/fuzzers/libfuzzer/fuzzer.py#L27 for option 2? Alternatively can you give me the flags to set and I'll do it.

Thanks!

morehouse:
I'm pretty sure we just need `-fsanitize=fuzzer-no-link -fno-sanitize-coverage=trace-cmp`.

jonathanmetzman:
https://www.fuzzbench.com/reports/2020-06-12/index.html
I think the findings here are similar to @vanhauser-thc 's on comparison instrumentation.
In most cases not doing comparison instrumentation seems very slightly better (probably because of improved speed).
But in some cases like libpcap_fuzz_both, it makes an extremely large difference.

jonathanmetzman:
CC @dokyungs 
I think we've done this already. Closing.

PR #432, opened 2020-06-12
[libfuzzer] Add libfuzzer variant that doesn't use cmp instrumentation

jonathanmetzman:
Related to #427.

### DIFF
--- a/fuzzers/libfuzzer_nocmp/fuzzer.py
+++ b/fuzzers/libfuzzer_nocmp/fuzzer.py
@@ -1,0 +1,42 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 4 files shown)


jonathanmetzman:
CC @morehouse

fuzzers/libfuzzer_nocmp/fuzzer.py
```diff
@@ -0,0 +1,41 @@
...
+# limitations under the License.
+"""Integration code for libFuzzer fuzzer."""
+
+import os
+
+from fuzzers import utils
+from fuzzers.libfuzzer import fuzzer as libfuzzer_fuzzer
+
+
+def build():
```
inferno-chromium, line 22: nit: similar to run_fuzzer, you can have a run_build with extra_cflags argument. i guess we will do more libfuzzer experiments, so good to keep libfuzzer's fuzzer.py that way.

jonathanmetzman: I think we plan on using runtime variants for this.

Issue #292, opened 2020-05-01
Commandline fuzzer variants break the builder integration tests


jonathanmetzman:
It breaks the tests test_build_changed_fuzzers/test_build_changed_benchmarks/build_measurers_benchmarks.
Those tests don't have the experiment config dir because they are run outside of an experiment.

If there's not an easy fix for this, I think we should just get rid of this feature. It hasn't been used, checked-in variants is sufficient, and I don't think the complexity is worth it. Not good for random parts of fuzzbench to be coupled with this feature.

jonathanmetzman:
The reason why this test isn't run anywhere (but my desktop :-) is because it tests building on GCB.
We can't really do this from most users machines and not from GitHub actions, so we just use local builds. I use this test to verify my changes to the gcb files work.

PR #436, opened 2020-06-15
Use fuzzer variants for run-time flags only.

inferno-chromium:
Remove complex and unmaintainable build argument support.
For fuzzers needing different build time variants should
be just use a different fuzzer directory e.g.
"fuzzer_build_config_1"

### DIFF
--- a/experiment/run_experiment.py
+++ b/experiment/run_experiment.py
@@ -157,7 +157,7 @@ def validate_fuzzer(fuzzer: str):
 
 def validate_fuzzer_config(fuzzer_config):
     """Validate |fuzzer_config|."""
-    allowed_fields = ['name', 'fuzzer_environment', 'build_arguments', 'fuzzer']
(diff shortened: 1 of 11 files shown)


inferno-chromium:
Just a sec, i might have to bring UNDERLYING_FUZZER back

inferno-chromium:
Not ready for review, need to fix generate_makefile.py

experiment/run_experiment.py
```diff
@@ -157,21 +157,16 @@ def validate_fuzzer(fuzzer: str):
 
 def validate_fuzzer_config(fuzzer_config):
```
jonathanmetzman, line 158: I really want to get rid of command line fuzzer variants: https://github.com/google/fuzzbench/issues/292

inferno-chromium: Will remove tomorrow.

PR #437, opened 2020-06-15
Important performance fix

vanhauser-thc:
When I integrated AFLfast 13 months ago I blindly copied the code and that activated a change that had a lasting performance impact - for every single fuzzing attempt there would be a hash performed on the result map and compared to all queue entries. this resulted in an 10-15% performance descrease, also if non AFLfast schedules were used.
This performance impact is the reason why AFLfast does not perform as well as AFL here on fuzzbench.
Today I finally found that cause and fixed it to only be done on AFfast schedules. And now afl++ is faster than afl :)
@mboehme


vanhauser-thc:
will redo.

PR #438, opened 2020-06-15
afl++ performance fix

vanhauser-thc:
When I integrated AFLfast 13 months ago I blindly copied the code and that activated a change that had a lasting performance impact - for every single fuzzing attempt there would be a hash performed on the result map and compared to all queue entries. this resulted in an 10-15% performance descrease, also if non AFLfast schedules were used.

This performance impact is the reason why AFLfast does not perform as well as AFL here on fuzzbench.

Today I finally found that cause and fixed it to only be done on AFfast schedules. And now afl++ is faster than afl :)

@mboehme

### DIFF
--- a/fuzzers/aflplusplus/builder.Dockerfile
+++ b/fuzzers/aflplusplus/builder.Dockerfile
@@ -24,7 +24,7 @@ RUN apt-get update && \
 # Set AFL_NO_X86 to skip flaky tests.
 RUN git clone https://github.com/AFLplusplus/AFLplusplus.git /afl && \
     cd /afl && git checkout dev && \
-    git checkout fc26001b50d27a276d2d50af1dbcd4dfa3886de5 && \
+    git checkout 9c293b5b7b941d8046e77989f100d084a516d029 && \
     unset CFLAGS && unset CXXFLAGS && \
     AFL_NO_X86=1 CC=clang PYTHON_INCLUDE=/ make && \
     cd llvm_mode && make
--- a/fuzzers/aflplusplus_optimal/builder.Dockerfile
+++ b/fuzzers/aflplusplus_optimal/builder.Dockerfile
@@ -37,7 +37,7 @@ RUN apt-get update && \
 # Set AFL_NO_X86 to skip flaky tests.
 RUN git clone https://github.com/AFLplusplus/AFLplusplus.git /afl && \
     cd /afl && \
-    git checkout fc26001b50d27a276d2d50af1dbcd4dfa3886de5 && \
+    git checkout 9c293b5b7b941d8046e77989f100d084a516d029 && \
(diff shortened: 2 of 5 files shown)


mboehme:
Nice detective work @vanhauser-thc! Great to see how AFL++ has grown since FuzzBench was released.

vanhauser-thc:
@jonathanmetzman 
since the last run that just finished there were 3 major improvements:
 * fix shmem (hopefully for the final time ;-)
 * switch hashing and random algorithm which are faster, better and dont have futexes
 * fix the default scheduler that had aflfast additions included that cost up to 10% performance

So for the normal variants I except a 5-10% performance increase and for the shmem variants another 5-10% performance increase on top.

So can you please run the afl++ variants again?
I am doing a fuzzing training on Tuesday and also present on the changes - and to have updated results for that would be great!

jonathanmetzman:
I'll run another experiment today.
By the way, I started another experiment on Sunday that included your last shmem fixes:
http://commondatastorage.googleapis.com/fuzzbench-reports/2020-06-14-test5/index.html

(I was testing some of my own changes so that's why the report isn't in the standard location).

vanhauser-thc:
> I'll run another experiment today.

thanks!

> By the way, I started another experiment on Sunday that included your last shmem fixes:
> http://commondatastorage.googleapis.com/fuzzbench-reports/2020-06-14-test5/index.html

is that one completed? because I see a few being at 23h but most at between 0:30 and 1:45.
Those which completed with 23h have good results for shmem :)

jonathanmetzman:
It's not completed, still going.

jonathanmetzman:
Sorry I didn't get a chance to run the experiment last night, I will do so within the hour. 
I did notice this from the previous experiment http://commondatastorage.googleapis.com/fuzzbench-data/2020-06-14-test5/experiment-folders/curl_curl_fuzzer_http-aflplusplus_optimal/trial-162231/results/fuzzer-log.txt

aflplusplus might be segfaulting on some runs.

vanhauser-thc:
uh thanks for the heads up. looks like there is a bug in cmplog ... will investigate.

vanhauser-thc:
@jonathanmetzman I have started the same test case (curl_curl_fuzzer_http-aflplusplus_optimal) and letting it run in a gdb. I dont know how reliable this happens though.

I just implemented a workaround in afl++ to a very fresh bug in llvm-11 - maybe its the same thing. (IRB.Createcall(func, args, "debug name") does interpret "debug name" as a return variable name - even if the function was defined as void ....". that bug is just a 1-2 weeks old)

just in case - is it possible to preserve core files of afl-fuzz (in case I cannot reproduce it)?

vanhauser-thc:
Found the bug. it was in the new MOpt code - that is already replaced in the afl++ repo but the commit for the test is old. whelp.

Issue #251, opened 2020-04-22
Support running FuzzBench without Google Cloud


jonathanmetzman:
Local experiments are currently not truly local because they depend on Google Cloud Storage (GCS).
We should break this dependency.
We use gsutil which has bash-like commands for managing files on GCS (e.g. cp, rm etc).
Instead we may need to switch to a library that supports other s3-like solutions and/or the local filesystem.
Local support will also break after the queue branch is merged into master and will need to be fixed again.

jonathanmetzman:
@zchai please link to this issue on every pull request you do for this feature.

jonathanmetzman:
PRs that were already done:
https://github.com/google/fuzzbench/pull/407
https://github.com/google/fuzzbench/pull/406
https://github.com/google/fuzzbench/pull/402
https://github.com/google/fuzzbench/pull/394
The initial PR: https://github.com/google/fuzzbench/pull/8

jonathanmetzman:
Let's keep this open until there is nothing left to be done here.

zchcai:
> Let's keep this open until there is nothing left to be done here.

Sounds good.

inferno-chromium:
Basic support is in - https://google.github.io/fuzzbench/running-a-local-experiment

jonathanmetzman:
This is a tracking bug and should be kept open as long as there is work to do here.
Reopening since there is still a ton of stuff to do here.

inferno-chromium:
See also https://github.com/google/fuzzbench/issues/777 for the need to support local docker registry.

PR #441, opened 2020-06-16
set related environ var for local running

zchcai:
## Description

When running locally, the following `filestore_utils.cp` and `filestore_utils.rsync` rely on two environment variables value to switch on `local_filestore` to use:

https://github.com/google/fuzzbench/blob/69feb9c327094033ced66ec2aea362c54bbbbb2a/experiment/run_experiment.py#L299-L304

where the logic is calling `_using_gsutil`
https://github.com/google/fuzzbench/blob/69feb9c327094033ced66ec2aea362c54bbbbb2a/common/filestore_utils.py#L19-L23

And then `get_experiment_filestore_path()` will need to get `os.environ['EXPERIMENT']` and `os.environ['EXPERIMENT_FILESTORE']`:
https://github.com/google/fuzzbench/blob/69feb9c327094033ced66ec2aea362c54bbbbb2a/common/experiment_utils.py#L45-L48
https://github.com/google/fuzzbench/blob/69feb9c327094033ced66ec2aea362c54bbbbb2a/common/experiment_utils.py#L35-L37

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

It will not change the behavior for cloud running, when `experiment_filestore` starts with `gs://`.

### DIFF
--- a/experiment/run_experiment.py
+++ b/experiment/run_experiment.py
@@ -287,6 +287,7 @@ def copy_resources_to_bucket(config_dir: str, config: Dict):
             return None
         return tar_info
 
+    # Set environment variables to use corresponding filestore_utils.
     os.environ['EXPERIMENT_FILESTORE'] = config['experiment_filestore']
     os.environ['EXPERIMENT'] = config['experiment']
     experiment_filestore_path = experiment_utils.get_experiment_filestore_path()
--- a/experiment/run_experiment.py
+++ b/experiment/run_experiment.py
@@ -287,8 +287,10 @@ def copy_resources_to_bucket(config_dir: str, config: Dict):
             return None
         return tar_info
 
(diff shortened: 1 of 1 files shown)


zchcai:
This PR is related to #251, #8, #399, #405 and #422.

experiment/run_experiment.py
```diff
@@ -287,8 +287,10 @@ def filter_file(tar_info):
             return None
         return tar_info
 
-    experiment_filestore_path = os.path.join(config['experiment_filestore'],
-                                             config['experiment'])
+    os.environ['EXPERIMENT_FILESTORE'] = config['experiment_filestore']
```
jonathanmetzman, line 290: Leave a comment about why this is needed please.

zchcai: Done.

PR #442, opened 2020-06-16
Fix/improve queue-based measuring

jonathanmetzman:
1. Fix some bugs in naming templates/groups.
2. Use an experiment-specific queue so that multiple experiments can run at the same time
without interfering with one another.
3. (temporary) Move measurer worker scheduling into the measure because the scheduler usually
quits before the measurer is finished.
4. Try to measure next cycle after previous cycle is measured successfully. This reduces
bottlenecks. Since 3 can make some workers live longer than they need to, this reduces
the amount of time they are idle by ensuring we don't wait around for the last few tasks
to finish before reducing the number of tasks.
4. Rename FAIL_WAIT_SECONDS to WAIT_SECONDS in scheduler as it is done even on success.
5. Use a separate process for the scheduler.

### DIFF
--- a/experiment/measurer/measure_manager.py
+++ b/experiment/measurer/measure_manager.py
@@ -150,9 +150,10 @@ def measure_all_trials(experiment_config: dict, queue) -> bool:  # pylint: disab
     # now because the scheduler quits too early.
     schedule_measure_workers.schedule(experiment_config, queue)
 
-    # Poll the queue for snapshots and save them in batches until the workers
-    # are done processing each unmeasured snapshot. Then save any remaining
-    # snapshots.
+    # Poll the queue for snapshots and save them in batches, for each snapshot
+    # that was measured, schedule another task to measure the next snapshot for
+    # that trial. Do this until there are no more tasks and then save any
+    # remaining snapshots.
     snapshots = []
     snapshots_measured = False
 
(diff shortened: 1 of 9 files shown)


jonathanmetzman:
Currently we use the queue in a similar manner to how we did measuring before, we query the db and then put tasks for snapshots that need measuring on the queue. We should definitely change this to something fancier, like putting all measurement tasks on the queue at once with dependencies accounted for and/or taking advantage of delayed tasks. 
However, I think it is best to land this version now since it seems to work and we can improve incrementally. Part of my reasoning for this is that the queue branch changes so much that waiting to land this will block other work (such as @Shadoom7 's work on coverage or @zchcai 's work on local support and queue based arch).

I plan on merging the queue branch with master after this PR is landed.

I think queue based measuring might not work properly with experiments that fill up or almost fill up the quota, but I think these fixes are fine to add later.

experiment/measurer/measure_manager.py
```diff
@@ -88,33 +83,77 @@ def get_job_timeout():
...
+
+    jobs = enqueue_measure_jobs_for_unmeasured(experiment_config, queue)
+    if not jobs:
+        # If we didn't enqueue any jobs, then there is nothing to measure.
+        return False
+
+    initial_jobs = {job.id: job for job in jobs}
+
+    # TODO(metzman): Move this to scheduler or it's own process. It's here for
+    # now because the scheduler quits too early.
```
lszekeres, line 150: Consider removing this TODO, as we might be able to skip these steps, and get rid of the scheduler completely (as in queue arch proposal)

jonathanmetzman: I need to comment on that. A lot of the scheduler now involves controlling resource consumption, I assume it will be more of a refactor than a "remove" (much like how measurer wasn't removed but was split in two), but maybe you have something else in mind then me.

jonathanmetzman: I think there isn't harm in leaving the todo for now.

PR #444, opened 2020-06-16
local experiment support

zchcai:
## Description

In this PR, to enable a pure local running, we add some environment variables settings and other integration required editing.
It can run and generate a desired report, although there are some warnings showing up.

To run it locally, we need to set `local_experiment` as `true` in the experiment-config.yaml, which will set to `--experiment-config` option, and provide local absolute path for shared filestore.
Here is a sample `experiment-config.yaml`:
```
trials: 4
max_total_time: 400
cloud_project: fuzzbench
experiment_filestore: /usr/local/home/alice/fuzzbench/fuzzbench-data
report_filestore: /usr/local/home/alice/fuzzbench/fuzzbench-reports
local_experiment: true
```

## Motivation and Context

This is related to #251. After this, we can have a local support running.
It also starts the beginning for next architecture, the scaleble queue-based on-premise running.

## How Has This Been Tested?

Currently, we need an end-to-end test for it.

It will not affect cloud running behavior.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] Add end-to-end test for local running.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.


common/experiment_utils.py
```diff
@@ -18,7 +18,7 @@
 
 from common import environment
 
-DEFAULT_SNAPSHOT_SECONDS = 15 * 60  # Seconds.
+DEFAULT_SNAPSHOT_SECONDS = 15 * 6  # Seconds.
```
jonathanmetzman: Remove debug code.

experiment/dispatcher.py
```diff
@@ -35,7 +35,7 @@
 from experiment import reporter
 from experiment import scheduler
 
-LOOP_WAIT_SECONDS = 5 * 60
+LOOP_WAIT_SECONDS = 5 * 6
```
jonathanmetzman: Debug code

experiment/measurer.py
```diff
@@ -52,7 +52,7 @@
 
 NUM_RETRIES = 3
 RETRY_DELAY = 3
-FAIL_WAIT_SECONDS = 30
+FAIL_WAIT_SECONDS = 3
```
jonathanmetzman: Debug code

experiment/measurer.py
```diff
@@ -578,7 +578,10 @@ def set_up_coverage_binaries(pool, experiment):
     ]
     coverage_binaries_dir = build_utils.get_coverage_binaries_dir()
     if not os.path.exists(coverage_binaries_dir):
-        os.makedirs(coverage_binaries_dir)
+        try:
+            os.makedirs(coverage_binaries_dir)
+        except Exception as err:
+            logger.warning(str(err))
```
jonathanmetzman, line 584: What happens if you don't catch the exception?
I'm fine with not logging anything since this seems to be expected.

experiment/measurer.py
```diff
@@ -588,15 +591,21 @@ def set_up_coverage_binary(benchmark):
     coverage_binaries_dir = build_utils.get_coverage_binaries_dir()
     benchmark_coverage_binary_dir = coverage_binaries_dir / benchmark
     if not os.path.exists(benchmark_coverage_binary_dir):
-        os.mkdir(benchmark_coverage_binary_dir)
+        try:
+            os.mkdir(benchmark_coverage_binary_dir)
+        except Exception as err:
```
jonathanmetzman, line 596: Are you sure this exception also needs to be caught?

experiment/measurer.py
```diff
@@ -588,15 +591,21 @@ def set_up_coverage_binary(benchmark):
...
+            logger.warning(str(err))
     archive_name = 'coverage-build-%s.tar.gz' % benchmark
     archive_filestore_path = exp_path.gcs(coverage_binaries_dir / archive_name)
     filestore_utils.cp(archive_filestore_path,
                        str(benchmark_coverage_binary_dir))
     archive_path = benchmark_coverage_binary_dir / archive_name
-    tar = tarfile.open(archive_path, 'r:gz')
-    tar.extractall(benchmark_coverage_binary_dir)
-    os.remove(archive_path)
+    try:
```
jonathanmetzman, line 603: It's hard for me to beleive that all of these exceptions you catch only happen when running locally and cause things to break but don't happen when I run it on the cloud.

experiment/run_experiment.py
```diff
@@ -237,7 +237,7 @@ def set_up_fuzzer_config_files(fuzzer_configs):
 def start_experiment(experiment_name: str, config_filename: str,
                      benchmarks: List[str], fuzzer_configs: List[dict]):
     """Start a fuzzer benchmarking experiment."""
-    check_no_local_changes()
+    # check_no_local_changes()
```
jonathanmetzman: Debug code.
Although frankly I find this check very annoying. Esp for development. Maybe we should allow silencing it with an env var @inferno-chromium ?

experiment/run_experiment.py
```diff
@@ -364,6 +369,10 @@ def start(self):
             '/var/run/docker.sock:/var/run/docker.sock',
             '-v',
             shared_volume_volume_arg,
```
jonathanmetzman, line 371: Is this still needed?

experiment/scheduler.py
```diff
@@ -39,9 +39,9 @@
 # Give the trial runner a little extra time to shut down and account for how
 # long it can take to actually start running once an instance is started. 5
 # minutes is an arbitrary amount of time.
-GRACE_TIME_SECONDS = 5 * 60
+GRACE_TIME_SECONDS = 5 * 6
```
jonathanmetzman: debug

experiment/scheduler.py
```diff
@@ -39,9 +39,9 @@
...
 
-FAIL_WAIT_SECONDS = 10 * 60
+FAIL_WAIT_SECONDS = 1 * 60
```
jonathanmetzman: debug

experiment/scheduler.py
```diff
@@ -130,7 +130,7 @@ def end_expired_trials(experiment_config: dict):
     if not expired_instances:
         return
 
-    if not delete_instances(expired_instances, experiment_config):
+    if not experiment_utils.is_local_experiment() and not delete_instances(expired_instances, experiment_config):
```
jonathanmetzman, line 133: line is almost certainly too long.
I think nesting here is fine. Or do `if local: bulksave; return;`

experiment/resources/runner-startup-script-template.sh
```diff
@@ -28,19 +28,19 @@ do
...
 
-docker run {% if local_experiment %}-v {{host_gcloud_config}}:/root/.config/gcloud {% endif %}\
---privileged --cpus=1 --rm \
+docker run --privileged --cpus=1 --rm \
 -e INSTANCE_NAME={{instance_name}} \
 -e FUZZER={{fuzzer}} \
 -e BENCHMARK={{benchmark}} \
 -e EXPERIMENT={{experiment}} \
 -e TRIAL_ID={{trial_id}} \
 -e MAX_TOTAL_TIME={{max_total_time}} \
```
jonathanmetzman, line 39: Are you giving the containers access to the filestores? If not why not?

experiment/run_experiment.py
zchcai: Will test this later.

experiment/run_experiment.py
zchcai: Please refresh. I decomment things and check format now.

experiment/resources/runner-startup-script-template.sh
zchcai: We give. Sorry not make all changes pushed when I create the PR.

experiment/stop_experiment.py
```diff
@@ -26,9 +26,14 @@
 
 def stop_experiment(experiment_name, experiment_config_filename):
     """Stop the experiment specified by |experiment_config_filename|."""
+    experiment_config = yaml_utils.read(experiment_config_filename)
+
+    if experiment_config.get('local_experiment', False):
```
lszekeres, line 31: can we use `experiment_utils.is_local_experiment()` here too?

experiment/measurer.py
```diff
@@ -588,15 +591,21 @@ def set_up_coverage_binary(benchmark):
     coverage_binaries_dir = build_utils.get_coverage_binaries_dir()
     benchmark_coverage_binary_dir = coverage_binaries_dir / benchmark
     if not os.path.exists(benchmark_coverage_binary_dir):
-        os.mkdir(benchmark_coverage_binary_dir)
+        try:
+            os.makedirs(benchmark_coverage_binary_dir)
+        except Exception as err:  # pylint: disable=broad-except
+            logger.warning(str(err))
```
lszekeres, line 597: like above, why catch the exception?

experiment/measurer.py
lszekeres: +1 not clear why we catch exceptions here and turn them into warnings

experiment/measurer.py
zchcai: If there is not catching, the measure_loop process will end and then there will be no measurement and no report generation.

experiment/measurer.py
zchcai: Yes. Based on the previous initial experiment #399.

zchcai: Basically, these 3 catching are related to the multiple initialization for benchmark coverage binaries preparation. And some race conditions happen locally.
I think the cloud running should also meet similar problem. The possible reason why it does not show up for cloud running, could be:
1. The different database implementation return different values in this function:
https://github.com/google/fuzzbench/blob/69feb9c327094033ced66ec2aea362c54bbbbb2a/experiment/measurer.py#L572-L582
Specifically, this line should only do one preparation for one benchmark. However, currently it will do the same amount as how many trials in total.
https://github.com/google/fuzzbench/blob/69feb9c327094033ced66ec2aea362c54bbbbb2a/experiment/measurer.py#L582
2. `gsutil` will handle them internally. However, I guess this is not the reason since some behavior only relate to `os` module.

experiment/measurer.py
zchcai: Please see https://github.com/google/fuzzbench/pull/444#discussion_r441080002

experiment/stop_experiment.py
zchcai: I prefer use the self-contained information within a function, rather than a global setting. wdyt?

experiment/measurer.py
zchcai: Like above, explained.

experiment/measurer.py
jonathanmetzman: >1. The different database implementation return different values in this function:

[This may be possible](https://stackoverflow.com/questions/17223174/returning-distinct-rows-in-sqlalchemy-with-sqlite).
Can you try removing distinct and the catch and replacing it with `set()`?

experiment/measurer.py
jonathanmetzman: Please try to investigate why these functions are being called once per trial instead of once per benchmark as intentended (log the benchmarks on 585).

zchcai:
Basically, these 3 catching are related to the multiple initialization for benchmark coverage binaries preparation. And some race conditions happen locally.
I think the cloud running should also meet similar problem. The possible reason why it does not show up for cloud running, could be:
1. The different database implementation return different values in this function:
https://github.com/google/fuzzbench/blob/69feb9c327094033ced66ec2aea362c54bbbbb2a/experiment/measurer.py#L572-L582
Specifically, this line should only do one preparation for one benchmark. However, currently it will do the same amount as how many trials in total.
https://github.com/google/fuzzbench/blob/69feb9c327094033ced66ec2aea362c54bbbbb2a/experiment/measurer.py#L582
2. `gsutil` will handle them internally. However, I guess this is not the reason since some behavior only relate to `os` module.
@lszekeres @jonathanmetzman

zchcai:
If we don't catch those exceptions, then the measure_loop will terminate and there will be no measurement before report generation.

experiment/run_experiment.py
zchcai: Still needed in `local_build.py`.
https://github.com/google/fuzzbench/blob/master/experiment/build/local_build.py

zchcai:
> A lot of this CL looks like debug code. Please remove those.

Thanks! `git stash` and `git stash pop` really helps.

experiment/measurer.py
```diff
@@ -588,7 +588,7 @@ def set_up_coverage_binary(benchmark):
     coverage_binaries_dir = build_utils.get_coverage_binaries_dir()
     benchmark_coverage_binary_dir = coverage_binaries_dir / benchmark
     if not os.path.exists(benchmark_coverage_binary_dir):
-        os.mkdir(benchmark_coverage_binary_dir)
+        os.makedirs(benchmark_coverage_binary_dir)
```
lszekeres, line 591: consider using common.filesystem.create_dir()

zchcai: Good suggestion! Then the `non exist` check can also be deleted.
wdyt? @jonathanmetzman

zchcai: Done.

zchcai:
Two things wait to do but it seems we can do both of them in another two PRs separately:
1. An end-to-end test for local running: I prefer to add it in this PR.
2. The documentation update for local running: a later doc-only PR.
wdyt? @lszekeres @jonathanmetzman

experiment/test_data/local-experiment-config.yaml
```diff
@@ -0,0 +1,23 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+experiment: test-experiment
+trials: 4
+max_total_time: 86400
+cloud_project: fuzzbench
```
lszekeres, line 18: consider adding a # TODO comment to remove cloud_project

experiment/measurer.py
```diff
@@ -571,14 +571,13 @@ def measure_snapshot_coverage(fuzzer: str, benchmark: str, trial_num: int,
 
 def set_up_coverage_binaries(pool, experiment):
     """Set up coverage binaries for all benchmarks in |experiment|."""
-    benchmarks = [
+    benchmarks = {
         trial.benchmark for trial in db_utils.query(models.Trial).distinct(
```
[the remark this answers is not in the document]
zchcai: Maybe keep the `distinct` now. I'm not sure whether it will affect cloud running.

experiment/test_data/local-experiment-config.yaml
zchcai: OK

experiment/run_experiment.py
zchcai: Sure. Will do it in another PR.
Related to https://github.com/google/fuzzbench/issues/457.

experiment/measurer.py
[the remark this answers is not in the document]
zchcai: Related to https://github.com/google/fuzzbench/issues/458.

zchcai:
Though some issues exist, they have no harm for functionality in local running. As there is an increasing demand for local support, we will merge this PR and fix those issues one by one.
Also, add documentations and end-to-end run test in separate PRs.

experiment/measurer.py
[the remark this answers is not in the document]
jonathanmetzman: it won't affect cloud running, using set and distinct is redundant.

experiment/scheduler.py
```diff
@@ -585,23 +586,28 @@ def schedule_loop(experiment_config: dict):
...
             try:
-                if not handle_preempted and not any_pending_trials(experiment):
-                    # Only start handling preempted instances once every initial
-                    # trial was started. This ensures that .
-                    handle_preempted = True
-
                 schedule(experiment_config, pool)
-                if handle_preempted:
-                    trial_instance_manager.handle_preempted_trials()
+                if not local_experiment:
```
jonathanmetzman, line 602: I didn't quite mean this, sorry for being unclear.
Let's clean it up a bit. Let's change this so that handle_preempted never becomes True when running locally

zchcai: Got the idea. Let me fix it tonight.

zchcai: Fixed in #464

PR #450, opened 2020-06-17
Add documentation for run time configurations.

inferno-chromium:


jonathanmetzman:
I'm not sure I want runtime variants or if I do, what we should do with them. I need to think on this.

inferno-chromium:
> I'm not sure I want runtime variants or if I do, what we should do with them. I need to think on this.

Sure, think on this. I dont feel this is any complicated now, since we use same images across variants, and only env variables get additionally passed. Take your time! 
+ @lszekeres for thoughts too.

lszekeres:
> > I'm not sure I want runtime variants or if I do, what we should do with them. I need to think on this.
> 
> Sure, think on this. I dont feel this is any complicated now, since we use same images across variants, and only env variables get additionally passed. Take your time!
> 
> * @lszekeres for thoughts too.

I'd support variants that are small changes in the `fuzzer.py` (i.e., `fuzz()` and `build()` could take optional `variant` name, based on which small changes can be made). E.g., 
```python
...
def build(variant='default'):
  ...
  if variant = 'shmem':
    os.environ['AFL_LLVM_USE_TRACE_PC'] = '1'

def fuzz(variant='default'):
  ...
```

inferno-chromium:
We are leaning towards removing run-time variants for now, so closing this.

PR #452, opened 2020-06-17
Remove command line fuzzer variants.

### LINKED ISSUE #292: Commandline fuzzer variants break the builder integration tests

inferno-chromium:
Fixes #292.
Regular fuzzer variants only differ by the environment variables
that they set during runtime. So, test_build_changed_fuzzers
should be fine building all the fuzzers. Remove customizable
fuzzer variants through command line.

### DIFF
--- a/common/fuzzer_config_utils.py
+++ b/common/fuzzer_config_utils.py
@@ -13,26 +13,8 @@
 # limitations under the License.
 """Configuration helper functions."""
 
-import os
-
 from common import experiment_path as exp_path
 from common import yaml_utils
-
-
-def get_by_variant_name(fuzzer_variant_name):
-    """Get a fuzzer config based on a fuzzer's display name."""
-    config_directory = get_fuzzer_configs_dir()
-    for config_filename in os.listdir(config_directory):
-        config_absolute_filename = config_directory / config_filename
-        if fuzzer_variant_name == get_fuzzer_name(config_absolute_filename):
-            return yaml_utils.read(config_absolute_filename)
-
-    return None
-
-
-def get_underlying_fuzzer_name(fuzzer_name):
-    """Get the underlying fuzzer name from the configuration's display name."""
-    return get_by_variant_name(fuzzer_name)['fuzzer']
 
 
 def get_dir():
--- a/common/fuzzer_utils.py
+++ b/common/fuzzer_utils.py
@@ -173,3 +173,12 @@ def get_fuzzer_configs(fuzzers=None):
             names.add(name)
 
     return fuzzer_configs
+
+
+def get_by_variant_name(fuzzer_variant_name):
+    """Get a fuzzer config based on a fuzzer's display name."""
+    fuzzer_configs = get_fuzzer_configs(fuzzers=[fuzzer_variant_name])
+    if not fuzzer_configs:
+        return None
+
+    return fuzzer_configs[0]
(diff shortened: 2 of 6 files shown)


jonathanmetzman:
> Fixes #292.

Did you check this? You have to explicitly enable these tests with env vars.
I would think this doesn't fix it because it doesn't get rid of this https://github.com/google/fuzzbench/pull/452/files#diff-86be9ce6aca645968dc413c144eb6452L251
This is the annoying part of command line variants that breaks things I think, that experiments can't just look in fuzzers/ directory for fuzzers they need to look at config/fuzzers/$BLAH.

> Regular fuzzer variants only differ by the environment variables
> that they set during runtime. So, test_build_changed_fuzzers
> should be fine building all the fuzzers. Remove customizable
> fuzzer variants through command line.

inferno-chromium:
> > Fixes #292.
> 
> Did you check this? You have to explicitly enable these tests with env vars.
> I would think this doesn't fix it because it doesn't get rid of this https://github.com/google/fuzzbench/pull/452/files#diff-86be9ce6aca645968dc413c144eb6452L251
> This is the annoying part of command line variants that breaks things I think, that experiments can't just look in fuzzers/ directory for fuzzers they need to look at config/fuzzers/$BLAH.

Should be now fixed with latest changes. PTAL.
There is an unrelated failure due to our use of gs://experiment-data bucket in fuzzer-evaluator project, probably should change to fuzzbench-data (can do in another cl) ?
```
ERROR    root:logs.py:161 Executed command: "gsutil cp /tmp/tmp1rgpo2pg gs://experiment-data/test-experiment/build-logs/oss-fuzz-curl-fuzzer-libfuzzer-hash-fc0c75e2d3457542b9f498c50990d688eebc8dbd3cc77581ef00e72454485b9d.txt" returned: 1. Extras: {'output': 'Copying file:///tmp/tmp1rgpo2pg [Content-Type=application/octet-stream]...\nAccessDeniedException: 403 The project to be billed is associated with an absent billing account.\n/ [0 files][    0.0 B/ 71.3 KiB]```

**Comment by jonathanmetzman** [comment]
It's still broken if you do 
`PYTHONPATH=. python experiment/build/builder.py --fuzzers afl --benchmarks zlib_zlib_uncompress_fuzzer`

The feature is still here and will still break random things. I really would like to get rid of it.

**Comment by jonathanmetzman** [comment]
Same issue if you use the unnitest "TEST_BUILD_CHANGED_FUZZERS=afl PYTHONPATH=. pytest -k test_build_changed_fuzzers":

```
E       FileNotFoundError: [Errno 2] No such file or directory: '/work/config/fuzzer-configs'
```

jonathanmetzman:
Actually, if this wasn't landed then no wonder it is "broken" when I tried it.
Can you check if it's broken when doing the commands I did and then land please.

inferno-chromium:
> Actually, if this wasn't landed then no wonder it is "broken" when I tried it.
> Can you check if it's broken when doing the commands I did and then land please.

Yes it works, only failure is at storing the build logs step where WORK and EXPERIMENT_FILESTORE are not set. i think those should be set by tests.
File "/build/fuzzbench/experiment/build/build_utils.py", line 33, in store_build_logs

PR #456, opened 2020-06-17
MOpt fix

vanhauser-thc:
@jonathanmetzman all _optimal* targets with MOpt will crash, because at that commit a non-working addition from upstream was ported in the code.
I dont think the just started testrun make sense - better abort it and run it again with this PR applied.
sorry!


jonathanmetzman:
OK, i'll do another experiment later today. Tied up ATM.

Issue #445, opened 2020-06-16
make `cloud_project` config not required for local


zchcai:
Once #444 merged, `cloud_project` is actually a required configuration in both local and cloud running settings.
Suggest to make two changes:
1. Add required checking for it in `read_and_validate_input_config`.
2. Rename it to a general name, without term `local` or `cloud`. Since `project` is a bit common and may conflict to other things, I propose as `project_name`.

@lszekeres @jonathanmetzman

lszekeres:
I'd suggest the following: make GCE-specific fields (such as `cloud_project`, `cloud_compute_zone`, `preemptible_runners`, etc) sub-fields of say a `gce` field in the yaml config. e.g.:
```
gce:
  cloud_project: fuzzbench
  cloud_compute_zone: us-central1-a
  preemptible_runners: true
```

and `gce` should be optional (and unused in the local case). wdyt?

zchcai:
Currently, the logic runs that:
1. If `local_experiment` is not set or False, then cloud_* are required and `*_filestore` should start with `gs://`.
2. If `local_experiment` is True, then `*_filestore` should start with `/`.

I think the problem is, in fact, `cloud_project` is needed for local running in current implementation, thus we change its name and make it required no matter `local_experiment` is true or not.

zchcai:
In short, `cloud_project` is not optional and since it is not optional and not specific to cloud running, we should change its name.

lszekeres:
`cloud_project` is the name/identifier of the used Google Cloud project: a strictly Google Cloud specific property. we do not need it in the local case.

zchcai:
> `cloud_project` is the name/identifier of the used Google Cloud project: a strictly Google Cloud specific property. we do not need it in the local case.

If it is in this case, we may need to remove the usages of it when running locally.

jonathanmetzman:
> I'd suggest the following: make GCE-specific fields (such as `cloud_project`, `cloud_compute_zone`, `preemptible_runners`, etc) sub-fields of say a `gce` field in the yaml config. e.g.:
> 
> ```
> gce:
>   cloud_project: fuzzbench
>   cloud_compute_zone: us-central1-a
>   preemptible_runners: true
> ```
> 
> and `gce` should be optional (and unused in the local case). wdyt?

Interesting idea. I suppose this change would impact any code that used these fields as well.

jonathanmetzman:
[...]

PR #459, opened 2020-06-17
[doc] local running

zchcai:
The documentation update for users to run local experiments on their own machines.

Related to #251.

This will open for review once there will be no configuration option change.

Currently, it will be blocked until #445 resolved.


jonathanmetzman:
Heh, not sure if this should land before the queue stuff which will set local experiments a few weeks back
:)

zchcai:
> Heh, not sure if this should land before the queue stuff which will set local experiments a few weeks back
> :)

Maybe we can try to let both working together. :)

jonathanmetzman:
> > Heh, not sure if this should land before the queue stuff which will set local experiments a few weeks back
> > :)
> 
> Maybe we can try to let both working together. :)

Heh, wouldn't count on it.

zchcai:
After #461 merged, this can be open for merging.

docs/running-a-local-experiment/running_a_local_experiment.md
```diff
@@ -51,9 +51,11 @@ max_total_time: 86400
 # docker_registry: lab-server:5000
 
 # The local experiment folder that will store most of the experiment data.
+# Please use absolute path.
```
jonathanmetzman, line 54: nit: "Please use an absolute path".

zchcai:
Discussion offline: though we already add `docker_registry` option for local running (#461), we will not publish our local support documentation until we have at least a basic resource control option for local running.
Blocking until further enhancement. Probably after `queue` branch merged, `queue` branch works on local, and, works for builders and runners together with measures.
For other internal users, they can checkout [this commit](https://github.com/google/fuzzbench/commit/b16c35cb5e819c8d025a6857e7266b457ae5b889) and refer to here for a local run.

inferno-chromium:
@jonathanmetzman - can you take a final review pass and commit this.

inferno-chromium:
actually i will review it and add that line on resource control.

PR #461, opened 2020-06-18
remove `cloud_project` requirement for local run; add `docker_registry` as a required option for both settings

### LINKED ISSUE #445: make `cloud_project` config not required for local

zchcai:
Fix #445, related to #251.


experiment/run_experiment.py
```diff
@@ -66,14 +66,17 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
     and returns it."""
     config = yaml_utils.read(config_filename)
     filestore_params = {'experiment_filestore', 'report_filestore'}
-    cloud_config = {'cloud_compute_zone'}
-    string_params = cloud_config.union(filestore_params)
+    cloud_config = {'cloud_compute_zone', 'cloud_project'}
+    local_config = {'docker_registry'}
```
lszekeres, line 70: `docker_registry` will be used for local and cloud too, no? why `local_config =`?

zchcai: Here means, `docker_registry` is required for local.
For cloud, since `cloud_project` is required and then `docker_registry` is not required.

experiment/scheduler.py
```diff
@@ -715,8 +715,14 @@ def render_startup_script_template(instance_name: str, fuzzer: str,
     """Render the startup script using the template and the parameters
     provided and return the result."""
     fuzzer_config = fuzzer_config_utils.get_by_variant_name(fuzzer)
+    local_experiment = experiment_utils.is_local_experiment()
+    if not local_experiment:
+        experiment_config['docker_registry'] = (
+            experiment_utils.get_base_docker_tag(
+                experiment_config['cloud_project']))
```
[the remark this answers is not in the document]
zchcai: I prefer not to make `docker_registry` as a required one for cloud running. @jonathanmetzman wdyt?
Do we need to make `docker_registry` also required option while it can be generated by `cloud_project`, where `cloud_project` is already required.

zchcai:
I prefer not to make docker_registry as a required one for cloud running.
docker_registry can be generated by cloud_project, where cloud_project is already required.

experiment/scheduler.py
```diff
@@ -715,8 +715,14 @@ def render_startup_script_template(instance_name: str, fuzzer: str,
     """Render the startup script using the template and the parameters
     provided and return the result."""
     fuzzer_config = fuzzer_config_utils.get_by_variant_name(fuzzer)
+    local_experiment = experiment_utils.is_local_experiment()
```
jonathanmetzman, line 718: I want to minimize the number of times we branch on is_local_experiment.
Let's make docker_registry mandatory for every experiment.

experiment/run_experiment.py
```diff
@@ -382,14 +389,14 @@ def start(self):
...
-            set_cloud_project_arg,
-            '-e',
             sql_database_arg,
             '-e',
             set_experiment_filestore_arg,
             '-e',
             set_report_filestore_arg,
             '-e',
+            set_docker_registry_arg,
+            '-e',
```
jonathanmetzman, line 399: I don't think we need to set this env var. Really the only config vars that should be set in the env are ones that are used in startup-dispatcher.sh.

zchcai: Need to test. Probably we should keep it. Since cloud_project is passed to runner docker containers level, we should make sure the only usage of them is not related to docker_registry construction logic in cloud settings.

experiment/scheduler.py
zchcai: This reason sounds reasonable.

zchcai: By the way, this one can be removed. Actually I added the `dcoker_registry` here is because of unit tests.

experiment/scheduler.py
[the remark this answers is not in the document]
zchcai: Right, I already set that in the run_experiment.py.
Here is for unit test passing.
I will remove this and enhance the unit test environment.

inferno-chromium:
I don't like docker_registry field at all, we would need cloud_project for other things in future and from a user perspective, it is easier to know the cloud project they use vs a gcr.io link to put in attribute.

lszekeres:
> I don't like docker_registry field at all, we would need cloud_project for other things in future and from a user perspective, it is easier to know the cloud project they use vs a gcr.io link to put in attribute.

There's more context in https://github.com/google/fuzzbench/issues/445. tl;dr:

- We are introducing `docker_registry` because we will need it for on-prem.
- I think it wouldn't hurt the cloud case either, as it would make it possible to use say `docker.io` instead of `gcr.io`.
- We are not removing `cloud_project`, but make it unnecessary to be defined in the on-prem case, as it is not meaningful in that case. 

Wdyt?

experiment/run_experiment.py
inferno-chromium: Probably add a comment on this below, that "if not local_experiment"

experiment/run_experiment.py
```diff
@@ -66,14 +66,17 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
...
-    cloud_config = {'cloud_compute_zone'}
-    string_params = cloud_config.union(filestore_params)
+    cloud_config = {'cloud_compute_zone', 'cloud_project'}
+    local_config = {'docker_registry'}
+    string_params = cloud_config.union(filestore_params).union(local_config)
     int_params = {'trials', 'max_total_time'}
     required_params = int_params.union(filestore_params)
 
     local_experiment = config.get('local_experiment', False)
     if not local_experiment:
```
inferno-chromium, line 76: nit: don't need the not, can switch the order, if local_experiment: do that else:

experiment/run_experiment.py
```diff
@@ -120,6 +123,12 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
 
     if not valid:
         raise ValidationError('Config: %s is invalid.' % config_filename)
+
+    config['local_experiment'] = local_experiment
+    if not local_experiment and 'docker_registry' not in config:
+        config['docker_registry'] = experiment_utils.get_base_docker_tag(
+            config['cloud_project'])
```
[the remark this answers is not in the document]
inferno-chromium: So, this makes dockerr_registry even configurable for cloud project, that seems good ?

experiment/run_experiment.py
inferno-chromium: Will defer to Jonathan on this part of review.

experiment/scheduler.py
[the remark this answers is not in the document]
inferno-chromium: +1 to enhancing test!

inferno-chromium:
> > I don't like docker_registry field at all, we would need cloud_project for other things in future and from a user perspective, it is easier to know the cloud project they use vs a gcr.io link to put in attribute.
> 
> There's more context in #445. tl;dr:
> 
> * We are introducing `docker_registry` because we will need it for on-prem.
> * I think it wouldn't hurt the cloud case either, as it would make it possible to use say `docker.io` instead of `gcr.io`.
> * We are not removing `cloud_project`, but make it unnecessary to be defined in the on-prem case, as it is not meaningful in that case.
> 
> Wdyt?

Sorry, i looked at cl more closely now and it all makes sense!

experiment/run_experiment.py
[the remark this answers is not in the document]
zchcai: Yes! So for cloud running, in default, we use "gcr.io/`cloud_project`", but we can also specify a different `docker_registry` for use.
In fact, this PR is more like detaching the construction of the docker registry from `cloud_project`.

experiment/run_experiment.py
zchcai: Hi, I tried one experiment running without setting it but failed to build images.
Probably it is because `experiment/dispatcher.py` imports `experiment/build/builder.py`, which imports `experiment/build/local_build.py`, where `docker_registry` env is used.

Maybe we keep it as this way now?

experiment/run_experiment.py
zchcai: Done. Add a TODO comment above it.

zchcai: Basically, we can accept docker_registry for both settings, but only required in local setting.

Maybe in the future, we can make it required in both settings, especially when we know whether `cloud_project` can be removed or not. If `cloud_project` can be removed for cloud running, then it will be a good time to make `docker_registry` required for cloud setting, too.

lszekeres: I think the question whether `cloud_project` can be removed or not is orthogonal to the question of making `docker_registry` a mandatory common option for all platforms. The benefit of making `docker_registry` a common option - instead of deriving it in the cloud case - would be simplicity: 

(1) It'd be clear in all cases what it is set to (ie. to the value in the config file), and 

(2) we wouldn't need these additional branches in the code here and below (like `if local_experiment:` and `if not local_experiment and 'docker_registry' not in config:`).

At this point of the project, in my opinion, simplicity is higher priority than backwards compatibility.

wdyt?

zchcai: I'm fine for this suggestion. Let me change it accordingly.

docs/running-your-own-experiment/running_an_experiment.md
```diff
@@ -56,6 +56,9 @@ trials: 5
 # 1 day = 24 * 60 * 60 = 86400
 max_total_time: 86400
 
+# The docker registry for your fuzzbench experiments.
```
jonathanmetzman, line 59: Make the example `gcr.io/$FUZZBENCH` and mention that this is the default if using Google cloud.

experiment/run_experiment.py
```diff
@@ -424,8 +425,7 @@ def start(self):
         gcloud.robust_begin_gcloud_ssh(self.instance_name,
                                        self.config['cloud_compute_zone'])
 
-        base_docker_tag = experiment_utils.get_base_docker_tag(
-            self.config['cloud_project'])
+        base_docker_tag = self.config['docker_registry']
```
jonathanmetzman, line 428: Let's call this var `docker_registry`.

experiment/run_experiment.py
```diff
@@ -435,6 +435,7 @@ def start(self):
             '-e INSTANCE_NAME="{instance_name}" '
             '-e EXPERIMENT="{experiment}" '
             '-e CLOUD_PROJECT="{cloud_project}" '
+            '-e DOCKER_REGISTRY="{base_docker_tag}" '
```
jonathanmetzman, line 438: `{docker_registry}

experiment/run_experiment.py
jonathanmetzman: What's the error?
It's clear https://github.com/google/fuzzbench/pull/461/files#diff-b6cb9bdd02ffa26364e101706ffa0fc1R71 will break unless it is modified. The questions is whether we should modify it to support config file or not.
I only set configs as env vars if they are needed in the startup script otherwise I read them from the config file and pass them through to any functions that need them. This breaks that policy.
It may be fine to change this policy (we could make the code that reads the config set env vars for each config var) but if we do that it should be an intentional decision and we should do it all at once, not just in one place.

jonathanmetzman:
If youre a local user what would you even use for a docker registry? Just gcr.io/fuzzbench? If this is the case it might be best to make this optional, and if not specified set it to gcr.io/+$CLOUD_PROJECT for nonlocal or gcr.io/fuzzbench for local.

jonathanmetzman:
didn't mean to close.

experiment/run_experiment.py
lszekeres: > @lszekeres and @inferno-chromium I'd like your opinion on this: Should we continue the policy that (almost all) config variables are read from the config file and passed to functions that use them? Or could we change things to so that we set env vars for each config var. I think the latter would make code easier to write and shorter to read but possibly harder to understand.

Yes, in general, whenever possible, global state such as environment vars should be avoided, as it makes the code harder to reason about and maintain.

(Re. this particular line of code, i don't have enough context, so don't know what is the best thing to do here).

zchcai:
> If youre a local user what would you even use for a docker registry? Just gcr.io/fuzzbench? If this is the case it might be best to make this optional, and if not specified set it to gcr.io/+$CLOUD_PROJECT for nonlocal or gcr.io/fuzzbench for local.

it's useful for the multiple local machine case.

docs/running-your-own-experiment/running_an_experiment.md
```diff
@@ -56,6 +56,10 @@ trials: 5
...
+# In default, we use gcr.io/fuzzbench if using Google Cloud.
```
lszekeres, line 60: I think what we mean to say here is something like:
```
# If you use Google Cloud, this can be gcr.io/$PROJECT_NAME
```
Also, i think it should be 

docker_registry: gcr.io/$PROJECT_NAME
instead of 
docker_registry: gcr.io/$FUZZBENCH

no?

zchcai: Not very sure but I guess you are right. @jonathanmetzman wdyt?

common/benchmark_utils.py
```diff
@@ -45,21 +44,18 @@ def get_fuzz_target(benchmark):
     return fuzzer_utils.DEFAULT_FUZZ_TARGET_NAME
 
 
-def get_runner_image_url(benchmark, fuzzer, cloud_project):
+def get_runner_image_url(benchmark, fuzzer, docker_registry):
```
zchcai, line 47: All usage of `get_runner_image_url` has been changed accordingly:
```
~/fuzzbench$ rg get_runner_image_url -A 2
common/benchmark_utils.py
47:def get_runner_image_url(benchmark, fuzzer, docker_registry):
48-    """Get the URL of the docker runner image for fuzzing the benchmark with
49-    fuzzer."""

common/test_benchmark_utils.py
62:def test_get_runner_image_url(benchmark, expected_url, oss_fuzz_benchmark):
63-    """Test that we can get the runner image url of a benchmark."""
64:    assert benchmark_utils.get_runner_image_url(benchmark, 'fuzzer',
65-                                                DOCKER_REGISTRY) == expected_url
66-

experiment/scheduler.py
720:    docker_image_url = benchmark_utils.get_runner_image_url(
721-        benchmark, fuzzer_config['fuzzer'],
722-        experiment_config['docker_registry'])
~/fuzzbench$
```

common/benchmark_utils.py
```diff
@@ -45,21 +44,18 @@ def get_fuzz_target(benchmark):
...
-    base_tag = experiment_utils.get_base_docker_tag(cloud_project)
-    return '{base_tag}/runners/{fuzzer}/{benchmark}'.format(base_tag=base_tag,
-                                                            fuzzer=fuzzer,
-                                                            benchmark=benchmark)
+    return '{docker_registry}/runners/{fuzzer}/{benchmark}'.format(
+        docker_registry=docker_registry, fuzzer=fuzzer, benchmark=benchmark)
 
 
-def get_builder_image_url(benchmark, fuzzer, cloud_project):
+def get_builder_image_url(benchmark, fuzzer, docker_registry):
```
zchcai, line 54: All usage of `get_builder_image_url` has been changed accordingly:
```
~/fuzzbench$ rg get_builder_image_url -A 1
common/benchmark_utils.py
54:def get_builder_image_url(benchmark, fuzzer, docker_registry):
55-    """Get the URL of the docker builder image for fuzzing the benchmark with

experiment/build/local_build.py
70:    builder_image_url = benchmark_utils.get_builder_image_url(
71-        benchmark, 'coverage', environment.get('DOCKER_REGISTRY'))
~/fuzzbench$
```

common/test_benchmark_utils.py
```diff
@@ -19,7 +19,7 @@
 
 # pylint: disable=invalid-name,unused-argument
 
-CLOUD_PROJECT = 'fuzzbench'
+DOCKER_REGISTRY = 'gcr.io/fuzzbench'
```
zchcai, line 22: Only one usage of CLOUD_PROJECT in this file. And changed to DOCKER_REGISTRY.

experiment/run_experiment.py
zchcai: > > @lszekeres and @inferno-chromium I'd like your opinion on this: Should we continue the policy that (almost all) config variables are read from the config file and passed to functions that use them? Or could we change things to so that we set env vars for each config var. I think the latter would make code easier to write and shorter to read but possibly harder to understand.
> 
> Yes, in general, whenever possible, global state such as environment vars should be avoided, as it makes the code harder to reason about and maintain.
> 
> (Re. this particular line of code, i don't have enough context, so don't know what is the best thing to do here).

TODO added and issues created for this #484.
Thanks for this discussion.

zchcai:
> We should probably add
> `docker_registry: gcr.io/fuzzbench`
> to `service/experiment-config.yaml` as well.

This is related to cloud running configuration. I guess you are right.

PR #464, opened 2020-06-18
[NFC] `handle_preemptible` never True for local running

zchcai:
A format refactoring and make the code more readability.
Related to #444 and #251.


experiment/scheduler.py
```diff
@@ -598,16 +598,18 @@ def schedule_loop(experiment_config: dict):
...
             try:
+                if (not local_experiment and not handle_preempted and
+                        not any_pending_trials(experiment)):
+                    # This ensures that:
+                    # 1. handle_preempted will not becomes True when running
+                    #    locally.
+                    # 2. Only start handling preempted instances once every
+                    #    initial trial was started.
+                    handle_preempted = True
+
```
lszekeres, line 609: consider:
`handle_preempted = not(handle_preempted or local_experiment or any_pending_trials(experiment))`

jonathanmetzman: Don't think this works.
Let's say handle_preempted is True.
Then next time handle preempted is False.

zchcai: > consider:
> `handle_preempted = not(handle_preempted or local_experiment or any_pending_trials(experiment))`

Logically it is equivalent, but it makes harder for readers. The current way is better for readability.

PR #465, opened 2020-06-18
[test] local end to end test on pull request

zchcai:
## Description

The end-to-end test for local running.
We check whether a local run can generate report and experimental data within a time limit.

## Motivation and Context

This test is required for local support related pull request handling. Related to #251 and #444.


zchcai:
We may also put this test under ./experiment/test*. But I guess this is more general than a unit test, where ./experiment/test* is more of a place for unit test.

jonathanmetzman:
Didn't we discuss how this should be written as a unittest?

zchcai:
> Didn't we discuss how this should be written as a unittest?

Yes we mentioned that offline.

zchcai:
Some discussion offline. Maybe it will be better to provide another check as `make local-integration` to do future integration end-to-end tests, which is separate with current unit test check in `make presubmit` and is not required when submitting a PR.
@jonathanmetzman @lszekeres  any other comments?

jonathanmetzman:
I want it to be a test so that has a familiar interface for using locally. But we will also run it on PRs.

zchcai:
> I want it to be a test so that has a familiar interface for using locally. But we will also run it on PRs.
Right. I will create another PR for this feature.

PR #467, opened 2020-06-19
Entropic: Disable 5min Restarts

mboehme:
Currently, in fork mode, LibFuzzer runs a new instance (job) with a 5min timeout, merges the generated corpus into the main corpus, and starts a new instance with a 5min timeout, recurrently. Each LF instance starts with a very small subset of the main corpus with bias to the last seed files (not sure "last" according to which ordering).

In this PR, in fork mode, LibFuzzer runs a new instance only when a crash is encountered and merges the current corpus into the main corpus every five minutes. The new LF instance (job) always starts with the entire main corpus. Some seed files might have been deleted before the merge. So, there is a thread-unsafe quick check whether the file still exists before it is merged. Also, I don't think that RunOneMergeJob can be run concurrently (e.g., `-fork=10`), but it should be fine for the FuzzBench setting where `-fork=1`.

Curious about the impact on Entropic results.

### DIFF
--- a/fuzzers/entropic/patch.diff
+++ b/fuzzers/entropic/patch.diff
@@ -510,9 +510,9 @@ index d9e6b79..9a28e7b 100644
    while (auto Job = FuzzQ->Pop()) {
      // Printf("WorkerThread: job %p\n", Job);
 +    Job->Executing = true;
-+    int Sleep_ms = 300000;
++    int Sleep_ms = 5 * 60 * 1000;
 +    std::thread([=]() {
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Thanks!

jonathanmetzman:
I was planning on switching entropic to the libFuzzer option, but I guess we don't want to do this anymore? I think it definitely makes sense to use patching if you want to test tweaks.

mboehme:
Sure, why not. We can patch in and test those tweaks on the most recent LF source, as well.

PR #466, opened 2020-06-19
Add `expect_zero` back for filestore_utils.cp

zchcai:
This fix #499. There will be no `ERROR:root:Executed command: "cp ...`
This pattern follows the pattern in old implementation, https://github.com/google/fuzzbench/tree/ec27e1711cd4c4904e3c0b853f046c1d98b7b189
https://github.com/google/fuzzbench/blob/ec27e1711cd4c4904e3c0b853f046c1d98b7b189/experiment/measurer.py#L398-L402

https://github.com/google/fuzzbench/blob/ec27e1711cd4c4904e3c0b853f046c1d98b7b189/experiment/measurer.py#L536-L541

PR #475, opened 2020-06-22
[NFC] #466 minor comment fix

zchcai:
Fix https://github.com/google/fuzzbench/pull/466#discussion_r443660959


common/filestore_utils.py
```diff
@@ -39,8 +39,8 @@ def get_impl():
 
 def cp(source, destination, recursive=False, parallel=False, expect_zero=True):  # pylint: disable=invalid-name
     """Copies |source| to |destination|. If |expect_zero| is True then it can
-    raise subprocess.CalledProcessError. |parallel| is only effective for
-    gsutil."""
+    raise subprocess.CalledProcessError. |parallel| is only used when by the
```
jonathanmetzman, line 42: Shoot, i had a typo in my comments to you on the last PR.
Here is the correct version:
`|parallel| is only used by the gsutil implementation.`

Sorry!

common/filestore_utils.py
```diff
@@ -71,7 +71,7 @@ def rsync(  # pylint: disable=too-many-arguments
         options=None,
         parallel=False):
     """Syncs |source| and |destination| folders. |gsutil_options| and |parallel|
-    are only used in gsutil."""
+    are only used when by the gsutil implementation."""
```
jonathanmetzman, line 74: `|gsutil_options| and |parallel| are only used by the gsutil implementation.`

zchcai: Nvm. :)

Issue #479, opened 2020-06-22
clear containers or images - fuzzbench only or experiment only support


zchcai:
## *Who* is affected?
Machines have other docker containers running background, or have other images caches, not related to fuzzbench project.
Especially local runners and debuggers.

## *When* does this need occur?

Run `make clear-cache`.
https://github.com/google/fuzzbench/blob/ca24c304439466eab8331b1ba5bd5aaad76a58e8/Makefile#L48-L52
Also, there is need when doing local testing runs.

## Expected behavior (i.e. solution)

Several thoughts and anticipations:
1. Separate stoping/removing docker *containers* with removing docker *images* caches.
2. Only clear/remove fuzzbench related docker images or containers.
3. Can provide options for experiment-specific docker containers stopping. This is important for local end-to-end testing, where docker runners have been started while that test is failed. We want to stop and remove docker containers only related to that test experiment.

Possible solution is similar to ones in #470. We may need to specify the experiment name or other identifiers when creating docker containers to run.

## Other Comments
Related to #365 and #470.

zchcai:
#470 vs #479
#470 is for local stopping logic. It may include more than stopping unfinished docker containers.
#479 is for a user-friendly option as `make clear-cache`. We may split that into two, and support experiment-level or fuzzbench-level only cleaning.

Issue #470, opened 2020-06-20
local stop logic - clean/kill docker containers for failed local experiment running


zchcai:
## *Who* is the bug affecting?

The local running users, especially testing new changes in code. Probably, it also applies to Google Cloud based running.
Also, related to implement end-to-end local test running (#465)
.

## *When* does this occur?

When the experiment running has some problems in the middle and not gracefully stopped, the runners docker containers are still running until the preset `max_total_time` reaches.

## *Where* on the platform does it happen?

Linux 5.2.17-1rodete3-amd64, CPU: 16 cores, RAM: 62.9 G
Docker version 19.03.5, build 633a0ea838

## *How* do we replicate the issue?

Run an experiment, and wait for the scheduler runs trial runners docker containers. Like this:
```
...
INFO:root:Done building fuzzer benchmarks.
INFO:root:Starting scheduler.
INFO:root:Finding trials to schedule.
INFO:root:Starting trials.
INFO:root:Start measure_loop.
INFO:root:Start trial 43.
INFO:root:Start trial 52.
INFO:root:Start trial 53.
INFO:root:Start trial 45.
INFO:root:Start trial 33.
INFO:root:Start trial 28.
INFO:root:Start trial 34.
...
```
Then type `Ctrl+C` several times to kill the process and exit to the bash for dispatch-container:
```
...
KeyboardInterrupt
^CException ignored in: <module 'threading' from '/usr/local/lib/python3.7/threading.py'>
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/threading.py", line 1307, in _shutdown
    lock.acquire()
KeyboardInterrupt
^CError in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/multiprocessing/util.py", line 277, in _run_finalizers
    finalizer()
  File "/usr/local/lib/python3.7/multiprocessing/util.py", line 201, in __call__
    res = self._callback(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.7/multiprocessing/pool.py", line 581, in _terminate_pool
    cls._help_stuff_finish(inqueue, task_handler, len(pool))
  File "/usr/local/lib/python3.7/multiprocessing/pool.py", line 566, in _help_stuff_finish
    inqueue._rlock.acquire()
KeyboardInterrupt
^C
[...]

PR #480, opened 2020-06-22
Give trial runner docker container a name; add new Makefile options

zchcai:
## Description

This PR gives a name as `fuzzbench-r-<experiment_name>-<trial_id>` and then provide fuzzbench-wide and experiment-wide docker cleaner option in `make`:
1) `make stop-trial-runners`: fuzzbench-wide stop and remove.
2) `make stop-trial-runners-experiment experiment=<experiment_name>`: experiment-level stop and remove runners docker containers.

## Motivation and Context

Related to #470 and #479: helpful options for local users and preparations for end-to-end local tests.

## How Has This Been Tested?

I started up two experiments and killed them after their trial runners started.
Then I tried to stop one experiment-related docker containers and then tried the fuzzbench-wide clean. Both work.

## Other
Unit tests fix to comply with the new format of runners' docker startup scripts.


Makefile
```diff
@@ -50,3 +50,11 @@ clear-cache:
 	docker rm -vf $$(docker ps -a -q) 2>/dev/null ; \
 	docker rmi -f $$(docker images -a -q) 2>/dev/null ; \
 	docker volume rm $$(docker volume ls -q) 2>/dev/null ; true
+
+stop-trial-runners:
```
inferno-chromium, line 54: Why not a simpler command stop-trials ?
and just one command, if experiment is set then cancels that experiment one, otherwise cancels all.

inferno-chromium: Will people be running multiple experiments at once ? Will that even work, if not, just a stop-experiment seems better ?

laurentsimon: is there an assumption that the runs are running in the background and don't block the terminal? For example, when I run `make run-XXX`, it blocks the terminal and I have to Ctl+C to stop it. I'm wondering it it'd make sense to have a dedicated utility - rather than make - for updating/controlling running experiments/tasks, instead of cluttering the Makefile. I guess now we don't have that many utility function so it
so fine :)

lszekeres: > I'm wondering it it'd make sense to have a dedicated utility - rather than make - for updating/controlling running experiments/tasks, instead of cluttering the Makefile. 

Yes, I'm hoping that this would get much cleaner when we switch to the queue based architecture. In that case each task will be a single `docker run` (or `docker build`) command, each worker would work on one task at a time, and we wouldn't even need to run background jobs.

Makefile
```diff
@@ -50,3 +50,11 @@ clear-cache:
...
+	docker stop $$(docker ps -f "name=fuzzbench-r-*" -q) 2>/dev/null ; \
```
lszekeres, line 55: i think `docker stop` sends SIGTERM. is that always enough, or we might need `docker kill` sometimes, that sends SIGKILL?

Makefile
```diff
@@ -50,3 +50,11 @@ clear-cache:
...
+	docker rm -vf $$(docker ps -f "name=fuzzbench-r-*" -q) 2>/dev/null ; true
```
lszekeres, line 56: Have you considered using a single `docker rm --force` instead of `docker stop` and then `docker rm`?
https://docs.docker.com/engine/reference/commandline/rm/#force-remove-a-running-container

Makefile
lszekeres: > Why not a simpler command stop-trials ?

or even something like `stop-all-fuzzbench-containers`, if it stops not just runners but others too (like measurer)?

Makefile
zchcai: Thanks! `docker rm --force` is cleaner and what we really want to achieve.

Makefile
zchcai: Right. SIGTERM is not always enough and SIGKILL is more powerful and suitable here if we intend to remove those containers completely.
Let's combine `docker stop` and `docker rm` into `docker rm --force` as suggested in another comment. Thanks.

zchcai:
As suggested above, one command as `make stop-trials` may be cleaner. If `experiment` is set, then we remove all related to that experiment only. Otherwise, remove all trials.

Also, use `docker rm --force` directly, instead two steps of `stop` and `rm`.

Since there will also be `dispatcher-container` running with one experiment, another command `make stop-experiment` will do `make stop-trials` first and clean the related `dispatcher-container` or all.

For measurers, currently it is a process under `dispatcher-container`. Let's update or separate that logic later when measurers have separate implementation.

zchcai:
> > I'm wondering it it'd make sense to have a dedicated utility - rather than make - for updating/controlling running experiments/tasks, instead of cluttering the Makefile.
> 
> Yes, I'm hoping that this would get much cleaner when we switch to the queue based architecture. In that case each task will be a single `docker run` (or `docker build`) command, each worker would work on one task at a time, and we wouldn't even need to run background jobs.

I agree. To that time, we should have a way to show current jobs running status. Also, we need to provide a way for pause an experiment and resume again, also adding or removing workers for running jobs.

To that time, the way by`make *` to control the experiment status should consider to update then.

Makefile
```diff
@@ -50,3 +50,11 @@ clear-cache:
 	docker rm -vf $$(docker ps -a -q) 2>/dev/null ; \
 	docker rmi -f $$(docker images -a -q) 2>/dev/null ; \
 	docker volume rm $$(docker volume ls -q) 2>/dev/null ; true
+
+experiment=*
```
lszekeres, line 54: conventional style would be `EXPERIMENT := "*"`

Makefile
```diff
@@ -50,3 +50,11 @@ clear-cache:
...
+
+stop-trials:
```
lszekeres, line 56: what's "-r-" and "-d-" in the name? consider avoiding such abbreviations in the code for readability

Makefile
```diff
@@ -50,3 +50,11 @@ clear-cache:
...
+	docker rm -vf $$(docker ps -f "name=fuzzbench-r-$(experiment)-[0-9]+$$" -q) 2>/dev/null ; true
```
lszekeres, line 57: consider using full flags in code e.g. `rm --force --verbose (??)`, and avoid abbreviations for readability. that way the reader don't need to look up e.g. what `-v` is

Makefile
jonathanmetzman, line 56: Please put this in stop_experiment.py.
I don't really think this should have a make target either frankly, but that's a separate point.

Makefile
lszekeres, line 57: why is the `; true` part necessary? why would the command fail, and why would we not terminate / hide the error code?

if it is necessary, the proper way to do this in a Makefile is to [prefix the command](https://www.gnu.org/software/make/manual/html_node/Errors.html) with "-", ie.:
```
  - docker rm ...
```

experiment/resources/runner-startup-script-template.sh
```diff
@@ -43,4 +43,5 @@ docker run \
 -e LOCAL_EXPERIMENT={{local_experiment}} \
 {{additional_env}} {% if not local_experiment %}--name=runner-container {% endif %}\
 --cap-add SYS_NICE --cap-add SYS_PTRACE \
+--name fuzzbench-{{instance_name}} \
```
jonathanmetzman, line 46: Let's get rid of the `if not local_experiment` part on line 44.
Also drop `fuzzbench-` prefix please.

experiment/test_scheduler.py
```diff
@@ -113,6 +113,7 @@ def test_create_trial_instance(benchmark, expected_image, expected_target,
 -e LOCAL_EXPERIMENT=False \\
 -e C1=custom -e C2=custom2 --name=runner-container \\
 --cap-add SYS_NICE --cap-add SYS_PTRACE \\
+--name fuzzbench-r-test-experiment-9 \\
```
jonathanmetzman, line 116: Yeah we definitely don't want two names.

Makefile
zchcai: `-r-` is for runners, `-d-` is for dispatchers. Does it make sense?

jonathanmetzman:
I think part of the issue here is that the scheduler can start docker containers and GCE instances but can only delete GCE instances and not containers. Teaching gcloud.delete_instances to stop containers would work fix this problem and make stop_experiment work properly.

>Also, we need to provide a way for pause an experiment and resume again, also adding or removing workers for running jobs

There's no way to pause a trial so other than restarting parts of an experiment (e.g. the measurer) this is not possible.

What happens when someone types in `make` without specifying targets? Does it just build the `all` target or does it build all of them? Because some of our targets like `clear-cache` would produce weird results if "built" in this scenario.

experiment/resources/runner-startup-script-template.sh
zchcai: Could you give some reasons to delete that prefix?

I think, to distinguish other docker containers in local machines, `fuzzbench-` should preserve.
For example, I also have some machine learning projects and run some docker containers.
Adding a prefix is more convenient to distinguish and, also more importantly, prevent delete other non-fuzzbench docker containers.

zchcai:
> There's no way to pause a trial so other than restarting parts of an experiment (e.g. the measurer) this is not possible.

Could you give some details why this is not possible?

Makefile
lszekeres: yes, but the suggestion was to avoiding such abbreviations, so the next person don't need to ask this question and/or spend time figuring this out. so could we please use eg `runner` and `dispatcher` instead of `r` and `d`?

experiment/resources/runner-startup-script-template.sh
jonathanmetzman: I think it's unlikely someone would have other containers on their machine named `r-$EXPERIMENT-$TRIAL_ID`, and I'd rather not have this so that we don't have to deal with complexity, we can just use the instance naming regex we use in stop_experiment.py to stop the containers as well.

jonathanmetzman:
> > There's no way to pause a trial so other than restarting parts of an experiment (e.g. the measurer) this is not possible.
> 
> Could you give some details why this is not possible?

It actually might be somewhat doable using `docker pause` and `unpause`. But I think there's little guarantee this works cleanly since fuzzers could be using things like timers that will be incorrect if unpaused same with our tasks in the queue. I could be wrong but I would guess this feature is very complicated to get right.

Makefile
jonathanmetzman: The reason why we do r- and d- is because there is a relatively short limit on gce instance names since we want to include the experiment name in the instance name we abbreviate so that the experiment name isn't as restricted.

zchcai: It is also the already instance name when creating docker containers. I think this helps to maintain a consistency.

lszekeres: > there is a relatively short limit on gce instance names 

how many characters is the limit? do a few characters matter that much?

jonathanmetzman: I think you end up with about 50 that can be used for the experiment name. I'd rather use the abbreviations so if we end up with something with a long name "measurer-worker" for example we don't have to be inconsistent or really make the experiment name tight.

zchcai: > The reason why we do r- and d- is because there is a relatively short limit on gce instance names since we want to include the experiment name in the instance name we abbreviate so that the experiment name isn't as restricted.

Don't know this. We can use a full name for dockers' name like runner or dispatcher here.

zchcai: Since the name limit for docker is 128 chars, just to make sure there arrives a consensus: @jonathanmetzman @lszekeres
Plan 1:
- we use `runner-<experiment-name>-<trial_id>` for trial runners containers.
- we use `dispatcher-<experiment-name>` for the dispatcher container.
- we use `measurer-worker-<experiment-name>_<work_id>` for the measurer workers I guess?
Plan 2:
- we use `r-<experiment-name>-<trial_id>` for trial runners containers.
- we use `d-<experiment-name>` for the dispatcher container.
- we use `m-<experiment-name>_<work_id>` for the measurer workers I guess?

lszekeres: I'm fine anything, I'm just curious about these limitations and numbers -- I haven't found anything about them.

> I think you end up with about 50 
> Since the name limit for docker is 128 chars

Do you happen to have a reference? I was looking a bunch and I didn't find anything about any length limit on docker container names.

jonathanmetzman: The limit is for google cloud instance names since these names are used for them as well. I don't have a reference but you can verify the limit is 63 by trying to create an instance. we also need 3 chars for hyphens and ~7 for for trial ids.

lszekeres: Got it, thanks! (For docker i don't think there's a limit: i just tried running `docker run --name <600+ character long name> ...` and it was happy to run it)

experiment/resources/runner-startup-script-template.sh
zchcai: I'm fine with this.

Makefile
zchcai: OK. My fault. I didn't run for a check but just search from Google. [This link](https://docs.docker.com/engine/reference/commandline/tag/#:~:text=A%20name%20component%20may%20not,a%20maximum%20of%20128%20characters.) says a limit for tag, not for name.

zchcai: Let's choose plan 2, which is shorter.

Makefile
zchcai: When there is no containers to remove, there is some error showing up.
Adding those will help to get a clean console I guess.
```
docker rm --volumes --force $(docker ps --filter "name=r-*-[0-9]+$" -q)
"docker rm" requires at least 1 argument.
See 'docker rm --help'.

Usage:  docker rm [OPTIONS] CONTAINER [CONTAINER...]

Remove one or more containers
make: *** [Makefile:57: stop-trials] Error 1
```
If we add `-` in front of it, the error report is the same but adding one ignore at last:
```
docker rm --volumes --force $(docker ps --filter "name=r-*-[0-9]+$" -q)
"docker rm" requires at least 1 argument.
See 'docker rm --help'.

Usage:  docker rm [OPTIONS] CONTAINER [CONTAINER...]

Remove one or more containers
make: [Makefile:58: stop-trials] Error 1 (ignored)
```
If we add only `2>/dev/null` then
```
docker rm --volumes --force $(docker ps --filter "name=r-*-[0-9]+$" -q) 2>/dev/null
make: *** [Makefile:57: stop-trials] Error 1
```
Thus, we add `true` at last to get rid of all error reporting.

Makefile
zchcai: Some concerns at first since EXPERIMENT may be used by the fuzzbench experiment.
Later I found, EXPERIMENT is exactly the name we want actually. So we can use all capital letters here.

Makefile
lszekeres: Thanks! I don't see why hiding error messages and masking the status code is useful in this case: if I run `make kill-trials`, and no trials were killed, I'd like to know it.

Makefile
lszekeres: My understanding was that this "stop-trials" is not the official way to stop an experiment, but a quick temporary fix only for developers to clean up stuck/stale/leftover/zombie trials that often happen currently due to some issues. Is that correct?

If yes, we might consider also calling this accordingly, e.g. as `kill-trials`, and leave a comment explaining the purpose and when to use it.

zchcai: Sure. Added to it and waiting for a last local test.
I think the existing of `make clear-cache` letting us believe that it is fine to have other cleaning options. If `stop-trials` is not needed or not proper, it would be better remove `clear-cache` as well.

zchcai:
Hi, anyone has some ideas on this weird behavior?
```
~/fuzzbench$ docker ps -f "name=r-*-[0-9]+$"
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
~/fuzzbench$ docker ps -f "name=-*-[0-9]+$"
CONTAINER ID        IMAGE                                                           COMMAND                  CREATED             STATUS              PORTS               NAMES
751a7a887381        gcr.io/fuzzbench/runners/fairfuzz/zlib_zlib_uncompress_fuzzer   "/bin/sh -c $ROOT_DI…"   14 minutes ago      Up 13 minutes                           r-workingsee-22
34ea57f1418b        gcr.io/fuzzbench/runners/aflfast/zlib_zlib_uncompress_fuzzer    "/bin/sh -c $ROOT_DI…"   14 minutes ago      Up 13 minutes                           r-workingsee-17
7e114bc325cf        gcr.io/fuzzbench/runners/aflfast/jsoncpp_jsoncpp_fuzzer         "/bin/sh -c $ROOT_DI…"   14 minutes ago      Up 13 minutes                           r-workingsee-19
c9c54f0359f5        gcr.io/fuzzbench/runners/fairfuzz/jsoncpp_jsoncpp_fuzzer        "/bin/sh -c $ROOT_DI…"   14 minutes ago      Up 13 minutes                           r-workingsee-23
878d472e46c5        gcr.io/fuzzbench/runners/fairfuzz/zlib_zlib_uncompress_fuzzer   "/bin/sh -c $ROOT_DI…"   14 minutes ago      Up 13 minutes                           r-workingsee-21
88d1ef35e6e5        gcr.io/fuzzbench/runners/aflfast/jsoncpp_jsoncpp_fuzzer         "/bin/sh -c $ROOT_DI…"   14 minutes ago      Up 13 minutes                           r-workingsee-20
522c0b1258ef        gcr.io/fuzzbench/runners/fairfuzz/jsoncpp_jsoncpp_fuzzer        "/bin/sh -c $ROOT_DI…"   14 minutes ago      Up 13 minutes                           r-workingsee-24
785f34788e97        gcr.io/fuzzbench/runners/aflfast/zlib_zlib_uncompress_fuzzer    "/bin/sh -c $ROOT_DI…"   14 minutes ago      Up 13 minutes                           r-workingsee-18
```

Makefile
zchcai: Looks like a valid need for potential users. I thought no error reporting is more smooth in general. 

Shall we also change it for letting users know there are no trials were killed for `make clear-cache` too?
https://github.com/google/fuzzbench/blob/7d8527cbded8d0280cbed461058a7b89beb935cd/Makefile#L48-L52

Makefile
zchcai: > My understanding was that this "stop-trials" is not the official way to stop an experiment, but a quick temporary fix only for developers to clean up stuck/stale/leftover/zombie trials that often happen currently due to some issues. Is that correct?

This is for one purpose. Another is for end-to-end testing, where it will cause trials not finished it the experiment has problems.
Let me add some comments for this make target.

zchcai: > kill-trials

Shall we avoid words like `kill`? I think `stop` or `terminate` should be fine and also warmer and sounds peaceful. :)

experiment/stop_experiment.py
```diff
@@ -29,8 +30,12 @@ def stop_experiment(experiment_name, experiment_config_filename):
     experiment_config = yaml_utils.read(experiment_config_filename)
 
     if experiment_config.get('local_experiment', False):
-        raise NotImplementedError(
-            'Local experiment stop logic is not implemented.')
+        new_process.execute(
+            ['make', 'stop-experiment', "EXPERIMENT=" + str(experiment_name)],
+            expect_zero=False)
```
zchcai, line 35: This is to avoid subprocess exception when the experiment normally stops since we don't make `make stop-expeirment` stop with no error when there is no docker to remove.

experiment/run_experiment.py
```diff
@@ -401,7 +401,9 @@ def start(self):
             '${WORK}/src/experiment/dispatcher.py || '
             '/bin/bash'  # Open shell if experiment fails.
         ]
-        return new_process.execute(command, write_to_stdout=True)
+        return new_process.execute(command,
+                                   write_to_stdout=True,
+                                   expect_zero=False)
```
zchcai, line 406: When it opens shell, it means the experiment fails. Then if we type `exit` it will return non-zero code for this return and it affects the later cleaning up logic execution in stop_experiment.stop_experiment.

Makefile
jonathanmetzman: I'd be happy to remove `clear-cache` I think it was a mistake to add it frankly.

lszekeres: As discussed offline, let's
1) remove `stop-trials` from Makefile
2) move the `stop-experiment` logic to `stop_experiment.py` (feel free to keep the `stop-experiment` make target, if you think it's useful, that just calls the `experiments/stop_experiment.py` script)
3) also remove the `clear-cache` from Makefile.

experiment/stop_experiment.py
```diff
@@ -29,8 +30,15 @@ def stop_experiment(experiment_name, experiment_config_filename):
     experiment_config = yaml_utils.read(experiment_config_filename)
 
     if experiment_config.get('local_experiment', False):
-        raise NotImplementedError(
-            'Local experiment stop logic is not implemented.')
+        subprocess.Popen([
```
zchcai, line 33: Choose this instead of `common/new_process` because, when directly calling `experiment/stop_experiment.py`, it will have problems for module `logs`. Similarly, `logger.info` cannot be executed normally.
I guess it is related to the initialization when calling stop_experiment.py directly. Any thoughts? Don't know the behavior for a cloud running stopping. Have you seen similar error logs when directly calling experiment/stop_experiment.py outside to stop a cloud running? @jonathanmetzman 

Below is the error message if we use `new_process.execute`:
```
Fatal Python error: Cannot recover from stack overflow.

Thread 0x00007fd49e6ba700 (most recent call first):
    File "/usr/lib/python3.7/threading.py", line 296 in wait
    File "/usr/lib/python3.7/queue.py", line 170 in get
    File
    "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/google/cloud/logging/handlers/transports/background_thread.py",
    line 65 in _get_many
    File
    "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/google/cloud/logging/handlers/transports/background_thread.py",
    line 142 in _thread_main
    File "/usr/lib/python3.7/threading.py", line 870 in run
    File "/usr/lib/python3.7/threading.py", line 926 in _bootstrap_inner
    File "/usr/lib/python3.7/threading.py", line 890 in _bootstrap

    Current thread 0x00007fd4a138b740 (most recent call first):
        File "/usr/lib/python3.7/email/utils.py", line 57 in
        _has_surrogates
        File "/usr/lib/python3.7/email/_policybase.py", line 287 in
        _sanitize_header
        File "/usr/lib/python3.7/email/_policybase.py", line 316 in
        header_fetch_parse
        File "/usr/lib/python3.7/email/message.py", line 471 in
        get
        File "/usr/lib/python3.7/email/message.py", line 578 in
        get_content_type
        File "/usr/lib/python3.7/email/feedparser.py", line
        256 in _parsegen
        File "/usr/lib/python3.7/email/feedparser.py", line
        180 in _call_parse
        File "/usr/lib/python3.7/email/feedparser.py",
        line 176 in feed
        File "/usr/lib/python3.7/email/parser.py", line
        56 in parse
        File "/usr/lib/python3.7/email/parser.py",
        line 67 in parsestr
        File "/usr/lib/python3.7/http/client.py",
        line 224 in parse_headers
        File "/usr/lib/python3.7/http/client.py",
        line 330 in begin
        File
        "/usr/lib/python3.7/http/client.py",
        line 1344 in getresponse
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/urllib3/connectionpool.py",
        line 421 in _make_request
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/urllib3/connectionpool.py",
        line 677 in urlopen
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/requests/adapters.py",
        line 449 in send
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/requests/sessions.py",
        line 643 in send
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/requests/sessions.py",
        line 530 in request
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/google/auth/transport/requests.py",
        line 181 in __call__
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/google/auth/compute_engine/_metadata.py",
        line 137 in get
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/google/auth/compute_engine/_metadata.py",
        line 219 in
        get_service_account_info
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/google/auth/compute_engine/credentials.py",
        line 77 in
        _retrieve_info
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/google/auth/compute_engine/credentials.py",
        line 96 in refresh
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/google/auth/credentials.py",
        line 124 in
        before_request
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/google/auth/transport/requests.py",
        line 442 in
        request
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/google/cloud/_http.py",
        line 315 in
        _do_request
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/google/cloud/_http.py",
        line 277 in
        _make_request
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/google/cloud/_http.py",
        line 419
        in
        api_request
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/google/cloud/logging/_http.py",
        line 168
        in
        write_entries
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/google/cloud/logging/logger.py",
        line
        132 in
        _do_log
        File
        "/usr/local/home/fuzzbench/.venv/lib/python3.7/site-packages/google/cloud/logging/logger.py",
        line
        189
        in
        log_struct
        File
        "/usr/local/home/fuzzbench/common/logs.py",
        line
        176
        in
        log
        File
        "/usr/local/home/fuzzbench/common/retry.py",
        line
        79
        in
        _wrapper
        File
        "/usr/local/home/fuzzbench/common/logs.py",
        line
        211
        in
        info
        File
        "/usr/local/home/fuzzbench/common/retry.py",
        line
        64
        in
        handle_retry
        File
        "/usr/local/home/fuzzbench/common/retry.py",
        line
        88
        in
        _wrapper
        File
        "/usr/local/home/fuzzbench/common/logs.py",
        line
        211
        in
        info
        File
        "/usr/local/home/fuzzbench/common/retry.py",
        line
        64
        in
        handle_retry
...
```

jonathanmetzman: This seems like a FORCE_NOT_LOCAL issue.
Could you add some logic to logs.py to try to catch this error and recommned FORCE_NOT_LOCAL instead of stack overflowing?

zchcai: > This seems like a FORCE_NOT_LOCAL issue.

Does it mean, when I directly run this stop_experiment.py, `FORCE_NOT_LOCAL=true` in default?

jonathanmetzman: It means you need to set `FORCE_NOT_LOCAL=True` in the environment when you run this script (and every other in FuzzBench).

experiment/stop_experiment.py
```diff
@@ -29,8 +30,15 @@ def stop_experiment(experiment_name, experiment_config_filename):
...
+            '/bin/bash', '-c', 'docker rm --volumes --force '
+            '$(docker ps -f "name=r-' + str(experiment_name) + '-[0-9]+$" -q)'
+        ])
```
lszekeres, line 36: why not subprocess.run()?

is /bin/bash -c necessary?

experiment/stop_experiment.py
zchcai: OK. That makes sense. I got it. Will have a test later.

experiment/stop_experiment.py
zchcai: > why not subprocess.run()?

I guess we prefer to the wrapper version in this code repo for a consistency, `common/new_process`, which adopts `subprocess` already. The current way is a solution to avoid some errors. But maybe we can add `FORCE_LOCAL` in the Makefile in default and use `new_process` back.

> is /bin/bash -c necessary?

Yes. Otherwise, it will look at the `$(docker ps ...)` as an argument, rather than executing it first and getting results.

PR #487, opened 2020-06-25
[docs] Improve docs-serve and fix link

jonathanmetzman:
1. Make sure dependencies are installed when using docs-serve
2. Fix link to customizing reports from report reference page.


jonathanmetzman:
The link this fixed had a trailing "/" which currently causes a 404 (try: https://google.github.io/fuzzbench/developing-fuzzbench/custom_analysis_and_reports/)
A better fix would be to make our docs accept this.

PR #488, opened 2020-06-25
again: afl++ improvements

vanhauser-thc:
making especially aflplusplus and aflplusplus_shmem even faster.
also removes aflplusplus_qemu.

if all benchmarks build without errors please run a test.


vanhauser-thc:
uh it is building noticeable faster, great job!

jonathanmetzman:
Thanks! I think qemu mode might not be a bad idea. That way blackbox techniques have a better baseline to compare against.
What do you think of adding it back?

vanhauser-thc:
> Thanks! I think qemu mode might not be a bad idea. That way blackbox techniques have a better baseline to compare against.
> What do you think of adding it back?

If you wish I can do that.

IMHO it is not very useful though. it is the only greybox fuzzer in the tests. and greybox fuzzing works differently, they need more hands-on to be effective (and fair in the comparison). e.g. winafl is great but you need to setup the target API function address etc. and also afl++ qemu_mode has a persistent mode that equally needs to be set up - and if done so improves speed by x2 to x5 - so it is really important to do that.
So basically to test greybox fuzzers effectively you need targets that do not change for every benchmark and kept static.

further complication is that several greybox fuzzers require specific or very old kernel versions (e.g. for intel pt optimizations or because they use pin 2.14).

I would love to see a greybox fuzzing benchmark. It would be more work on the fuzzer setup side (because the binaries of the target are not changed), but I think this would be amazing for further progress in this field.

vanhauser-thc:
P.S. there was no advancement in the qemu_mode hence that would not need to be re-run ... so please start the existing aflplusplus ones :) (and whatever other fuzzer that changed too :)

jonathanmetzman:
>IMHO it is not very useful though. it is the only greybox fuzzer in the tests. 

Eclipser too. And if we don't have any I think it is discouraging for people to add more :-)

>and also afl++ qemu_mode has a persistent mode that equally needs to be set up - and if done so improves speed by x2 to x5 - so it is really important to do that.

Ah is it not using persistent mode in FuzzBench? Is it very hard to automatically set up? Since unlike in the real world the right function for all of our targets is LLVMFuzzerTestOneInput.

>P.S. there was no advancement in the qemu_mode hence that would not need to be re-run ... so please start the existing aflplusplus ones :) (and whatever other fuzzer that changed too :)

Starting one now.

vanhauser-thc:
the benchmark revealed a bug in the shmem implementation when together with LTO autodict feature, which only affected libxml2.
overall shmem is better than the original so I will switch this over to be the default in another PR.
can you fix this benchmark for aflplusplus_optimal_shmem for libxml2 to have the same value as aflplusplus_optimal? or restart just that one once the qemu PR is in?

jonathanmetzman:
I'll land the qemu PR and restart that fuzzer.

Issue #361, opened 2020-05-20
Use a startup script for the dispatcher instead of using SSH


jonathanmetzman:
SSH can fail sometimes and is frankly unnecessary now that FuzzBench has proper logging.

Issue #486, opened 2020-06-24
SSH time out


Shadoom7:
ssh time out when trying to run experiment on gcp.

zchcai:
> ssh time out when trying to run experiment on gcp.

Could you provide more details? For [example](https://github.com/zchcai/github-issue-templates/blob/master/questions-answers/ISSUE_TEMPLATE.md).

Shadoom7:
When trying to run experiment on gcp, the error occurs like this:

$ PYTHONPATH=../fuzzbench python3 ../fuzzbench/experiment/run_experiment.py -e ruikang2 -c ./fuzzbench-test.yaml --fuzzers afl libfuzzer --benchmarks libpng-1.2.56 mbedtls_fuzz_dtlsclient
INFO:root:GCP instance isn't ready yet. Rerunning SSH momentarily.
INFO:root:GCP instance isn't ready yet. Rerunning SSH momentarily.
INFO:root:GCP instance isn't ready yet. Rerunning SSH momentarily.
INFO:root:GCP instance isn't ready yet. Rerunning SSH momentarily.
INFO:root:GCP instance isn't ready yet. Rerunning SSH momentarily.
INFO:root:GCP instance isn't ready yet. Rerunning SSH momentarily.
INFO:root:GCP instance isn't ready yet. Rerunning SSH momentarily.
INFO:root:GCP instance isn't ready yet. Rerunning SSH momentarily.
INFO:root:GCP instance isn't ready yet. Rerunning SSH momentarily.
INFO:root:GCP instance isn't ready yet. Rerunning SSH momentarily.
Traceback (most recent call last):
  File "../fuzzbench/experiment/run_experiment.py", line 549, in <module>
    sys.exit(main())
  File "../fuzzbench/experiment/run_experiment.py", line 541, in main
    args.benchmarks, fuzzer_configs)
  File "../fuzzbench/experiment/run_experiment.py", line 260, in start_experiment
    start_dispatcher(config, CONFIG_DIR)
  File "../fuzzbench/experiment/run_experiment.py", line 272, in start_dispatcher
    dispatcher.start()
  File "../fuzzbench/experiment/run_experiment.py", line 425, in start
    self.config['cloud_compute_zone'])
  File "/home/ruikangshi/work/fuzzbench/common/gcloud.py", line 62, in robust_begin_gcloud_ssh
    raise Exception('Couldn\'t SSH to instance.')
Exception: Couldn't SSH to instance.

[...]

PR #489, opened 2020-06-26
Use a startup script for launching the dispatcher

jonathanmetzman:
Using ssh is hacky and error prone. Use a startup script instead.
This will also help with servicification as run_experiment.py will finish immediately after an experiment is started.

Fixes https://github.com/google/fuzzbench/issues/361 and https://github.com/google/fuzzbench/issues/486

### DIFF
--- a/experiment/dispatcher.py
+++ b/experiment/dispatcher.py
@@ -159,6 +159,7 @@ def dispatcher_main():
     logs.info('Dispatcher finished.')
     scheduler_loop_thread.join()
     measurer_loop_process.join()
+    return experiment.config
 
 
 def main():
@@ -168,10 +169,14 @@ def main():
     })
 
     try:
-        dispatcher_main()
+        experiment_config = dispatcher_main()
     except Exception as error:
         logs.error('Error conducting experiment.')
         raise error
+    stop_experiment.stop_experiment(experiment_config['experiment'],
+                                    experiment_config)
+
+    return 0
 
 
 if __name__ == '__main__':
--- a/experiment/dispatcher.py
+++ b/experiment/dispatcher.py
@@ -34,6 +34,7 @@ from experiment.build import builder
 from experiment import measurer
 from experiment import reporter
 from experiment import scheduler
+from experiment import stop_experiment
 
 LOOP_WAIT_SECONDS = 5 * 60
 
--- a/experiment/dispatcher.py
+++ b/experiment/dispatcher.py
@@ -177,11 +177,11 @@ def main():
                                                'experiment.yaml')
 
     experiment_utils.is_local_experiment()
-    if stop_experiment.stop_experiment(experiment_config['experiment'],
-                                       experiment_config):
+    if stop_experiment.stop_experiment(experiment_utils.get_experiment_name(),
+                                       experiment_config_file_path):
         return 0
-    else:
-        return 1
+
+    return 1
 
 
 if __name__ == '__main__':
--- a/experiment/dispatcher.py
+++ b/experiment/dispatcher.py
@@ -176,7 +176,9 @@ def main():
     experiment_config_file_path = os.path.join(fuzzer_config_utils.get_dir(),
                                                'experiment.yaml')
 
-    experiment_utils.is_local_experiment()
+    if experiment_utils.is_local_experiment():
+        return 0
+
     if stop_experiment.stop_experiment(experiment_utils.get_experiment_name(),
                                        experiment_config_file_path):
         return 0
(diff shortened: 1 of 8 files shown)


jonathanmetzman:
This still needs testing and unittests.

jonathanmetzman:
@Shadoom7 Can you please try this out locally and see if it works?
It will still be a nuisance if you can't ssh directly from your machine into GCP and we should figure out why this happens.
But I think this should work around it. Worst case, if we can't figure this out, you can ssh from the cloud interface if you need to debug.

jonathanmetzman:
@Shadoom7 I made some fixes I can't work on this more today, but I'm testing now. I'll let you know if it should work. The last one had a silly bug that would terminate the experiment after it started.

Shadoom7:
> @Shadoom7 I made some fixes I can't work on this more today, but I'm testing now. I'll let you know if it should work. The last one had a silly bug that would terminate the experiment after it started.

Thank you so much! I will also try to solve this ssh issue in the meantime.

jonathanmetzman:
Stopping experiment is still probably broken (in particular we try to delete all experiment instances without realizing that we might delete the dispatcher first).

experiment/run_experiment.py
```diff
@@ -410,59 +407,42 @@ def start(self):
...
+                (cloud_sql_instance_connection_name),
+            'docker_registry': self.config['docker_registry'],
+        }
+        return template.render(**kwargs)
+
+    def write_startup_script(self, startup_script_file):
+        """Get the startup script to start the experiment on the dispatcher."""
+        startup_script = self._render_startup_script()
+        startup_script_file.write(startup_script)
+        startup_script_file.flush()
```
lszekeres, line 445: `close()` instead of `flush()`?

experiment/run_experiment.py
```diff
@@ -60,6 +61,13 @@
 
 CONFIG_DIR = 'config'
 
+RESOURCES_DIR = os.path.join(utils.ROOT_DIR, 'experiment', 'resources')
+
+JINJA_ENV = jinja2.Environment(
+    undefined=jinja2.StrictUndefined,
+    loader=jinja2.FileSystemLoader(RESOURCES_DIR),
+)
+
```
lszekeres, line 70: does this need to be a global?

experiment/run_experiment.py
jonathanmetzman: Flush is needed. Closing the file deletes it.

experiment/dispatcher.py
```diff
@@ -172,6 +173,15 @@ def main():
     except Exception as error:
         logs.error('Error conducting experiment.')
         raise error
+    experiment_config_file_path = os.path.join(fuzzer_config_utils.get_dir(),
+                                               'experiment.yaml')
+
+    experiment_utils.is_local_experiment()
```
inferno-chromium, line 179: Did you mean to return here.

experiment/run_experiment.py
```diff
@@ -267,11 +275,8 @@ def start_dispatcher(config: Dict, config_dir: str):
     """Start the dispatcher instance and run the dispatcher code on it."""
     dispatcher = get_dispatcher(config)
     # Is dispatcher code being run manually (useful for debugging)?
-    manual_experiment = os.getenv('MANUAL_EXPERIMENT')
-    if not manual_experiment:
-        dispatcher.create_async()
     copy_resources_to_bucket(config_dir, config)
-    if not manual_experiment:
+    if not os.getenv('MANUAL_EXPERIMENT'):
```
inferno-chromium, line 279: Can this be a helper in experiment_utils.py ?

experiment/run_experiment.py
inferno-chromium: +1, can be just local.

experiment/run_experiment.py
```diff
@@ -410,59 +407,42 @@ def start(self):
...
-            '/work/startup-dispatcher.sh'
-        ).format(
-            instance_name=self.instance_name,
-            postgres_password=os.environ['POSTGRES_PASSWORD'],
-            experiment=self.config['experiment'],
+        kwargs = {
+            'instance_name': self.instance_name,
+            'postgres_password': os.environ['POSTGRES_PASSWORD'],
+            'experiment': self.config['experiment'],
             # TODO(metzman): Create a function that sets env vars based on
```
inferno-chromium, line 430: Is this comment still applicable?

jonathanmetzman: Removed.

experiment/run_experiment.py
jonathanmetzman: I don't see a need for this, it's used pretty infrequently.

experiment/dispatcher.py
jonathanmetzman: Yes

PR #496, opened 2020-06-30
[NFC] Cleanup (mostly comments) in fuzzer integrations.

jonathanmetzman:
Integrations get copied so stale comments and other small nits
multiply.


jonathanmetzman:
lafintel failure looks spurious (network/docker issue), aflcc looks like it is failing on perl which is an experimental benchmark and not really supported. eclipser failure is timeout.

PR #501, opened 2020-07-06
change defaults, new temporary tests

vanhauser-thc:
this moves the _shmem variants to be the normal defaults. also adds new tests variants. these have been in before however there were several changes so I expect differences. also the autodict feature is a new test.


vanhauser-thc:
a few benchmarks e.g. curl might fail as they take very long and can therefore trigger timeouts. I would not care in the experience if that happens.

jonathanmetzman:
sounds good to me. Should I wait for CI to finish or go ahead and merge?

vanhauser-thc:
ah lets wait. maybe I overlooked a benchmark that needs the dynamic flag for LTO - was changed it now would need it

vanhauser-thc:
but can you please merge reports 2020-06-30 into 2020-06-26 (overwriting the aflplusplus_optimal_shmem from ...-26) and the merge this to 2020-05-24? thanks!

the one you just started have not the _shmem variant which is the new default :)

vanhauser-thc:
btw the first oss-fuzz completed successfully from the LTO variants so all is fine and it can be merged.

vanhauser-thc:
and one last thing before going to bed - afl had several updates, commit id is 82b5e359463238d790cadbe2dd494d6a4928bff3 and also mopt, commit id is 61cdee22a9dc0f638d2a200c0d460c699e8500e0

jonathanmetzman:
We're running a large scale experiment using clang coverage right now so it might be a day or so before I get to run this.

jonathanmetzman:
I merged the reports as you asked. See https://www.fuzzbench.com/reports/2020-06-30-and-05-24/index.html and https://www.fuzzbench.com/reports/2020-06-30/index.html

vanhauser-thc:
> I merged the reports as you asked. See https://www.fuzzbench.com/reports/2020-06-30-and-05-24/index.html and https://www.fuzzbench.com/reports/2020-06-30/index.html

Thank you! This is for the update of the paper

vanhauser-thc:
> We're running a large scale experiment using clang coverage right now so it might be a day or so before I get to run this.

as long as it is finished before the 24th ... no hurry ... :)
(it is also for the paper)

is the "clang coverage" new coverage measurement (the gcov style instead of sanitize-coverage)?

jonathanmetzman:
> > We're running a large scale experiment using clang coverage right now so it might be a day or so before I get to run this.
> 
> as long as it is finished before the 24th ... no hurry ... :)
> (it is also for the paper)

I think this should be doable.

> 
> is the "clang coverage" new coverage measurement (the gcov style instead of sanitize-coverage)?

Yes.

PR #503, opened 2020-07-07
Fix bugs in queue implementation

jonathanmetzman:
Fix these two bugs in the queue implementation:
1. Sometimes trials will fail and not produce any more snapshots.
2. Sometimes trials fail to archive a corpus for a cycle or mark it is unchanged, but will produce later ones.

Fix this by allowing the workers to specify to the manager that a cycle cannot be measured and whether there are later cycles for the trial that can be measured.
This allows us to handle the second problem gracefully by skipping the missing cycle and allows us to stop measuring a trial when it has ended but has nothing more to measure (previously the measurer would retry measuring it forever).

We solved these problems by doing the following:
1. Adding a class MeasureManager that is stateful for enqueuing measure jobs. This class will decide to skip measuring a cycle or an entire trial if needed.
2. Changing the response of a measure task from Snapshot or None to SnapshtoMeasureResponse. This can be used by the worker to communicate to the manager what can be measured next in the trial.
3. Allowing the worker to deal with skipping cycles in a trial.

Also, simplify handling of unchanged-cycles. Reduce the complexity of this by allowing copying of the same file more than once per task.

Also do the following:

1. Allow users of retry.wrap decorator to opt-out of logging on each retry. This is useful because our logging functions are wrapped so we end up with a situation where logging can fail (because of quota limits for example) and every time we retry logging, we log this.
2. Add some more options to filestore functions to avoid unnecessary error logging.


jonathanmetzman:
Before landing the queue I need to make it deal with an experiment where there will be competition between the measurer and the trial for quota. This is somewhat of a blocking problem but should only be temporary until we switch to the queue for trials.
I don't think the measurer starving trials is a big problem because it shouldn't need more than N instances for an extended period of time (where N is the number of trials) and because instance groups are limited by quota to 1000 anyway. The bigger problem would be the measurer being starved by the trials. I think I will solve this by creating a minimum measurer group size of 5% of the number of trials.

common/filestore_utils.py
```diff
@@ -35,11 +35,12 @@ def _using_gsutil():
     from common import gsutil as filestore_utils_impl
 
 
-def cp(source, destination, recursive=False, parallel=False):  # pylint: disable=invalid-name
+def cp(source, destination, recursive=False, must_exist=True, parallel=False):  # pylint: disable=invalid-name
```
zchcai, line 38: I guess this and related changes can be separate into another PR and makes this PR easier to focus on queue implementation bug fixing.

jonathanmetzman: I think it makes sense to keep here.

inferno-chromium:
Is this complexity worth it? this code is getting crazy complicated, why dont we just drop the trial ? How many cycles should be allow skipping, why just 1 or more ?

common/filestore_utils.py
jonathanmetzman: I'm actually merging with master, I think this change will go away when I'm done.

jonathanmetzman:
>Is this complexity worth it?

I'm not 100% sure about this. I think yes for now. I think I can reduce the complexity by using exceptions, but I don't want to get into those right now.

>why dont we just drop the trial

Much of the code needed to figure out whether to drop the trial is the same as what I have now. The only difference is the code I use to skip a particular cycle which although a bit hacky, I don't think adds much complexity and is relatively isolated.

> How many cycles should be allow skipping, why just 1 or more ?

I don't really understand this question. But if cycle 1,2,3 miss but 4 is there, we will measure 4.

experiment/measurer/measure_worker.py
```diff
@@ -37,21 +37,35 @@
...
     'SnapshotMeasureRequest', ['fuzzer', 'benchmark', 'trial_id', 'cycle'])
 
+SnapshotMeasureResponse = collections.namedtuple('SnapshotMeasureResponse',
+                                                 ['snapshot', 'next_cycle'])
 
-def initialize_logs():
+MEASURED_FILES_STATE_NAME = 'measured-files'
+COVERED_PCS_STATE_NAME = 'covered-pcs'
+
+CORPUS_ARCHIVE_CYCLE_REGEX = re.compile(r'.*\/corpus-archive-(\d{4})\.tar\.gz$')
```
lszekeres, line 51: are these internal? if yes consider prepending name with _

PR #506, opened 2020-07-08
Reduce the min fraction of alive trials for picking a snapshot time.

### LINKED ISSUE #481: Dynamic report generation
Though we already provide things like ipython notebooks to generate reports, I think we should lower the bar so that people can modify the report from the experiment from the report page.
I can think of two ways to do this:
1. Allow dynamic report generation using some kind of webapp.
2. Having a link to an ipython notebook that uses the data for the report on each report page.

Option 1 is a bit of a challenge because we generate reports server side and they take a while to do.
Option 2 is not ideal because I assume there is startup time for ipython notebooks and because it puts a greater burden on users.

lszekeres:
The "snapshot time" is the point in time during the trials where we the take the 
coverage values from to compare the fuzzers of a benchmark. The new 0.5 value 
means that we pick t=24h as the snapshot time if at least half of the trials of a
benchmark reached the end of the experiment (24 h). We only pick an earlier time
if less than half of them reached 24 hours.

Also rename the constant name to be less confusing.

Fixes #481 for real this time.

### DIFF
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -100,18 +100,17 @@ def filter_max_time(experiment_df, max_time):
 
 # Creating "snapshots" (see README.md for definition).
 
-_DEFAULT_BENCHMARK_SAMPLE_NUM_THRESHOLD = 1.0
+_MIN_FRACTION_OF_ALIVE_TRIALS_AT_SNAPSHOT = 0.5
 
 
 def get_benchmark_snapshot(benchmark_df,
-                           threshold=_DEFAULT_BENCHMARK_SAMPLE_NUM_THRESHOLD):
-    """Finds the latest time where all trials were still running
-    (or |threshold| * trials were still running). In most cases, this is the end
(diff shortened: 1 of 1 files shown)


analysis/data_utils.py
```diff
@@ -100,18 +100,17 @@ def filter_max_time(experiment_df, max_time):
 
 # Creating "snapshots" (see README.md for definition).
 
-_DEFAULT_BENCHMARK_SAMPLE_NUM_THRESHOLD = 1.0
+_MIN_FRACTION_OF_ALIVE_TRIALS_AT_SNAPSHOT = 0.5
```
jonathanmetzman, line 103: I'm confused, where is this variable used?

lszekeres: right below (just renamed this to be less confusing).

jonathanmetzman: ACK

PR #507, opened 2020-07-08
[NFC] help RQ worker reuse this code

zchcai:
This minor change will not affect the current functionality but make initialization easier to setup.
The previous code assumes `experiment_filestore/<experiment_name>` already exists. Since we want to directly use `build/builder.py` in local setting, this assumption may not hold for a RQ worker.


jonathanmetzman:
LGTM

PR #510, opened 2020-07-09
Add "merge_with_nonprivate" and "private" experiment configs.

jonathanmetzman:
In the experiment config when private is set to True, the "private"
field in experiment database entity will be set to True.

If merge_with_nonprivate is set to True in the experiment config,
the final report generated for the experiment will be generated using
merge_with_clobber on all experiments that are not marked private.

This will be used by our experiment service. Instead of users requesting
that a FuzzBench maintainer merge their report with others, it will be
done automatically.

One edge case this does not deal with is if another experiment is
started after the current one. In that case, that experiment will
be merged into this one (with this one give priority) which I think
is unintuitive behavior.


jonathanmetzman:
Laszlo, I think we discussed how the fact that we use pandas for merging+clobbering may become problematic since it will eventually operate on a lot of data. I think you said you could look into doing the clobbering on the SQL server side (so that we aren't querying all data and then filtering it in python). Could you take a look when you get a chance?

analysis/queries.py
```diff
@@ -34,3 +34,19 @@ def get_experiment_data(experiment_names):
...
     return pd.read_sql_query(snapshots_query.statement, db_utils.engine)
+
+
+def add_nonprivate_experiments_for_merge_with_clobber(experiment_names):
+    """Returns a new list containing experiment names preeceeded by a list of
+    nonprivate experiments in the order in which they were run. This is useful
+    if you want to combine reports from |experiment_names| and all nonprivate
+    experiments."""
+    nonprivate_experiments = db_utils.query(
+        Experiment.name).filter(~Experiment.private).filter(
```
lszekeres, line 45: nit: instead of .filter(A).filter(B), you can also use .filter(A, B)

analysis/queries.py
```diff
@@ -34,3 +34,19 @@ def get_experiment_data(experiment_names):
...
+            ~Experiment.name.in_(experiment_names)).order_by(
+                Experiment.time_created)
+    nonprivate_experiment_names = [
+        result[0] for result in nonprivate_experiments
```
lszekeres, line 49: nit: you might also use `result for result, in ...`

lszekeres:
> Laszlo, I think we discussed how the fact that we use pandas for merging+clobbering may become problematic since it will eventually operate on a lot of data. I think you said you could look into doing the clobbering on the SQL server side (so that we aren't querying all data and then filtering it in python). Could you take a look when you get a chance?

Sure, added a TODO for myself https://github.com/google/fuzzbench/issues/512

experiment/test_reporter.py
```diff
@@ -16,24 +16,37 @@
...
-    """Test that output_report writes the report and rsyncs it to the web
-    bucket."""
-    web_bucket = 'gs://web-bucket/experiment'
+def test_output_report_filestore(fs, experiment):
+    """Test that output_report writes the report and rsyncs it to the report
+    filestore."""
+    # Get the config.
+    config_filepath = os.path.join(os.path.dirname(__file__), 'test_data',
+                                   'experiment-config.yaml')
+    fs.add_real_file(config_filepath)
```
inferno-chromium, line 33: Do you need this, since we are reading the experiment config yaml content below.

analysis/generate_report.py
```diff
@@ -97,6 +97,13 @@ def get_arg_parser():
               'get a report of all trials in "A" except for afl++ and all the '
               'trials from "B". "Later experiments" are those whose names come '
               'later when passed to this script.'))
+    mutually_exclusive_group.add_argument(
+        '-p',
+        '--merge-with-clobber-nonprivate',
+        action='store_true',
+        default=False,
+        help=('Does merge-with-clobber but includes all experiments that are '
```
inferno-chromium, line 105: Can you expand more on description, unsure if someone will understand this completely.

jonathanmetzman: I don't want to repeat the help string for merge with clobber so I reference it explicitly now and tell people to read it.
And frankly, I don't expect anyone to use this option. It's mainly for me to debug what goes wrong with it since 
reporter.py will use this functionality.

analysis/queries.py
jonathanmetzman: I didn't know about that syntax thanks!
But I think I'll keep it the way it is, I think it's clearer in some ways (though your suggestion is more explicit about there being one item in the tuple).

experiment/test_reporter.py
jonathanmetzman: I do need this because we are using the fakefs and can't read it otherwise.

Issue #515, opened 2020-07-10
Makefile uses Python3 even though project doesn't support Python3.8


jonathanmetzman:
This has become a problem for me personally because python3.8 is now the default python3 on my machine (don't know why this is or if this happened recently).
The result of this is that pytype fails to install or work when make presubmit is run and the test `test_build_all_measurers` in `test_builder.py` hangs forever.

I guess the fix for this could be to specify Python3.7, but we might want to consider making the project work with other versions of Python.

inferno-chromium:
i ran into that hang as well, had to create a python3.7 virtualenv. but yeah 3.8 should be made to work.

jonathanmetzman:
I think we fixed this.

PR #516, opened 2020-07-10
Remove some blockers for Python3.8 support.

jonathanmetzman:
1. Only use pytype on 3.7 since it doesn't work on 3.8+
2. Skip a test that hangs if not on Python3.7.

Also, mark a test as "slow" instead of "long" and register the marker
to avoid getting the warnings.


presubmit.py
```diff
@@ -210,6 +210,13 @@ def lint(paths: List[Path]) -> bool:
 def pytype(paths: List[Path]) -> bool:
     """Run pytype on |path| if it is a python file. Return False if it fails
     type checking."""
+    # Check that we are on Python3.7.
+    if not sys.version.startswith('3.7'):
+        logs.error(
+            'Python version is: "%s". You should be using 3.7.'
+            'Not running pytype.', sys.version)
+        return False
```
jonathanmetzman, line 218: Returning False means presubmit will always fail. Not sure if that's the correct behavior if it should return True.

presubmit.py
```diff
@@ -210,6 +210,13 @@ def lint(paths: List[Path]) -> bool:
 def pytype(paths: List[Path]) -> bool:
     """Run pytype on |path| if it is a python file. Return False if it fails
     type checking."""
+    # Check that we are on Python3.7.
```
inferno-chromium, line 213: I think we should fail presubmit.py on anything other than python 3 (sys.version_info.major != 3).
and this should check sys.version_info.minor <= 7 and reference
https://github.com/google/pytype/issues/440 (pytype is not supported on 3.8 yet).

presubmit.py
inferno-chromium: i thought your intention here was to skip pytype on anything > 3.7. if that is case, then should return true.

experiment/build/test_builder.py
```diff
@@ -62,6 +63,8 @@ def get_benchmarks_or_fuzzers(benchmarks_or_fuzzers_directory, filename,
     ]
 
 
+@pytest.mark.skipif(not sys.version.startswith('3.7.7'),
```
inferno-chromium, line 66: Why need 3.7.7 exact, should check major =3 and minor =7. it is easy to go higher version in 3.7.

presubmit.py
jonathanmetzman: That is my intention. Not sure why it should return True though? They are using the wrong version so I'm reporting it as a failure since I can't check instead of reporting at as a success when I can't check.

jonathanmetzman:
Thanks for the last review, I didn't know about `version_info`. This is better than string matching.
The other option here is not to land this patch and to just force Python3.7 in Make.
That may turn away some users but it saves on complexity.

jonathanmetzman:
Related: https://github.com/google/fuzzbench/issues/515

experiment/build/test_builder.py
jonathanmetzman: Done.

presubmit.py
jonathanmetzman: >I think we should fail presubmit.py on anything other than python 3 (sys.version_info.major != 3).

I'll do it to make the code look nice but it is pointless since presubmit.py can't even start executing in Python2. It is not a syntactically valid Python2 file. Also people aren't likely to use python2 (or 4 whenever it comes out) because Make doesn't specify it.

jonathanmetzman: Done.

experiment/build/test_builder.py
jonathanmetzman: Done.

jonathanmetzman:
It's not clear to me that pytype will support 3.8 any time soon so if we want to switch to 3.8 officially we will need to use a different typechecker (or get rid of it entirely).

presubmit.py
jonathanmetzman: Just wanted to make my reasoning more clear but if you think it should return True I can change it.

inferno-chromium: i mean does it work if you just type "make presubmit", it does fail on pytype, but still continues the other checks. i remember you implemented that way to do all tests. then it is fine.

presubmit.py
```diff
@@ -210,6 +210,15 @@ def lint(paths: List[Path]) -> bool:
 def pytype(paths: List[Path]) -> bool:
     """Run pytype on |path| if it is a python file. Return False if it fails
     type checking."""
+    # Pytype isn't supported on Python3.8+. See
+    # https://github.com/google/pytype/issues/440.
+    assert sys.version_info.major == 3
```
inferno-chromium, line 215: Shouldn't this assert be in main(), we shouldn't run presubmit if we are not on version 3, also add a message at end
assert sys.version_info.major == 3, "Needs python 3"

presubmit.py
lszekeres: Sorry, top level question: what do we want to use 3.7 or 3.8?

Should we set the version we want for our venv? E.g., if we want to use 3.7, we can update the Makefile to initialize the venv with:
```
${VENV_ACTIVATE}: requirements.txt
	python3.7 -m venv .venv
```
which should ensure that the presubmit runs with the right version.

jonathanmetzman: I think offline we (Abhishek and I) decided we didn't want to force a Python minor version on users.

presubmit.py
jonathanmetzman: This assertion is purely for show so that the check for ".7" doesn't look weird. As I mentioned, until Python 4 comes around there's no danger of using the wrong Python major version since this file is syntactically invalid under Python2.

jonathanmetzman:
I return True now as we discussed offline. This way we can test the other parts using python3.8 in CI.

presubmit.py
lszekeres: That's fine, but it'd still be useful to decide which one is our chosen/preferred/recommended version and indicate that, so people can know with which version they can expect to have most things working. E.g., what is more important: to be able to use 3.8 features or pytype. If we prefer say 3.7, in the Makefile we could eg try to initialize venv with python3.7 first, and if that fails use python3

jonathanmetzman: 3.7 is the preferred one. I think having the makefile try python3.7 before python3 is a good idea.

lszekeres: wanna add it in this pr?

jonathanmetzman: I'd rather land this and unbreak fuzzbench development on my machine and figure out how to improve this later.

PR #522, opened 2020-07-13
Add end-to-end integration test skeleton for new architecture.

lszekeres:
The PR sets up the infrastructure for the continuous integration testing of the new architecture we're developing.
 
- Add a skeleton of tests/checks that will run after a test experiment run, as part of the integration test.
- Add Docker Compose specification for the end to end integration test. The spec extends the main docker compose specification with a container that runs the test script mentioned above.
- Add Makefile target for running the e2e integration test.

The [next PR](https://github.com/google/fuzzbench/pull/523) will set this up as a CI test at prepsubmit (with GH actions), so we can validate each step of development of the new architecture. See the output of the test at https://github.com/google/fuzzbench/pull/523/checks?check_run_id=867126894, under presumbit / Run end to end CI test.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,15 @@ run-experiment:
 
 # Development.
 
+run-end-to-end-test: export COMPOSE_PROJECT_NAME := e2e-test
+run-end-to-end-test: export COMPOSE_FILE := compose/fuzzbench.yaml:compose/ci.yaml
+run-end-to-end-test:
+	docker-compose build
+	docker-compose up --detach queue-server
+	docker-compose up --scale worker=3 run-experiment worker
+	docker-compose run ci-tests; STATUS=$$?; \
+	docker-compose down; exit $$STATUS
+
 include docker/build.mk
 include docker/generated.mk
 
--- a/compose/ci.yaml
+++ b/compose/ci.yaml
@@ -1,0 +1,9 @@
+version: "3"
+
+services:
+
+  ci-tests:
+    image: fuzzbench
+    links:
+      - queue-server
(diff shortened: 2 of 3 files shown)


docker-compose.ci.yml
```diff
@@ -0,0 +1,9 @@
+version: "3"
```
jonathanmetzman, line 1: Can this go in the docker/directory?

fuzzbench/ci_test.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2020 Google LLC
```
jonathanmetzman, line 1: Most of our tests start with the word "test_".
I don't really think this is a good place to change it so that they end with "_test".

docker-compose.ci.yml
```diff
@@ -0,0 +1,9 @@
...
+
+services:
+
+  ci-tests:
+    image: fuzzbench
+    links:
+      - queue-server
+    command: python3 -m pytest -vv ci_test.py
```
jonathanmetzman, line 9: I would do `pytest -vv -k ci_test`

docker-compose.ci.yml
lszekeres: Moved it under compose/ as discussed in https://github.com/google/fuzzbench/pull/521

fuzzbench/ci_test.py
lszekeres: Done

docker-compose.ci.yml
lszekeres: done

PR #532, opened 2020-07-14
support zlib

laurentsimon:


fuzzers/klee/fuzzer.py
```diff
@@ -210,8 +210,13 @@ def covert_seed_inputs(ktest_tool, input_klee, input_corpus):
         if not os.path.isfile(seedfile):
             continue
 
-        if os.path.getsize(seedfile) > get_size_for_benchmark():
-            continue
+        # Truncate the seed to the max size for the benchmark
+        fsize = os.path.getsize(seedfile)
```
jonathanmetzman, line 214: It's a fuzzer file so I'm more lax, but we usually would use the names "file_size" and "benchmark_size" for variables instead of fsize and bsize in this project.

jonathanmetzman:
Argh another formatting error. I'm going to merge this. Can you make another PR after this with a dummy change to fuzzer.py, run `make presubmit` remove dummy change and the put PR up for review? thanks!

PR #535, opened 2020-07-14
Add short description of new directories to doc.

lszekeres:


lszekeres:
> I think this just creates more confusion since this code is in very early stage, can probably keep it as-is for now and commit after 2-3 weeks.

yea, i'd lean towards doing that too. the pr is following up on https://github.com/google/fuzzbench/pull/521#pullrequestreview-448286174 though -- @jonathanmetzman let us know your preference

docs/getting-started/contributing_code.md
```diff
@@ -19,10 +19,12 @@ FuzzBench.
 * **analysis/** - code for analyzing fuzzer performance using statistical tests.
 * **benchmarks/** - benchmarks integrated in the FuzzBench platform (e.g. OpenSSL).
 * **common/** - common helper modules (e.g. logging, new process handling, etc).
+* **compose/** - Docker Compose files (not used in production yet).
```
jonathanmetzman, line 22: Can you add something indicating this is for the new arch as well?

docs/getting-started/contributing_code.md
```diff
@@ -19,10 +19,12 @@ FuzzBench.
...
 * **database/** - database handling code.
 * **docker/** - Dockerfiles for infra images (does not include fuzzer images).
 * **docs/** - this documentation.
 * **experiment/** - code for running FuzzBench experiments.
+* **fuzzbench/** - code for new architecture (not used in production yet).
```
jonathanmetzman, line 27: Maybe also mention the new arch is a Work in progress (I think that's kinda implied but I'd like to be explicit).

jonathanmetzman:
I think if we are clear that it's for a new feature being developed it wouldn't be confusing. Anyone disagree?

My thoughts is it is very confusing to have a `fuzzbench` directory given that it isn't used right now so documenting its purpose is a good idea. I think actually having a short README in fuzzbench/ since not everyone will read the docs is a good idea.

lszekeres:
> I think if we are clear that it's for a new feature being developed it wouldn't be confusing. Anyone disagree?
> 
> My thoughts is it is very confusing to have a `fuzzbench` directory given that it isn't used right now so documenting its purpose is a good idea. I think actually having a short README in fuzzbench/ since not everyone will read the docs is a good idea.



> I think if we are clear that it's for a new feature being developed it wouldn't be confusing. Anyone disagree?
> 
> My thoughts is it is very confusing to have a `fuzzbench` directory given that it isn't used right now so documenting its purpose is a good idea. I think actually having a short README in fuzzbench/ since not everyone will read the docs is a good idea.

makes sense, i can add a README pointing out that it's WIP

PR #537, opened 2020-07-15
Add more source files to fuzzbench image.

lszekeres:
Also fix pytest invocation for e2e test by specifying which module to run.

### DIFF
--- a/compose/ci.yaml
+++ b/compose/ci.yaml
@@ -6,4 +6,4 @@ services:
     image: fuzzbench
     links:
       - queue-server
-    command: python3 -m pytest -vv -k test_e2e_run
+    command: python3 -m pytest -vv fuzzbench/test_e2e_run.py
--- a/docker/fuzzbench/Dockerfile
+++ b/docker/fuzzbench/Dockerfile
@@ -22,9 +22,16 @@ RUN wget https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_
 
 WORKDIR /fuzzbench
 
-ADD requirements.txt /fuzzbench/requirements.txt
+COPY requirements.txt requirements.txt
 RUN pip3 install -r requirements.txt
 
-ADD fuzzbench/*.py /fuzzbench/
+COPY alembic.ini alembic.ini
+COPY analysis analysis
+COPY benchmarks benchmarks
+COPY common common
+COPY database database
+COPY docker docker
+COPY fuzzbench fuzzbench
+COPY fuzzers fuzzers
 
-CMD python3 -u run_experiment.py
+CMD PYTHONPATH=. python3 -u fuzzbench/run_experiment.py
--- a/fuzzbench/run_experiment.py
+++ b/fuzzbench/run_experiment.py
@@ -18,7 +18,7 @@ import time
 import redis
 import rq
 
-import fake_jobs
+from fuzzbench import fake_jobs
 
 
 def run_experiment():


jonathanmetzman:
> fine for now, but please do think of a better name for fuzzbench dir, that directory name should mean a purpose, e.g. controller, task_management, queue_management, or something better.

@lszekeres can speak on this but I think this is going to be the main directory that contains python files in this project. So now fuzzbench will be a python package.

I think naming the docker image fuzzbench is a mistake though. Dispatcher (or something else conveying it is the brain of the experiment) seems good to me.

compose/ci.yaml
```diff
@@ -6,4 +6,4 @@ services:
     image: fuzzbench
     links:
       - queue-server
-    command: python3 -m pytest -vv -k test_e2e_run
+    command: python3 -m pytest -vv fuzzbench/test_e2e_run.py
```
jonathanmetzman, line 9: Did the `-k` format not work?

docker/fuzzbench/Dockerfile
```diff
@@ -22,9 +22,16 @@ RUN wget https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_
 
 WORKDIR /fuzzbench
 
-ADD requirements.txt /fuzzbench/requirements.txt
+COPY requirements.txt requirements.txt
 RUN pip3 install -r requirements.txt
 
-ADD fuzzbench/*.py /fuzzbench/
+COPY alembic.ini alembic.ini
```
jonathanmetzman, line 28: I think alembic.ini is probably not needed as it is used only in development (it is used for schema migrations in SQL).

compose/ci.yaml
lszekeres: It did not. (ran all the other tests in the new files that the pr adds)

docker/fuzzbench/Dockerfile
lszekeres: Right, we'd need it to initialize the db schema for new db instances, e.g., at the beginning of the e2e test we'd use an ephemeral postgres container, or when setting up a postgres server for local experiments.

lszekeres:
> > fine for now, but please do think of a better name for fuzzbench dir, that directory name should mean a purpose, e.g. controller, task_management, queue_management, or something better.
> 
> @lszekeres can speak on this but I think this is going to be the main directory that contains python files in this project. So now fuzzbench will be a python package.

Yea the idea is to migrate to a conventional python project tree, with a "python package" (ie. directory with the project name) inside the repo.

> I think naming the docker image fuzzbench is a mistake though. Dispatcher (or something else conveying it is the brain of the experiment) seems good to me.

Pls see a prior discussion here: https://github.com/google/fuzzbench/pull/521#discussion_r454078302

docker/fuzzbench/Dockerfile
jonathanmetzman: Are we sure about the db being ephemeral? I thought we seed this is TBD (my personal thought is that this isn't a good idea (unlike redis which can be ephemeral)).

Also, I don't think you need it for initializing a new db. Just for migrating between schemas. This is how local experiments work. In dispatcher.py there is code to initialize the db that can be done by sqlalchemy and without alembic.

(I'm not horribly concerned about this file polluting the image btw, just trying to clarify things :-)

lszekeres: > Are we sure about the db being ephemeral? I thought we said this is TBD (my personal thought is that this isn't a good idea (unlike redis which can be ephemeral) I don't see how it would work with the fuzzbench service also).

The service of course would not use an ephemeral db, but eg. the e2e test running in ci (ie. it'd create a fresh postgres docker container with e2e-test docker compose config).

> Also, I don't think you need it for initializing a new db. Just for migrating between schemas. This is how local experiments work. In dispatcher.py there is code to initialize the db that can be done by sqlalchemy and without alembic.

Ah cool! So that'd create the tables too in a fresh postgres instance (eg used during the e2e test), before inserting rows?

jonathanmetzman:
On Wed, Jul 15, 2020 at 10:59 PM Laszlo Szekeres <notifications@github.com>
wrote:

> *@lszekeres* commented on this pull request.
> ------------------------------
>
> In docker/fuzzbench/Dockerfile
> <https://github.com/google/fuzzbench/pull/537#discussion_r455527310>:
>
> >  RUN pip3 install -r requirements.txt
>
> -ADD fuzzbench/*.py /fuzzbench/
> +COPY alembic.ini alembic.ini
>
> Are we sure about the db being ephemeral? I thought we said this is TBD
> (my personal thought is that this isn't a good idea (unlike redis which can
> be ephemeral) I don't see how it would work with the fuzzbench service
> also).
>
> The service of course would not use an ephemeral db, but eg. the e2e test
> running in ci (ie. it'd create a fresh postgres docker container with
> e2e-test docker compose config).
>

Got it.



> Also, I don't think you need it for initializing a new db. Just for
> migrating between schemas. This is how local experiments work. In
> dispatcher.py there is code to initialize the db that can be done by
> sqlalchemy and without alembic.
>
> Ah cool! So that'd create the tables too in a fresh postgres instance (eg
> used during the e2e test), before inserting rows?
>

Yeah. I think this does it:
https://github.com/google/fuzzbench/blob/master/conftest.py#L44

PR #538, opened 2020-07-15
Run test_e2e_run only in the context of an end to end test.

lszekeres:
This test is not a unit test, but is for checking the results of an end-to-end
test. With this change we avoid it running during regular pytest unit test runs.

### DIFF
--- a/compose/ci.yaml
+++ b/compose/ci.yaml
@@ -6,4 +6,6 @@ services:
     image: fuzzbench
     links:
       - queue-server
+    environment:
+      E2E_INTEGRATION_TEST: 1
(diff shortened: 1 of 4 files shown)


compose/ci.yaml
```diff
@@ -6,4 +6,6 @@ services:
     image: fuzzbench
     links:
       - queue-server
+    environment:
```
jonathanmetzman, line 9: I don't love that this file is called CI since it is really about end-to-end tests and we use CI for other things as well.

lszekeres: Good point, let me rename it!

PR #540, opened 2020-07-15
Remove variants, simplify code.

### LINKED ISSUE #539: Remove code complexity with unused variants code.

inferno-chromium:
Fixes #539.

### DIFF
--- a/common/fuzzer_utils.py
+++ b/common/fuzzer_utils.py
@@ -59,11 +59,6 @@ class FuzzerDirectory:
         return os.path.join(self.directory, 'builder.Dockerfile')
 
     @property
-    def variants_yaml(self):
-        """Returns the path to the variants.yaml file in fuzzer directory."""
-        return os.path.join(self.directory, 'variants.yaml')
-
-    @property
     def dockerfiles(self):
         """Returns a list of paths to the runner and builder dockerfiles in the
         fuzzer directory."""
@@ -121,10 +116,10 @@ def validate(fuzzer):
         return False
 
 
-def get_fuzzer_from_config(fuzzer_config: dict) -> str:
+def get_fuzzer_from_config(fuzzer_config: dict) -> Optional[str]:
     """Returns the fuzzer of |fuzzer_config| for a non-variant fuzzer or returns
     the name for a fuzzer variant."""
-    return fuzzer_config.get('name', fuzzer_config['fuzzer'])
+    return fuzzer_config.get('fuzzer')
 
 
 def get_fuzzer_names():
@@ -134,13 +129,8 @@ def get_fuzzer_names():
 
 def get_fuzzer_configs(fuzzers=None):
     """Returns the list of all fuzzer and variant configurations."""
-    # Import it here to avoid yaml dependency in runner.
-    # pylint: disable=import-outside-toplevel
-    from common import yaml_utils
-
     fuzzers_dir = os.path.join(utils.ROOT_DIR, 'fuzzers')
     fuzzer_configs = []
-    names = set()
     for fuzzer in os.listdir(fuzzers_dir):
         if not os.path.isfile(os.path.join(fuzzers_dir, fuzzer, 'fuzzer.py')):
             continue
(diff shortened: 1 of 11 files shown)


common/fuzzer_utils.py
```diff
@@ -121,10 +116,10 @@ def validate(fuzzer):
         return False
 
 
-def get_fuzzer_from_config(fuzzer_config: dict) -> str:
+def get_fuzzer_from_config(fuzzer_config: dict) -> Optional[str]:
```
jonathanmetzman, line 119: I'd like to get rid of the command line version of variants too frankly.
Why can this return None? If it does return None does that break callers?

experiment/run_experiment.py
```diff
@@ -168,31 +168,6 @@ def validate_fuzzer(fuzzer: str):
         raise Exception('Fuzzer "%s" does not exist.' % fuzzer)
 
 
-def validate_fuzzer_config(fuzzer_config):
```
jonathanmetzman, line 171: If we are no longer validating fuzzer configs maybe we should stop supporting them?

common/fuzzer_utils.py
inferno-chromium: This is just to satisfy linter, it will never return None. This is just return 'fuzzer' attribute which should always exist.

Command line variants are all removed already - https://github.com/google/fuzzbench/pull/452
This fuzzer config is just a way of passing in old code, it is just a dict with one fuzzer attribute [there are no env, name attribute anymore, see they are removed in this cl]. i can remove fuzzer-config entirely and use fuzzer name only, but that should in another cl, this cl is already complicated.

docker/generate_makefile.py
```diff
@@ -14,8 +14,6 @@
 """Simple generator for local Makefile rules."""
```
jonathanmetzman, line 14: I'm pretty sure generate_makefile was only introduced for variants so in theory we can get rid of it.
Since we plan on putting more build logic in python we could reuse some of this. But frankly, I think starting fresh is a better idea.

common/fuzzer_utils.py
jonathanmetzman: It is confusing to me why we even have parts of the code using fuzzer configs even though we don't support them anymore.

experiment/run_experiment.py
inferno-chromium: yes we dont support them at all, not accepted through command line. we just create fuzzer configs files with fuzzer attribute to pass to dispatcher.

docker/generate_makefile.py
inferno-chromium: Yes i would let @zchcai do this as part of redesign.

common/fuzzer_utils.py
jonathanmetzman: Sorry didn't see your last comment before I left (the rest of) my review.
I'm fine with doing it in another CL.

experiment/run_experiment.py
jonathanmetzman: Even this I'd like to get rid of if possible then. There's no reason to support it anymore.
This can come later if it would be easier.

inferno-chromium:
Regarding @jonathanmetzman "It is confusing to me why we even have parts of the code using fuzzer configs even though we don't support them anymore."
- Old code create fuzzer config files to put in GCS right so that dispatcher knows which fuzzers to run. that continues to be the case, we can instead put a fuzzer list in a file, but i think this should go into next CL. thoughts ?

experiment/run_experiment.py
inferno-chromium: Yes next CL, i move faster in increments rather than gigantic cls.

jonathanmetzman:
> Regarding @jonathanmetzman "It is confusing to me why we even have parts of the code using fuzzer configs even though we don't support them anymore."
> 
> * Old code create fuzzer config files to put in GCS right so that dispatcher knows which fuzzers to run. that continues to be the case, we can instead put a fuzzer list in a file, but i think this should go into next CL. thoughts ?

Yeah fine to do in another CL. Putting the fuzzer list in the config file that gets past to the dispatcher (as we do with the benchmarks) sgtm.

zchcai:
Does this PR replace the `config/fuzzer-configs` folder by a list of fuzzers in `config/experiment.yaml`?

jonathanmetzman:
> Does this PR replace the `config/fuzzer-configs` folder by a list of fuzzers in `config/experiment.yaml`?

It does not, though this is definitely a goal of mine at some point.

inferno-chromium:
> > Does this PR replace the `config/fuzzer-configs` folder by a list of fuzzers in `config/experiment.yaml`?
> 
> It does not, though this is definitely a goal of mine at some point.

Next one, let me see what is required. Will still keep list of fuzzers accepted through command line, but instead of fuzzer-configs, it will be a single file with list of fuzzers.

jonathanmetzman:
LGTM

> > > Does this PR replace the `config/fuzzer-configs` folder by a list of fuzzers in `config/experiment.yaml`?
> > 
> > 
> > It does not, though this is definitely a goal of mine at some point.
> 
> Next one, let me see what is required. Will still keep list of fuzzers accepted through command line, but instead of fuzzer-configs, it will be a single file with list of fuzzers.

i think it can just go in the config file we pass to the dispatcher. We already add the benchmarks to it.

inferno-chromium:
> LGTM
> 
> > > > Does this PR replace the `config/fuzzer-configs` folder by a list of fuzzers in `config/experiment.yaml`?
> > > 
> > > 
> > > It does not, though this is definitely a goal of mine at some point.
> > 
> > 
> > Next one, let me see what is required. Will still keep list of fuzzers accepted through command line, but instead of fuzzer-configs, it will be a single file with list of fuzzers.
> 
> i think it can just go in the config file we pass to the dispatcher. We already add the benchmarks to it.

aha, thanks for the tip, will do it now!

inferno-chromium:
> Does this PR replace the `config/fuzzer-configs` folder by a list of fuzzers in `config/experiment.yaml`?

Done in https://github.com/google/fuzzbench/pull/542

Issue #536, opened 2020-07-15
Use BuildKit for better docker build caching


lszekeres:
By default `docker build` can only use the local cache. Currently, when we build images, in order to take advantage of the local cache, we pull images from the registry first. These steps can be eliminated and our build process can be simplified by using BuildKit. (This includes the local `make` based building, the GCB based building, and it will make the new distributed queue-based build process much simpler too.)

BuildKit is an advanced docker image building backend, which (among other things) can use the (remote) registry as a cache. I.e., with BuilKit enabled, when we build images like this: 
```shell
docker build -t fuzzbench/base-runner --build-arg BUILDKIT_INLINE_CACHE=1 ...
docker push fuzzbench/base-runner
```
Then the caching information is inlined and uploaded to the registry, so when we build an image on another machine, and use the --cache-from flag:
```shell
docker build --cache-from fuzzbench/base-runner ...
```
BuildKit will check if the there's a cache hit and automatically pulls data from the registry, if there is. Also, it downloads only the parts of the image that are actually useful.

BuildKit can be enabled by setting the `DOCKER_BUILDKIT=1` env var (before running `docker build`). See:
https://docs.docker.com/engine/reference/commandline/build/#specifying-external-cache-sources

While BuildKit is part of Docker now, it can also be used as a standalone tool:
https://github.com/moby/buildkit
https://github.com/genuinetools/img

Alternative tools we could use that support using the registry as a remote cache:
https://github.com/GoogleContainerTools/kaniko
https://github.com/uber/makisu

zchcai:
If I understand correctly, these two will only be used once at the beginning, for initializing the registry.
> docker build -t gcr.io/fuzzbench/base-runner --build-arg BUILDKIT_INLINE_CACHE=1
docker push gcr.io/fuzzbench/base-runner

And the following is our logic in new arch:
> DOCKER_BUILDKIT=1 docker build --cache-from gcr.io/fuzzbench/base-runner

lszekeres:
There'd be just one build step for each image, that uses the registry as cache and inlines the cache info in newly built images, ie:
```
export DOCKER_BUILDKIT=1
docker build --cache-from $REGISTRY/base-runner --build-arg BUILDKIT_INLINE_CACHE=1 ...
docker push ...
```
(Note that in the single machine scenario we don't need to push and we don't need a registry, we need this central cache when we're doing a distributed build with multiple machines.)

inferno-chromium:
[...]

PR #543, opened 2020-07-16
Initial prototype of build logic consolidation

lszekeres:
Running a FuzzBench experiment involves building hundreds of docker images for all the different fuzzer-benchmark pairs. Currently, we build these images using Google Cloud Build in prod, and using Makefiles locally. The logic of building the images and the dependencies between them is currently encoded in two places, ie., in docker/gcb/*.yaml and docker/generate_makefile.py. This is difficult to maintain and [very error prone](https://github.com/google/fuzzbench/issues/336). To complicate things further, we are also implementing distributed builds as part of a new queue-based architecture (https://github.com/google/fuzzbench/pull/521).

To eliminate the above described redundancy and make the FuzzBench build system simple and easy to maintain, we introduce a `get_images_to_build(fuzzers, benchmarks)` function that returns the set of buildable images with their properties describing how to build them, as well as their dependencies (as python objects). This can be used to simply generate the input for different build tools, ie. the 1) GCB build spec, 2) the Makefile, and 3) the (rq) jobs to enqueue in the new architecture.

The PR includes the initial prototype of:
- the `get_images_to_build` function, which relies on
- an easy to read/modify config file describing the image types and their dependencies
- generator for Cloud Build spec
- generator for Makefile rules

### DIFF
--- a/docker/image_types.yaml
+++ b/docker/image_types.yaml
@@ -1,0 +1,67 @@
+# Defines the different types of docker images that FuzzBench uses and their
+# dependency graph.
+
+'base-image':
+  tag: 'base-image'
+  context: 'docker/base-image'
+
+'base-builder':
+  tag: 'base-builder'
+  context: 'docker/base-builder'
+  depends_on:
+    - 'base-image'
+
+'base-runner':
+  tag: 'base-runner'
+  context: 'docker/base-runner'
+  depends_on:
+    - 'base-image'
+
+'coverage-builder':
+  tag: 'builders/coverage'
+  context: 'fuzzers/coverage'
+  dockerfile: 'fuzzers/coverage/builder.Dockerfile'
+  depends_on:
+    - 'base-builder'
+
+'coverage-{benchmark}-builder':
+  tag: 'builders/coverage/{benchmark}'
+  context: '.'
+  dockerfile: 'docker/benchmark-builder/Dockerfile'
+  depends_on:
+    - 'coverage-builder'
+
+'{fuzzer}-builder':
+  tag: 'builders/{fuzzer}'
+  context: 'fuzzers/{fuzzer}'
+  dockerfile: 'fuzzers/{fuzzer}/builder.Dockerfile'
+  depends_on:
+    - 'base-builder'
+
+'{fuzzer}-{benchmark}-builder':
+  tag: 'builders/{fuzzer}/{benchmark}'
+  context: '.'
+  dockerfile: 'docker/benchmark-builder/Dockerfile'
+  build_arg:
+    - 'fuzzer={fuzzer}'
+    - 'benchmark={benchmark}'
+  depends_on:
+    - '{fuzzer}-builder'
+
+'{fuzzer}-{benchmark}-intermediate-runner':
+  tag: 'runners/{fuzzer}/{benchmark}-intermediate'
+  context: 'fuzzers/{fuzzer}'
+  dockerfile: 'fuzzers/{fuzzer}/runner.Dockerfile'
+  depends_on:
+    - 'base-runner'
+
+'{fuzzer}-{benchmark}-runner':
+  tag: 'runners/{fuzzer}/{benchmark}'
+  context: '.'
+  dockerfile: 'docker/benchmark-runner/Dockerfile'
+  build_arg:
+    - 'fuzzer={fuzzer}'
+    - 'benchmark={benchmark}'
+  depends_on:
+    - '{fuzzer}-{benchmark}-builder'
+    - '{fuzzer}-{benchmark}-intermediate-runner'
--- a/experiment/build/docker_images.py
+++ b/experiment/build/docker_images.py
@@ -1,0 +1,56 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Provides the set of buildable images and their dependencies."""
+
+import yaml
+
+
+def subst(template, fuzzer, benchmark):
+    """Replaces {fuzzer} or {benchmark} with |fuzzer| or |benchmark| in
+    |template| string."""
+    return template.format(fuzzer=fuzzer, benchmark=benchmark)
+
+
+def instantiate_image_obj(name_template, obj_template, fuzzer, benchmark):
+    """Instantiates an image object from a template for a |fuzzer| - |benchmark|
+    pair."""
+    name = subst(name_template, fuzzer, benchmark)
+    obj = obj_template.copy()
+    for key in obj:
+        if key in ('build_arg', 'depends_on'):
+            obj[key] = [subst(item, fuzzer, benchmark) for item in obj[key]]
+        else:
+            obj[key] = subst(obj[key], fuzzer, benchmark)
+    return name, obj
+
+
+def get_image_type_templates():
+    """Loads the image types config that contains "templates" describing how to
+    build them and their dependencies."""
+    with open('docker/image_types.yaml') as config_file:
+        templates = yaml.load(config_file, yaml.SafeLoader)
+        return templates
+
+
+def get_images_to_build(fuzzers, benchmarks):
+    """Returns the set of buildable images."""
+    images = {}
+    templates = get_image_type_templates()
+    for fuzzer in fuzzers:
+        for benchmark in benchmarks:
+            for name_templ, obj_templ in templates.items():
+                name, obj = instantiate_image_obj(name_templ, obj_templ, fuzzer,
+                                                  benchmark)
+                images[name] = obj
+    return images
--- a/experiment/build/docker_images.py
+++ b/experiment/build/docker_images.py
@@ -16,7 +16,7 @@
 from common import yaml_utils
 
 
-def _subst(template, fuzzer, benchmark):
+def _substitute(template, fuzzer, benchmark):
     """Replaces {fuzzer} or {benchmark} with |fuzzer| or |benchmark| in
     |template| string."""
     return template.format(fuzzer=fuzzer, benchmark=benchmark)
@@ -25,13 +25,15 @@ def _subst(template, fuzzer, benchmark):
 def _instantiate_image_obj(name_template, obj_template, fuzzer, benchmark):
     """Instantiates an image object from a template for a |fuzzer| - |benchmark|
     pair."""
-    name = _subst(name_template, fuzzer, benchmark)
+    name = _substitute(name_template, fuzzer, benchmark)
     obj = obj_template.copy()
     for key in obj:
         if key in ('build_arg', 'depends_on'):
-            obj[key] = [_subst(item, fuzzer, benchmark) for item in obj[key]]
+            obj[key] = [
+                _substitute(item, fuzzer, benchmark) for item in obj[key]
+            ]
         else:
-            obj[key] = _subst(obj[key], fuzzer, benchmark)
+            obj[key] = _substitute(obj[key], fuzzer, benchmark)
     return name, obj
 
 
--- a/experiment/build/docker_images.py
+++ b/experiment/build/docker_images.py
@@ -13,44 +13,42 @@
 # limitations under the License.
 """Provides the set of buildable images and their dependencies."""
 
-import yaml
+from common import yaml_utils
 
 
-def subst(template, fuzzer, benchmark):
+def _subst(template, fuzzer, benchmark):
     """Replaces {fuzzer} or {benchmark} with |fuzzer| or |benchmark| in
     |template| string."""
     return template.format(fuzzer=fuzzer, benchmark=benchmark)
 
 
-def instantiate_image_obj(name_template, obj_template, fuzzer, benchmark):
+def _instantiate_image_obj(name_template, obj_template, fuzzer, benchmark):
     """Instantiates an image object from a template for a |fuzzer| - |benchmark|
     pair."""
-    name = subst(name_template, fuzzer, benchmark)
+    name = _subst(name_template, fuzzer, benchmark)
     obj = obj_template.copy()
     for key in obj:
         if key in ('build_arg', 'depends_on'):
-            obj[key] = [subst(item, fuzzer, benchmark) for item in obj[key]]
+            obj[key] = [_subst(item, fuzzer, benchmark) for item in obj[key]]
         else:
-            obj[key] = subst(obj[key], fuzzer, benchmark)
+            obj[key] = _subst(obj[key], fuzzer, benchmark)
     return name, obj
 
 
-def get_image_type_templates():
+def _get_image_type_templates():
     """Loads the image types config that contains "templates" describing how to
     build them and their dependencies."""
-    with open('docker/image_types.yaml') as config_file:
-        templates = yaml.load(config_file, yaml.SafeLoader)
-        return templates
+    return yaml_utils.read('docker/image_types.yaml')
 
 
 def get_images_to_build(fuzzers, benchmarks):
     """Returns the set of buildable images."""
     images = {}
-    templates = get_image_type_templates()
+    templates = _get_image_type_templates()
     for fuzzer in fuzzers:
         for benchmark in benchmarks:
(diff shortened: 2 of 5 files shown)


docker/image_types.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+'base-image':
+  tag: 'base-image'
+  context: 'docker/base-image'
+
+'base-builder':
+  tag: 'base-builder'
+  context: 'docker/base-builder'
+  depends_on:
+    - 'base-image'
```
zchcai, line 12: It also needs `pull-base-clang` from `gcr.io/oss-fuzz-base` according to current `docker/build.mk`. Does BuiltKit will remove this dependency?

experiment/build/docker_images.py
```diff
@@ -0,0 +1,56 @@
+# Copyright 2020 Google LLC
```
zchcai, line 1: How to deal with `pull-image` jobs?

docker/image_types.yaml
lszekeres: Yes

experiment/build/docker_images.py
lszekeres: BuildKit (https://github.com/google/fuzzbench/issues/536) should take care of that. 
Also please note that this PR is just a prototype showing how https://github.com/google/fuzzbench/issues/336 can be approached, but @Tanq16 would work on implementing it end to end.

lszekeres:
> The old architecture build logic involves three things: (see experiment/dispatcher.py: build_images_for_trials)
> 
> 1. base images: base-image, base-runner, base-builder
> 2. benchmark measurers/coverage images + copy coverage binaries
> 3. trial images for (fuzzer, benchmark) pairs
> 
> We put (2) in front of (3) because we don't need to build trial images for those benchmark failed in (2).
> However, this pattern is more like an optimization, not the strict dependency logic.
> 
> Shall we remove (2) and put it into the place when we prepare measure jobs later?
> Or we code this dependency here too, along with the dependency in docker/image_types.yaml?

As you say that's an "optimization" not a real dependency between images, so we should not encode that here, but in the rq job creation logic.

experiment/build/docker_images.py
zchcai: Sounds good, thanks.

zchcai:
> > The old architecture build logic involves three things: (see experiment/dispatcher.py: build_images_for_trials)
> > 
> > 1. base images: base-image, base-runner, base-builder
> > 2. benchmark measurers/coverage images + copy coverage binaries
> > 3. trial images for (fuzzer, benchmark) pairs
> > 
> > We put (2) in front of (3) because we don't need to build trial images for those benchmark failed in (2).
> > However, this pattern is more like an optimization, not the strict dependency logic.
> > Shall we remove (2) and put it into the place when we prepare measure jobs later?
> > Or we code this dependency here too, along with the dependency in docker/image_types.yaml?
> 
> As you say that's an "optimization" not a real dependency between images, so we should not encode that here, but in the rq job creation logic.

Yes. This makes more sense.

experiment/build/test_docker_images.py
```diff
@@ -0,0 +1,28 @@
...
+def test_dependencies_exist():
+    """Tests that if an image has a dependency, then the dependency exist among
+    the images."""
+    fuzzers = ['afl', 'libfuzzer']
+    benchmarks = ['libxml', 'libpng']
+    all_images = docker_images.get_images_to_build(fuzzers, benchmarks)
+    for image in all_images.values():
+        if 'depends_on' in image:
+            for dep in image['depends_on']:
+                assert dep in all_images
```
inferno-chromium, line 28: nit: in tests, usually nice to match against the exact list/set to all_images.

experiment/build/generate_makefile.py
```diff
@@ -0,0 +1,50 @@
...
+
+    for name, image in buildable_images.items():
+        print(name + ':', end='')
+        if 'depends_on' in image:
+            for dep in image['depends_on']:
+                print(' ' + dep, end='')
+        print()
+        print('\tdocker build \\')
+        print('    --tag ' + image['tag'] + ' \\')
+        print('    --cache-from gcr.io/fuzzbench/' + image['tag'] + ' \\')
```
inferno-chromium, line 31: Do we need the project name here instead of hardcoded fuzzbench, like image['project'] or something ?

experiment/build/generate_cloudbuild.py
```diff
@@ -0,0 +1,66 @@
...
+    cloud_build_spec['steps'] = []
+    cloud_build_spec['images'] = []
+
+    for name, image in buildable_images.items():
+        step = {}
+        step['id'] = name
+        step['name'] = 'gcr.io/cloud-builders/docker'
+        step['args'] = []
+        step['args'] += ['--tag', image['tag']]
+        step['args'] += ['--cache-from', 'gcr.io/fuzzbench/' + image['tag']]
```
inferno-chromium, line 35: project name shouldn't be hardcoded, can we get the right one somehow.

experiment/build/docker_images.py
```diff
@@ -0,0 +1,56 @@
...
+            obj[key] = [subst(item, fuzzer, benchmark) for item in obj[key]]
+        else:
+            obj[key] = subst(obj[key], fuzzer, benchmark)
+    return name, obj
+
+
+def get_image_type_templates():
+    """Loads the image types config that contains "templates" describing how to
+    build them and their dependencies."""
+    with open('docker/image_types.yaml') as config_file:
```
inferno-chromium, line 41: use yaml_utils to read.

experiment/build/docker_images.py
```diff
@@ -0,0 +1,56 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Provides the set of buildable images and their dependencies."""
+
+import yaml
+
+
+def subst(template, fuzzer, benchmark):
```
inferno-chromium, line 19: prefix these helpers for this file with _
_subst
_instantiate_image_obj

docker/image_types.yaml
```diff
@@ -0,0 +1,67 @@
...
+'{fuzzer}-{benchmark}-runner':
+  tag: 'runners/{fuzzer}/{benchmark}'
+  context: '.'
+  dockerfile: 'docker/benchmark-runner/Dockerfile'
+  build_arg:
+    - 'fuzzer={fuzzer}'
+    - 'benchmark={benchmark}'
+  depends_on:
+    - '{fuzzer}-{benchmark}-builder'
+    - '{fuzzer}-{benchmark}-intermediate-runner'
```
inferno-chromium, line 67: in future, we should think of simplying and seeing if this intermediate runner image can be removed.

experiment/build/generate_cloudbuild.py
```diff
@@ -0,0 +1,66 @@
+# Copyright 2020 Google LLC
```
inferno-chromium, line 1: Probably want a simple test for this. can come in next cl.

experiment/build/generate_makefile.py
```diff
@@ -0,0 +1,50 @@
+# Copyright 2020 Google LLC
```
inferno-chromium, line 1: Probably want a simple test for this. can come in next cl.

docker/image_types.yaml
```diff
@@ -0,0 +1,67 @@
...
+  tag: 'base-builder'
+  context: 'docker/base-builder'
+  depends_on:
+    - 'base-image'
+
+'base-runner':
+  tag: 'base-runner'
+  context: 'docker/base-runner'
+  depends_on:
+    - 'base-image'
```
zchcai, line 18: As suggested in #548 by @inferno-chromium , shall we change this to depend on `base-builder` here?

experiment/build/docker_images.py
```diff
@@ -0,0 +1,56 @@
...
+        else:
+            obj[key] = subst(obj[key], fuzzer, benchmark)
+    return name, obj
+
+
+def get_image_type_templates():
+    """Loads the image types config that contains "templates" describing how to
+    build them and their dependencies."""
+    with open('docker/image_types.yaml') as config_file:
+        templates = yaml.load(config_file, yaml.SafeLoader)
```
zchcai, line 42: Shall we use common/yaml_utils.read() here?

experiment/build/docker_images.py
lszekeres: done

experiment/build/docker_images.py
lszekeres: done

experiment/build/docker_images.py
lszekeres: done

experiment/build/generate_cloudbuild.py
lszekeres: left a TODO

experiment/build/generate_cloudbuild.py
lszekeres: done

experiment/build/generate_makefile.py
lszekeres: added TODO

experiment/build/generate_makefile.py
lszekeres: done. made registry a flag.

experiment/build/test_docker_images.py
lszekeres: done. added another test for exact set.

docker/image_types.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+'coverage-builder':
+  tag: 'builders/coverage'
+  context: 'fuzzers/coverage'
+  dockerfile: 'fuzzers/coverage/builder.Dockerfile'
+  depends_on:
+    - 'base-builder'
+
+'coverage-{benchmark}-builder':
```
jonathanmetzman, line 27: Does this need to be seperate from fuzzer-benchmark-builder?

experiment/build/generate_cloudbuild.py
```diff
@@ -0,0 +1,77 @@
...
+    cloud_build_spec = {}
+    cloud_build_spec['steps'] = []
+    cloud_build_spec['images'] = []
+
+    for name, image in buildable_images.items():
+        step = {}
+        step['id'] = name
+        step['name'] = 'gcr.io/cloud-builders/docker'
+        step['args'] = []
+        step['args'] += ['--tag', image['tag']]
```
jonathanmetzman, line 36: I think we need two tags here as the YAML files have.
This is so that (for example) base-image can be pushed to gcr.io/fuzzbench-test/base-image and base-builder can inherit from gcr.io/fuzzbench/base-image instead of dealing with the mess of using arguments to decide what to inherit from.

jonathanmetzman, line 36: Should this handle the tagging feature i just added (https://github.com/google/fuzzbench/pull/557) or can that come later?

docker/image_types.yaml
lszekeres: It doesn't need to be, but it makes things cleaner and simpler:
- otherwise, when we call `get_images_to_build(fuzzers, benchmark)`, then `fuzzers` would need to include `coverage` -- while it's not a fuzzer
- if we considered "coverage" as a fuzzer, `get_images_to_build` would also generate a coverage-benchmark-runner "image prototype" due to the '{fuzzer}-{benchmark}-runner' type -- and we don't have / need such image

experiment/build/generate_cloudbuild.py
jonathanmetzman: :-( I suspect this trick won't work with buildkit since I assume buildkit will be "smart" by pulling from gcr.io/fuzzbench/base-image when it should really pull from gcr.io/fuzzbench-test/base-image if the repo we are using is gcr.io/fuzzbench-test.

docker/image_types.yaml
```diff
@@ -0,0 +1,67 @@
...
+  depends_on:
+    - '{fuzzer}-builder'
+
+'{fuzzer}-{benchmark}-intermediate-runner':
+  tag: 'runners/{fuzzer}/{benchmark}-intermediate'
+  context: 'fuzzers/{fuzzer}'
+  dockerfile: 'fuzzers/{fuzzer}/runner.Dockerfile'
+  depends_on:
+    - 'base-runner'
+
```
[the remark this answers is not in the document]
lszekeres: @Tanq16 has signed up for adding that and finalize the consolidation so we can replace docker/gcb/*, docker/generate_makefile.py, and other old scripts.

experiment/build/generate_makefile.py
```diff
@@ -0,0 +1,61 @@
...
+
+import argparse
+
+from experiment.build import docker_images
+
+
+# TODO: Add unit test for this.
+def print_makefile(buildable_images, docker_registry):
+    """Prints the generated makefile to stdout."""
+    print('export DOCKER_BUILDKIT := 1')
```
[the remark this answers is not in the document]
lszekeres: I'd stick to printing, because it's simpler to do this:
```
docker/generated.mk: experiment/build/generate_makefile.py
	python3 $< > $@
```

experiment/build/generate_cloudbuild.py
lszekeres: let's add it later @Tanq16

docker/image_types.yaml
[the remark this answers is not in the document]
Tanq16: The upcoming draft PR built on this will have the OSS-Fuzz images as well.

experiment/build/generate_makefile.py
```diff
@@ -0,0 +1,61 @@
...
+
+    for name, image in buildable_images.items():
+        print(name + ':', end='')
```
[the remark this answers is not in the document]
lszekeres: a simple format string wouldn't be enough due to the conditionals and lists. not sure if a jinja template would be more readable. will leave it up to @Tanq16: feel free to change this as you find it best.

experiment/build/docker_images.py
```diff
@@ -0,0 +1,54 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Provides the set of buildable images and their dependencies."""
+
+from common import yaml_utils
+
+
+def _subst(template, fuzzer, benchmark):
```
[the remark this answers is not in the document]
lszekeres: done

experiment/build/generate_cloudbuild.py
Tanq16: +1

experiment/build/generate_makefile.py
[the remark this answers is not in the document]
Tanq16: +1

experiment/build/generate_cloudbuild.py
lszekeres: Afaik, we can tell buildkit to `--cache-from` any docker registry/image.
(I'm not sure though why'd we need a separate gcr.io/fuzzbench and gcr.io/fuzzbench-test registy). 
Anyhow, if we need that i'd have a separate `--push-docker-registry` and `--cache-docker-registry` flag.

experiment/build/generate_makefile.py
```diff
@@ -0,0 +1,61 @@
...
+                print('    --build-arg ' + arg + ' \\')
+        if 'dockerfile' in image:
+            print('    --file ' + image['dockerfile'] + ' \\')
+        print('    ' + image['context'])
+        print()
+
+
+def main():
+    """Generates Makefile with docker image build rules."""
+    parser = argparse.ArgumentParser(description='GCB spec generator.')
```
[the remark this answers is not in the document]
lszekeres: good catch! done

experiment/build/generate_makefile.py
```diff
@@ -0,0 +1,61 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Generates Makefile containing docker image targets."""
+
+import argparse
+
+from experiment.build import docker_images
+
+
+# TODO: Add unit test for this.
```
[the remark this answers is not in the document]
lszekeres: done

experiment/build/generate_cloudbuild.py
```diff
@@ -0,0 +1,77 @@
...
+def main():
+    """Generates Cloud Build specification."""
+    parser = argparse.ArgumentParser(description='GCB spec generator.')
+    parser.add_argument('-r',
+                        '--docker-registry',
+                        default='gcr.io/fuzzbench/',
+                        help='Docker registry to use.')
+    args = parser.parse_args()
+
+    # TODO: Create fuzzer/benchmark list dynamically.
```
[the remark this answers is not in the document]
lszekeres: done

PR #548, opened 2020-07-16
Job dependency test

zchcai:
This PR adds one end-to-end test for new arch.

For new arch end-to-end test writing, we are not monitoring an experiment when it is ongoing, but check the outcome after it finishes.

For example, the jobs status, the filestore changes, the database changes and so on. This PR checks the starting time and end time for correlated jobs.


zchcai:
Here is the last several lines of output for `make run-end-to-end-test`.
```
Attaching to e2e-test_run-experiment_1, e2e-test_worker_2, e2e-test_worker_3, e2e-test_worker_1
run-experiment_1  | Initializing the job queue.
run-experiment_1  |   fuzzbench.fake_jobs.build_image('base-image',) : queued
run-experiment_1  |   fuzzbench.fake_jobs.build_image('base-runner',) : deferred
run-experiment_1  |   fuzzbench.fake_jobs.build_image('base-builder',) : deferred
worker_2          | 23:09:07 Worker rq:worker:56de4e40e05649e4a73ee6af2cc07017: started, version 1.4.3
worker_2          | 23:09:07 *** Listening on default...
worker_2          | 23:09:07 Cleaning registries for queue: default
worker_2          | 23:09:07 default: fuzzbench.fake_jobs.build_image('base-image') (base-image)
worker_1          | 23:09:07 Worker rq:worker:666b6bbf0d07434dbb081e455fcfff14: started, version 1.4.3
worker_1          | 23:09:07 *** Listening on default...
worker_1          | 23:09:07 Worker rq:worker:666b6bbf0d07434dbb081e455fcfff14: done, quitting
worker_3          | 23:09:07 Worker rq:worker:a8d990cf381041db9313a8b28ab4a612: started, version 1.4.3
worker_3          | 23:09:07 *** Listening on default...
worker_3          | 23:09:07 Worker rq:worker:a8d990cf381041db9313a8b28ab4a612: done, quitting
worker_2          | Docker version 19.03.12, build 48a66213fe
e2e-test_worker_3 exited with code 0
e2e-test_worker_1 exited with code 0
run-experiment_1  |   fuzzbench.fake_jobs.build_image('base-image',) : started
run-experiment_1  |   fuzzbench.fake_jobs.build_image('base-runner',) : deferred
run-experiment_1  |   fuzzbench.fake_jobs.build_image('base-builder',) : deferred
worker_2          | 23:09:10 default: Job OK (base-image)
worker_2          | 23:09:10 Result is kept for 500 seconds
worker_2          | 23:09:10 default: fuzzbench.fake_jobs.build_image('base-runner') (base-runner)
worker_2          | Docker version 19.03.12, build 48a66213fe
run-experiment_1  |   fuzzbench.fake_jobs.build_image('base-image',) : finished
run-experiment_1  |   fuzzbench.fake_jobs.build_image('base-runner',) : started
run-experiment_1  |   fuzzbench.fake_jobs.build_image('base-builder',) : queued
worker_2          | 23:09:13 default: Job OK (base-runner)
worker_2          | 23:09:13 Result is kept for 500 seconds
worker_2          | 23:09:13 default: fuzzbench.fake_jobs.build_image('base-builder') (base-builder)
worker_2          | Docker version 19.03.12, build 48a66213fe
run-experiment_1  |   fuzzbench.fake_jobs.build_image('base-image',) : finished
run-experiment_1  |   fuzzbench.fake_jobs.build_image('base-runner',) : finished
run-experiment_1  |   fuzzbench.fake_jobs.build_image('base-builder',) : started
worker_2          | 23:09:16 default: Job OK (base-builder)
worker_2          | 23:09:16 Result is kept for 500 seconds
worker_2          | 23:09:16 Worker rq:worker:56de4e40e05649e4a73ee6af2cc07017: done, quitting
e2e-test_worker_2 exited with code 0
run-experiment_1  |   fuzzbench.fake_jobs.build_image('base-image',) : finished
run-experiment_1  |   fuzzbench.fake_jobs.build_image('base-runner',) : finished
run-experiment_1  |   fuzzbench.fake_jobs.build_image('base-builder',) : finished
run-experiment_1  | All done!
e2e-test_run-experiment_1 exited with code 0
docker-compose run run-tests; STATUS=$?; \
docker-compose down; exit $STATUS
Starting e2e-test_queue-server_1 ... done
============================ test session starts =============================
platform linux -- Python 3.7.8, pytest-5.3.5, py-1.9.0, pluggy-0.13.1 -- /usr/local/bin/python3
cachedir: .pytest_cache
rootdir: /fuzzbench
plugins: pyfakefs-3.7.1
collected 5 items

fuzzbench/test_e2e_run.py::TestEndToEndRunResults::test_jobs_dependency PASSED [ 20%]
fuzzbench/test_e2e_run.py::TestEndToEndRunResults::test_all_jobs_finished_sucessfully PASSED [ 40%]
fuzzbench/test_e2e_run.py::TestEndToEndRunResults::test_measurement_jobs_were_started_before_trial_jobs_finished PASSED [ 60%]
fuzzbench/test_e2e_run.py::TestEndToEndRunResults::test_db_contains_experiment_results PASSED [ 80%]
fuzzbench/test_e2e_run.py::TestEndToEndRunResults::test_experiment_report_is_generated PASSED [100%]

============================= 5 passed in 0.08s ==============================
Stopping e2e-test_queue-server_1 ... done
Removing e2e-test_run-tests_run_2f2358075682 ... done
Removing e2e-test_worker_3                   ... done
Removing e2e-test_worker_1                   ... done
Removing e2e-test_worker_2                   ... done
Removing e2e-test_run-experiment_1           ... done
Removing e2e-test_queue-server_1             ... done
Removing network e2e-test_default

File: fuzzbench/test_e2e_run.py [hunk 1]
```diff
@@ -26,6 +28,13 @@
 class TestEndToEndRunResults:
     """Checks the result of a test experiment run."""
 
+    def test_jobs_dependency(self):
+        """Tests that jobs dependency preserves during working."""
+        jobs = Job.fetch_many(['base-image', 'base-runner', 'base-builder'],
+                              connection=redis.Redis(host='queue-server'))
+        assert jobs[0].ended_at <= jobs[1].started_at
```
lszekeres, line 35: can we refer to jobs with their name (eg jobs['base-image'].ended_at <= jobs['base_runer'].started_at .. to make this more readable?

zchcai: Good suggestion!

fuzzbench/test_e2e_run.py
```diff
@@ -26,6 +28,15 @@
...
     """Checks the result of a test experiment run."""
 
+    def test_jobs_dependency(self):
+        """Tests that jobs dependency preserves during working."""
+        jobs = {
+            name: Job.fetch(name, connection=redis.Redis(host='queue-server'))
+            for name in ['base-image', 'base-runner', 'base-builder']
+        }
+        assert jobs['base-image'].ended_at <= jobs['base-runner'].started_at
+        assert jobs['base-image'].ended_at <= jobs['base-builder'].started_at
```
inferno-chromium, line 38: base-runner started_at <= base-builder's started at as well.

fuzzbench/run_experiment.py
```diff
@@ -26,8 +26,18 @@ def run_experiment():
...
     queue = rq.Queue()
     jobs = []
-    for i in range(6):
-        jobs.append(queue.enqueue(fake_jobs.build_image, 'something-%d' % i))
+    jobs.append(
+        queue.enqueue(fake_jobs.build_image, 'base-image', job_id='base-image'))
+    jobs.append(
+        queue.enqueue(fake_jobs.build_image,
+                      'base-runner',
+                      job_id='base-runner',
```
inferno-chromium, line 34: i think base-runner should depend on base-builder first. you cant run without build finished.

fuzzbench/run_experiment.py
```diff
@@ -26,8 +26,18 @@ def run_experiment():
     print('Initializing the job queue.')
     queue = rq.Queue()
     jobs = []
-    for i in range(6):
-        jobs.append(queue.enqueue(fake_jobs.build_image, 'something-%d' % i))
+    jobs.append(
+        queue.enqueue(fake_jobs.build_image, 'base-image', job_id='base-image'))
```
inferno-chromium, line 30: Should we rename fake_jobs module, why not just jobs module.

fuzzbench/test_e2e_run.py
zchcai: According to current build dependency logic, it is not required.
Wdyt? @lszekeres @jonathanmetzman 
We can add this dependency logic into both #543 and here for consistency.

fuzzbench/run_experiment.py
zchcai: Done.

fuzzbench/fake_jobs.py
```diff
@@ -15,13 +15,19 @@
...
     print('Building', name)
-    subprocess.run(['docker', '--version'], check=True)
-    time.sleep(3)
+    if name.startswith('base'):
+        if name in ['base-image', 'base-builder', 'base-runner']:
+            build_jobs.build_base_images(name)
+    else:
+        subprocess.run(['docker', '--version'], check=True)
+        time.sleep(3)
     return True
```
lszekeres, line 31: Let's just have a jobs.py instead of fake_jobs and build_jobs and use the build_image function currently in build_jobs

fuzzbench/build_jobs.py
```diff
@@ -0,0 +1,32 @@
...
+from rq.job import Job
+
+
+BASE_TAG = 'gcr.io/fuzzbench'
+
+
+def build_base_images(name):
+    """Builds a Docker image."""
+    image_tag = os.path.join(BASE_TAG, name)
+    return subprocess.run(['docker', 'build',
```
lszekeres, line 29: How long does it take to build the 3 images you're building with an empty cache?

I think we should pull before build to warm up the cache. (We'll only need the explicit pull until we switched to buildkit. once the images in the registry were built with buildkit and the cache info is inlined in the images, we won't need to explicitly pull, but --cache-from will pull automatically relevant parts.)

fuzzbench/run_experiment.py
lszekeres: +1

fuzzbench/build_jobs.py
zchcai: > How long does it take to build the 3 images you're building with an empty cache?

For `make run-end-to-end-test` command, building run-experiment takes 3 mins, base-image 5 mins, base-builder 1 and base-runner 1 min.

> pull before build to warm up the cache

Does this mean, we first do all pull related tasks and then do build jobs with `--cache-from`?

zchcai:
If we add a pull before the build, as docker pull gcr.io/fuzzbench/base-[image,builder,runner], the time is similar to not adding this line.

fuzzbench/build_jobs.py
zchcai: See #558.

zchcai:
> If we add a pull before the build, as docker pull gcr.io/fuzzbench/base-[image,builder,runner], the time is similar to not adding this line.

See #558.

fuzzbench/jobs.py
```diff
@@ -11,25 +11,41 @@
...
+    """Builds a Docker image."""
     print('Building', name)
-    subprocess.run(['docker', '--version'], check=True)
-    time.sleep(3)
+    if name.startswith('base'):
+        if name in ['base-image', 'base-builder', 'base-runner']:
+            build_base_images(name)
+    else:
+        subprocess.run(['docker', '--version'], check=True)
+        time.sleep(3)
```
lszekeres, line 40: Can we please merge these two functions? It's fine if for now it only works for base images, we can start add support for fuzzer-benchmark runners once https://github.com/google/fuzzbench/pull/543 lands

fuzzbench/jobs.py
```diff
@@ -11,25 +11,42 @@
...
+BASE_TAG = 'gcr.io/fuzzbench'
 
-def build_image(name):
-    """Build a Docker image."""
+
+def build_base_images(name: str):
+    """Builds a Docker image."""
+    image_tag = os.path.join(BASE_TAG, name)
+    subprocess.run(['docker', 'pull', image_tag], check=True)
+    return subprocess.run(
```
lszekeres, line 26: this returns a `CompletedProcess` object? is that what we want?

do we actually need to / want to return any value? it might be fine not to return anything (for now) as due to the `check=True` if the function returns that already means "success". wdyt?

fuzzbench/test_e2e_run.py
```diff
@@ -26,6 +28,15 @@
...
+        """Tests that jobs dependency preserves during working."""
+        jobs = {
+            name: Job.fetch(name, connection=redis.Redis(host='queue-server'))
+            for name in ['base-image', 'base-builder', 'base-runner']
+        }
+        assert jobs['base-image'].ended_at <= jobs['base-builder'].started_at
+        assert jobs['base-builder'].ended_at <= jobs['base-runner'].started_at
+
     def test_all_jobs_finished_sucessfully(self):
         """Fake test to be implemented later."""
```
lszekeres, line 41: Consider implementing this check as well

fuzzbench/jobs.py
zchcai: > this returns a `CompletedProcess` object? is that what we want?

Yes this returns a `CompletedProcess` object. I'm not sure whether this is what we want, but this way we keep information.

> do we actually need to / want to return any value?

That is a good question. I think we should at least return this is success or failed for a run.

> it might be fine not to return anything (for now) as due to the `check=True` if the function returns that already means "success". wdyt?

check=True may need to remove in the future and that is not in default for every job.
For now, I agree, because for build jobs we should expect success.

Let's just return True here. And if we need to add other information in the returned value, we change it.

fuzzbench/test_e2e_run.py
zchcai: Good!

fuzzbench/test_e2e_run.py
```diff
@@ -26,9 +28,21 @@
 class TestEndToEndRunResults:
     """Checks the result of a test experiment run."""
 
-    def test_all_jobs_finished_sucessfully(self):
-        """Fake test to be implemented later."""
-        assert True
+    def test_jobs_dependency(self):
+        """Tests that jobs dependency preserves during working."""
+        jobs = {
+            name: Job.fetch(name, connection=redis.Redis(host='queue-server'))
```
lszekeres, line 34: nit: consider using a fixture for the redis_connection, ie something like:
```python
@pytest.fixture(scope="module")
def redis_connection():
  return redis.Redis(host='queue-server')
```
See https://docs.pytest.org/en/stable/fixture.html#fixture

zchcai: Sounds good! Let's put it at a separate PR to implement then.

fuzzbench/jobs.py
```diff
@@ -11,25 +11,30 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Fake jobs."""
+"""Jobs."""
```
jonathanmetzman, line 14: docstring

fuzzbench/jobs.py
```diff
@@ -11,25 +11,30 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
```
jonathanmetzman, line 12: I don't think it's a good idea to have all jobs/tasks in one module. They will do totally different things and this file will become a mess.

fuzzbench/test_e2e_run.py
jonathanmetzman: I don't see why it shouldn't be added in this PR.

fuzzbench/jobs.py
zchcai: How about "Defines all required jobs for one experiment."?

fuzzbench/jobs.py
```diff
@@ -11,25 +11,30 @@
...
+BASE_TAG = 'gcr.io/fuzzbench'
 
-def build_image(name):
-    """Build a Docker image."""
-    print('Building', name)
-    subprocess.run(['docker', '--version'], check=True)
-    time.sleep(3)
+
+def build_image(name: str):
+    """Builds a Docker image."""
```
jonathanmetzman, line 23: Document what it returns.

fuzzbench/jobs.py
```diff
@@ -11,25 +11,30 @@
...
+    image_tag = os.path.join(BASE_TAG, name)
```
jonathanmetzman, line 24: This function is not going to be sufficient for building every image. Is that fine?

fuzzbench/jobs.py
zchcai: > lszekeres yesterday Member
Let's just have a jobs.py instead of fake_jobs and build_jobs and use the build_image function currently in build_jobs

I agree with you. But as @lszekeres suggested above, we can separate the logic later and then keep it simple now.

fuzzbench/jobs.py
zchcai: >This function is not going to be sufficient for building every image.

No, it is not sufficient.

>Is that fine?

Currently, it should be fine. We may add more logic when we need.

fuzzbench/test_e2e_run.py
zchcai: Sounds good too.

fuzzbench/jobs.py
zchcai: Done.

fuzzbench/jobs.py
lszekeres: (for more context: as mentioned before once https://github.com/google/fuzzbench/pull/543 lands we can use that here without re-implementing the build logic)

fuzzbench/jobs.py
lszekeres: once we implement all these functions (build/run/measure), it will definitely make sense to separate them

fuzzbench/conftest.py
```diff
@@ -11,25 +11,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Fake jobs."""
+"""A pytest conftest.py file that defines fixtures."""
+import redis
```
jonathanmetzman, line 16: no need for blank line.
And pytest should come before redis. Both are 3rd party modules.

fuzzbench/conftest.py
```diff
@@ -11,25 +11,13 @@
...
-    return True
-
-
-def measure_corpus_snapshot():
-    """Measure a corpus snapshot."""
-    return True
+@pytest.fixture
+def redis_connection():
+    """Returns the default redis server connection."""
+    return redis.Redis(host='queue-server')
```
jonathanmetzman, line 23: This works? It doesn't need to be an IP? 
I guess the device you are running this on knows an ip that queue-server maps to?

fuzzbench/test_e2e_run.py
jonathanmetzman: Yeah I don't think this makes sense to test for. They are built in parallel.

fuzzbench/conftest.py
zchcai: >This works? 

Yes
>It doesn't need to be an IP?

I think here it matches the pattern in fuzzbench/run_experiment.py

https://github.com/google/fuzzbench/blob/5db871f65c312d4b2cedf16bd7ac2c6311987b26/fuzzbench/run_experiment.py#L42-L46

>I guess the device you are running this on knows an ip that queue-server maps to?

I think it uses the docker container here:
https://github.com/google/fuzzbench/blob/5db871f65c312d4b2cedf16bd7ac2c6311987b26/compose/fuzzbench.yaml#L26-L27

fuzzbench/test_e2e_run.py
zchcai: So, shall we change it back? Say, base-runner depends on base-image, not base-builder?

fuzzbench/conftest.py
zchcai: Done.

fuzzbench/test_e2e_run.py
```diff
@@ -26,9 +27,21 @@
...
-        """Fake test to be implemented later."""
-        assert True
+    def test_jobs_dependency(self, redis_connection):
+        """Tests that jobs dependency preserves during working."""
+        jobs = {
+            name: Job.fetch(name, connection=redis_connection)
+            for name in ['base-image', 'base-builder', 'base-runner']
+        }
+        assert jobs['base-image'].ended_at <= jobs['base-builder'].started_at
+        assert jobs['base-builder'].ended_at <= jobs['base-runner'].started_at
```
zchcai, line 37: >Yeah I don't think this makes sense to test for. They are built in parallel.

So, shall we change it back? Say, base-runner depends on base-image, not base-builder?

fuzzbench/conftest.py
lszekeres: do we need a separate conftest.py file for this?

can we put this fixture into test_e2e_run.py?

can we make it's scope modul / class level (ie. run it only once before calling the test_... methods)?

fuzzbench/test_e2e_run.py
lszekeres: yes, 
base-builder depends on base-image, and 
base-runner depends on base-image

fuzzbench/conftest.py
zchcai: > do we need a separate conftest.py file for this?

I follow the pattern of current code. We will have specific conftest.py in each folder.

> can we put this fixture into test_e2e_run.py?

I guess put it aside will be better and clearer? wdyt? @jonathanmetzman 

> can we make it's scope modul / class level (ie. run it only once before calling the test_... methods)?

Shall we do that @jonathanmetzman ?

fuzzbench/test_e2e_run.py
```diff
@@ -18,17 +18,38 @@
 import os
 
 import pytest
+import redis
+from rq.job import Job
+
+
+@pytest.fixture(scope='module')
```
zchcai, line 25: Not sure which scope level should be used here.
function or class or module or nothing needed?

fuzzbench/test_e2e_run.py
zchcai: OK. Change it back now. Done.

fuzzbench/test_e2e_run.py
lszekeres: i'd use `class` and use it with `TestEndToEndRunResults`

fuzzbench/conftest.py
jonathanmetzman: I don't really think this is an intended pattern. If a fixture will be used in multiple files then put it in conftest.py otherwise can go in the file it is used.

fuzzbench/jobs.py
jonathanmetzman: Fine for now

fuzzbench/test_e2e_run.py
jonathanmetzman: Yes base-runner does not depend on base-builder.

fuzzbench/conftest.py
zchcai: Yes. So we already put it into the e2e-test since this fixture is only used here now. Done.

fuzzbench/test_e2e_run.py
zchcai: Sounds good! Done.

PR #549, opened 2020-07-16
Use libFuzzer that supports interceptors without ASan

dokyungs:
libFuzzer now supports intercepting memcmp-like functions on its own, without ASan. This patch updates the commit number of the llvm project that is used to setup libFuzzer.


dokyungs:
@morehouse @hctim FYI

dokyungs:
@jonathanmetzman Would you be able to start a new benchmark run that includes honggfuzz and libFuzzer? I see that all the existing libFuzzer (oss-fuzz and standard) checks went fine with this libFuzzer update. So I assume we can run the benchmarks to see how latest libFuzzer performs. Let me know if you have any concern!

PR #550, opened 2020-07-17
Add default end to end test config file for testing; prepare for config validation refactoring

zchcai:
## What is this PR?
This PR helps for later development, when we need to have fixed end-to-end testing environment. In short, we need to run a unique experiment with a specified/default configuration for our end-to-end test.

This PR helps with this and starts the refactoring of reading config files, `common/config_utils.py`. We will move the validation logic from `experiment/run_experiment.py` to `common/config_utils.py` in later PR.

## How to test?
We modify several files to work together and achieve that:
1. `Makefile` and `compose/e2e-test.yaml`: The end-to-end test will start with `test-run-expeirment`, which is no different with `run-experiment` excepting additional environment variable: `E2E_INTEGRATION_TEST = 1`.
2. `docker/fuzzbench/Dockerfile` and `config.yaml`: copy the user-specified configuration as default behavior (which is not for end-to-end testing).
3. `fuzzbench/run_experiment.py`: it will first use the `config.yaml`. However, if `E2E_INTEGRATION_TEST` is set, then we use the default config for end-to-end test. Also, if `config['end_to_end_test']=true`, then we enqueue a fake job into the queue with `id='e2e_test'`.
4. `fuzzbench/test_e2e_run.py`: we check that a job with `id='e2e_test'` should exist.


fuzzbench/end-to-end-test-config.yaml
```diff
@@ -0,0 +1,13 @@
+benchmarks:
+    - freetype2-2017
+    - bloaty_fuzz_target
+docker_registry: gcr.io/fuzzbench
+end_to_end_test: true
```
lszekeres, line 5: can you remove the "end_to_end_test" option from the config? the end to end test should work as a real experiment run, but with an experiment config that only runs very few and short things, e.g., builds just a few images. but it should use the same config options such as fuzzers and benchmarks, so something like "end_to_end_test" should not be part of the experiment config file

fuzzbench/fake_jobs.py
```diff
@@ -33,3 +33,8 @@ def run_trial():
 def measure_corpus_snapshot():
     """Measure a corpus snapshot."""
     return True
+
+
+def fake_job():
+    """A signal fake job function."""
+    return True
```
lszekeres, line 40: why do we need another fake job function? why don't we build images?

fuzzbench/run_experiment.py
```diff
@@ -13,21 +13,25 @@
...
-def run_experiment():
+def run_experiment(config):
     """Main experiment logic."""
     print('Initializing the job queue.')
     queue = rq.Queue()
     jobs = []
     for i in range(6):
         jobs.append(queue.enqueue(fake_jobs.build_image, 'something-%d' % i))
+    if config.get('end_to_end_test', False):
+        jobs.append(queue.enqueue(fake_jobs.fake_job, job_id='e2e-test'))
```
lszekeres, line 34: why are we doing this? can we build an image instead? this code shouldn't behave differently when we're running an e2e test. the e2e test is for validating the normal behavior. only the configuration should be different (see other comment)

fuzzbench/test_e2e_run.py
```diff
@@ -41,3 +45,12 @@ def test_db_contains_experiment_results(self):
...
         assert True
+
+    def test_default_config_is_used(self):
+        """Tests whether the default config is used for end to end test."""
+        try:
+            fake_job = Job.fetch('e2e-test',
+                                 connection=redis.Redis(host='queue-server'))
+        except rq.exceptions.NoSuchJobError:
+            assert False
+        assert fake_job is not None
```
lszekeres, line 56: Not sure what's the purpose of this check. The purpose of the e2e test is to do an as realistic as possible experiment run then check the result with these test_ functions. E.g. build a few images, then check here if the images were built successfully. 

If you'd like to check that current fake_jobs.bulid_image() jobs were successful, you can do that, eg under `test_all_jobs_finished_sucessfully()`.

compose/e2e-test.yaml
```diff
@@ -9,3 +9,26 @@ services:
...
+    environment:
+      RQ_REDIS_URL: redis://queue-server
+    command: rq worker --burst
+    volumes:
+      # Allow access to the host's docker daemon.
+      - /var/run/docker.sock:/var/run/docker.sock
+    links:
+      - queue-server
+    depends_on:
+      - test-run-experiment
```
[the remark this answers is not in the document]
zchcai: I see! Let me try this. Thanks.

common/config_utils.py
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Provides helper functions to obtain configurations."""
+
+
+def validate_and_expand(config):
+    """Validates |config| and returns the enlarged one."""
```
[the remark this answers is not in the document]
zchcai: Right. Maybe this will not happen in the new arch. In the old arch, some arguments will be assigned with a default one and added to the dictionary. Like this:

https://github.com/google/fuzzbench/blob/140e69107d725611290a05d6369f159337d61bdf/experiment/run_experiment.py#L79

https://github.com/google/fuzzbench/blob/140e69107d725611290a05d6369f159337d61bdf/experiment/run_experiment.py#L135-L136

Also,

https://github.com/google/fuzzbench/blob/140e69107d725611290a05d6369f159337d61bdf/experiment/run_experiment.py#L214

fuzzbench/test_e2e_run.py
zchcai: > Not sure what's the purpose of this check.

**This check is to see whether this end-to-end test use the correct/desired config file for experiment running.** The point lies that, we may hardcode some image name or jobs name and properties ahead in the test functions. If we are not certain this end-to-end test using our desired config or not, we are not sure the correctness to write other test functions.

> The purpose of the e2e test is to do an as realistic as possible experiment run then check the result with these test_ functions.

Yes! That is why we want to use a specified config file for an experiment run, other than making up the environment in the test class itself. Maybe `test_default_config_is_used` is not related to the real experiment running, however, it is a required step and give us confidence to write other end-to-end test. wdyt?

>However this check (test_default_config_is_used) doesn't really test "real code", but the "test code" that you added under the "if end-to-end test" branch.

The implementation is ugly, I admit. But the point is to let the test function to test whether we are using the desired test yaml.

> We can check that the current fake_jobs.bulid_image() jobs were successful on the other hand. Consider adding that to test_all_jobs_finished_sucessfully().

Good suggestion to look through.
By the way, I doubt whether we really need `test_all_jobs_finished_successfully` test. Should all jobs finish at last? This is more like an optimized desired behavior, other than a logic requirement/specification. Consider the coverage building task status in the future. We may want to pause those trial runners if some coverage build jobs fail.

Instead, I guess the test in #548 makes more sense since we assert the dependency logic for jobs working.

fuzzbench/fake_jobs.py
zchcai: > why do we need another fake job function?

Since this function is not related to experiment logic, I think another different name will be helpful for readability.

> why don't we build images?

As said, build image fake job will be confusing here.

fuzzbench/run_experiment.py
zchcai: > why are we doing this? can we build an image instead? this code shouldn't behave differently when we're running an e2e test. the e2e test is for validating the normal behavior. only the configuration should be different (see other comment)

I think the main point lying in that, do we need to test the end-to-end testing is using the desired config file. This comment argument point is based on we actually don't need this test, which means we already know the end-to-end test will use the desired config file.

I think this test included in the end-to-end test is needed to check. This is more like, we need to first sure run on an experiment we desire and then we check experimental behavior.

Otherwise, how to know the modification in `compose/e2e-test.yaml` is correct? wdyt

fuzzbench/end-to-end-test-config.yaml
zchcai: >can we remove the "end_to_end_test" option from the config?

This is a way to know we are using the specified config.yaml for end-to-end testing.

Again, I think the main point lying in that, **do we need to test the end-to-end testing is using the desired config file.** This comment argument point is based on we actually don't need this test, which means we already know the end-to-end test will use the desired config file.

I think this test included in the end-to-end test is needed to check. This is more like, we need to first sure run on an experiment we desire and then we check experimental behavior.

Otherwise, how to know the modification in compose/e2e-test.yaml is correct? wdyt

fuzzbench/end-to-end-test-config.yaml
```diff
@@ -0,0 +1,13 @@
+benchmarks:
```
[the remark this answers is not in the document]
zchcai: Let's finalize others issues in this PR and then refactor this. Good refactoring suggestion!

config.yaml
```diff
@@ -0,0 +1,11 @@
...
+max_total_time: 3600
+benchmarks:
+    - freetype2-2017
+    - bloaty_fuzz_target
+fuzzers:
+    - afl
+    - libfuzzer
+docker_registry: gcr.io/fuzzbench
+experiment_filestore: <your-experiment-filestore>
+report_filestore: <your-report-filestore>
```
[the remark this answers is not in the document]
zchcai: > we can add these when we'll need them, no?

Yes, we can. If looking from this PR, there is no need to putting anything into this file.
Shall we remove these default and required config options in this default file?

fuzzbench/test_e2e_run.py
```diff
@@ -41,3 +45,12 @@ def test_db_contains_experiment_results(self):
     def test_experiment_report_is_generated(self):
         """Fake test to be implemented later."""
         assert True
+
+    def test_default_config_is_used(self):
+        """Tests whether the default config is used for end to end test."""
+        try:
+            fake_job = Job.fetch('e2e-test',
+                                 connection=redis.Redis(host='queue-server'))
+        except rq.exceptions.NoSuchJobError:
```
inferno-chromium, line 54: No need of try except, exception raised in test is perfectly fine and is a failure.

fuzzbench/test_e2e_run.py
inferno-chromium, line 56: Just being not None is not the right way to test things, why not have one job in queue and test that exact one.

fuzzbench/run_experiment.py
```diff
@@ -42,8 +45,13 @@ def run_experiment():
 def main():
     """Set up Redis connection and start the experiment."""
     redis_connection = redis.Redis(host="queue-server")
+    config_path = environment.get('EXPERIMENT_CONFIG', 'config.yaml')
```
inferno-chromium, line 48: Shouldn't this use experiment_utils to get config file. Why use env var ? for tests, we just mock var.

fuzzbench/run_experiment.py
inferno-chromium: yes tests thing cannot be in main code.

fuzzbench/test_e2e_run.py
zchcai: This is cool!

fuzzbench/test_e2e_run.py
zchcai: > Just being not None is not the right way to test things, why not have one job in queue and test that exact one.

Right. This looks unclear. My thoughts were, if it can fetch this job, then it means this experiment is running with the desired test-config.yaml.

fuzzbench/run_experiment.py
zchcai: > yes tests thing cannot be in main code.

For end-to-end test, since we cannot monitor the experiment but only check the outcome of the experiment, thus I need to have some specific outcome for this experiment. Otherwise, we cannot check whether it is using the desired config file if there is no special thing happen.

I think currently we should first try to agree or disagree on, whether we should test whether this e2e test is using the desired config file. Then, we can discuss how to implement this goal. 

wdyt?

fuzzbench/run_experiment.py
```diff
@@ -18,16 +19,22 @@
...
     jobs = []
-    for i in range(6):
-        jobs.append(queue.enqueue(fake_jobs.build_image, 'something-%d' % i))
+    jobs.append(queue.enqueue(fake_jobs.build_image, 'base-images'))
+    for benchmark in config.get('benchmarks'):
+        jobs.append(queue.enqueue(fake_jobs.build_image, benchmark))
+    for fuzzer in config.get('fuzzers'):
+        for benchmark in config.get('benchmarks'):
+            jobs.append(queue.enqueue(fake_jobs.build_image,
+                                      fuzzer + benchmark))
```
lszekeres, line 37: Are you planning to get https://github.com/google/fuzzbench/pull/548 in first, and then update this part?

fuzzbench/test_e2e/end-to-end-test-config.yaml
```diff
@@ -0,0 +1,6 @@
+benchmarks:
+    - freetype2-2017
+    - bloaty_fuzz_target
+fuzzers:
+    - afl
+    - libfuzzer
```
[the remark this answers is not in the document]
zchcai: For building `run-experiment`, it needs 3 mins.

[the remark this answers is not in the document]
zchcai: For the base-image building, it needs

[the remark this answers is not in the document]
zchcai: For all base-images, it needs <waiting>.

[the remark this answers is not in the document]
zchcai: For other build images, we have no logic for building them now. It will be tested in a later PR.

fuzzbench/run_experiment.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
zchcai: Depends on whether we need to run it as `./fuzzbench/run_experiment.py` directly.
This follows the pattern in `experiment/run_experiment.py`.
Looks fine to remove. wdyt?

fuzzbench/run_experiment.py
zchcai: Yes, probably.

fuzzbench/run_experiment.py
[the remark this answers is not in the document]
lszekeres: Yes, we can add it when we'll need it.

fuzzbench/run_experiment.py
lszekeres: SG! Let's get back to this after that's done (you can re-request review then). Thanks!

fuzzbench/run_experiment.py
zchcai: > Shouldn't this use experiment_utils to get config file.

Do you suggest that we put the config getting logic under `common/experiment_utils.py` rather than `common/config_utils.py`?

> Why use env var ? for tests, we just mock var.

For end-to-end test, we cannot monitor the experiment and can only check the outcome when it finishes. However, to let our end-to-end test run with a specified config, it should be set in the compose/ folder. Thus, we choose to use a global environment variable to do this.

config.yaml
```diff
@@ -0,0 +1,6 @@
+benchmarks:
+    - freetype2-2017
+    - bloaty_fuzz_target
```
zchcai, line 3: bloaty_fuzz_target is a oss-fuzz benchmark.
The next group of commits will resolve this issue.
If we change this into libpng-1.2.56. Then it should work now.

docker/fuzzbench/Dockerfile
```diff
@@ -31,7 +31,10 @@ COPY benchmarks benchmarks
 COPY common common
 COPY database database
 COPY docker docker
+COPY experiment/build experiment/build
+COPY experiment/*.py experiment/
```
zchcai, line 35: We are not going to copy all files in `experiment`.
Maybe in the future, these files should move out of `experiment`, where should be the old skeleton, independent with the new arch. And both archs base on same group of code.

fuzzbench/run_experiment.py
```diff
@@ -17,37 +17,72 @@
...
+            unqueued_build_images.append((name, obj))
+            continue
+
+        jobs_list.append(
+            queue.enqueue(jobs.build_image,
+                          tag=obj['tag'],
+                          context=obj['context'],
+                          dockerfile=obj.get('dockerfile', None),
+                          buildargs=obj.get('build_arg', None),
+                          job_timeout=600,
```
zchcai, line 58: Need to think about the proper timeout limit for each build task.

fuzzbench/jobs.py
```diff
@@ -19,14 +19,17 @@
...
+def build_image(tag: str, context: str, dockerfile=None, buildargs=None):
     """Builds a Docker image and returns whether it succeeds."""
-    image_tag = os.path.join(BASE_TAG, name)
+    image_tag = os.path.join(BASE_TAG, tag)
     subprocess.run(['docker', 'pull', image_tag], check=True)
-    subprocess.run(
-        ['docker', 'build', '--tag', image_tag,
-         os.path.join('docker', name)],
-        check=True)
+    command = ['docker', 'build', '--tag', image_tag, context]
```
zchcai, line 26: Shall we add `--cpus=1` as default?

fuzzbench/test_e2e/end-to-end-test-config.yaml
```diff
@@ -0,0 +1,6 @@
+benchmarks:
+    - freetype2-2017
+    - libpng-1.2.56
```
zchcai, line 3: These two benchmarks are both non-oss-fuzz benchmarks.

zchcai:
The CI test should be much longer than before and I will edit this comment to reflect the time usage when it finishes.

config.yaml
lszekeres: yes, let's change it

fuzzbench/jobs.py
lszekeres: yes

fuzzbench/jobs.py
```diff
@@ -19,14 +19,17 @@
 BASE_TAG = 'gcr.io/fuzzbench'
 
 
-def build_image(name: str):
+def build_image(tag: str, context: str, dockerfile=None, buildargs=None):
```
lszekeres, line 22: consider passing in the object that was returned by `get_images_to_build`

fuzzbench/run_experiment.py
```diff
@@ -17,37 +17,72 @@
...
-                      depends_on='base-image'))
-    jobs_list.append(
-        queue.enqueue(jobs.build_image,
-                      'base-runner',
-                      job_timeout=600,
-                      job_id='base-runner',
-                      depends_on='base-image'))
+    unqueued_build_images = []
+    for name, obj in images_to_build.items():
+        if name in ['base-image']:
```
lszekeres, line 39: why does base image need a separate branch?

zchcai: >why does base image need a separate branch?

This image is the only one not depending on other images. We can also write as
```
if obj.get('depends_on', None) is None:
```
Wdyt? Now it seems we change to this way.

fuzzbench/jobs.py
zchcai: OK.
I guess the benefit is we can make the function signature shorter. Also, we can move the logic from run_experiment to here.

Any other benefits?

fuzzbench/jobs.py
zchcai: Cool!

config.yaml
zchcai: Done.

fuzzbench/run_experiment.py
```diff
@@ -17,37 +17,72 @@
...
+                queue.enqueue(jobs.build_image,
+                              tag=obj['tag'],
+                              context=obj['context'],
+                              job_timeout=600,
+                              job_id=name))
+            continue
+
+        if len(obj['depends_on']) > 1:
+            unqueued_build_images.append((name, obj))
+            continue
```
zchcai, line 50: Since `rq` doesn't support multi-dependency and the discussion has been existing for 7 years, I choose a simple way to implement this.
Simply, log those multi-depend jobs down, and add them during an iteration where checking whether all its depended jobs finish.

Any other thoughts? Or better suggestions?

fuzzbench/run_experiment.py
lszekeres: Does it actually matter what order do we enqueue jobs? E.g., would that be an issue if we happen to first
- enqueue('base-builder', depends_on='base_image), and then
- enqueue('base_image)

?

fuzzbench/jobs.py
lszekeres: seems cleaner and simpler to me, is there any benefit of the current approach?

fuzzbench/jobs.py
```diff
@@ -19,14 +19,19 @@
...
     """Builds a Docker image and returns whether it succeeds."""
-    image_tag = os.path.join(BASE_TAG, name)
+    image_tag = os.path.join(BASE_TAG, tag)
     subprocess.run(['docker', 'pull', image_tag], check=True)
-    subprocess.run(
-        ['docker', 'build', '--tag', image_tag,
-         os.path.join('docker', name)],
-        check=True)
+    command = ['docker', 'build', '--tag', image_tag, context]
+    cpu_options = ['--cpu-period', '100000', '--cpu-quota', '100000']
```
zchcai, line 27: As there is no `--cpus` option for `docker build`, we use this way, which is equivalent to `--cpus=1`.
refer: https://docs.docker.com/config/containers/resource_constraints/#configure-the-default-cfs-scheduler

fuzzbench/run_experiment.py
zchcai: Good question! Let's me have a try for this.

fuzzbench/jobs.py
lszekeres: Looks good, thanks!

docker/image_types.yaml
```diff
@@ -5,12 +5,19 @@
   tag: 'base-image'
   context: 'docker/base-image'
```
zchcai, line 7: We hugely update this file according to existing `gcb/*` rules.

zchcai:
Currently, coverage-{benchmark}-builder will fail in local `make run-end-to-end-test`. I will debug this and then start another run.

fuzzbench/run_experiment.py
zchcai: We resolve this by removing multi-dependency.

fuzzbench/jobs.py
zchcai: Done.

zchcai:
Still these two failed now. Will take a look.
```
run-experiment_1  |     queued: 0
run-experiment_1  |     started:        0
run-experiment_1  |     deferred:       0
run-experiment_1  |     finished:       22
run-experiment_1  |     failed: 2
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (base-image)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (dispatcher-image)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (coverage-freetype2-2017-builder-intermediate)
run-experiment_1  |   fuzzbench.jobs.build_image : failed       (coverage-freetype2-2017-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (freetype2-2017-project-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (afl-freetype2-2017-builder-intermediate)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (afl-freetype2-2017-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (afl-freetype2-2017-runner-intermediate)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (afl-freetype2-2017-runner)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (coverage-libpng-1.2.56-builder-intermediate)
run-experiment_1  |   fuzzbench.jobs.build_image : failed       (coverage-libpng-1.2.56-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (libpng-1.2.56-project-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (afl-libpng-1.2.56-builder-intermediate)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (afl-libpng-1.2.56-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (afl-libpng-1.2.56-runner-intermediate)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (afl-libpng-1.2.56-runner)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (libfuzzer-freetype2-2017-builder-intermediate)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (libfuzzer-freetype2-2017-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (libfuzzer-freetype2-2017-runner-intermediate)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (libfuzzer-freetype2-2017-runner)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (libfuzzer-libpng-1.2.56-builder-intermediate)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (libfuzzer-libpng-1.2.56-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (libfuzzer-libpng-1.2.56-runner-intermediate)
run-experiment_1  |   fuzzbench.jobs.build_image : finished     (libfuzzer-libpng-1.2.56-runner)

**Comment by zchcai** [comment]
It says `clang-11: error: no such file or directory: '/usr/lib/libFuzzer.a'`.
```
worker_2          | ar: `u' modifier ignored since `D' is the default (see `U')
worker_2          | libtool: link: ranlib /src/freetype2/objs/.libs/libfreetype.a
worker_2          | libtool: link: ( cd "/src/freetype2/objs/.libs" && rm -f "libfreetype.la" && ln -s "../libfreetype.la" "libfreetype.la" )
worker_2          | + clang++ -stdlib=libc++ -pthread -Wl,--no-as-needed -Wl,-ldl -Wl,-lm -Wno-unused-command-line-argument -O3 -fprofile-instr-generate -fcoverage-mapping -std=c++11 -I include -I . src/tools/ftfuzzer/ftfuzzer.cc objs/.libs/libfreetype.a /usr/lib/libFuzzer.a -larchive -lz -o /out/fuzz-target
worker_2          | clang-11: error: no such file or directory: '/usr/lib/libFuzzer.a'
worker_2          | Traceback (most recent call last):
worker_2          |   File "<string>", line 1, in <module>
worker_2          |   File "/src/fuzzer.py", line 31, in build
worker_2          |     utils.build_benchmark()
worker_2          |   File "/src/fuzzers/utils.py", line 49, in build_benchmark
worker_2          |     subprocess.check_call(['/bin/bash', '-ex', build_script], env=env)
worker_2          |   File "/usr/local/lib/python3.7/subprocess.py", line 363, in check_call
worker_2          |     raise CalledProcessError(retcode, cmd)
worker_2          | subprocess.CalledProcessError: Command '['/bin/bash', '-ex', '/src/build.sh']' returned non-zero exit status 1.
worker_2          | The command '/bin/sh -c PYTHONPATH=$SRC python3 -u -c "from fuzzers import utils; utils.initialize_flags(); import fuzzer; fuzzer.build()"' returned a non-zero code: 1
worker_2          | 19:55:55 subprocess.CalledProcessError: Command '['docker', 'build', '--tag', 'gcr.io/fuzzbench/builders/coverage/freetype2-2017', '.', '--cpu-period', '100000', '--cpu-quota', '100000', '--file', 'docker/benchmark-builder/Dockerfile', '--build-arg', 'parent_image=gcr.io/fuzzbench/builders/coverage/freetype2-2017-intermediate', '--build-arg', 'fuzzer=coverage', '--build-arg', 'benchmark=freetype2-2017', '--build-arg', 'parent_image=gcr.io/fuzzbench/builders/benchmark/freetype2-2017']' returned non-zero exit status 1.

fuzzbench/test_e2e/test_e2e_run.py
```diff
@@ -31,17 +41,29 @@ def redis_connection():
...
+        all_images = docker_images.get_images_to_build(config['fuzzers'],
+                                                       config['benchmarks'])
+        jobs = {
+            name: rq.job.Job.fetch(name, connection=redis_connection)
+            for name in all_images
+        }
+        for name, image in all_images.items():
+            if 'depends_on' in image:
+                for dep in image['depends_on']:
+                    assert jobs[dep].ended_at <= jobs[name].started_at
```
zchcai, line 59: Need to figure out the differences among:
- enqueued_at
- started_at
- ended_at

zchcai:
It ended with `##[error]No space left on device`: (Any thoughts? @lszekeres  @jonathanmetzman Just use one pair of fuzzer and benchmark, instead of 4 pairs?)
```
run-experiment_1  | 	queued:	3
run-experiment_1  | 	started:	3
run-experiment_1  | 	deferred:	10
run-experiment_1  | 	finished:	8
run-experiment_1  | 	failed:	0
run-experiment_1  |   fuzzbench.jobs.build_image : finished	(base-image)
run-experiment_1  |   fuzzbench.jobs.build_image : finished	(dispatcher-image)
run-experiment_1  |   fuzzbench.jobs.build_image : finished	(coverage-freetype2-2017-builder-intermediate)
run-experiment_1  |   fuzzbench.jobs.build_image : started	(coverage-freetype2-2017-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : finished	(freetype2-2017-project-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : finished	(afl-freetype2-2017-builder-intermediate)
run-experiment_1  |   fuzzbench.jobs.build_image : queued	(afl-freetype2-2017-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(afl-freetype2-2017-intermediate-runner)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(afl-freetype2-2017-runner)
run-experiment_1  |   fuzzbench.jobs.build_image : finished	(coverage-libpng-1.2.56-builder-intermediate)
##[error]No space left on device

**Comment by jonathanmetzman** [comment]
Do we do 4 pairs because some benchmarks are OSS-Fuzz and other is standard?
maybe do afl-oss-fuzz-benchmark, libfuzzer-standard-benchmark
But frankly it isn't really necessary to test both anymore as they should be the same

File: fuzzbench/test_e2e/test_e2e_run.py (hunk 24 above)
Comment (zchcai) [reply fuzzbench/test_e2e/test_e2e_run.py:59]: We also need to try putting a job first and put the its dependency later.
To check whether:
1. it will be located at the deferred_job_registry and wait for its dependency.
2. it will directly go into the queue for running.

File: fuzzbench/test_e2e/end-to-end-test-config.yaml (hunk 18 above)
Comment (lszekeres) [reply fuzzbench/test_e2e/end-to-end-test-config.yaml:2]: With the recent changes there should be no difference between standard and oss-fuzz benchmarks, both should work.

File: fuzzbench/test_e2e/end-to-end-test-config.yaml [hunk 25]
```diff
@@ -0,0 +1,5 @@
+benchmarks:
+    - libpng-1.2.56
+fuzzers:
+    - libfuzzer
+max_total_time: 3600
```
lszekeres, line 5: can we remove max_total_time?

fuzzbench/run_experiment.py
```diff
@@ -18,20 +18,39 @@
...
-                      job_id='base-image'))
+    for name, obj in images_to_build.items():
+        if 'depends_on' not in obj:
+            jobs_list.append(
+                queue.enqueue(jobs.build_image,
+                              job_obj=obj,
+                              job_timeout=1800,
+                              result_ttl=config['max_total_time'],
+                              job_id=name))
+            continue
```
lszekeres, line 45: can we remove this part, and have a single enqueue() call in the for loop?

fuzzbench/jobs.py
```diff
@@ -19,14 +19,21 @@
 BASE_TAG = 'gcr.io/fuzzbench'
 
 
-def build_image(name: str):
+def build_image(job_obj):
```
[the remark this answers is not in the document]
zchcai: Done.

fuzzbench/test_e2e/end-to-end-test-config.yaml
zchcai: Yes. I change it back to one normal and one oss-fuzz.

fuzzbench/test_e2e/end-to-end-test-config.yaml
zchcai: Done.

fuzzbench/run_experiment.py
zchcai: Done.

fuzzbench/run_experiment.py
```diff
@@ -18,20 +18,39 @@
 import redis
 import rq
 
+from common import config_utils, environment, yaml_utils
+from experiment.build import docker_images
 from fuzzbench import jobs
 
+redis_connection = redis.Redis(host="queue-server")  # pylint: disable=invalid-name
```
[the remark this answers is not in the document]
zchcai: I remove this. Previously, we need it in two functions. Now it is redundant.

fuzzbench/jobs.py
```diff
@@ -19,14 +19,21 @@
...
-         os.path.join('docker', name)],
-        check=True)
+    command = ['docker', 'build', '--tag', image_tag, job_obj['context']]
+    cpu_options = ['--cpu-period', '100000', '--cpu-quota', '100000']
+    command.extend(cpu_options)
+    dockerfile = job_obj.get('dockerfile', None)
+    if dockerfile:
+        command.extend(['--file', dockerfile])
+    buildargs = job_obj.get('build_arg', None)
+    if buildargs:
```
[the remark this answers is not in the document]
zchcai: Done.

fuzzbench/jobs.py
```diff
@@ -19,14 +19,21 @@
...
     subprocess.run(['docker', 'pull', image_tag], check=True)
-    subprocess.run(
-        ['docker', 'build', '--tag', image_tag,
-         os.path.join('docker', name)],
-        check=True)
+    command = ['docker', 'build', '--tag', image_tag, job_obj['context']]
+    cpu_options = ['--cpu-period', '100000', '--cpu-quota', '100000']
+    command.extend(cpu_options)
+    dockerfile = job_obj.get('dockerfile', None)
+    if dockerfile:
```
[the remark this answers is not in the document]
zchcai: Done.

config.yaml
```diff
@@ -0,0 +1,7 @@
+benchmarks:
+    - freetype2-2017
+    - libpng-1.2.56
+fuzzers:
+    - afl
+    - libfuzzer
+max_total_time: 3600
```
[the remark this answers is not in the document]
zchcai: Yes. Done.

fuzzbench/run_experiment.py
```diff
@@ -18,20 +18,29 @@
...
-                      'base-image',
-                      job_timeout=600,
-                      job_id='base-image'))
+    for name, obj in images_to_build.items():
+        depends = obj.get('depends_on', None)
+        jobs_list.append(
+            queue.enqueue(jobs.build_image,
+                          image=obj,
+                          job_timeout=1800,
+                          result_ttl=-1,
```
zchcai, line 41: reference: https://python-rq.org/docs/results/
```
q.enqueue(foo, result_ttl=-1)  # result never expires--you should delete jobs manually

File: fuzzbench/test_e2e/end-to-end-test-config.yaml (hunk 12 above)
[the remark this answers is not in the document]
Comment (zchcai) [reply fuzzbench/test_e2e/end-to-end-test-config.yaml:6]: From previous running, if one fuzzer and one benchmark, it takes ~30 mins in Github CI running.

**Comment by zchcai** [comment]
I guess we may still use one fuzzer and one benchmark. Otherwise,

```
Unhandled exception. System.IO.IOException: No space left on device
   at System.IO.FileStream.WriteNative(ReadOnlySpan`1 source)
   at System.IO.FileStream.FlushWriteBuffer()
   at System.IO.FileStream.Flush(Boolean flushToDisk)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at System.Diagnostics.TraceSource.Flush()
   at GitHub.Runner.Common.TraceManager.Dispose(Boolean disposing)
   at GitHub.Runner.Common.TraceManager.Dispose()
   at GitHub.Runner.Common.HostContext.Dispose(Boolean disposing)
   at GitHub.Runner.Common.HostContext.Dispose()
   at GitHub.Runner.Worker.Program.Main(String[] args)
System.IO.IOException: No space left on device
   at System.IO.FileStream.WriteNative(ReadOnlySpan`1 source)
   at System.IO.FileStream.FlushWriteBuffer()
   at System.IO.FileStream.Flush(Boolean flushToDisk)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at GitHub.Runner.Common.HostTraceListener.WriteHeader(String source, TraceEventType eventType, Int32 id)
   at GitHub.Runner.Common.HostTraceListener.TraceEvent(TraceEventCache eventCache, String source, TraceEventType eventType, Int32 id, String message)
   at System.Diagnostics.TraceSource.TraceEvent(TraceEventType eventType, Int32 id, String message)
   at GitHub.Runner.Worker.Worker.RunAsync(String pipeIn, String pipeOut)
   at GitHub.Runner.Worker.Program.MainAsync(IHostContext context, String[] args)
System.IO.IOException: No space left on device
   at System.IO.FileStream.WriteNative(ReadOnlySpan`1 source)
   at System.IO.FileStream.FlushWriteBuffer()
   at System.IO.FileStream.Flush(Boolean flushToDisk)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at GitHub.Runner.Common.HostTraceListener.WriteHeader(String source, TraceEventType eventType, Int32 id)
   at GitHub.Runner.Common.HostTraceListener.TraceEvent(TraceEventCache eventCache, String source, TraceEventType eventType, Int32 id, String message)
   at System.Diagnostics.TraceSource.TraceEvent(TraceEventType eventType, Int32 id, String message)
   at GitHub.Runner.Common.Tracing.Error(Exception exception)
   at GitHub.Runner.Worker.Program.MainAsync(IHostContext context, String[] args)

zchcai:
One fuzzer and one benchmark is acceptable. No errors in presubmit CI test.

fuzzbench/test_e2e/test_e2e_run.py
zchcai: > We also need to try putting a job first and put the its dependency later.
> To check whether:
> 
> 1. it will be located at the deferred_job_registry and wait for its dependency.
> 2. it will directly go into the queue for running.

2. it will directly go into the queue for running.

zchcai: - enqueued_at: when the job enqueued to the queue
- started_at: when the job is worked by a worker
- ended_at: when the job is finished

Here is the execution result for 
```
jobs_list.append(queue.enqueue(jobs.run_trial, name='first', job_id='first'))
jobs_list.append(queue.enqueue(jobs.run_trial, name='second', job_id='second', depends_on='first'))
```
where one job is just a sleeping 10 seconds:
```
run-experiment_1  | All done!
run-experiment_1  |   fuzzbench.jobs.run_trial : finished       (first)
run-experiment_1  |    enqueued_at:  2020-08-03 18:53:29.681580
run-experiment_1  |    started_at:  2020-08-03 18:53:31.139904
run-experiment_1  |    ended_at:  2020-08-03 18:53:41.152006
run-experiment_1  |   fuzzbench.jobs.run_trial : finished       (second)
run-experiment_1  |    enqueued_at:  2020-08-03 18:53:41.155340
run-experiment_1  |    started_at:  2020-08-03 18:53:41.167681
run-experiment_1  |    ended_at:  2020-08-03 18:53:51.172041

Comment (zchcai) [reply fuzzbench/test_e2e/test_e2e_run.py:59]: > > We also need to try putting a job first and put the its dependency later.
> > To check whether:
> > 
> > 1. it will be located at the deferred_job_registry and wait for its dependency.
> > 2. it will directly go into the queue for running.
> 
> it will directly go into the queue for running.

Here is the execution result for
```
jobs_list.append(queue.enqueue(jobs.run_trial, name='second', job_id='second', depends_on='first'))
jobs_list.append(queue.enqueue(jobs.run_trial, name='first', job_id='first'))
```
where one job is just a sleeping 10 seconds:
```
run-experiment_1  | All done!
run-experiment_1  |   fuzzbench.jobs.run_trial : finished       (first)
run-experiment_1  |    enqueued_at:  2020-08-03 18:57:50.513644
run-experiment_1  |    started_at:  2020-08-03 18:57:52.592289
run-experiment_1  |    ended_at:  2020-08-03 18:58:02.601616
run-experiment_1  |   fuzzbench.jobs.run_trial : finished       (second)
run-experiment_1  |    enqueued_at:  2020-08-03 18:57:50.512880
run-experiment_1  |    started_at:  2020-08-03 18:57:52.484071
run-experiment_1  |    ended_at:  2020-08-03 18:58:02.515156

zchcai: The above execution can be done through this branch: https://github.com/zchcai/fuzzbench/tree/dep-test-branch

config.yaml
```diff
@@ -0,0 +1,6 @@
+benchmarks:
```
lszekeres, line 1: nit: can we call this file `local-experiment-config.yaml` ?

fuzzbench/test_e2e/test_e2e_run.py
```diff
@@ -19,7 +19,17 @@
...
 import redis
-from rq.job import Job
+import rq
+
+from common import config_utils, yaml_utils
+from experiment.build import docker_images
+
+
+@pytest.fixture(scope='class')
+def config():
```
lszekeres, line 29: nit: consider making this a bit more descriptive eg by calling it `experiment_config`

docker/fuzzbench/Dockerfile
```diff
@@ -31,7 +31,10 @@ COPY benchmarks benchmarks
...
 COPY fuzzbench fuzzbench
 COPY fuzzers fuzzers
+COPY config.yaml config.yaml
```
lszekeres, line 38: nit: can we make this more descriptive by calling it local-experiment-config.yaml?

fuzzbench/run_experiment.py
```diff
@@ -18,20 +18,29 @@
...
+                          job_id=name,
+                          depends_on=None if not depends else depends[0]))
```
lszekeres, line 43: consider: `depends[0] if `depends_on` in image else None`

fuzzbench/test_e2e/test_e2e_run.py
lszekeres: Thanks. So it sounds like we need to do a topological sort before enqueuing jobs. Currently, they happen to be in order, but please add a TODO to enforce this.

fuzzbench/run_experiment.py
```diff
@@ -18,20 +18,29 @@
...
-        queue.enqueue(jobs.build_image,
-                      'base-image',
-                      job_timeout=600,
-                      job_id='base-image'))
+    for name, obj in images_to_build.items():
+        depends = obj.get('depends_on', None)
+        jobs_list.append(
+            queue.enqueue(jobs.build_image,
+                          image=obj,
+                          job_timeout=1800,
```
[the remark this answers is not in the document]
zchcai: Done.

fuzzbench/run_experiment.py
```diff
@@ -18,20 +18,29 @@
...
+
+    images_to_build = docker_images.get_images_to_build(config['fuzzers'],
+                                                        config['benchmarks'])
     jobs_list = []
-    jobs_list.append(
-        queue.enqueue(jobs.build_image,
-                      'base-image',
-                      job_timeout=600,
-                      job_id='base-image'))
+    for name, obj in images_to_build.items():
```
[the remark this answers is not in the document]
zchcai: Good suggestion!

fuzzbench/test_e2e/test_e2e_run.py
zchcai: There is already a `experiment_config` and `local_experiment_config` in experiment/conftest.py for old architecture.
Maybe we can leave it as it is and later refactor/unify all of these three to a place.
wdyt?

docker/fuzzbench/Dockerfile
zchcai: Done.

fuzzbench/run_experiment.py
zchcai: Sure. Done.

zchcai: Sg!

docker/fuzzbench/Dockerfile
zchcai: Yes. Done.

fuzzbench/test_e2e/test_e2e_run.py
lszekeres: as discussed offline, there's no conflict with the other file

zchcai:
All checks are passed! @jonathanmetzman Maybe we can merge it now! :)

local-experiment-config.yaml
```diff
@@ -0,0 +1,6 @@
+benchmarks:
```
jonathanmetzman, line 1: Actually, I don't want this in the root directory. Please move.

jonathanmetzman:
> All checks are passed! @jonathanmetzman Maybe we can merge it now! :)

heh, I didn't mean *you* needed to change this, just whoever was taking over this work. Thanks a ton! But please don't feel obligated to do anything more.

local-experiment-config.yaml
zchcai: Sounds good! Done.

zchcai:
> > All checks are passed! @jonathanmetzman Maybe we can merge it now! :)
> 
> heh, I didn't mean _you_ needed to change this, just whoever was taking over this work. Thanks a ton! But please don't feel obligated to do anything more.

:) let's merge it now. Thanks.

lszekeres:
> > > All checks are passed! @jonathanmetzman Maybe we can merge it now! :)
> > 
> > 
> > heh, I didn't mean _you_ needed to change this, just whoever was taking over this work. Thanks a ton! But please don't feel obligated to do anything more.
> 
> :) let's merge it now. Thanks.

Thank YOU! :)

zchcai:
Just an update from rq: their 1.8.0 version support multi-dependency now!
https://github.com/rq/rq/issues/260#issuecomment-850986832

jonathanmetzman:
Cool! Thanks for updating us on this.

PR #551, opened 2020-07-19
remove temporary fuzzers, update commit id, typos

vanhauser-thc:
This removes the temporary variants, updates the commit it - plus fixes typos in the report template.


jonathanmetzman:
Do you need another experiment for this change? I assume not.

vanhauser-thc:
I have a long list :)
some I can do with our test server, but others I cannot.
However I am pretty busy this week so designing the next test case will take some time.
Next will be most likely new mutation options.

vanhauser-thc:
With the initial 7 hours of 2020-07-17 it might maybe be a good idea to do a full run again and see if the results are same as previous runs or if there is a difference as it looks in this one?

jonathanmetzman:
Yeah I'll rerun when I get a chance. There are some other other things in the queue, but I should get to it pretty soon.

jonathanmetzman:
> With the initial 7 hours of 2020-07-17 it might maybe be a good idea to do a full run again and see if the results are same as previous runs or if there is a difference as it looks in this one?

Just kicked off an experiment for this. Should appear on http://commondatastorage.googleapis.com/fuzzbench-reports/oss-fuzz-benchmarks/index.html in a couple hours.

vanhauser-thc:
the results look "normal", not like the other run :)

PR #553, opened 2020-07-20
[service] Make service ready for external users

jonathanmetzman:
1. Document service for external users.
2. Document service for FuzzBench maintainers.
3. Include a copy of the service's crontab.
4. Output stdout/stderr of cron job to a file.

### DIFF
--- a/service/README.md
+++ b/service/README.md
@@ -1,0 +1,41 @@
+# The FuzzBench Service
+
+*** note
+**Note:** This document and most of the `service/` directory is only intendended
+for use by FuzzBench maintainers. It will contain hardcoded values and
+references to things that don't make sense for other users.
+***
+
+## Overview
+
+This document discusses the FuzzBench service. The way the service works is when
+a user wants a new experiment they add the experiment to
+`experiment-requests.yaml`. Once a day at 12 PM PT (19:00 UTC) a cron job on the
+`service` instance will execute the script `run.bash`. `run.bash` will clone
+FuzzBench and then execute `automatic_run_experiment.py` which starts newly
+requested experiments. Since Jonathan Metzman setup this cron job on `service`
+the cron job runs under his user account on `service`.
+
+## Setting up an instance to run an experiment
+
+This shouldn't be necessary, but here are instructions in case the current
+instance is lost.
+1. Run `setup.bash`. This will build and install a supported Python version,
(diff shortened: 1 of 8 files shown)


jonathanmetzman:
The presubmit errors happened because I modified (comments) in experiment-requests.yaml.
Whenever experiment-requests.yaml is modified, the last experiment is validated. The last experiment no longer passes validation because it includes fuzzers that were deleted since the experiment ran. I'm not sure how to fix this in the general case.

service/README.md
```diff
@@ -0,0 +1,40 @@
+# The FuzzBench Service
```
jonathanmetzman, line 1: Maybe this README should be in the normal place for docs (I can create a fuzzbench maintainers section to put it in)?

service/README.md
```diff
@@ -0,0 +1,40 @@
...
+***
+
+## Overview
+
+This document discusses the FuzzBench service. The way the service works is when
+a user wants a new experiment they add the experiment to
+`experiment-requests.yaml`. Once a day at 12 PM PT (19:00 UTC) a cron job on the
+`service` instance will execute the script `run.bash`. `run.bash` will clone
+FuzzBench and then execute `automatic_run_experiment.py` which starts newly
+requested experiments. Since Jonathan Metzman setup this cron job on `service`
```
inferno-chromium, line 16: "Since Jonathan Metzman" line is not needed.

service/README.md
inferno-chromium: Yes should be in main docs.

service/run.bash
```diff
@@ -16,6 +16,10 @@
 # Use this script to clone a new copy of fuzzbench and run a requested
 # experiment.
 
+# Append stdout and stderr to /tmp/fuzzbench-service.logs. Redirecting the
+# stdout/stderr of this script in the cron command didn't work for some reason.
+exec &>> /tmp/fuzzbench-service.logs
```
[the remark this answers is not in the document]
jonathanmetzman: I tried this and the most intuitive solutions `&>` and they didn't seem to work. Maybe they didn't work because I tried them while fixing other issues and those other issues prevented this from working. But I can't spend more time on this now, so I think this solution, imperfect as it may be is good enough for now.

jonathanmetzman:
I made the requested changes. Please take another look.

Issue #555, opened 2020-07-20
uncertain lint error reporting


zchcai:
Uncertain behavior for `make lint`.
Sometimes it will give this error, which doesn't make sense as the later execution does call fake_jobs module.
```
source .venv/bin/activate && python3 presubmit.py lint
************* Module run_experiment
fuzzbench/run_experiment.py:21:0: E0611: No name 'fake_jobs' in module 'fuzzbench' (no-name-in-module)
```

Any clues or ideas to fix this?

jonathanmetzman:
You might need to add an `__init__.py` to `fuzzbench`.

zchcai:
An interesting behavior is, google/master works and if I switch from google/master to one pre-failed branch, it works too.

zchcai:
> An interesting behavior is, google/master works and if I switch from google/master to one pre-failed branch, it works too. (Without adding **init**.py)

I think this is just fake success since we don't change any source files. Then the lint doesn't work.

zchcai:
As suggested offline, we may need to change the folder name for the inside `fuzzbench`. See #559. This error is related to the name.

zchcai:
As suggested offline, we may want to try the latest pylint 2.5.3, as we're using 2.4.4 now.

jonathanmetzman:
Feel free to update it in requirements.txt

zchcai:
> As suggested offline, we may want to try the latest pylint 2.5.3, as we're using 2.4.4 now.

This probably not the reason.
When I change it, there is additional errors:
```
~/fuzzbench$ make lint
source .venv/bin/activate && python3 presubmit.py lint
************* Module run_experiment
fuzzbench/run_experiment.py:23:0: E0401: Unable to import 'common' (import-error)
fuzzbench/run_experiment.py:24:0: E0611: No name 'fake_jobs' in module 'fuzzbench' (no-name-in-module)

------------------------------------------------------------------
Your code has been rated at 7.78/10 (previous run: 7.78/10, +0.00)

ERROR: lint failed, see errors above.
Failed.
make: *** [Makefile:60: lint] Error 1
```
Any thoughts?

lszekeres:
> > As suggested offline, we may want to try the latest pylint 2.5.3, as we're using 2.4.4 now.
> 
> This probably not the reason.
> When I change it, there is additional errors:
> 
> ```
> ~/fuzzbench$ make lint
> source .venv/bin/activate && python3 presubmit.py lint
[...]

PR #556, opened 2020-07-20
Add __init__.py to fuzzbench

zchcai:


jonathanmetzman:
Does this fix the issue? Please link

zchcai:
This PR may fix #555. Under testing now.

jonathanmetzman:
TO test you might need to make a dummy change to the file that failed linting since we only lint changed files (which is incorrect behavior, but a different issue).

zchcai:
> TO test you might need to make a dummy change to the file that failed linting since we only lint changed files (which is incorrect behavior, but a different issue).

Oh, I see. Let me try again.

zchcai:
> TO test you might need to make a dummy change to the file that failed linting since we only lint changed files (which is incorrect behavior, but a different issue).

Well, adding `__init__.py` under fuzzbench/fuzzbench does not help. I check this by modifying a comment in fuzzbench/run_experiment.py:
```
************* Module fuzzbench.fuzzbench.run_experiment

-----------------------------------
Your code has been rated at 7.92/10

ERROR: lint failed, see errors above.
Failed.
make: *** [Makefile:60: lint] Error 1
```

zchcai:
As we can see, the lint check fails now.

zchcai:
By the way, despite #555, shall we add this `__init__.py` here?

jonathanmetzman:
Sure. It doesn't really matter to me.

lszekeres:
> By the way, despite #555, shall we add this `__init__.py` here?

I think empty __init__.py files are not necessary from python 3.3. why add them if we don't need them?

jonathanmetzman:
> > By the way, despite #555, shall we add this `__init__.py` here?
> 
> I think empty `__init__.py` files are not necessary from python 3.3. why add them if we don't need them?

I've definitely seen cases where not having them has been an issue. in src_analysis I had to write my own tooling because Python's `modulefinder` library. And I feel like I've seen cases where not having them messes up linting although I can't say for sure this is the case. At this point I think it's better to have them since there is no cost to doing so, but not having them can have a cost.

zchcai:
According to #566, this PR should be closed now to fix #555

Issue #289, opened 2020-05-01
Ensure docker image built on GCB is the one we use in an experiment


jonathanmetzman:
Currently FuzzBench uses GCB to build docker images and then tries to pull the latest version of that image. However if you run multiple experiments at once, one experiment can clobber the "latest" image leaving the other experiment using the wrong image. Instead we should get a hash from the build and ensure we use that hash. This could help with reproducing a very old experiment if we saved that hash somewhere.

jonathanmetzman:
This would have prevented #288.

jonathanmetzman:
Fixed by https://github.com/google/fuzzbench/pull/557

PR #557, opened 2020-07-20
Tag docker images by experiment

jonathanmetzman:
Previously docker images were untagged when built and pulled.
Previously: if I start experiment A and then start experiment B while A is still running, A might use some builds produced by B.
This makes it so that experiment A will only use the builds produced by A.
The most thorough fix would be to get the image digests and use those to pull.
But this solution is cleaner and should work as well so long as there are never two experiments with the same name but different builds started at the same time.

### DIFF
--- a/common/benchmark_utils.py
+++ b/common/benchmark_utils.py
@@ -44,11 +44,12 @@ def get_fuzz_target(benchmark):
     return fuzzer_utils.DEFAULT_FUZZ_TARGET_NAME
 
 
-def get_runner_image_url(benchmark, fuzzer, docker_registry):
+def get_runner_image_url(experiment, benchmark, fuzzer, docker_registry):
     """Get the URL of the docker runner image for fuzzing the benchmark with
     fuzzer."""
-    return '{docker_registry}/runners/{fuzzer}/{benchmark}'.format(
-        docker_registry=docker_registry, fuzzer=fuzzer, benchmark=benchmark)
+    return '{docker_registry}/runners/{fuzzer}/{benchmark}:{experiment}'.format(
+        docker_registry=docker_registry, fuzzer=fuzzer, benchmark=benchmark,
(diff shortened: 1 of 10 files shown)


jonathanmetzman:
Note this PR will break local experiments. I don't know the right way to fix them since they use make which doesn't use this tagging scheme (since most of the time, make is run outside of experiments).

jonathanmetzman:
Landing this today would be very helpful as there are a few experiments that need running in the queue right now. Running them concurrently would be nice.

jonathanmetzman:
CC @zchcai

jonathanmetzman:
This fixes #289

zchcai:
> This makes it so that experiment A will only use the builds produced by A.

Does this mean we can lose benefit from other experiment building? Or slower in general?

jonathanmetzman:
> > This makes it so that experiment A will only use the builds produced by A.
> 
> Does this mean we can lose benefit from other experiment building? Or slower in general?

I don't think it means you lose caching benefit. It means you won't accidentally use the wrong image in an experiment.

zchcai:
> > > This makes it so that experiment A will only use the builds produced by A.
> > 
> > 
> > Does this mean we can lose benefit from other experiment building? Or slower in general?
> 
> I don't think it means you lose caching benefit. It means you won't accidentally use the wrong image in an experiment.

I see. Then we can do parallel running. Thanks.

jonathanmetzman:
Previously docker images were untagged when built and pulled.
Previously: if I start experiment A and then start experiment B while A is still running, A might use some builds produced by B.
This makes it so that experiment A will only use the builds produced by A.
The most thorough fix would be to get the image digests and use those to pull.
But this solution is cleaner and should work as well so long as there are never two experiments with the same name but different builds started at the same time.

PR #559, opened 2020-07-21
[Do not merge] rename fuzzbench

### LINKED ISSUE #555: uncertain lint error reporting

zchcai:
Probably one way to fix #555 .
This PR is to show that #555 relates the name.
But we still want to keep this name and not change it.


zchcai:
Note: this PR's CI test in github will not fail in the `make lint`.

PR #561, opened 2020-07-21
[GCB] Fix image tagging

jonathanmetzman:
Don't add two tags in a single command. One is sufficient and two
is an error.


jonathanmetzman:
PTAL.

jonathanmetzman:
TBR

PR #566, opened 2020-07-21
Fix pylint

### LINKED ISSUE #368: Linting/typechecking in presubmit incorrectly assumes unchanged files cannot have new errors
This is assumption is wrong. Changing the function signature for example can introduce errors in unchanged files if their calls need to be updated. Deleting modules as well could have a similar effect.
The current system hasn't worked too badly but we should look into how quickly linting/typechecking can be done on every file and try this instead.

### LINKED ISSUE #555: uncertain lint error reporting

lszekeres:
Fixes #555 (pylint breakage) and fixes #368 (linting everything).

Linting all fuzzbench code at presubmit time takes ~2 mins:

```sh
$ time make lint
-------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 9.99/10, +0.01)

Success.

real	0m43.754s
user	1m50.426s
sys	0m5.777s
```


jonathanmetzman:
presubmit is still failing. Does this do what it is intended to do?

zchcai:
> presubmit is still failing. Does this do what it is intended to do?

This is really uncertain. We may need to dig much deeper for #555 now.

jonathanmetzman:
https://github.com/google/fuzzbench/issues/555#issuecomment-662158832 might help, just treat the outer fuzzbench dir as the package.

lszekeres:
> [#555 (comment)](https://github.com/google/fuzzbench/issues/555#issuecomment-662158832) might help, just treat the outer fuzzbench dir as the package.

Replied on that thread, PTAL

presubmit.py
```diff
@@ -194,16 +194,25 @@ def test_changed_integrations(paths: List[Path]):
...
-    if not paths:
-        return True
+def lint(_: List[Path]) -> bool:
+    """Run python's linter on all python code. Return False if it fails
+    linting."""
+
+    to_check = [
+        'analysis',
+        'common',
+        'database',
```
jonathanmetzman, line 204: What about docker?

jonathanmetzman: Also need `.github/workflows/build_and_test_run_fuzzer_benchmarks.py`

jonathanmetzman:
I think to trulyf ix #368 this needs to do the same behavior for pytype.

jonathanmetzman:
> I think to truly fix #368 this needs to do the same behavior for pytype.

Actually don't do that please. pytype is running single threaded because of some stability issues. Running it on every file in the project is prohibitively expensive IMO

presubmit.py
lszekeres: done

.pylintrc
```diff
@@ -9,7 +9,9 @@ extension-pkg-whitelist=
 
 # Add files or directories to the blacklist. They should be base names, not
 # paths.
-ignore=
+#
+# Files under alembic are nonconform.
```
jonathanmetzman, line 13: nit: `Files under alembic are generated and are nonconforming.`

lszekeres: done

PR #568, opened 2020-07-22
Fix presubmit.

lszekeres:


experiment/test_run_experiment.py
```diff
@@ -193,6 +193,7 @@ def test_validate_experiment_name_invalid(experiment_name):
 def test_copy_resources_to_bucket(tmp_path):
     """Tests that copy_resources_to_bucket copies the correct resources."""
     # Do this so that Ctrl-C doesn't pollute the repo.
+    cwd = os.getcwd()
```
jonathanmetzman, line 196: I'm surprised it isn't reset. But I think you need a try-finally here otherwise on failure you won't chdir back

lszekeres: done. (why'd it reset?)
also i'm not sure i understand why are we changing to a temp dir here.
at any rate, we should try to avoid chdir()-ing in tests.

jonathanmetzman: I had some idea that tests were run in different processes but it's definitely not true.
Because this test will write to cwd so it pollutes the directory otherwise.

jonathanmetzman:
I merged this to unbreak CI.

PR #569, opened 2020-07-22
Transparent oss-fuzz benchmarks

jonathanmetzman:


jonathanmetzman:
I generated [this report](http://commondatastorage.googleapis.com/fuzzbench-reports/oss-fuzz-benchmarks/index.html) using data from using the old (current) system for OSS-Fuzz benchmarks (using this command: `python3 analysis/generate_report.py -q -f aflplusplus_optimal honggfuzz entropic lafintel aflcc fastcgs_lm aflplusplus_qemu mopt aflplusplus afl aflsmart libfuzzer aflfast ankou eclipser fairfuzz afl_qemu manul honggfuzz_qemu -b bloaty_fuzz_target curl_curl_fuzzer_http jsoncpp_jsoncpp_fuzzer libpcap_fuzz_both mbedtls_fuzz_dtlsclient openssl_x509 sqlite3_ossfuzz php_php-fuzz-parser systemd_fuzz-link-parser zlib_zlib_uncompress_fuzzer -e 82800 -m -d /tmp/nontransparent-benchmarks 2020-05-24 2020-06-26 2020-06-30 2020-07-09 2020-07-13`).


They seem pretty comparable. So I think this can land.

docker/oss-fuzz-builder/checkout_commit.py
```diff
@@ -0,0 +1,72 @@
+# Copyright 2020 Google LLC
```
lszekeres, line 1: brought this up [before](https://github.com/google/fuzzbench/pull/306#discussion_r445719758) but not sure if there was a conclusion: have you considered using explicit `git checkout <commit>` commands in the Dockerfile-s instead of using this script?

Tanq16: Yes, the same has a TODO after your suggestion. Afaik, it can be done but still requires more work to figure out the name for the directory where the checkout needs to be done, specifically in case of multiple clone commands and when the name of the directory is not same as that of the repository (as opposed to finding it using a loop currently).

Tanq16: Update: The suggestion is to have the integrator must manually edit the Dockerfile to checkout the required commit as opposed to this being done by the script. This would be relatively cleaner. New TODO reflects the same.

jonathanmetzman:
Damn it. Looks like I didn't do http://commondatastorage.googleapis.com/fuzzbench-reports/oss-fuzz-benchmarks/index.html with the new benchmark system. I can tell because https://github.com/google/fuzzbench/commits/2dfe596dcdb3666eeaeddf3fe975cb1ea9554eb4 is not transparent-oss-fuzz-benchmarks branch

jonathanmetzman:
The experiment that I ran `oss-fuzz-benchmarks` was with some weird version that had files we added but still used builder hashes which seems wrong. @Tanq16 Can you please figure out what went wrong here.

PR #571, opened 2020-07-22
[libFuzzer] Add -fno-builtin flags to disable optimizing certain libcalls

dokyungs:
libFuzzer implements hooks to `memcmp`-like functions to get past data-flow dependent branches more easily.

There is an effort to disable optimizing calls to such functions when `-fsanitize=fuzzer-no-link` is given (see https://reviews.llvm.org/D83987), but getting this patch land and finally appear in the base docker image would take some time.

In the meantime, it would be useful to see if libFuzzer performs differently with this set of `-fno-builtin` flags applied. I will remove these lines once they become redundant.

### DIFF
--- a/fuzzers/libfuzzer/fuzzer.py
+++ b/fuzzers/libfuzzer/fuzzer.py
@@ -25,6 +25,18 @@ def build():
     # /usr/lib/libFuzzer.a as the FUZZER_LIB for the main fuzzing binary. This
     # allows us to link against a version of LibFuzzer that we specify.
     cflags = ['-fsanitize=fuzzer-no-link']
+
+    # Can be removed once the patch https://reviews.llvm.org/D83987 lands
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
Can you request an experiment as documented [here](https://google.github.io/fuzzbench/getting-started/adding-a-new-fuzzer/#requesting-an-experiment).

Please let me know if there is anything unclear about this process since you are the first person to use it (other than me).

dokyungs:
I edited `experiment-requests.yaml` (by following the instruction you pointed me to) to request an experiment for this change, and updated the patch with it.

dokyungs:
All checks have passed. Can somebody help merge this patch?

jonathanmetzman:
heh, the crontab that will run this service had a 17 instead of 0 for minutes, so this should run in a few minutes.

PR #577, opened 2020-07-23
Add missing build_args.

lszekeres:


docker/image_types.yaml
```diff
@@ -28,6 +28,9 @@
   tag: 'builders/coverage/{benchmark}'
```
zchcai, line 28: Probably, the name and tag should add `fuzzer` too.

lszekeres: discussed offline

Issue #573, opened 2020-07-23
OSS-Fuzz benchmarks failing to build in production


jonathanmetzman:
They fail with this error:

```
KeyError: 'oss_fuzz_builder_hash' Message: Failed to build measurer for php_php-fuzz-parser.
at get_oss_fuzz_builder_hash (/work/src/common/benchmark_utils.py:68)
at _build_oss_fuzz_project_coverage (/work/src/experiment/build/gcb_build.py:104)
at build_coverage (/work/src/experiment/build/gcb_build.py:51)
at build_measurer (/work/src/experiment/build/builder.py:62)
```
(https://pantheon.corp.google.com/errors/CI21rqXmirXkTw?time=P7D&project=fuzzbench&organizationId=433637338589)
Surprising because I think it worked in past experiments. Maybe a merge messed things up
CC @Tanq16

jonathanmetzman:
#578 fixed this.

PR #578, opened 2020-07-23
Fix GCB builds

jonathanmetzman:
1. Experiment based tagging is broken. Previously the experiment tag was added to the tag argument that used ${_REPO} and the one that used gcr.io/fuzzbench, but this was a mistake. It should have only been done to ${_REPO} because we only want to do this before pushing.
2. GCB builds for OSS-Fuzz benchmarks did not take into account the new build system for oss-fuzz benchmarks until now.

### DIFF
--- a/common/benchmark_utils.py
+++ b/common/benchmark_utils.py
@@ -61,14 +61,6 @@ def get_builder_image_url(benchmark, fuzzer, docker_registry):
         docker_registry=docker_registry, fuzzer=fuzzer, benchmark=benchmark)
 
 
-def get_oss_fuzz_builder_hash(benchmark):
-    """Get the specified hash of the OSS-Fuzz builder for the OSS-Fuzz project
-    used by |benchmark|."""
-    if is_oss_fuzz(benchmark):
-        return oss_fuzz.get_config(benchmark)['oss_fuzz_builder_hash']
-    raise ValueError('Can only get project on OSS-Fuzz benchmarks.')
-
-
 def validate(benchmark):
     """Return True if |benchmark| is a valid fuzzbench fuzzer."""
     if VALID_BENCHMARK_REGEX.match(benchmark) is None:
--- a/common/test_benchmark_utils.py
+++ b/common/test_benchmark_utils.py
@@ -33,18 +33,6 @@ def test_is_oss_fuzz(benchmark, expected_result, oss_fuzz_benchmark):
     assert benchmark_utils.is_oss_fuzz(benchmark) == expected_result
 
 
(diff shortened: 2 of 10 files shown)


jonathanmetzman:
Fixes #573

lszekeres:
Two things to consider:

1) It's difficult to see what's the difference between `gcr.io/fuzzbench` and `${_REPO}` and why they are needed. Would it make sense to use something like `${CACHE_FROM_REGISTRY}` and `${PUSH_TO_REGISTRY}` respectively instead? 

2) When we are pushing a new image tagged with `:${EXPERIMENT}`, should we tag it with `:latest` as well? That way we could use the `:latest` image to cache from.

inferno-chromium:
> Two things to consider:
> 
> 1. It's difficult to see what's the difference between `gcr.io/fuzzbench` and `${_REPO}` and why they are needed. Would 
it make sense to use something like `${CACHE_FROM_REGISTRY}` and `${PUSH_TO_REGISTRY}` respectively instead?

This should probably come in the rewrite, as i think Jonathan is trying to unblock Tanishq CL.

> 2. When we are pushing a new image tagged with `:${EXPERIMENT}`, should we tag it with `:latest` as well? That way we could use the `:latest` image to cache from.

+1 i had the same question, but i think this is already taken care of with e.g. (i think latest is default).
```
'--tag',
  'gcr.io/fuzzbench/builders/${_FUZZER}/${_BENCHMARK}',
``

jonathanmetzman:
>It's difficult to see what's the difference between gcr.io/fuzzbench and ${_REPO} and why they are needed. Would it make sense to use something like ${CACHE_FROM_REGISTRY} and ${PUSH_TO_REGISTRY} respectively instead?

I'm not sure if those suffice. But the difference is that the gcr.io/fuzzbench tags are for later images builds to use. So if I build base-image, I want base-builder to inherit from it. Since base-builder inherits from gcr.io/fuzzbench/base-image:latest, I tag my base-image build as this (:latest is implicit).

${_REPO}/fuzzbench/...:${_EXPERIMENT} tags on the other hand for pushing to docker registry. This is needed because 1. When we pull the runner to run a trial, we want to make sure we pull the runner that was built for the current experiment. 2. Because people using this code might not be allowed to push to gcr.io/fuzzbench.

jonathanmetzman:
>+1 i had the same question, but i think this is already taken care of with e.g. (i think latest is default).
>```
>'--tag',
>  'gcr.io/fuzzbench/builders/${_FUZZER}/${_BENCHMARK}',
>```

This is correct.

PR #584, opened 2020-07-24
Get rid of distinction between OSS-Fuzz and standard benchmarks

jonathanmetzman:
One of the more annoying sources of complexity in FuzzBench is the difference between OSS-Fuzz benchmarks and standard benchmarks. Builds already have distinctions between GCB and make, as well as differences between coverage builds and non coverage builds. This means that when changing how builds work, one might need to write the same code in 8 different places. The distinction between these types of benchmarks also was noticed by fuzzer contributors who sometimes noticed subtle differences between the images used to build the benchmarks (and some not so subtle ones like different working directories).

This PR eliminates the distinction for everyone except those integrating new benchmarks.
With https://github.com/google/fuzzbench/pull/569 OSS-Fuzz benchmarks became benchmarks that inherit from gcr.io/oss-fuzz/base-builder and have a Dockerfile and build.sh.
With this PR, "standard benchmarks" become the same.
The only difference is that when integrating an OSS-Fuzz benchmark, one uses our script to copy over the Dockerfile and build.sh, for standard benchmarks, one creates these manually.

Some changes this PR makes:
1. Eliminates dockerfiles for standard benchmarks (benchmark-builder).
2. Eliminates the process used for building standard benchmarks (in make and in GCB) and replaces it with the existing OSS-Fuzz build process.
3. Renames things related to the OSS-Fuzz build process. Since the OSS-Fuzz build process is now the only one, we remove "OSS-Fuzz" from names.
4. Creates OSS-Fuzz-style benchmarks for all standard benchmarks. All of these were verified with make using test-run. Some of these were verified to build on GCB.
5. Rename oss-fuzz.yaml to benchmark.yaml.

PR #585, opened 2020-07-24
change worker mode

### LINKED ISSUE #579: worker mode change
Currently, we use brust mode for all workers.
However, this may let some workers quickly down when there is nothing to do temporarily.
For example, when one job is worked by worker A, and all other jobs are deferred or finished, then all other workers will exit.

I guess, we still need these workers on until all jobs finish.

zchcai:
Fix #579.

In general, we want the workers work on some jobs or stay online, until the queue is empty.
_burst mode_ doesn't help since if there is no available jobs to start, a worker can still exit even the queue is not empty.

[One suggestion](https://github.com/rq/rq/pull/1305) has been made to rq.

However, we also don't wish the worker to be stopped from outside logic: we want it able to stop on his own by watching the status of the queues. Specifically, we want that
```
while(1):
    if queue.empty():
        time.sleep(30)
        if queue.empty():
            worker.exit()
    time.sleep(WAIT_TO_NEXT_CHECK)
```
And this logic is separate the internal logic for a worker.

Currently, we need to think a way to achieve this. Welcome to all technical ideas or suggestion to implement. Thanks.


zchcai:
Currently, it will stuck at here for `make run-end-to-end-test`:
```
...
run-experiment_1  |   fuzzbench.jobs.build_image('base-image',) : finished
run-experiment_1  |   fuzzbench.jobs.build_image('base-builder',) : finished
run-experiment_1  |   fuzzbench.jobs.build_image('base-runner',) : finished
run-experiment_1  | All done!
e2e-test_run-experiment_1 exited with code 0

```

lszekeres:
> In general, we want the workers work on some jobs or stay online, until the queue is empty.

To keep the worker alive until the queue is empty, can you do something like this?
```
while queue is not empty:
  worker.work(burst=True)
```

and if you also want to wait a bit before exiting:

```
while queue is not empty:
  worker.work(burst=True)
  time.sleep(few mins)
```

However, ideally we should create jobs in a way that relying on timing is not necessary.

fuzzbench/worker.py
```diff
@@ -11,3 +11,20 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""Self-defined worker module."""
+import redis
+import rq
+from rq import Queue, Worker
```
lszekeres, line 17: we typically import modules, not classes. can we just keep the import rq above and use rq.Queue and rq.Worker in the code?

zchcai: Done.

zchcai:
From [this suggestion], we know that, deferred jobs are not actually enqueued into the queue when we `enqueue` them.
Thus, to determine whether there are jobs to do, we cannot use `len(queue)` to decide whether there exist more jobs to start later.

In other words, if we think the jobs are enqueued successfully, then `len(queue)` will not reflect them until their status become _queued_, rather than _deferred_.

zchcai:
I guess we probably need to use [specific job registries](https://python-rq.org/docs/job_registries/) to know whether there still exists deferred or unfinished jobs.

zchcai:
If we use the following logic, the worker docker container will not stop for that `return`.
So basically, we need to use `worker.work(burst=True)` rather than `worker.work()`.

```
MAX_TIME_LIMIT = 3600
REASSIGN_GAP_TIME = 5


def no_pending_jobs(queue):
    """Checks whether the queue has unfinished jobs."""
    return (queue.deferred_job_registry.count +
        queue.scheduled_job_registry.count +
        queue.started_job_registry.count) == 0


def main():
    """Sets up Redis connection and starts the worker."""
    redis_connection = redis.Redis(host="queue-server")
    with rq.Connection(redis_connection):
        queue = rq.Queue('build_n_run_queue')
        worker = rq.Worker([queue], connection=redis_connection)
        worker.work()

        start_time = time.time()
        while time.time() - start_time < MAX_TIME_LIMIT:
            if no_pending_jobs(queue):
                time.sleep(REASSIGN_GAP_TIME)
                if no_pending_jobs(queue):
                    return
            time.sleep(REASSIGN_GAP_TIME)


if __name__ == '__main__':
    main()

**Comment by zchcai** [comment]
For this PR, we shall add  the following to `.github/workflows/presubmit.yml` to let the github run a CI test.
```
    - name: Run end to end CI test
      run: |
        make run-end-to-end-test
```

File: fuzzbench/worker.py [hunk 2]
```diff
@@ -0,0 +1,50 @@
...
+
+        start_time = time.time()
+        while time.time() - start_time < MAX_TIME_LIMIT:
+            if not no_pending_jobs(queue):
+                worker.work(burst=True)
+            else:
+                time.sleep(REASSIGN_GAP_TIME)
+                if no_pending_jobs(queue):
+                    return
+            time.sleep(REASSIGN_GAP_TIME)
```
lszekeres, line 46: would something like this work?
```
while len(queue) or len(queue.deferred_job_registry):
  worker.work(burst=True)
  time.sleep(5)
```

fuzzbench/worker.py
```diff
@@ -0,0 +1,50 @@
...
+import rq
+
+MAX_TIME_LIMIT = 3600
+REASSIGN_GAP_TIME = 5
+
+
+def no_pending_jobs(queue):
+    """Checks whether the queue has unfinished jobs."""
+    return (queue.deferred_job_registry.count +
+            queue.scheduled_job_registry.count +
```
lszekeres, line 27: isn't the `scheduled_jobs_registry` contain jobs that were scheduled to be enqueued at a later time by with `enqueue_at()` or `enqueue_in()`? if yes, then we don't need to check that as we only `enqueue()` jobs, right?

fuzzbench/worker.py
zchcai: This is interesting. I will try this later.
It should work if there is no multi-dependency. But since the current implementation has to resolve multi-denpendency, it probably will be broken when all old jobs finished but new jobs have not been scheduled.

fuzzbench/worker.py
zchcai: > the scheduled_jobs_registry contain jobs that were scheduled to be enqueued at a later time by with enqueue_at() or enqueue_in()? 

Yes. I think so. Also refer to https://github.com/google/fuzzbench/pull/588#discussion_r461162784
> if yes, then we don't need to check that as we only enqueue() jobs, right?

Right. Let's change this to `len(queue)`.

zchcai: Done.

fuzzbench/worker.py
zchcai: If we eliminate it as this plan:
```
Eliminate base-runner and the double-dependency of the benchmark-runner image
Eliminate base-runner
by moving the steps of base-runner into benchmark-runner ({fuzzer}-{benchmark}-runner)
This would allow builds of benchmark-runner to only depend on runner.Dockerfile builds which will depend on builder.Dockerfile but inherit from base-image. This is fine because builder.Dockerfile inherits/depends on base-image because 
Make {fuzzer}-{benchmark}-intermediate-runner depend on {fuzzer}-{benchmark}-builder
Make {fuzzer}-{benchmark}-runner only depend on {fuzzer}-{benchmark}-intermediate-runner 
```
Then it works for your proposed way.
```
while len(queue) or len(queue.deferred_job_registry):
  worker.work(burst=True)
  time.sleep(5)
```
Check this PR: https://github.com/zchcai/fuzzbench/pull/10.

lszekeres: Thanks! Should we check-in the elimination of the double dependency first, and then this PR?

zchcai: Sounds good to me. I think the possible order would be: #584, then eliminate double dependency (which may have conflicts with #584), then #585 (this PR but use the way in zchcai#10).

lszekeres: SG!

fuzzbench/worker.py
```diff
@@ -0,0 +1,36 @@
...
+REASSIGN_GAP_TIME = 5
+
+
+def main():
+    """Sets up Redis connection and starts the worker."""
+    redis_connection = redis.Redis(host="queue-server")
+    with rq.Connection(redis_connection):
+        queue = rq.Queue('build_n_run_queue')
+        worker = rq.Worker([queue], connection=redis_connection)
+
```
lszekeres, line 29: can you drop `connection=redis_connection` here, given that context is set with `with rq.Connection(redis_connection)`?

fuzzbench/worker.py
```diff
@@ -0,0 +1,36 @@
...
+        while queue.count + queue.deferred_job_registry.count > 0:
+            worker.work(burst=True)
+            time.sleep(REASSIGN_GAP_TIME)
```
lszekeres, line 32: i think `REASSIGN_GAP_TIME` just raises more questions for the reader than answers, i'd probably just `time.sleep(5)` here. i think we want to sleep here a bit no matter what, otherwise this (potentially empty) loop  might eat up the CPU

fuzzbench/run_experiment.py
```diff
@@ -24,7 +24,7 @@
 def run_experiment():
     """Main experiment logic."""
     print('Initializing the job queue.')
-    queue = rq.Queue()
+    queue = rq.Queue('build_n_run_queue')
```
lszekeres, line 27: i was just thinking that 'build_n_run_queue' can be misunderstood as a queue that contains tasks that "build and run" at the same time. maybe we should drop the "n" and just call it "build_run_queue"? this is not really important, but feel free to change it if you think it makes sense

zchcai: Like this point. I think we can make a comment before this line, saying
```
# Create the queue for scheduling build jobs and run jobs.
```
wdyt?

fuzzbench/worker.py
zchcai: Sounds good! Done.

fuzzbench/worker.py
zchcai: OK. Done.

fuzzbench/run_experiment.py
lszekeres: sg

PR #588, opened 2020-07-26
[NFC] update current jobs status reporting info

zchcai:
Use queue's job registries to reflect the actual status of that queue. This will help to have a glance at what is going on now.
Minor format changing makes it look neat and aligned.

Reference: https://python-rq.org/docs/job_registries/

The final result will look like this:
```
...
Attaching to e2e-test_run-experiment_1, e2e-test_worker_2, e2e-test_worker_1, e2e-test_worker_3
run-experiment_1  | Initializing the job queue.
run-experiment_1  | 	started:	1
run-experiment_1  | 	deferred:	15
run-experiment_1  | 	finished:	0
run-experiment_1  | 	scheduled:	0
run-experiment_1  | 	failed:	0
run-experiment_1  |   fuzzbench.jobs.build_image : started	(base-image)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(base-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(base-runner)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(coverage-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(coverage-freetype2-2017-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(afl-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(afl-freetype2-2017-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(afl-freetype2-2017-intermediate-runner)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(coverage-libpng-1.2.56-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(afl-libpng-1.2.56-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(afl-libpng-1.2.56-intermediate-runner)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(libfuzzer-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(libfuzzer-freetype2-2017-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(libfuzzer-freetype2-2017-intermediate-runner)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(libfuzzer-libpng-1.2.56-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred	(libfuzzer-libpng-1.2.56-intermediate-runner)
...

### DIFF
--- a/fuzzbench/run_experiment.py
+++ b/fuzzbench/run_experiment.py
@@ -25,6 +25,8 @@ def run_experiment():
     """Main experiment logic."""
     print('Initializing the job queue.')
     queue = rq.Queue()
+    # Initialization.
+    queue.empty()
     jobs_list = []
     jobs_list.append(
         queue.enqueue(jobs.build_image,
@@ -46,6 +48,7 @@ def run_experiment():
 
     while True:
         print('Current status of jobs:')
+        print('\tqueued:\t%d' % len(queue))
         print('\tstarted:\t%d' % queue.started_job_registry.count)
         print('\tdeferred:\t%d' % queue.deferred_job_registry.count)
         print('\tfinished:\t%d' % queue.finished_job_registry.count)
--- a/fuzzbench/run_experiment.py
+++ b/fuzzbench/run_experiment.py
@@ -46,8 +46,14 @@ def run_experiment():
 
     while True:
         print('Current status of jobs:')
+        print('\tstarted:\t%d' % queue.started_job_registry.count)
+        print('\tdeferred:\t%d' % queue.deferred_job_registry.count)
+        print('\tfinished:\t%d' % queue.finished_job_registry.count)
+        print('\tscheduled:\t%d' % queue.scheduled_job_registry.count)
+        print('\tfailed:\t%d' % queue.failed_job_registry.count)
         for job in jobs_list:
-            print('  %s%s : %s' % (job.func_name, job.args, job.get_status()))
+            print('  %s : %s\t(%s)' % (job.func_name, job.get_status(), job.id))
+
         if all([job.result is not None for job in jobs_list]):
             break
         time.sleep(3)
--- a/fuzzbench/run_experiment.py
+++ b/fuzzbench/run_experiment.py
@@ -25,8 +25,6 @@ def run_experiment():
     """Main experiment logic."""
     print('Initializing the job queue.')
     queue = rq.Queue()
-    # Initialization.
-    queue.empty()
     jobs_list = []
     jobs_list.append(
         queue.enqueue(jobs.build_image,
--- a/fuzzbench/run_experiment.py
+++ b/fuzzbench/run_experiment.py
@@ -46,11 +46,10 @@ def run_experiment():
 
     while True:
         print('Current status of jobs:')
-        print('\tqueued:\t%d' % len(queue))
+        print('\tqueued:\t%d' % queue.count)
         print('\tstarted:\t%d' % queue.started_job_registry.count)
         print('\tdeferred:\t%d' % queue.deferred_job_registry.count)
         print('\tfinished:\t%d' % queue.finished_job_registry.count)
-        print('\tscheduled:\t%d' % queue.scheduled_job_registry.count)
         print('\tfailed:\t%d' % queue.failed_job_registry.count)
         for job in jobs_list:
             print('  %s : %s\t(%s)' % (job.func_name, job.get_status(), job.id))

### CODE REVIEW

**Comment by jonathanmetzman** [comment]
Long term I think this printing isn't a good idea as it will be extremely spammy given all the tasks that happen

File: fuzzbench/run_experiment.py [hunk 1]
```diff
@@ -46,8 +46,14 @@ def run_experiment():
...
+        print('\tfinished:\t%d' % queue.finished_job_registry.count)
+        print('\tscheduled:\t%d' % queue.scheduled_job_registry.count)
+        print('\tfailed:\t%d' % queue.failed_job_registry.count)
         for job in jobs_list:
-            print('  %s%s : %s' % (job.func_name, job.args, job.get_status()))
+            print('  %s : %s\t(%s)' % (job.func_name, job.get_status(), job.id))
+
         if all([job.result is not None for job in jobs_list]):
             break
         time.sleep(3)
```
zchcai, line 59: > Long term I think this printing isn't a good idea as it will be extremely spammy given all the tasks that happen

Yes. In the long term, these printing will be too much.
This change is more for helping understand that, when we write `queue.enqueue()`, a job is not immediately gotten into the queue, but first get registered in a registry.

Btw, in the long term, I think we shall change 3 seconds to a longer one. And then it is not that spammy.

fuzzbench/run_experiment.py
```diff
@@ -46,8 +46,14 @@ def run_experiment():
 
     while True:
         print('Current status of jobs:')
+        print('\tstarted:\t%d' % queue.started_job_registry.count)
+        print('\tdeferred:\t%d' % queue.deferred_job_registry.count)
+        print('\tfinished:\t%d' % queue.finished_job_registry.count)
+        print('\tscheduled:\t%d' % queue.scheduled_job_registry.count)
```
lszekeres, line 52: isn't the scheduled_jobs_registry contain jobs that were scheduled to be enqueued at a later time by with enqueue_at() or enqueue_in()?

zchcai: Good question! From offline discussion, this will be confirmed by not running workers and check whether there is only one scheduled job in this registry, saying `base-image`.
I will check this later.

zchcai: `len(queue)` is not `queue.scheduled_job_registry.count`.

> the scheduled_jobs_registry contain jobs that were scheduled to be enqueued at a later time by with enqueue_at() or enqueue_in()

I think yes according to this doc link: https://python-rq.org/docs/scheduling/#scheduling-jobs-for-execution

```
run-experiment_1  |     queued: 1
run-experiment_1  |     started:        0
run-experiment_1  |     deferred:       15
run-experiment_1  |     finished:       0
run-experiment_1  |     scheduled:      0
run-experiment_1  |     failed: 0
run-experiment_1  |   fuzzbench.jobs.build_image : queued       (base-image)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred     (base-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred     (base-runner)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred     (coverage-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred     (coverage-freetype2-2017-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred     (afl-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred     (afl-freetype2-2017-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred     (afl-freetype2-2017-intermediate-runner)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred     (coverage-libpng-1.2.56-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred     (afl-libpng-1.2.56-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred     (afl-libpng-1.2.56-intermediate-runner)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred     (libfuzzer-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred     (libfuzzer-freetype2-2017-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred     (libfuzzer-freetype2-2017-intermediate-runner)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred     (libfuzzer-libpng-1.2.56-builder)
run-experiment_1  |   fuzzbench.jobs.build_image : deferred     (libfuzzer-libpng-1.2.56-intermediate-runner)
```

fuzzbench/run_experiment.py
```diff
@@ -25,6 +25,8 @@ def run_experiment():
     """Main experiment logic."""
     print('Initializing the job queue.')
     queue = rq.Queue()
+    # Initialization.
+    queue.empty()
```
lszekeres, line 29: Why is this necessary? We create a fresh brand redis instance every time, no?

fuzzbench/run_experiment.py
lszekeres: So, given that we don't "schedule" jobs, should we remove the  
`print('\tscheduled:\t%d' % queue.scheduled_job_registry.count)`
line?

fuzzbench/run_experiment.py
```diff
@@ -46,8 +48,15 @@ def run_experiment():
 
     while True:
         print('Current status of jobs:')
+        print('\tqueued:\t%d' % len(queue))
```
lszekeres, line 51: consider using either `.count` or `len()` everywhere for consistency

fuzzbench/run_experiment.py
```diff
@@ -46,8 +48,15 @@ def run_experiment():
...
+        print('\tstarted:\t%d' % queue.started_job_registry.count)
+        print('\tdeferred:\t%d' % queue.deferred_job_registry.count)
+        print('\tfinished:\t%d' % queue.finished_job_registry.count)
+        print('\tscheduled:\t%d' % queue.scheduled_job_registry.count)
+        print('\tfailed:\t%d' % queue.failed_job_registry.count)
```
lszekeres, line 56: Do we want to print both the stats and the list of jobs?

By "stats" i mean this part:
```
run-experiment_1  |     queued: 1
run-experiment_1  |     started:        0
run-experiment_1  |     deferred:       15
run-experiment_1  |     finished:       0
run-experiment_1  |     scheduled:      0
run-experiment_1  |     failed: 0
```

If the stats are enough, we can remove the list of jobs.

If we want to keep the list of jobs, do we still need the stats, given that we print the status of the jobs (which makes the stats redundant)?

fuzzbench/run_experiment.py
zchcai: No. If some jobs finished and the their results don't expired, we can still have those information in the queue, i.e., finished_job_registry.
For others, the same for unexpected termination of experiment.

fuzzbench/run_experiment.py
zchcai: Sounds good.

fuzzbench/run_experiment.py
zchcai: Good suggestions! I use this form from the documentation. I'm not sure they share a same way for requesting length/amount. Will check it later.

fuzzbench/run_experiment.py
zchcai: > Do we want to print both the stats and the list of jobs?

In the long term, I think this should configured by user for information printing. Also the time period to iteration printing.

Short termly, we use this information to understand why len(queue) is zero but there are deferred jobs "enqueued" by `queue.enqueue()` statement. The fact is, deferred jobs are not in the queue, but in deferred_jobs_registry.

My opinion is to keep both of them currently.

> If the stats are enough, we can remove the list of jobs.

I would not say stats are enough.

> If we want to keep the list of jobs, do we still need the stats, given that we print the status of the jobs (which makes the stats redundant)?

I think it depends on what information we want. For stats, we can have a glance on what's going on. Just a list of status of jobs will be hard to know the numbers.
For list, that is also important and we can know which image is building now. Further, which trial is running now.

I think the question is, "we" are developers and debuggers, and we need more information.
However, if we think from the "users", we cannot assume which level of information will be better to them. We need to provide flexibility for information printing. Like mentioned above, status or stats, 3 seconds or 30 seconds, etc.

fuzzbench/run_experiment.py
zchcai: By the way, I think we still have reasons to keep it as it will print all registries' info, no missing, which is complete.
wdyt?

fuzzbench/run_experiment.py
lszekeres: As mentioned above, when we e.g. `make run-experiment` or `make run-end-to-end-test`, we instantiate a new redis container (with docker compose), so there shouldn't be any existing data in it.

fuzzbench/run_experiment.py
lszekeres: i personally don't see a reason the print "scheduled: 0" every few seconds, when we never schedule jobs

fuzzbench/run_experiment.py
lszekeres: i'd personally eliminate redundant information, but leaving it up to you

fuzzbench/run_experiment.py
zchcai: unfortunately. no. So this is necessary.

zchcai: Could you try it on your side? For example, start `make run-end-to-end-test` and then stop it some time after.
Then starts it again.

fuzzbench/run_experiment.py
zchcai: I agree. But I guess the meaning of "redundant" depends on who runs this experiment.

fuzzbench/run_experiment.py
lszekeres: Right, if your redis is not empy, that means that the experiment run wasn't shut down properly (ie., the `docker-compose down` command didn't run). This will happen e.g., if the experiment was interrupted. Since the containers were not shut down, when you run docker compose up again, you'll just use the running containers. If you manually interrupt the experiment run, make sure you "docker compose down" before running again.

zchcai: Good suggestions!

zchcai: https://github.com/google/fuzzbench/pull/600 Done.

fuzzbench/run_experiment.py
zchcai: Done. We use `.count` for consistency.

fuzzbench/run_experiment.py
zchcai: I eliminate the `scheduled_job_registry`. If finally we use that, we can add it at that time.

fuzzbench/run_experiment.py
zchcai: OK. Done for removal.

PR #589, opened 2020-07-26
Fix exceptions with clang source code coverage while measuring.

inferno-chromium:


inferno-chromium:
Without this, trials ends with exceptions due to subprocess.CalledProcessError triggering on expect_zero=True in new_process.
There is still main bug which is profraw is 0 bytes when coverage run fails (aka libFuzzer crashes, e.g. with timeout), filed in https://github.com/google/fuzzbench/issues/590

experiment/measurer.py
```diff
@@ -387,6 +387,13 @@ def get_current_coverage(self) -> int:
 
     def generate_profdata(self, cycle: int):
         """Generate .profdata file from .profraw file."""
+        if not os.path.exists(self.profraw_file):
```
jonathanmetzman, line 390: So this will log the same error about there not being a profraw file twice? Once in generate_profdata and once in generate_summary? Maybe we should have a helper method that checks for these files and returns early if they don't exist, otherwise it calls both of these functions. This way we can log error once.

inferno-chromium: Yes i will fix that in next CL, at first i am debugging other things. Basically all runner exits get fixed after this. Now, the next challenge is some coverage merge failures - see http://commondatastorage.googleapis.com/fuzzbench-reports/aarya-2020-07-26/index.html

Issue #392, opened 2020-05-31
Ankou's integration seems buggy


jonathanmetzman:
In https://www.fuzzbench.com/reports/2020-05-28/index.html I tested new the new preemptible strategy in an experiment comparing ankou to other fuzzers. However, ankou seems to fail a lot in ways that have nothing to do with the preemptible strategy. For example, in freetype2-2017, only 6 out of 20 trials of ankou that weren't terminated before 23 hours, actually produced corpus for 23 hours. When I took a look at the logs of one of the trials that lasted 23 hours (using this command `gsutil cat gs://fuzzbench-data/2020-05-28/experiment-folders/freetype2-2017-ankou/trial-147317/results/fuzzer-log.txt`) I see the following:
```
[run_fuzzer] Running target with Ankou
[run_fuzzer] Running command: ./Ankou -app /out/fuzz-target -i /out/seeds -o /out/corpus
2020/05/28 22:37:30 Invalid fuzz duration: time: invalid duration .
Persistent mode detected.
Deferred fork server detected.
len(inputs) = 2
len(seedPts) = 2
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/out/fuzzer.py", line 48, in fuzz
    subprocess.check_call(command)
  File "/usr/local/lib/python3.7/subprocess.py", line 363, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['./Ankou', '-app', '/out/fuzz-target', '-i', '/out/seeds', '-o', '/out/corpus']' died with <Signals.SIGKILL: 9>.
```

jonathanmetzman:
@zchcai

jonathanmetzman:
Actually, I'm not sure this is the integration's fault. It might be Ankou's since we do have corpora for about 45 minutes of fuzzing. But @zchcai it would probably help get this fixed if you had a reproducer for this.

zchcai:
Just check the trials running towards end (should be 20). I think the main issues are due to the trials for ankou with freetype2-2017 and harfbuzz-1.3.2.

  | ankou-82800
-- | --
bloaty_fuzz_target | 20
curl_curl_fuzzer_http | 19
freetype2-2017 | 6
harfbuzz-1.3.2 | 3
jsoncpp_jsoncpp_fuzzer | 19
lcms-2017-03-21 | 19
libpng-1.2.56 | 20
libjpeg-turbo-07-2017 | 20
[...]

PR #593, opened 2020-07-27
Remove ankou temporarily.

inferno-chromium:
Remove ankou as it consumes a very large amount of memory and fails
on runner bots for most benchmarks. Remove it till the failures
are fixed and avoid using it in full experiments.
Tracked in https://github.com/google/fuzzbench/issues/392


jonathanmetzman:
Related: #392

jonathanmetzman:
CC @Jiliac

Issue #590, opened 2020-07-26
No coverage is captured due to profile version mismatch in old OSS-Fuzz benchmarks (bloaty, curl).


inferno-chromium:
Profraw file is non-empty.
Can be easily tested by
gs://fuzzbench-data/aarya-2020-07-26%2Fcoverage-binaries/coverage-build-bloaty_fuzz_target.tar.gz

```
mkdir crashes
LLVM_PROFILE_FILE=abc.profraw time ./fuzz_target -merge=1 -dump_coverage=1 -artifact_prefix=crashes/ -timeout=5 -rss_limit_mb=2048 -max_total_time=900
<path-to-fuzzbench>/experiment/test_data/llvm_tools/llvm-profdata merge -sparse abc.profraw -o abc.profdata
```

See error:
```
warning: abc.profraw: Unsupported instrumentation profile format version
error: No profiles could be merged.
```

Happening for bloaty, curl. Checking others.

inferno-chromium:
E.g. run

```
WARNING: Failed to find function "__sanitizer_set_death_callback".
INFO: Seed: 1559695180
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 1048576 bytes
MERGE-INNER: using the control file '/tmp/libFuzzerTemp.Merge176090.txt'
MERGE-INNER: '/work/measurement-folders/jasper-1.701.0-ankou/trial-220668/corpus/2a08f7efcbd3dac4af41bf7242932e6ef8dd0c50' caused a failure at the previous merge step
MERGE-INNER: 41 total files; 29 processed earlier; will process 12 files now
fuzz-target: jpc_bs.c:195: long jpc_bitstream_getbits(jpc_bitstream_t *, int): Assertion `n >= 0 && n < 32' failed.
==176227== ERROR: libFuzzer: deadly signal
NOTE: libFuzzer has rudimentary signal handlers.
      Combine libFuzzer with AddressSanitizer or similar for better crash reports.
SUMMARY: libFuzzer: deadly signal
MS: 0 ; base unit: 0000000000000000000000000000000000000000
artifact_prefix='/work/measurement-folders/jasper-1.701.0-ankou/trial-220668/crashes/'; Test unit written to /work/measurement-folders/jasper-1.701.0-ankou/trial-220668/crashes/crash-ac28bfbedaa7f81bd9bc4641a8ce6415f991a158
MERGE-OUTER: attempt 31
[...]

PR #594, opened 2020-07-27
Upgrade curl, bloaty benchmarks for clang source code coverage.

inferno-chromium:
Fixes https://github.com/google/fuzzbench/issues/590

### DIFF
--- a/benchmarks/bloaty_fuzz_target/Dockerfile
+++ b/benchmarks/bloaty_fuzz_target/Dockerfile
@@ -15,7 +15,6 @@
 ################################################################################
 
 FROM gcr.io/oss-fuzz-base/base-builder@sha256:12125f1ecb9e4905a19c35700a3402c1eba43c82ea2d227f01c28833d75e3574
-MAINTAINER jhaberman@gmail.com
 RUN apt-get update && apt-get install -y cmake ninja-build g++
 RUN git clone --depth 1 https://github.com/google/bloaty.git bloaty
 WORKDIR bloaty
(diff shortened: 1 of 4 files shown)


jonathanmetzman:
Do we have any data on running an experiment with this new OSS-Fuzz benchmark system? The experiments I ran before were wrong, and it would be nicer to have that before making this change (and especially before landing #584)

inferno-chromium:
> Do we have any data on running an experiment with this new OSS-Fuzz benchmark system? The experiments I ran before were wrong, and it would be nicer to have that before making this change (and especially before landing #584)

Yes it works, see http://commondatastorage.googleapis.com/fuzzbench-reports/aarya-2020-07-27/index.html that is run only for bloaty and curl. And see the old one that fails - http://commondatastorage.googleapis.com/fuzzbench-reports/aarya-2020-07-26/index.html

jonathanmetzman:
> > Do we have any data on running an experiment with this new OSS-Fuzz benchmark system? The experiments I ran before were wrong, and it would be nicer to have that before making this change (and especially before landing #584)
> 
> Yes it works, see http://commondatastorage.googleapis.com/fuzzbench-reports/aarya-2020-07-27/index.html that is run only for bloaty and curl. And see the old one that fails - http://commondatastorage.googleapis.com/fuzzbench-reports/aarya-2020-07-26/index.html

Yes but this is incomplete and uses the new coverage. Still, the trend isn't good because now mopt is in first place which would mean one of our changes is affecting the benchmarking results. That doesn't mean the new results are wrong and the old ones right but isn't great for fuzzbench.

Shadoom7:
LGTM!

PR #600, opened 2020-07-28
add stop logic: docker-compose down

zchcai:
Fix https://github.com/google/fuzzbench/pull/588#discussion_r461247543

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,14 +34,10 @@ run-end-to-end-test:
 
 # Stop and clean running environment.
 
-stop-experiment: export COMPOSE_PROJECT_NAME := fuzzbench
-stop-experiment: export COMPOSE_FILE := compose/fuzzbench.yaml
 stop-experiment:
+	COMPOSE_PROJECT_NAME=fuzzbench COMPOSE_FILE=compose/fuzzbench.yaml \
 	docker-compose down
-
-stop-end-to-end-test: export COMPOSE_PROJECT_NAME := e2e-test
-stop-end-to-end-test: export COMPOSE_FILE := compose/fuzzbench.yaml:compose/e2e-test.yaml
-stop-end-to-end-test:
+	COMPOSE_PROJECT_NAME=e2e-test COMPOSE_FILE=compose/fuzzbench.yaml:compose/e2e-test.yaml \
 	docker-compose down
 
 include docker/build.mk
--- a/Makefile
+++ b/Makefile
@@ -14,17 +14,22 @@
 
 # Running experiments locally.
 
-run-experiment: export COMPOSE_PROJECT_NAME := fuzzbench
-run-experiment: export COMPOSE_FILE := compose/fuzzbench.yaml
+run-experiment stop-experiment: export COMPOSE_PROJECT_NAME := fuzzbench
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -32,6 +32,18 @@ run-end-to-end-test:
 	docker-compose run run-tests; STATUS=$$?; \
 	docker-compose down; exit $$STATUS
 
+# Stop and clean running environment.
+
+stop-experiment: export COMPOSE_PROJECT_NAME := fuzzbench
```
jonathanmetzman, line 37: Can you make this one target please?

Makefile
```diff
@@ -32,6 +32,18 @@ run-end-to-end-test:
...
+stop-experiment: export COMPOSE_FILE := compose/fuzzbench.yaml
+stop-experiment:
+	docker-compose down
+
+stop-end-to-end-test: export COMPOSE_PROJECT_NAME := e2e-test
```
jonathanmetzman, line 42: Same comment.

Makefile
zchcai: I guess different COMPOSE_PROJECT_NAME needs different targets?

zchcai: Done.

Makefile
```diff
@@ -32,6 +32,14 @@ run-end-to-end-test:
 	docker-compose run run-tests; STATUS=$$?; \
 	docker-compose down; exit $$STATUS
 
+# Stop and clean running environment.
+
+stop-experiment:
```
lszekeres, line 37: You can 1) re-use the env var settings above, and 2) mention that this is only necessary to run if the experiment was interrupted and the the down command above didn't run. I.e.:

```
run-experiment stop-experiment: export COMPOSE_PROJECT_NAME := fuzzbench
run-experiment stop-experiment: export COMPOSE_FILE := compose/fuzzbench.yaml
run-experiment:
	docker-compose up --build --scale worker=2 --detach
	docker-compose logs --follow run-experiment
	docker-compose down

# Running this is only necessary if `run-experiment` was interrupted and
# containers were not cleaned up.
stop-experiment:
	docker-compose down
```

zchcai: What about `stop end-to-end-test`? Shall we omit that?

lszekeres: You can add  a `stop-end-to-end-test` rule below `run-end-to-end-test` in a similar fashion, if you think that's useful to have. But please add the same comment saying that in general explicit stopping is not necessary.

zchcai: OK. @lszekeres  As suggested in https://github.com/google/fuzzbench/pull/600#discussion_r461709175
I think @jonathanmetzman  suggest only one target. :)

lszekeres: Not sure what that comment was referring to, but "run-experiment" and "run-end-to-end-test" are two separate setups. If we'd like to have a manual cleanup target for both of them, please let's add a corresponding "stop-experiment" and "stop-end-to-end-test" target, as described above.

jonathanmetzman: My comment was sort of a misunderstanding, I thought there were a bunch of lines listing dependencies for the stop-experiment target that could be merged into one line listing dependencies.

zchcai: I see. Done.

Issue #547, opened 2020-07-16
Make sure clang cov data is always dumped by libfuzzer


jonathanmetzman:
In https://github.com/google/fuzzbench/blob/master/fuzzers/coverage/patch.diff you can see that I added some code to make sure the coverage is always dumped by libFuzzer (I think the patch is to handle ooms and timeouts). I believe libFuzzer already did this for crashes.
DumpCoverage calls a sancov specific function, you'll probably need to add the clang-cov specific one. @Dor1s might know the name of this.

Dor1s:
I think you're looking for https://source.chromium.org/chromium/chromium/src/+/master:base/test/clang_profiling.cc

inferno-chromium:
Clang code coverage profraw is not generated when run_coverage fails on timeout.

inferno-chromium:
https://pantheon.corp.google.com/logs/viewer?resource=global&logName=projects%2Ffuzzbench%2Flogs%2Ferrors&expandAll=false&timestamp=2020-07-28T14:35:16.000000000Z&project=fuzzbench&src=ac&minLogLevel=0&customFacets=&limitCustomFacetWidth=true&advancedFilter=severity%3E%3DERROR&interval=JUMP_TO_TIME&scrollTimestamp=2020-07-28T10:22:10.004998190Z

inferno-chromium:
```
 jsonPayload: {
  component: "dispatcher"   
  coverage_binary: "/work/coverage-binaries/sqlite3_ossfuzz/ossfuzz"   
  experiment: "2020-07-27"   
  instance_name: "d-2020-07-27"   
  message: "Coverage run failed."   
  output: "WARNING: Failed to find function "__sanitizer_acquire_crash_state".
WARNING: Failed to find function "__sanitizer_print_stack_trace".
WARNING: Failed to find function "__sanitizer_set_death_callback".
INFO: Seed: 892174498
MERGE-OUTER: 296 files, 0 in the initial corpus, 0 processed earlier
MERGE-OUTER: attempt 1
WARNING: Failed to find function "__sanitizer_acquire_crash_state".
WARNING: Failed to find function "__sanitizer_print_stack_trace".
WARNING: Failed to find function "__sanitizer_set_death_callback".
INFO: Seed: 892267735
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 1048576 bytes
MERGE-INNER: using the control file '/tmp/libFuzzerTemp.Merge3332240.txt'
[...]

PR #602, opened 2020-07-28
Dump the coverage report even if it times out.

Shadoom7:
The current case is that clang source-based coverage measurer will dump an empty .profraw file if the measuring times out.
This patch will fix it.
Related to #547

### DIFF
--- a/fuzzers/coverage/builder.Dockerfile
+++ b/fuzzers/coverage/builder.Dockerfile
@@ -15,6 +15,7 @@
 ARG parent_image=gcr.io/fuzzbench/base-builder
 FROM $parent_image
 
+# The patch adds hook to dump clang coverage data when timeout.
 COPY patch.diff /
 
 # Use a libFuzzer version that supports clang source-based coverage.
--- a/fuzzers/coverage/builder.Dockerfile
+++ b/fuzzers/coverage/builder.Dockerfile
@@ -15,12 +15,13 @@
 ARG parent_image=gcr.io/fuzzbench/base-builder
 FROM $parent_image
 
+COPY patch.diff /
+
 # Use a libFuzzer version that supports clang source-based coverage.
 RUN git clone https://github.com/llvm/llvm-project.git /llvm-project && \
-    cd /llvm-project/ && \
-    git checkout d8981ce5b9f8caa567613b2bf5aa3095e0156130 && \
-    cd compiler-rt/lib/fuzzer && \
-    (for f in *.cpp; do \
-      clang++ -stdlib=libc++ -fPIC -O2 -std=c++11 $f -c & \
-    done && wait) && \
-    ar r /usr/lib/libFuzzer.a *.o
+    cd /llvm-project && \
+    git checkout 0b5e6b11c358e704384520dc036eddb5da1c68bf && \
+    patch -p1 < /patch.diff && \
+    cd /llvm-project/compiler-rt/lib/fuzzer && \
+    bash build.sh && \
+    cp libFuzzer.a /usr/lib
--- a/fuzzers/coverage/patch.diff
+++ b/fuzzers/coverage/patch.diff
@@ -1,0 +1,55 @@
+diff --git a/compiler-rt/lib/fuzzer/FuzzerDriver.cpp b/compiler-rt/lib/fuzzer/FuzzerDriver.cpp
+index 306e644277c..26db440ecd7 100644
+--- a/compiler-rt/lib/fuzzer/FuzzerDriver.cpp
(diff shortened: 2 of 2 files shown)


fuzzers/coverage/builder.Dockerfile
```diff
@@ -15,12 +15,13 @@
 ARG parent_image=gcr.io/fuzzbench/base-builder
 FROM $parent_image
 
+COPY patch.diff /
```
inferno-chromium, line 18: Add patch.diff file

Shadoom7: Done it.

fuzzers/coverage/patch.diff
```diff
@@ -0,0 +1,55 @@
...
+diff --git a/compiler-rt/lib/fuzzer/FuzzerExtFunctions.def b/compiler-rt/lib/fuzzer/FuzzerExtFunctions.def
+index bb15d68e0de..154b6239d53 100644
+--- a/compiler-rt/lib/fuzzer/FuzzerExtFunctions.def
++++ b/compiler-rt/lib/fuzzer/FuzzerExtFunctions.def
+@@ -48,3 +48,6 @@ EXT_FUNC(__sanitizer_dump_coverage, void, (const uintptr_t *, uintptr_t),
+ EXT_FUNC(__msan_scoped_disable_interceptor_checks, void, (), false);
+ EXT_FUNC(__msan_scoped_enable_interceptor_checks, void, (), false);
+ EXT_FUNC(__msan_unpoison, void, (const volatile void *, size_t size), false);
++
++//Clang source-based coverage functions
```
inferno-chromium, line 23: nit: space between // and Clang, end comment with period.

Shadoom7:
This version of libfuzzer along with the patch.diff works fine in my local environment for a small testing fuzz-target.
Currently testing it in real experiment.

fuzzers/coverage/builder.Dockerfile
inferno-chromium, line 18: Add comment that it adds hook to dump clang coverage data at crash.

inferno-chromium:
Patch looks good, adding @Dor1s @jonathanmetzman as well for quick review.

fuzzers/coverage/patch.diff
```diff
@@ -0,0 +1,55 @@
+diff --git a/compiler-rt/lib/fuzzer/FuzzerDriver.cpp b/compiler-rt/lib/fuzzer/FuzzerDriver.cpp
+index 306e644277c..26db440ecd7 100644
+--- a/compiler-rt/lib/fuzzer/FuzzerDriver.cpp
++++ b/compiler-rt/lib/fuzzer/FuzzerDriver.cpp
+@@ -737,6 +737,8 @@ int FuzzerDriver(int *argc, char ***argv, UserCallback Callback) {
+       F->SetMaxInputLen(kDefaultMaxMergeLen);
+     assert(Flags.merge_control_file);
+     F->CrashResistantMergeInternalStep(Flags.merge_control_file);
++    if (Options.DumpCoverage)
```
Dor1s, line 9: I'm not sure I understand this. If timeout happens inside `CrashResistantMergeInternalStep` and it crashes, coverage data inside if will be lost. Did you try to convert `.profraw` that you're getting locally into `.profdata`? Do they have non-zero counters? The best way to verify that is to check the number of execution on `LLVMFuzzerTestOneInput`

Shadoom7: Sorry, I'm not very familiar with the llvm-project structure. I will dig deeper to see what will happen if it times out inside CrashResistantMergeInternalStep. I did the DumpCoverage here because I saw @jonathanmetzman did the same for sanitizer coverage. And the .profdata works fine. It has non-zero counters and execution number on LLVMFuzzerTestOneInput is correct.

Shadoom7:
@inferno-chromium I did a tiny experiment with jasper in it.
Every measuring on jasper times out but there is no more 'empty profraw' warnings
[logs](https://pantheon.corp.google.com/logs/viewer?organizationId=433637338589&src=ac&project=fuzzbench-test&minLogLevel=0&expandAll=false&timestamp=2020-07-29T00:13:08.078000000Z&customFacets=&limitCustomFacetWidth=true&interval=NO_LIMIT&resource=global&scrollTimestamp=2020-07-28T21:14:02.599109605Z&advancedFilter=resource.type%3D%22global%22%0AjsonPayload.experiment%3D%22dump-test-1%22%0Aseverity%3E%3D%22ERROR%22) here.

PR #603, opened 2020-07-28
Increase per-unit coverage timeout from 5 to 10 sec.

inferno-chromium:


inferno-chromium:
Addresses https://github.com/google/fuzzbench/issues/547#issuecomment-665294785
if this is not sufficient, then will increase to 15 sec.

PR #607, opened 2020-07-29
Simplify standard benchmarks, remove common.sh dependency.

### LINKED ISSUE #583: Get rid of common.sh usage in build.sh files of standard benchmarks
I don't plan on fixing this in #584 but getting rid of this is high priority. It's leftover from the old kind of standard benchmarks and is one of two differences between OSS-Fuzz/standard benchmarks that exist at build time.

inferno-chromium:
Fixes #583


jonathanmetzman:
Fixes https://github.com/google/fuzzbench/issues/583

PR #608, opened 2020-07-29
Add new fuzzer configs for libFuzzer interceptor evaluation.

dokyungs:
This patch adds [entropic|libfuzzer]_interceptors fuzzer configs that use the latest interceptor support in upstream LLVM.

With this patch, existing entropic and libfuzzer fuzzer configs also use a recent version of libFuzzer in LLVM upstream, but slightly older version than entropic_interceptors and libfuzzer_interceptors, which does not have interceptor support.

Separating configs and updating LLVM versions of existing fuzzer configs is an effort to evaluate interceptors, while minimizing unrelated factors potentially in play.


dokyungs:
@morehouse @hctim Could you guys please also take a look at this patch, which is an effort to evaluate libFuzzer interceptors more accurately?

jonathanmetzman:
Do you want committer access btw?

jonathanmetzman:
Btw if you are sure the new variants pass CI if the originals do then this is fine, otherwise, it's better to add the fuzzers to .github/workflows/fuzzer.yml to ensure this.

dokyungs:
> Do you want committer access btw?

Yes, I think that would be helpful.

dokyungs:
> Btw if you are sure the new variants pass CI if the originals do then this is fine, otherwise, it's better to add the fuzzers to .github/workflows/fuzzer.yml to ensure this.

I think it's better to run CI also for the new fuzzer variants, since they use different versions of llvm-project/compiler-rt and there's a chance something may go wrong there. I added the new variants to CI just now.

jonathanmetzman:
> > Do you want committer access btw?
> 
> Yes, I think that would be helpful.

Still get maintainer review if you can. But I think this will come in handy in situations when I approve PR but don't merge because I'm waiting for CI. You can just merge yourself when it finishes.

fuzzers/libfuzzer_interceptors/fuzzer.py
```diff
@@ -0,0 +1,86 @@
...
+        '-print_final_stats=1',
+        # `close_fd_mask` to prevent too much logging output from the target.
+        '-close_fd_mask=3',
+        # Run in fork mode to allow ignoring ooms, timeouts, crashes and
+        # continue fuzzing indefinitely.
+        '-fork=1',
+        '-ignore_ooms=1',
+        '-ignore_timeouts=1',
+        '-ignore_crashes=1',
+
```
jonathanmetzman, line 72: Not relevant to this PR in particular but there is no timeout here since I prefer defaults are used. But libFuzzer's default here probably hurts it (1-hour) I think libFuzzer's default timeout should be lowered. Especially for fork mode where it doesn't mean the end of fuzzing. I assume the generous timeout was a result of libFuzzer's single threaded days.

dokyungs:
> > > Do you want committer access btw?
> > 
> > 
> > Yes, I think that would be helpful.
> 
> Done.
> Still get maintainer review if you can. But I think this will come in handy in situations when I approve PR but don't merge because I'm waiting for CI. You can just merge yourself when it finishes.

Thanks! Sounds good. I will make sure to get future PRs approved and to wait for CI before committing them.

fuzzers/libfuzzer_interceptors/fuzzer.py
```diff
@@ -0,0 +1,86 @@
...
+    utils.build_benchmark()
+
+
+def fuzz(input_corpus, output_corpus, target_binary):
+    """Run fuzzer. Wrapper that uses the defaults when calling
+    run_fuzzer."""
+    run_fuzzer(input_corpus, output_corpus, target_binary)
+
+
+def run_fuzzer(input_corpus, output_corpus, target_binary, extra_flags=None):
```
inferno-chromium, line 57: Why is this duplicated, please use libfuzzer.run_fuzzer ?

dokyungs: Done.

inferno-chromium: No need of two function fuzz and run_fuzzer, fuzz can directly call libfuzzer_fuzzer.run_fuzzer

inferno-chromium:
Skip the last comment since this is temporary fuzzer anyway.

inferno-chromium:
Please fix the presubmit failure, then we can merge

```
source .venv/bin/activate && python3 presubmit.py
************* Module fuzzer
fuzzers/libfuzzer_interceptors/fuzzer.py:16:0: W0611: Unused import subprocess (unused-import)
```

dokyungs:
> Please fix the presubmit failure, then we can merge
> 
> ```
> source .venv/bin/activate && python3 presubmit.py
> ************* Module fuzzer
> fuzzers/libfuzzer_interceptors/fuzzer.py:16:0: W0611: Unused import subprocess (unused-import)
> ```

Fixed now!

jonathanmetzman:
You may want to request experiment right?

dokyungs:
@inferno-chromium Thanks for merging! I was gonna merge too.

> You may want to request experiment right?

Oh yes, but they are not urgent. I will request an experiment tomorrow so that these new variants are included in tomorrow's scheduled run.

Issue #609, opened 2020-07-29
Move all git clones in build.sh for any benchmark to its Dockerfile


inferno-chromium:
This would reduce download resources and speed up building. This would also match OSS-Fuzz benchmarks.

PR #610, opened 2020-07-29
Move git cloning from benchmark's build.sh to Dockerfile.

inferno-chromium:
Fixes https://github.com/google/fuzzbench/issues/609

### DIFF
--- a/benchmarks/freetype2-2017/Dockerfile
+++ b/benchmarks/freetype2-2017/Dockerfile
@@ -24,4 +24,7 @@ RUN apt-get update &&  \
     libarchive-dev \
     zlib1g-dev
 
+RUN git clone git://git.sv.nongnu.org/freetype/freetype2.git
+RUN git clone https://github.com/unicode-org/text-rendering-tests.git TRT
+
 COPY * $SRC/
--- a/benchmarks/libpng-1.2.56/build.sh
+++ b/benchmarks/libpng-1.2.56/build.sh
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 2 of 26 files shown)


benchmarks/libpng-1.2.56/build.sh
```diff
@@ -23,5 +22,3 @@ make -j $(nproc)
 $CXX $CXXFLAGS -std=c++11 $SRC/target.cc .libs/libpng12.a $FUZZER_LIB -I . -lz \
     -o $OUT/fuzz-target
 cp -r /opt/seeds $OUT/
-wget -qO fuzz-target.dict \
```
inferno-chromium, line 26: This looks like a bug, wrong dict path :( yikes!

benchmarks/freetype2-2017/Dockerfile
```diff
@@ -24,4 +24,7 @@ RUN apt-get update &&  \
     libarchive-dev \
     zlib1g-dev
 
+RUN git clone git://git.sv.nongnu.org/freetype/freetype2.git
+RUN git clone https://github.com/unicode-org/text-rendering-tests.git TRT
+
```
lszekeres, line 29: Consider also moving the `git checkout <version we want>` commands right after the corresponding clone-s.

(later we might also consider doing the clone/checkout in a single RUN step with pruning the repo to save space)

inferno-chromium: Yes let me file a bug for this, both of these can be done in one go. This one saves a whole bunch of network bandwidth resources which we are wasting atm.

inferno-chromium: Filed https://github.com/google/fuzzbench/issues/611.

PR #612, opened 2020-07-30
[GCB] Fix parent_image outside of FuzzBench cloud project

jonathanmetzman:
GCB was broken on non-FuzzBench cloud projects because experiment
tagging meant that certain images were no longer tagged the way
they were passed as parent_image.


jonathanmetzman:
> How about
> docker/generate_makefile.py: --build-arg parent_image={base_tag}/builders/{fuzzer}/{benchmark}-intermediate \
> 
> getting really tired of these differences.

We don't have the crazy tagging situation in local builds (yet?) so it doesn't matter there.

PR #613, opened 2020-07-30
Request an experiment

dokyungs:
Requesting an experiment to evaluate the effectiveness of libFuzzer interceptors.


dokyungs:
> Frankly I don't thin you need afl and honggfuzz, since merging should take care of taking the latest runs from these fuzzers and combining with data from this report.

Gotcha. I removed afl/honggfuzz just now.

PR #614, opened 2020-07-31
Add temporary libfuzzer/entropic variants to evaluate the impact of keeping all seed inputs

dokyungs:
This patch adds temporary `libfuzzer` and `entropic` variants, namely `libfuzzer_keepseed` and `entropic_keepseed`, to evaluate the impact of keeping all seed inputs regardless of whether they find new coverage or not.

This MAY help SQLITE3 benchmarks where the seed inputs are not sound w.r.t. the input language that the fuzzer executable expects. Such seed inputs are likely discarded by libFuzzer as ill-formed inputs are rejected earlier by the fuzzer executable, not finding new coverage. However, they may still contain useful fragments, and thus be worth mutating or spliced with others.

The downside of this is that the corpus size (and likely individual input size too) grows, which may negatively impact the performance. A quick 30-min local testing says that fuzzing unfortunately does get slowed down, but the coverage may increase with Entropic schedule (but not with libFuzzer's vanilla schedule). I would like to see, with this patch and experiment request, how keeping all seed inputs affects libfuzzer and entropic's performance on a larger scale and with statistic significance both on SQLITE3 and other benchmarks.

### DIFF
--- a/service/experiment-requests.yaml
+++ b/service/experiment-requests.yaml
@@ -19,6 +19,28 @@
 # are still testing this feature. You should request an experiment by contacting
 # us as you normally do.
 
+- experiment: 2020-08-01
+  fuzzers:
+    - afl
+    - aflfast
+    - aflplusplus
+    - aflplusplus_optimal
+    - aflplusplus_qemu
+    - afl_qemu
+    - aflsmart
+    - eclipser
+    - entropic
+    - fairfuzz
+    - fastcgs_lm
+    - honggfuzz
+    - honggfuzz_qemu
+    - lafintel
+    - libfuzzer
+    - manul
+    - mopt
+    - libfuzzer_keepseed
+    - entropic_keepseed
+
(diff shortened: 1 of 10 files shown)


dokyungs:
@morehouse @hctim Could you guys also take a look at this patch, and see whether running this experiment makes sense?

service/experiment-requests.yaml
```diff
@@ -19,6 +19,11 @@
 # are still testing this feature. You should request an experiment by contacting
 # us as you normally do.
 
+- experiment: 2020-07-31
```
jonathanmetzman, line 22: When did you merge the last experiment request? It's probably better to combine them.

dokyungs: Last one was requested yesterday, and it's already running.

jonathanmetzman: Nevermind then!

PR #616, opened 2020-08-01
[DO NOT MERGE] Proof-of-concept report generation webapp

jonathanmetzman:
This is a webapp that allows users to specify fuzzers from specific experiments and will generate reports based on the selection.
webapp-demo

If I add boxes to select all fuzzers in an experiment I think this is close to the ideal interface for generating reports for users (only downside is some slowness in waiting for reports to generate).

Some not so ideal things that could prevent us from deploying this:
1. Do we want to deal with auth or should we allow any rando on the internet to use this? Maybe limiting to people with gmail accounts is sufficient to prevent abuse?
2. Writing will be hard. Appengine doesn't allow writing to disk but graph generation assumes writing to disk. We could just deploy this on a single server and periodically delete generated reports, but this feels fragile and not really up to best standards for webapps. We could also teach generate_report to write directly to GCS and use appengine. Or we could not use appengine and just copy the report to GCS after generation. I'm tempted to go with this solution.

I think the interface for selecting experiment-fuzzer pairs is ideal in this solution. But as an alternative we could tell people to run this webapp themselves and have the webapp download a public archive of the DB.

In any case, I think this is a good starting point for a discussion. I don't like the current solution and I don't love the proposed improvements. They seem more cumbersome and error prone than this interface (where if someone makes a mistake they can just regenerate the report). This solution also has the advantage of not changing reports. Thus the reports we publish on fuzzbench.com/reports/ will always contain just what was run in the experiment.

### DIFF
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -72,6 +72,27 @@ def clobber_experiments_data(df, experiments):
     return result
 
 
+def drop_private(experiment_df):
+    """Drop all snapshots from private experiments."""
+    return experiment_df[~experiment_df['private']]
+
+
+def filter_fuzzers_from_experiments(experiment_df, fuzzers_from_experiments):
+    """Drop snapshots whose fuzzer and experiment combination are not in
+    |fuzzers_from_experiments| from experiment_df and returns the result.
+    fuzzers_from_experiments is a list of tuples containing a fuzzer and an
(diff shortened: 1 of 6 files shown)


jonathanmetzman:
Changed it so that fuzzers not run an in an experiment can't be checked
disable

PR #617, opened 2020-08-03
Fix for dispatcher oooming out.

inferno-chromium:
Dispatcher docker oomed out on 2020-07-30, and some coverage
timeouts are also noticed on 2020-08-01. Trying n1-highmem-96
till we break measurer.


inferno-chromium:
TBR

PR #618, opened 2020-08-03
Generate coverage reports.

Shadoom7:
This CL generates coverage reports for each pair of benchmark and fuzzer. 
A demo report is [here](https://storage.googleapis.com/my-fuzzbench-reports/gen-cov-14/index.html).

Further development: need to do some calculation on profdata files to generate the coverage reports that shows the unique code covered by each fuzzer.


analysis/report_templates/default.html
```diff
@@ -232,6 +232,27 @@ <h5 class="center-align">Mann-Whitney U test</h4>
                 </li>
             </ul>
 
+            <div class="row">
+                <div class="col s12">
+                    <div class="card white darken-1">
+                        <div class="card-content black-text">
+                        <span class="card-title">Coverage reports</span>
```
inferno-chromium, line 239: UI needs to be changed. We need to have a drop down similar to "Sample statistics and statistical significance"
Make it like "Coverage reports" and when someone clicks it, have a list of buttons for fuzzers (no dropdown, why extra click).

analysis/report_utils.py
```diff
@@ -0,0 +1,115 @@
...
+from common import new_process
+from common import benchmark_utils
+from common import logs
+from experiment.build import build_utils
+from experiment import measurer
+
+logger = logs.Logger('reporter')  # pylint: disable=invalid-name
+
+
+def get_profdata_file_path(fuzzer, benchmark, trial_id):
```
inferno-chromium, line 30: This is all wrong place for this code, analysis only contains analysis and db query code. This should all be in some common/coverage_utils file.

docker/benchmark-builder/Dockerfile
```diff
@@ -49,3 +49,7 @@ RUN export CHECKOUT_COMMIT=$(cat /benchmark.yaml | tr -d ' ' | grep 'commit:' |
     python3 -u /opt/fuzzbench/checkout_commit.py $CHECKOUT_COMMIT $SRC
 
 RUN PYTHONPATH=$SRC python3 -u -c "from fuzzers import utils; utils.initialize_flags(); import fuzzer; fuzzer.build()"
+
+# Copy source code files of benchmark to $OUT for the report generation.
+RUN mkdir $OUT/src-files && \
+    cp -rL --parent $SRC $OUT/src-files
```
inferno-chromium, line 55: newline

analysis/report_utils.py
jonathanmetzman: I don't think putting it in common makes sense. Why not in experiment?

inferno-chromium: yes experiment seems fine too, just not analysis

analysis/generate_report.py
```diff
@@ -167,7 +170,29 @@ def generate_report(experiment_names,
...
+                                                coverage_report_directory)
+                report_utils.get_profdata_files(experiment_names[0],
+                                                benchmark_names, fuzzer_names,
+                                                coverage_report_directory)
+
+            # Generate coverage reports for each benchmark.
+            report_utils.generate_cov_reports(benchmark_names, fuzzer_names,
+                                              coverage_report_directory)
+        except Exception:  # pylint: disable=broad-except
+            logger.error('Failed when generating coverage reports.')
```
lszekeres: Currently, the input of `generate_report.py` is simply either a database to query or a CSV file (for custom report generation locally). With these change we seem to require binaries, source files, and profdata files as well as input. In order to keep our tools simpler, would it make sense to have a separate tool (eg `generate_clang_coverage_reports.py` for creating the clang html coverage reports? 
- This tool could perhaps take an optional flag pointing to the directory containing those reports to link them (if provided). 
- Alternatively, perhaps it could take exported coverage data in JSON (https://clang.llvm.org/docs/SourceBasedCodeCoverage.html#exporting-coverage-data) output by `generate_clang_coverage_reports.py`? 
- Later, the most ideal solution might be to save the clang coverage data into the DB (i.e., turn the structured JSON data into DB data). That way `generate_report.py` could still simply either take a database to query or a CSV file.

Wdyt?

Shadoom7: I think it's a good idea. We can also keep the front page simple by moving those differential graphs and reports to another html.
On the other hand, we couldn't avoid using the binaries, source files and profdata files in any case, since we need them to generate the coverage report. The json file can only provide information about covered regions. I generated the json files in the end of measurer and use them to get the differential graphs, this is in another CL.

inferno-chromium: And yes, how long this clang code coverage generation takes, since locally, it might take hours ?

docker/benchmark-builder/Dockerfile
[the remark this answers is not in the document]
jonathanmetzman: Does this need to be done in benchmark-builder? Why can't it be done in fuzzers/coverage/Dockerfile?

experiment/coverage_utils.py
```diff
@@ -0,0 +1,115 @@
...
+    """Fetch source files to |report_dir|."""
+    for benchmark in benchmarks:
+        coverage_binaries_dir = build_utils.get_coverage_binaries_dir()
+        src_path = os.path.join(coverage_binaries_dir, benchmark, 'src-files')
+        dst_dir = os.path.join(report_dir, benchmark, 'src-files')
+        filesystem.copytree(src_path, dst_dir)
+
+
+def get_profdata_files(experiment, benchmarks, fuzzers, report_dir):
+    """Get profdata files to |report_dir|."""
```
jonathanmetzman, line 59: I don't understand this comment.

experiment/coverage_utils.py
```diff
@@ -0,0 +1,115 @@
...
+                for trial_id in trial_ids
+            ]
+            dst_dir = os.path.join(report_dir, benchmark, fuzzer)
+            filesystem.create_directory(dst_dir)
+            dst_file = os.path.join(dst_dir, 'merged.profdata')
+            merge_profdata_files(files_to_merge, dst_file)
+
+
+def fetch_binary_files(benchmarks, report_dir):
+    """Get binary file for each benchmark."""
```
jonathanmetzman, line 74: Comment needs to be in descriptive form "gets" not imperative "get".
And I don't think the comment correctly describes the function.

experiment/coverage_utils.py
```diff
@@ -0,0 +1,115 @@
...
+    for benchmark in benchmarks:
+        src_file = measurer.get_coverage_binary(benchmark)
+        dst_dir = os.path.join(report_dir, benchmark)
+        filesystem.copy(src_file, dst_dir)
```
jonathanmetzman, line 78: Why do we copy the binaries into the report directory?

experiment/coverage_utils.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def generate_cov_reports(benchmarks, fuzzers, report_dir):
+    """Generate coverage reports for each benchmark and fuzzer."""
```
jonathanmetzman, line 82: "generates"

experiment/coverage_utils.py
```diff
@@ -0,0 +1,115 @@
...
+    logger.info('Start generating coverage report for benchmarks.')
+    with multiprocessing.Pool() as pool:
+        generate_cov_report_args = [(benchmark, fuzzer, report_dir)
+                                    for benchmark in benchmarks
+                                    for fuzzer in fuzzers]
+        pool.starmap_async(generate_cov_report, generate_cov_report_args)
```
jonathanmetzman, line 88: WHy do you use async?

experiment/coverage_utils.py
```diff
@@ -0,0 +1,115 @@
...
+        pool.close()
+        pool.join()
+    logger.info('Finished generating coverage report.')
+
+
+def generate_cov_report(benchmark, fuzzer, report_dir):
+    """Generate the coverage report for one pair of benchmark and fuzzer."""
+    logger.info('Generating coverage report for benchmark:{benchmark} \
```
jonathanmetzman, line 96: We don't use backslashes. Use implicit string joining.

analysis/report_templates/default.html
```diff
@@ -232,6 +232,27 @@ <h5 class="center-align">Mann-Whitney U test</h4>
                 </li>
             </ul>
 
+            <div class="row">
```
jonathanmetzman, line 235: Please link to a report in gcs and les us see it.

analysis/generate_report.py
```diff
@@ -167,7 +170,29 @@ def generate_report(experiment_names,
         experiment_df = data_utils.clobber_experiments_data(
             experiment_df, experiment_names)
 
+    # Generate coverge reports for each benchmark.
+    benchmark_names = experiment_df.benchmark.unique()
     fuzzer_names = experiment_df.fuzzer.unique()
+    coverage_report_directory = os.path.join(report_directory,
+                                             'coverage-reports')
+    if not in_progress:
```
jonathanmetzman, line 178: I don't see why we should use `in_progress` for this. Sure we don't actually need it while it's in progress but this feels like exposing details of the dispatcher to report generation that shouldn't be exposed. There should probably be a seperate variable.

experiment/coverage_utils.py
```diff
@@ -0,0 +1,115 @@
...
+    result = new_process.execute(command, expect_zero=False)
+    if result.retcode != 0:
+        logger.error('Profdata files merging failed.')
+
+
+def fetch_source_files(benchmarks, report_dir):
+    """Fetch source files to |report_dir|."""
+    for benchmark in benchmarks:
+        coverage_binaries_dir = build_utils.get_coverage_binaries_dir()
+        src_path = os.path.join(coverage_binaries_dir, benchmark, 'src-files')
```
jonathanmetzman, line 53: If you're going to reuse a string like this please make it a constant so that we can catch any typos in linting.

analysis/generate_report.py
```diff
@@ -167,7 +170,29 @@ def generate_report(experiment_names,
...
+        logger.info('Generating coverage reports.')
+        try:
+            if not from_cached_data:
```
jonathanmetzman, line 181: WHy not do this if form_cached_data?

experiment/coverage_utils.py
```diff
@@ -0,0 +1,115 @@
...
+
+def generate_cov_report(benchmark, fuzzer, report_dir):
+    """Generate the coverage report for one pair of benchmark and fuzzer."""
+    logger.info('Generating coverage report for benchmark:{benchmark} \
+                fuzzer:{fuzzer}.'.format(benchmark=benchmark, fuzzer=fuzzer))
+    dst_dir = os.path.join(report_dir, benchmark, fuzzer)
+    fuzz_target = benchmark_utils.get_fuzz_target(benchmark)
+    profdata_file_path = os.path.join(dst_dir, 'merged.profdata')
+    binary_file_path = os.path.join(report_dir, benchmark, fuzz_target)
+    source_file_path_prefix = os.path.join(report_dir, benchmark, 'src-files')
```
jonathanmetzman, line 102: This module seems to reimplement a lot of the same logic multiple times. Instead you should have helper functions for things like getting the path to source files for a particular benchmark so there can be one source of truth.

analysis/generate_report.py
jonathanmetzman: I think coverage reports which need binaries can go in their own module.
But I think producing metrics on coverage, like the uniqueness of each fuzzer's coverage should go in generate_report.py

Shadoom7: > And yes, how long this clang code coverage generation takes, since locally, it might take hours ?

I'm not sure how long it's gonna take to do a coverage generation for a full experiment locally, since I haven't ran a full experiment to get all the source files I need to do this. But I don't think it's going to take hours. I'm using multi-process and even if I don't, it takes approximately 1 sec to generate the report for one pair of fuzzer and benchmark. So I'd say 10 mins or less.

Shadoom7: > I think coverage reports which need binaries can go in their own module.
> But I think producing metrics on coverage, like the uniqueness of each fuzzer's coverage should go in generate_report.py

Yeah, I agree.

inferno-chromium: So, generate_report.py is a standalone script that anyone can execute locally to generate full report. With this code, that wont be possible, as it relies on measuring files to exist in specific locations of disk. I think this needs to change, we need to put stuff in DB in last cycle ? So that this generate_report.py works quickly as expected. @jonathanmetzman @lszekeres - thoughts ?

jonathanmetzman: Putting things in DB would be ideal but I have no idea if the data will be too large. @Shadoom7 it would be nice to have an estimate on how much it would cost to store the regions covered by each trial on the last cycle in the DB (or maybe even fuzzer-benchmark pair to save space).
I totally agree that specific locations on disk is a bad way to do this. I think a reasonable way to do it that wouldn't risk exploding the db is to store info about the GCS bucket in the db (on the experiment entity). Then generate_report can download the data it needs from there.

lszekeres: Yes, having the data in the DB would be most ideal, but even before we get there, we could do a simpler to implement alternative mentioned on the top of this thread. Not sure that comment was clear. What i meant was that we could have two separate tools: 1) the existing `generate_report.py`, and 2) a new `generate_clang_coverage_reports.py` that would have this new code this comment thread is on.
- `generate_clang_coverage_reports.py` could output the HTML coverage reports and the JSON coverage data files under some `reports/<experiment>/clang-coverage` directory, provided by a flag.
- `generate_report.py` would take this via an optional --clang-coverage-directory flag, and if provided, then the main report would link to the HTML coverage reports (just like in @Shadoom7's demo) and would create the differential plot.

To answer comments above directly:

> We can also keep the front page simple by moving those differential graphs and reports to another html.

There wouldn't be need for another HTML in this case

> On the other hand, we couldn't avoid using the binaries, source files and profdata files in any case, since we need them to generate the coverage report. 

Yes, the `generate_clang_coverage_reports.py` tool would use these, while `generate_report.py` would use the HTML reports and JSON files output by the former tool.

Having two tools will keep things simpler and will give us more flexibility, even before we put the clang cov data in the db.

Shadoom7: I think it is still convenient for users to generate report locally, since it syncs all the needed file to a single folder in gcs.
Currently, to generate the report locally. We need to download `gs://my-fuzzbench-reports/<experiment name>/data.csv.gz` to the local `report-dir` and then run generate_report.
With this patch, the only difference is that we need to download `gs://my-fuzzbench-reports/<experiment name>/coverage`, which contains all the source files and binaries.

analysis/report_templates/default.html
lszekeres: +1 on having a simple list, instead of a the current dropdown "fuzzer list" button

analysis/generate_report.py
Shadoom7: @lszekeres Oh I got you. That makes sense.

analysis/generate_coverage_report.py
```diff
@@ -0,0 +1,39 @@
...
+    """Generates coverage reports."""
+    logger.info('Generating coverage reports.')
+    coverage_report_directory = os.path.join(report_directory,
+                                             'coverage-reports')
+    filesystem.create_directory(coverage_report_directory)
+    try:
+        # Generate coverage reports for each benchmark.
+        coverage_utils.generate_cov_reports(experiment_names, benchmarks,
+                                            fuzzers, coverage_report_directory)
+    except Exception:  # pylint: disable=broad-except
```
inferno-chromium, line 38: Why broad exception, dont use unless needed.

analysis/generate_report.py
```diff
@@ -74,6 +77,11 @@ def get_arg_parser():
                         '--fuzzers',
                         nargs='*',
                         help='Names of the fuzzers to include in the report.')
+    parser.add_argument('-clang',
```
inferno-chromium, line 80: -cov

analysis/generate_report.py
```diff
@@ -74,6 +77,11 @@ def get_arg_parser():
...
+                        '--clang-coverage-reports',
```
inferno-chromium, line 81: --coverage-report without s

analysis/generate_report.py
```diff
@@ -167,14 +176,23 @@ def generate_report(experiment_names,
         experiment_df = data_utils.clobber_experiments_data(
             experiment_df, experiment_names)
 
+    # Generate coverge reports for each benchmark.
```
inferno-chromium, line 179: coverage

analysis/rendering.py
```diff
@@ -38,4 +39,5 @@ def render_report(experiment_results, template, in_progress):
     )
     template = environment.get_template(template)
     return template.render(experiment=experiment_results,
-                           in_progress=in_progress)
+                           in_progress=in_progress,
+                           clang_coverage_reports=clang_coverage_reports)
```
inferno-chromium, line 43: coverage_report

experiment/coverage_utils.py
```diff
@@ -0,0 +1,206 @@
...
+
+def generate_cov_report(experiments, benchmark, fuzzer, report_dir):
+    """Generate the coverage report for one pair of benchmark and fuzzer."""
+    logs.initialize()
+    logger.info('Generating coverage report for benchmark: {benchmark} \
+                fuzzer: {fuzzer}.'.format(benchmark=benchmark, fuzzer=fuzzer))
+    generator = CoverageReporter(fuzzer, benchmark, experiments, report_dir,
+                                 logger)
+    # Gets and merges all the profdata files.
+    generator.fetch_profdata_file()
```
inferno-chromium, line 55: fetch_profdata_files

experiment/coverage_utils.py
```diff
@@ -0,0 +1,206 @@
...
+    generator.generate_cov_report()
+
+    logger.info('Finished generating coverage report for '
+                'benchmark:{benchmark} fuzzer:{fuzzer}.'.format(
+                    benchmark=benchmark, fuzzer=fuzzer))
+
+
+def set_up_coverage_files(experiment_names, report_dir, benchmarks):
+    """Sets up coverage files for all benchmarks."""
+    with multiprocessing.Pool() as pool:
```
inferno-chromium, line 67: No need of multiprocessing here at all, these gcs file cp is fast, remove complexity.

experiment/coverage_utils.py
```diff
@@ -0,0 +1,206 @@
...
+                                               self.benchmark)
+            files_to_merge.extend([
+                self.get_profdata_file_path(experiment, trial_id)
+                for trial_id in trial_ids
+            ])
+        filesystem.create_directory(self.fuzzer_report_dir)
+        for file_path in files_to_merge:
+            filestore_utils.cp(file_path,
+                               self.fuzzer_report_dir,
+                               expect_zero=False)
```
inferno-chromium, line 175: Why expect_zero, lets not abuse this. These should all succeed.

experiment/coverage_utils.py
```diff
@@ -0,0 +1,206 @@
...
+        self.benchmark_report_dir = os.path.join(self.report_dir, benchmark)
+        self.fuzzer_report_dir = os.path.join(self.benchmark_report_dir, fuzzer)
+        self.merged_profdata_file = os.path.join(self.fuzzer_report_dir,
+                                                 'merged.profdata')
+        self.source_files = os.path.join(self.benchmark_report_dir, 'src-files')
+        fuzz_target = benchmark_utils.get_fuzz_target(self.benchmark)
+        self.binary_file = fuzzer_utils.get_fuzz_target_binary(
+            self.benchmark_report_dir, fuzz_target_name=fuzz_target)
+
+    def merge_profdata_files(self):
```
inferno-chromium, line 145: This function feels duplicated in measurer, why not share code as part of cl.

experiment/coverage_utils.py
```diff
@@ -0,0 +1,206 @@
...
+        experiment_filestore_dir = get_experiment_filestore_path(experiment)
+        profdata_file_path = posixpath.join(
+            experiment_filestore_dir, 'experiment-folders',
+            benchmark_fuzzer_trial_dir,
+            'data-{id}.profdata'.format(id=trial_id))
+        return profdata_file_path
+
+    def generate_cov_report(self):
+        """Generates the coverage report."""
+        dst_dir = self.fuzzer_report_dir
```
inferno-chromium, line 190: no need of these local vars. if needed, better name the self. vars.

experiment/coverage_utils.py
```diff
@@ -0,0 +1,206 @@
...
+        profdata_file_path = self.merged_profdata_file
+        binary_file_path = self.binary_file
+        source_file_path_prefix = self.source_files
+        command = [
+            'llvm-cov-11', 'show', '-format=html',
```
inferno-chromium, line 195: Why llvm-cov-11, why not just llvm-cov

experiment/dispatcher.py
```diff
@@ -145,7 +145,10 @@ def dispatcher_main():
             is_complete = not measurer_main_process.is_alive()
 
         # Generate periodic output reports.
-        reporter.output_report(experiment.config, in_progress=not is_complete)
+        clang_cov_reports = is_complete
```
inferno-chromium, line 148: use directly.

experiment/dispatcher.py
```diff
@@ -145,7 +145,10 @@ def dispatcher_main():
...
+        reporter.output_report(experiment.config,
+                               in_progress=not is_complete,
+                               clang_coverage_reports=clang_cov_reports)
```
inferno-chromium, line 151: coverage_report=

experiment/measurer.py
```diff
@@ -92,6 +92,34 @@ def get_coverage_infomation(coverage_summary_file):
         return json.loads(summary.readlines()[-1])
 
 
+def store_profdata_files(experiment_config: dict):
```
inferno-chromium, line 95: this function should be 
get_fuzzer_benchmark_trial_pairs
and then below function should be
store_coverage_profdata_files

experiment/measurer.py
```diff
@@ -92,6 +92,34 @@ def get_coverage_infomation(coverage_summary_file):
...
+    experiment = experiment_config['experiment']
+    for benchmark in benchmarks:
+        for fuzzer in fuzzers:
+            trial_ids = get_trial_ids(experiment, fuzzer, benchmark)
+            for trial_id in trial_ids:
+                store_profdata_file(fuzzer, benchmark, trial_id)
+    logger.info('Finished storing profdata files')
+
+
+def store_profdata_file(fuzzer, benchmark, trial_id):
```
inferno-chromium, line 109: This function should be in coverage_utils, you shouldn't depend more on SnapshotMeasurer, but SnapshotMeasurer should depend on a coverage_utils Class or something (like inherit from it).

experiment/measurer.py
```diff
@@ -92,6 +92,34 @@ def get_coverage_infomation(coverage_summary_file):
...
+    """Store profdata file for a pair of fuzzer and benchmark to gcs."""
+    snapshot_measurer = SnapshotMeasurer(fuzzer, benchmark, trial_id, logger)
+    src_profdata = snapshot_measurer.profdata_file
+    benchmark_fuzzer_trial_dir = experiment_utils.get_trial_dir(
+        fuzzer, benchmark, trial_id)
+    dst_profdata_exp_dir = os.path.join(get_experiment_folders_dir(),
+                                        benchmark_fuzzer_trial_dir)
+    dst_profdata_dir = exp_path.filestore(dst_profdata_exp_dir)
+    dst_profdata = os.path.join(dst_profdata_dir,
+                                'data-{id}.profdata'.format(id=trial_id))
```
inferno-chromium, line 119: this code is repeated, you need a helper in coverage_utils for get_profdata_path(coverage_dir, trial_id)

experiment/measurer.py
```diff
@@ -182,6 +210,7 @@ def measure_main(experiment_config):
 
     # Do the final measuring and store the coverage data.
     store_coverage_data(experiment_config)
+    store_profdata_files(experiment_config)
```
inferno-chromium, line 213: This call should be inside the store_coverage_data function.

experiment/reporter.py
```diff
@@ -32,7 +32,9 @@ def get_reports_dir():
     return exp_path.path('reports')
 
 
-def output_report(experiment_config: dict, in_progress=False):
+def output_report(experiment_config: dict,
+                  in_progress=False,
+                  clang_coverage_reports=False):
```
inferno-chromium, line 37: coverage_report

fuzzers/coverage/builder.Dockerfile
```diff
@@ -26,3 +26,7 @@ RUN git clone https://github.com/llvm/llvm-project.git /llvm-project && \
     cd /llvm-project/compiler-rt/lib/fuzzer && \
     bash build.sh && \
     cp libFuzzer.a /usr/lib
+
+# Copy source code files of benchmark to $OUT for the report generation.
+RUN mkdir $OUT/src-files && \
```
inferno-chromium, line 31: why not just call it src instead of src-files.

experiment/coverage_utils.py
```diff
@@ -0,0 +1,206 @@
...
+    return posixpath.join(experiment_filestore_dir, 'coverage-binaries',
+                          archive_name)
+
+
+def get_experiment_filestore_path(experiment_name):
+    """Returns the path of the storage folder for |experiment_name|."""
+    if 'EXPERIMENT_FILESTORE' in os.environ:
+        experiment_filestore = os.environ['EXPERIMENT_FILESTORE']
+    else:
+        experiment_filestore = 'gs://fuzzbench-data'
```
inferno-chromium, line 115: NO hardcode, just add helper in experiment_utils to get this and assume it is always there.

experiment/coverage_utils.py
```diff
@@ -0,0 +1,206 @@
...
+    return posixpath.join(experiment_filestore, experiment_name)
+
+
+def get_benchmark_report_dir(benchmark, report_dir):
```
inferno-chromium, line 119: Used in just one place, use directly.

experiment/coverage_utils.py
```diff
@@ -0,0 +1,206 @@
...
+        archive_exist = filestore_utils.ls(archive_filestore_path,
+                                           must_exist=False).retcode == 0
+        if archive_exist:
+            benchmark_report_dir = get_benchmark_report_dir(
+                benchmark, report_dir)
+            filesystem.create_directory(benchmark_report_dir)
+            filestore_utils.cp(archive_filestore_path,
+                               str(benchmark_report_dir))
+            print(benchmark_report_dir)
+            archive_name = 'coverage-build-%s.tar.gz' % benchmark
```
inferno-chromium, line 91: get_coverage_archive_name(benchmark) and use in all places.

experiment/coverage_utils.py
```diff
@@ -0,0 +1,206 @@
...
+    """Class used to generate coverage report for a pair of
+    fuzzer and benchmark."""
+
+    # pylint: disable=too-many-arguments
+    def __init__(self, fuzzer, benchmark, experiments, report_dir, cov_logger):
+        self.fuzzer = fuzzer
+        self.benchmark = benchmark
+        self.experiments = experiments
+        self.report_dir = report_dir
+        self.logger = cov_logger
```
inferno-chromium, line 135: Why cannot use the global cov_logger directly ?

experiment/coverage_utils.py
Shadoom7: Does it mean we should expect the user to set this environment variable when running `generate_report` locally? I thought that may be a burden for the user and most of them would expect to get the experiment-data from `gs://fuzzbench-data`; so I coded this way.

inferno-chromium:
ping here once this is ready for another round of review.

Shadoom7:
It's ready now. @inferno-chromium

inferno-chromium:
> It's ready now. @inferno-chromium

Can you fix these conflicting files, can you pull and rebase

Conflicting files
experiment/measurer.py
experiment/test_reporter.py

experiment/measurer.py
```diff
@@ -181,7 +70,9 @@ def measure_main(experiment_config):
     measure_loop(experiment, max_total_time)
 
     # Do the final measuring and store the coverage data.
-    store_coverage_data(experiment_config)
+    cov_utils.store_coverage_data(experiment_config)
+    cov_utils.generate_cov_reports(experiment_config)
```
inferno-chromium, line 74: use full function names, s/generate_cov_reports /generate_coverage_report
and next one s/update_coverage_to_bucket/upload_coverage_report_to_bucket

experiment/coverage_utils.py
```diff
@@ -0,0 +1,348 @@
...
+    return trial_ids
+
+
+def merge_profdata_files(src_files, dst_file):
+    """Uses llvm-profdata to merge |src_files| to |dst_files|."""
+    command = ['llvm-profdata', 'merge', '-sparse']
+    command.extend(src_files)
+    command.extend(['-o', dst_file])
+    result = new_process.execute(command, expect_zero=False)
+    return result.retcode
```
inferno-chromium, line 186: Probably just "return result" to mtach other style.

Shadoom7:
> > It's ready now. @inferno-chromium
> 
> Can you fix these conflicting files, can you pull and rebase
> 
> Conflicting files
> experiment/measurer.py
> experiment/test_reporter.py

Rebased it, now running a new test experiment to make sure it works.

experiment/coverage_utils.py
```diff
@@ -0,0 +1,348 @@
...
+                         'fuzzer: {fuzzer},benchmark: {benchmark}.'.format(
+                             fuzzer=self.fuzzer, benchmark=self.benchmark))
+
+
+def get_coverage_archive_name(benchmark):
+    """Gets the archive name for |benchmark|."""
+    return 'coverage-build-%s.tar.gz' % benchmark
+
+
+def get_experiment_folders_dir():
```
inferno-chromium, line 138: THis should be in experiment_utils.

experiment/coverage_utils.py
```diff
@@ -0,0 +1,348 @@
...
+
+def update_coverage_to_bucket():
+    """Copies the coverage reports to gcs bucket."""
+    report_dir = reporter.get_reports_dir()
+    src_dir = get_coverage_report_dir()
+    dst_dir = exp_path.filestore(report_dir)
+    filestore_utils.cp(src_dir, dst_dir, recursive=True, parallel=True)
+
+
+def generate_cov_reports(experiment_config: dict):
```
inferno-chromium, line 47: Use full names
generate_coverage_reports
generate_coverage_report
for functions

analysis/report_templates/default.html
```diff
@@ -232,7 +232,25 @@ <h5 class="center-align">Mann-Whitney U test</h4>
...
+            <ul class="collapsible">
+                <li>
+                    <div class="collapsible-header">
+                        Coverage reports for each fuzzer on this benchmark
+                    </div>
+                    <div class="collapsible-body">
+                        <div class="row">
+                            {% for fuzzer in benchmark.fuzzers %}
+                            <div class="col">
+                                <a class="waves-effect waves-light btn" href="https://storage.googleapis.com/{{ experiment_filestore_name }}/{{ experiment.name }}/reports/coverage/{{ benchmark.name }}/{{ fuzzer }}/index.html">{{ fuzzer }}</a>
```
inferno-chromium, line 245: Is there a reports sub-dir ? i dont think so.

Shadoom7: I created that sub-dir and put all the html there

Shadoom7:
I think the master is broken right now? My experiment in fuzzbench-test crashes even using the master.

jonathanmetzman:
> I think the master is broken right now? My experiment in fuzzbench-test crashes even using the master.

What was the error you saw?

Shadoom7:
> > I think the master is broken right now? My experiment in fuzzbench-test crashes even using the master.
> 
> What was the error you saw?

`failed to solve with frontend dockerfile.v0: failed to build LLB: failed to load cache key: docker.io/library/ubuntu:xenial not found
ERROR
ERROR: build step 0 "gcr.io/cloud-builders/docker" failed: step exited with non-zero status: 1`

jonathanmetzman:
Try again. Should be fixed by #632

Shadoom7:
> Try again. Should be fixed by #632

I still get get the same error :(

jonathanmetzman:
> > Try again. Should be fixed by #632
> 
> I still get the same error :(

How did you try again? Did you pull from master?

Shadoom7:
> > > Try again. Should be fixed by #632
> > 
> > 
> > I still get the same error :(
> 
> How did you try again? Did you pull from master?

Yeah, I confirmed it just now. Still failed with the same error.

jonathanmetzman:
> > > > Try again. Should be fixed by #632
> > > 
> > > 
> > > I still get the same error :(
> > 
> > 
> > How did you try again? Did you pull from master?
> 
> Yeah, I confirmed it just now. Still failed with the same error.

So is this an example of the failures you are talking about? https://console.cloud.google.com/cloud-build/builds/14c5ae22-9d6c-4d1a-9087-70aef5efc5a2;step=0?project=fuzzbench-test

I'll note that this is not the same error we were seeing in CI though there is some similarity.

jonathanmetzman:
> > > > > Try again. Should be fixed by #632
> > > > 
> > > > 
> > > > I still get the same error :(
> > > 
> > > 
> > > How did you try again? Did you pull from master?
> > 
> > 
> > Yeah, I confirmed it just now. Still failed with the same error.
> 
> So is this an example of the failures you are talking about? https://console.cloud.google.com/cloud-build/builds/14c5ae22-9d6c-4d1a-9087-70aef5efc5a2;step=0?project=fuzzbench-test
> 
> I'll note that this is not the same error we were seeing in CI though there is some similarity.

Damn it https://stackoverflow.com/a/58791983

CC @lszekeres

Shadoom7:
> > > > > Try again. Should be fixed by #632
> > > > 
> > > > 
> > > > I still get the same error :(
> > > 
> > > 
> > > How did you try again? Did you pull from master?
> > 
> > 
> > Yeah, I confirmed it just now. Still failed with the same error.
> 
> So is this an example of the failures you are talking about? https://console.cloud.google.com/cloud-build/builds/14c5ae22-9d6c-4d1a-9087-70aef5efc5a2;step=0?project=fuzzbench-test
> 
> I'll note that this is not the same error we were seeing in CI though there is some similarity.

Yes, this is the one.

Shadoom7:
The test experiment succeeded. Report [here](https://storage.googleapis.com/my-fuzzbench-reports/gen-cov-test-4-3/index.html) @inferno-chromium

analysis/report_templates/default.html
lszekeres, line 245: would it make sense to make this whole path (https://storage.googleapis.com/{{ experiment_filestore_name }}) a variable (eg `reports_path`) instead of just `experiment_filestore_name`, in order to make this more flexible (eg support local paths)

analysis/generate_report.py
```diff
@@ -74,6 +76,11 @@ def get_arg_parser():
                         '--fuzzers',
                         nargs='*',
                         help='Names of the fuzzers to include in the report.')
+    parser.add_argument('-cov',
+                        '--coverage-report',
```
lszekeres, line 80: consider something like `--link-coverage-reports` to indicate that they are not generated, just linked

analysis/report_templates/default.html
lszekeres: alternatively, we could use `experiment_filestore` from the experiement-config.yaml, and if it's a gs:// path then use https://storage.googleapis.com/, otherwise not.

PR #621, opened 2020-08-03
Create experiment entity before building.

inferno-chromium:


inferno-chromium:
> LGTM
> 
> Can you explain why we want to do this in the commit.
> This sort of breaks MANUAL_EXPERIMENT but that's only debug junk that I can fix next time I need it.

It is good for debugging that another experiment does not start so that logs dont get polluted, right now i cannot see when was the time it happened when that vm restarted as another experiment just started.

jonathanmetzman:
> > LGTM
> > Can you explain why we want to do this in the commit.
> > This sort of breaks MANUAL_EXPERIMENT but that's only debug junk that I can fix next time I need it.
> 
> It is good for debugging that another experiment does not start so that logs dont get polluted, right now i cannot see when was the time it happened when that vm restarted as another experiment just started and started quoting the logs. Also this is useful for manual experiment runs where someone chooses the same name as an existing service experiment (since dates are easy to conflict with). i think you meant rerunning experiment, but we can worry about it later since it is hacky anyway.

Right I understand this, I meant when committing it, to put that description in the message.

PR #622, opened 2020-08-03
Use -summary-only flag for coverage measuring

Shadoom7:
This patch will use `-summary-only` flag for the measuring process.

### DIFF
--- a/experiment/measurer.py
+++ b/experiment/measurer.py
@@ -545,7 +545,7 @@ class SnapshotMeasurer:  # pylint: disable=too-many-instance-attributes
             '-instr-profile=%s' % self.profdata_file
         ]
 
-        if summary:
+        if summary_only:
             command.append('-summary-only')
 
         with open(self.cov_summary_file, 'w') as output_file:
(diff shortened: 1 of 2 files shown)


experiment/measurer.py
```diff
@@ -536,21 +537,30 @@ def generate_profdata(self, cycle: int):
...
-    def generate_summary(self, cycle: int):
+    def generate_summary(self, cycle: int, summary=True):
         """Transform the .profdata file into json form."""
         coverage_binary = get_coverage_binary(self.benchmark)
         command = [
             'llvm-cov', 'export', '-format=text', coverage_binary,
             '-instr-profile=%s' % self.profdata_file
         ]
+
+        # Add summary-only flag to reduce memory use.
```
inferno-chromium, line 548: Remove this comment since it is not needed during trials even, this is bad for perf.

experiment/measurer.py
```diff
@@ -536,21 +537,30 @@ def generate_profdata(self, cycle: int):
             self.logger.error(
                 'Coverage profdata generation failed for cycle: %d.', cycle)
 
-    def generate_summary(self, cycle: int):
+    def generate_summary(self, cycle: int, summary=True):
```
inferno-chromium, line 540: s/summary/summary_only ?

Issue #627, opened 2020-08-04
Make issues: debug is broken, run is impossible to control-C out of


jonathanmetzman:
make debug- is broken. `make debug-libfuzzer-zlib_zlib_uncompress_fuzzer` fails with this error:
```
docker run \                                                                                                                                                                                                       
--cpus=1 \                                                                                                                                                                                                         
--cap-add SYS_NICE \                                                                                                                                                                                               
--cap-add SYS_PTRACE \                                                                                                                                                                                             
-e FUZZ_OUTSIDE_EXPERIMENT=1 \                                                                                                                                                                                     
-e FORCE_LOCAL=1 \                                                                                                                                                                                                 
-e TRIAL_ID=1 \                                                                                                                                                                                                    
-e FUZZER=libfuzzer \                                                                                                                                                                                              
-e BENCHMARK=zlib_zlib_uncompress_fuzzer \                                                                                                                                                                         
-e FUZZ_TARGET=zlib_uncompress_fuzzer \                                                                                                                                                                            
-entrypoint "/bin/bash" \                                                                                                                                                                                          
-it gcr.io/fuzzbench/runners/libfuzzer/zlib_zlib_uncompress_fuzzer                                                                                                                                                 
docker: invalid reference format.                                                                                                                                                                                  
See 'docker run --help'.                                                                                                                                                                                           
make: *** [docker/generated.mk:26568: debug-libfuzzer-zlib_zlib_uncompress_fuzzer] Error 125
```

make `run-libfuzzer-zlib_zlib_uncompress_fuzzer` runs forever and cannot be Ctrl-Ced out of.

jonathanmetzman:
cc @lszekeres

jonathanmetzman:
I don't have time to bisect this but I can confirm that ctrl-C was possible on cb44bac80f7ff1ff64f5c72c89466221c98c038e (from June 10th). Not sure that it's because of the recent makefile changes.

PR #628, opened 2020-08-04
Fix make debug-

jonathanmetzman:
Fixes a typo causing #627

### DIFF
--- a/docker/generate_makefile.py
+++ b/docker/generate_makefile.py
@@ -58,7 +58,7 @@ def _print_makefile_run_template(image):
         if run_type == 'test-run':
             print('\t-e MAX_TOTAL_TIME=20 \\\n\t-e SNAPSHOT_PERIOD=10 \\')
         if run_type == 'debug':
-            print('\t-entrypoint "/bin/bash" \\\n\t-it ', end='')
+            print('\t--entrypoint "/bin/bash" \\\n\t-it ', end='')
         else:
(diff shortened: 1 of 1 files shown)


docker/generate_makefile.py
```diff
@@ -58,7 +58,7 @@ def _print_makefile_run_template(image):
         if run_type == 'test-run':
             print('\t-e MAX_TOTAL_TIME=20 \\\n\t-e SNAPSHOT_PERIOD=10 \\')
         if run_type == 'debug':
-            print('\t-entrypoint "/bin/bash" \\\n\t-it ', end='')
+            print('\t--entrypoint "/bin/bash" \\\n\t-it ', end='')
```
jonathanmetzman, line 61: I don't like having tabs and newlines everywhere especially given how ugly escaping them is.
I'm not sure that if we didn't have them we could have spotted this, but it's possible.

lszekeres: we should actually have a tab only in front of each command of a rule, but not in front of lines after `\`-s. ie.
```
rule:
<tab>command ... \
<spc><spc>continuation
```
we could also remove the `\` newlines altogether

inferno-chromium: +1 to removing newlines completely. this is generated content, does not need to look pretty, and should be less error prone. this newlines and tab removal will be awesome.

Issue #357, opened 2020-05-19
make installing pip dependencies more resilient in CI


jonathanmetzman:
Right now we get a fair amount of failures just because we need to install pip dependencies in CI and sometimes this fails.
I can think of two ways to solve this:
1. [Using caching](https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows). I prefer this solution because it would also improve speed.
2. Using retries.

jonathanmetzman:
I haven't seen many failures since we added caching. I don't know if this is due to caching or because the failures happen as a result of infra changes in GitHub actions that are random we don't control (so they happen not to occur now). Either way, we can reopen this issue if failures happen frequently again.

PR #631, opened 2020-08-05
[CI] Cache Python packages

jonathanmetzman:
Transient pip install errors are a large source of spurious CI
failures. Pip install also takes > 1 minute, around half the time
it takes for presubmit job to run on CI so speeding it up will
be a nice improvement.

### DIFF
--- a/.github/workflows/fuzzers.yml
+++ b/.github/workflows/fuzzers.yml
@@ -66,7 +66,6 @@ jobs:
       with:
         python-version: 3.7
 
-    # test
     # Copied from:
     # https://docs.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions
     - name: Cache pip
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
The speedup is extremely modest it takes ~1:20 to install without caching. Caching+install with caching takes about ~1:05.

I still think this change is worth it to avoid the weird network errors we've been seeing in CI.

jonathanmetzman:
Fixes https://github.com/google/fuzzbench/issues/357

jonathanmetzman:
If we really want the speed up, we can probably cache .venv instead of the pip packages (which we cache by caching the files in ~/.cache/pip)

jonathanmetzman:
Caching all of .venv is probably a bad idea because we will cache link to Python binary as well. Will just cache the pip package cache instead.

Issue #629, opened 2020-08-05
Buildkit failing in CI


jonathanmetzman:
On [this CI run ](https://github.com/google/fuzzbench/pull/628/checks?check_run_id=947692658) AFL build fails with the message `failed to solve with frontend dockerfile.v0: failed to build LLB: mount already exists with name`.
I don't know what is causing this but I assume it's something to do with the buildkit changes.
I've never seen this happen before, I assume it was introduced by buildkit. Can you please look at other CI runs and see if it happens frequently?

jonathanmetzman:
Looks like this caused 6/8 failures here: https://github.com/google/fuzzbench/pull/625

jonathanmetzman:
From googling this, it seems like a buildkit error: https://github.com/moby/buildkit/issues/1460

jonathanmetzman:
More errors: https://github.com/google/fuzzbench/pull/633/checks?check_run_id=951608193

@tanq16 did you try copying the fuzzer and fuzzers/ directory from host instead? Does that work.
You can force CI to build everything by modifying runner.py But please don't do this too much or else there will be a backlog for CI bots.

Tanq16:
Fixed by #632

PR #632, opened 2020-08-05
Test fix - BuildKit Fails on CI

Tanq16:
This PR is to test CI and relates to #629 .


Tanq16:
I haven't seen the issue happen with this change. Since this is hard to test as the issue isn't easily reproducible, I did 2 parallel runs of CI on this PR and another ensuring that the issue happens on the other and not here.

jonathanmetzman:
If you commit and then push to the same PR you will also do two parallel ci runs.

Tanq16:
> If you commit and then push to the same PR you will also do two parallel ci runs.

Oh, I thought it cancels the first run. I'll do one more test this way then.

Tanq16:
Both the CI runs passed. I'll test once to build everything to be certain that the issue doesn't occur.

Tanq16:
Everything apart from the Eclipser and AFLplusplus_optimal pass. Those don't have the mount error. This should fix the buildkit issue.

jonathanmetzman:
OK hopefully this fixes it. Can you get rid of the unnecessary changes you did for testing and we'll merge this?

Tanq16:
Fixes #629 .

docker/benchmark-runner/Dockerfile
```diff
@@ -54,13 +54,15 @@ ENV WORKDIR /out
 RUN mkdir -p $WORKDIR
 WORKDIR $WORKDIR
 
+ARG fuzzer
```
jonathanmetzman, line 57: This line is unnecessary it already is at the top of the file.

Tanq16: It gets reset after that. I couldn't get it to work without this. If I remove it, it basically returns an empty string (at least on a local run).

jonathanmetzman: Are you sure? I find this behavior extremely bizarre. Especially given that we use the same variable twice on 37 and 41. If you're sure please just leave a comment explaining why you are doing this.

Tanq16: Yes I'm pretty sure this happens (at least on a local run). I'll add the comment.

Tanq16: I'll try a CI run without it just in case.

Tanq16: Can be seen in [this link](https://github.com/google/fuzzbench/runs/959551133?check_suite_focus=true).
Made changes.

PR #633, opened 2020-08-05
Support running experiments without seeds or dictionaries.

### LINKED ISSUE #127: Support dictionaries (mostly done, only need to add way to test dictionaryless experiment)
Many of our benchmarks have dictionaries, it would probably be interesting to see how using dictionaries affects results (since dictionaries are pretty widely used in the real world).
The downsides I see to supporting dictionaries are:
1. A wider interface for integrating fuzzers.
2. Some fuzzers (like eclipser I believe) don't support dictionaries, so in some sense the comparsion won't be apples to apples.

jonathanmetzman:
This adds two new flags to run_experiment: --no-seeds and --no-dictionaries.
When passed, they cause trials run as part of the experiment not to use seed corpus files or dictionaries (respectively) even if they are provided by the benchmark.

Fixes #127

### DIFF
--- a/experiment/run_experiment.py
+++ b/experiment/run_experiment.py
@@ -200,7 +200,8 @@ def get_git_hash():
 
 
 def start_experiment(experiment_name: str, config_filename: str,
-                     benchmarks: List[str], fuzzers: List[str]):
+                     benchmarks: List[str], fuzzers: List[str],
+                     no_seeds=False, no_dictionaries=False):
     """Start a fuzzer benchmarking experiment."""
     check_no_local_changes()
 
@@ -212,6 +213,8 @@ def start_experiment(experiment_name: str, config_filename: str,
     config['benchmarks'] = ','.join(benchmarks)
     config['experiment'] = experiment_name
     config['git_hash'] = get_git_hash()
+    config['no_seeds'] = no_seeds
+    config['no_dictionaries'] = no_dictionaries
 
     set_up_experiment_config_file(config)
 
@@ -448,28 +451,26 @@ def main():
                                required=False,
                                default=None,
                                choices=all_fuzzers)
-    fuzzers_group.add_argument('-cf',
-                               '--changed-fuzzers',
-                               help=('Use fuzzers that have changed since the '
-                                     'last experiment. The last experiment is '
-                                     'determined by the database your '
-                                     'experiment uses, not necessarily the '
-                                     'fuzzbench service'),
-                               action='store_true',
-                               required=False)
+    fuzzers_group.add_argument('-s',
+                               '--no-seeds',
+                               help='Should trials be conducted without seed corpora.',
+                               required=False,
+                               default=False,
+                               action='store_true')
+    fuzzers_group.add_argument('-d',
+                               '--no-dictionaries',
+                               help='Should trials be conducted without dictionaries.',
+                               required=False,
+                               default=False,
+                               action='store_true')
 
     args = parser.parse_args()
-
-    if args.changed_fuzzers:
-        fuzzers = experiment_changes.get_fuzzers_changed_since_last()
-        if not fuzzers:
-            logs.error('No fuzzers changed since last experiment. Exiting.')
-            return 1
-    else:
-        fuzzers = args.fuzzers or all_fuzzers
+    fuzzers = args.fuzzers or all_fuzzers
 
     start_experiment(args.experiment_name, args.experiment_config,
-                     args.benchmarks, fuzzers)
+                     args.benchmarks, fuzzers,
+                     no_seeds=args.no_seeds,
+                     no_dictionary=args.no_dictionaries)
     return 0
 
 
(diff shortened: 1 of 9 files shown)


experiment/run_experiment.py
```diff
@@ -448,28 +451,26 @@ def main():
...
-    fuzzers_group.add_argument('-cf',
-                               '--changed-fuzzers',
-                               help=('Use fuzzers that have changed since the '
-                                     'last experiment. The last experiment is '
-                                     'determined by the database your '
-                                     'experiment uses, not necessarily the '
-                                     'fuzzbench service'),
-                               action='store_true',
-                               required=False)
+    fuzzers_group.add_argument('-s',
```
inferno-chromium, line 454: -ns ?

experiment/run_experiment.py
```diff
@@ -448,28 +451,26 @@ def main():
...
+                               '--no-seeds',
+                               help='Should trials be conducted without seed corpora.',
+                               required=False,
+                               default=False,
+                               action='store_true')
+    fuzzers_group.add_argument('-d',
```
inferno-chromium, line 460: -nd ?

experiment/run_experiment.py
```diff
@@ -448,28 +451,26 @@ def main():
...
     args = parser.parse_args()
-
-    if args.changed_fuzzers:
-        fuzzers = experiment_changes.get_fuzzers_changed_since_last()
-        if not fuzzers:
-            logs.error('No fuzzers changed since last experiment. Exiting.')
-            return 1
-    else:
-        fuzzers = args.fuzzers or all_fuzzers
+    fuzzers = args.fuzzers or all_fuzzers
```
inferno-chromium, line 468: this needs to be reverted

experiment/runner.py
```diff
@@ -65,9 +65,15 @@
...
 def _clean_seed_corpus(seed_corpus_dir):
-    """Moves seed corpus files from sub-directories into the corpus directory
-    root. Also, deletes any files that exceed the 1 MB limit."""
-    if not os.path.exists(seed_corpus_dir):
+    """Prepares |seed_corpus_dir| for the trial. This ensures that it can be
+    used by AFL which is picky about the seed corpus. Moves seed corpus files
+    from sub-directories into the corpus directory root. Also, deletes any files
+    that exceed the 1 MB limit. If the NO_SEEDS env var is specified than the
+    seed corpus files are deleted."""
+    if environment.get('NO_SEEDS'):
```
inferno-chromium, line 73: Where are NO_SEEDS and NO_DICTIONARIES set ?

experiment/runner.py
```diff
@@ -65,9 +65,15 @@
...
+        logs.info('NO_SEEDS specified, deleting seed corpus files.')
+        shutil.rmtree(seed_corpus_dir)
```
inferno-chromium, line 75: Does seed_corpus_dir always exist ?

jonathanmetzman: It does, but I should probably added back this check right?

experiment/runner.py
jonathanmetzman: They are set in scheduler.py which creates a startup script that runs the container.
This is the same as most other env vars this file uses.

experiment/run_experiment.py
jonathanmetzman: I didn't include this in the description, but we don't need this code anymore. I can remove here or in another PR up to you.

experiment/run_experiment.py
jonathanmetzman: I don't like doing two char short flags. I think it's more normal to use an unrelated letter than to use a char short flag (like how `-d` in AFL means deterministic so the flag for dictionary is `-x`)

inferno-chromium: Here flag means the opposite, so it is very confusing. -d mean no dict is so confusing.

experiment/run_experiment.py
inferno-chromium: ok, can you remove the changed_fuzzers argument as well.

experiment/runner.py
inferno-chromium: Probably good to check or assert.

experiment/runner.py
inferno-chromium: You need to add it in startup script template file like
-e NO_SEEDS={{no_seeds}} \
-e NO_DICTIONARIES={{no_dictionaries}} \

Also, no_dictionaries can be shortened to no_dicts everywhere.

experiment/run_experiment.py
jonathanmetzman: Changed it.

experiment/run_experiment.py
jonathanmetzman: Which argument? I don't see it.

experiment/runner.py
jonathanmetzman: Ah good catch. Forgot to commit this.
I think the extra letters is fine, I'd rather use dictionaries since no-dictionaries will be user facing.

experiment/run_experiment.py
inferno-chromium: ah i didnt see it properly, you did remove -changed-fuzzers

PR #635, opened 2020-08-06
[benchmarks] Remove perl benchmark.

jonathanmetzman:
The benchmark doesn't work and it's better not to have brokenish
benchmarks laying around to trip up users.


jonathanmetzman:
@roachspray I'll be happy to add it back when it works. But keeping around confused it some of our maintainers who accidentally used this in experiments.

PR #636, opened 2020-08-06
Fix issue with presubmit in CI

jonathanmetzman:


jonathanmetzman:
Looks like github changed what they return to clients invoking the magic git commands we do. 
Don't really know why this fixes things but it seems to do so in local tests.

jonathanmetzman:
Failure can be seen here: https://github.com/google/fuzzbench/pull/632/checks?check_run_id=954303324

jonathanmetzman:
TBR

Issue #41, opened 2020-03-02
Report fuzzed speed (execs/sec) as well


lszekeres:
Suggestion by @andreafioraldi: https://twitter.com/andreafioraldi/status/1234578416336240642

jonathanmetzman:
As I [mentioned](https://twitter.com/metzmanj/status/1234584852634034176) I'm a bit hesitant to add this feature for fear that the fuzzer API becomes too wide. An opt-in or an AFL supported by default approach might mitigate this somewhat.

ammaraskar:
CPU usage + Memory usage would also be useful to note over time to make sure fuzzers are fully saturating hardware.

stevenagy:
This would be awesome for comparing fuzzers as a whole! But for specific optimization areas like instrumentation-side performance, using execs/sec across random trials is potentially flawed because the parent fuzzer may get bogged down on slow paths. 

What has worked well for us in such cases is isolating performance comparison on some fixed input corpora -- we pre-generate 24-hr input dumps and "replay" the same inputs to whatever instrumenters we are testing, logging their execution times on each input, and then compute statistics on the sums of all execution times per dump. Scaling everything to the performance of a "forkserver-only" configuration of AFL-Clang allows us to get a sense of what an instrumenter's overall fuzzing overhead is.

catenacyber:
I am not sure how to apply this but one feature of fuzzers is to find slow paths and timeouts (with smallest inputs) and we want to see which fuzzers can do that...

jonathanmetzman:
> I am not sure how to apply this but one feature of fuzzers is to find slow paths and timeouts (with smallest inputs) and we want to see which fuzzers can do that...

Interesting idea, but I don't that's a priority for now. First we want to figure out which fuzzers are best at finding exploitable bugs, other bugs come second to that.

ammaraskar:
Maybe % of total coverage, i.e what oss-fuzz reports as `edge_coverage` in the fuzzer states:

```
17.66% (30854/174672)
```

to help figure out how saturated a target is.

PR #648, opened 2020-08-10
Support reporting of fuzzing-time stats (by fuzzer.py)

jonathanmetzman:
Currently the only stats FuzzBench reports per cycle/snapshot are those recorded by the measurer. This means we cannot report some useful fuzzing-time stats such as executions/sec, stability, CPU/RAM or I/O usage etc.

This PR adds an optional API for fuzzer.py to report extra stats in a uniform way that allows fuzzers to be compared on these stats.
For now it only supports executions per second a float that we call avg_execs.
We can use the same pipeline in runner.py to report fuzz-time stats that are fuzzer independent such as RAM, disk, IO, CPU.
TODO before this is ready for review:
1. Rename avg_execs to execs_sec. It's more descriptive this way.
2. Reuse AFL's reporting API implementation in AFL-based fuzzers.
3. Implement stats reporting in (some) other fuzzers like libFuzzer.
4. Support at least one fuzzer-independent stat like CPU usage.
5. Document this API for external devs to use.
6. Figure out if we want each stat to be its own column or if one JSON column for all of them is fine. I think JSON field makes more sense since each stat is optional and adding new ones will be easier this way (no need for DB migrations). The downside is when generating reports we will have to collapse this field to create a dataframe where we have a column for each supported stat. It may also make querying harder for SQL wizards.

We also need to figure out how to report this. I think we could maybe link to another page where we generate stats in aggregate and for each benchmark. I think the exact way we report it can wait though.


jonathanmetzman:
I fixed a lot of some of the more important issues, others (like reusing the stats implementation in other AFL-based fuzzers) can be done later.
@inferno-chromium WDYT?

inferno-chromium:
> I fixed a lot of some of the more important issues, others (like reusing the stats implementation in other AFL-based fuzzers) can be done later.
> @inferno-chromium WDYT?

So, we wont have stats for libFuzzer and honggfuzz fuzzers ? Starting somewhere is fine, we need to add those as well over time.

jonathanmetzman:
> > I fixed a lot of some of the more important issues, others (like reusing the stats implementation in other AFL-based fuzzers) can be done later.
> > @inferno-chromium WDYT?
> 
> So, we wont have stats for libFuzzer and honggfuzz fuzzers ? Starting somewhere is fine, we need to add those as well over time.

Not in this patch. But they can definitely be added. This API is meant to be supportable by most fuzzers and is meant to be optional as well so we don't increase onboarding complexity.

inferno-chromium:
> > > I fixed a lot of some of the more important issues, others (like reusing the stats implementation in other AFL-based fuzzers) can be done later.
> > > @inferno-chromium WDYT?
> > 
> > 
> > So, we wont have stats for libFuzzer and honggfuzz fuzzers ? Starting somewhere is fine, we need to add those as well over time.
> 
> Not in this patch. But they can definitely be added. This API is meant to be supportable by most fuzzers and is meant to be optional as well so we don't increase onboarding complexity.

Sounds perfect, we definitely add this incrementally and not everything needs to integrate with this.

jonathanmetzman:
I think this is ready to land, I'm just going to test it with a local experiment first.

database/models.py
```diff
@@ -62,3 +68,4 @@ class Snapshot(Base):
     trial_id = Column(Integer, ForeignKey('trial.id'), primary_key=True)
     trial = sqlalchemy.orm.relationship('Trial', back_populates='snapshots')
     edges_covered = Column(Integer, nullable=False)
+    fuzzer_stats = Column(JSON, nullable=True)
```
lszekeres, line 71: Are we planning to keep this? If yes, what is the benefit of having a JSON here?

IMO having JSONs here will be a large source of complexity (e.g. in queries) and bugs.

Consider having a `number_of_executions` column (Integer, nullable) instead for this PR, which would be more normal form than `{'execs_per_sec': float}` JSONs. Later, other metrics should have their own column too (like CPU/mem usage at `time`).

Related:
https://www.thomascerqueus.fr/json-fields-relational-databases-postgresql/
https://stackoverflow.com/questions/15367696/storing-json-in-database-vs-having-a-new-column-for-each-key
https://softwareengineering.stackexchange.com/questions/215429/is-it-wise-to-store-a-big-lump-of-json-on-a-database-row

inferno-chromium: I think it depends on if we will have custom stats and whether we want to allow fuzzers to use those. If we decide with Laszlo approach, i dont think we should add all these columns in Snapshot table ? maybe some other table type. Should we have a FuzzerStats table object here and then we can add any columns to it over time ?

jonathanmetzman: >Are we planning to keep this?

I think so.

>If yes, what is the benefit of having a JSON here?

It offers a lot more flexibility. It means that stats can be added and removed without changing the SQL schema.

>IMO having JSONs here will be a large source of complexity (e.g. in queries) and bugs.

I suspect this won't actually be the case since we tend to most analysis on dataframes, so once we clean things up so that JSON fields work the same as other fields in dataframes, we wont have a problem.

In https://stackoverflow.com/questions/15367696/storing-json-in-database-vs-having-a-new-column-for-each-key they describe a case for JSON fields that I think is appicable:
```
Saving arbitrary key/value user preferences (where the value can be boolean, textual, or numeric, and you don't want to have separate columns for different data types)
```
I think I want to do something similar in the next iteration. There I envision allowing fuzzers to store arbitrary key/value as stats.

Personally I don't think I will ever want to use these stats in an SQL query. For me, using pandas is easier, and this is what most users will do as well. Thus I don't think we lose anything from using a JSON field.
We do gain flexibility though so on balance I think there is a stronger case for JSON fields.

jonathanmetzman:
> I think this is ready to land, I'm just going to test it with a local experiment first.

Local experiment worked fine.

database/models.py
jonathanmetzman: I think in-person you preferred using number of execs rather than avg execs since the latter can be derived from the former (which conveys more info). I think it's probably better to store stats in the format they will be consumed. So assuming avg execs is more useful to users than total execs (and maybe I'm wrong) it's better to store avg rather than store total and compute average before graphing. What do you think?

lszekeres: > It means that stats can be added and removed without changing the SQL schema.

That is what I think we should avoid. Would we allow arbitrary JSON, or enforce a schema? If the former, how are we going to use that? The PR suggests the latter, as it implements manual schema checking in `common/fuzzer_stats.py` (`SCHEMA = {'execs_per_sec': float}`). What's not clear is: why would we implement this manually, adding all this code and complexity, while this is what the DB schema gives us out of the box?

> > IMO having JSONs here will be a large source of complexity (e.g. in queries) and bugs.
> 
> I suspect this won't actually be the case since we tend to most analysis on dataframes, so once we clean things up so that JSON fields work the same as other fields in dataframes, we wont have a problem.

Yes, exactly: Using the DB schema, we get the dataframe automatically, no additional code needed. With JSON we need to write custom code validating, and converting arbitrary schemaless JSONs to dataframe columns manually, handling all possible errors and so on, which I think will be a massive source of complexity and bugs.

> I think I want to do something similar in the next iteration. There I envision allowing fuzzers to store arbitrary key/value as stats.

What do you mean by arbitrary? Does arbitrary mean that it's relevant only to a particular fuzzer? If yes, how would we use that? Or is it just for the developer of a particular engine? If it's a fuzzer engine specific stat and only for the developer of that engine to be used by their private scripts, and not by fuzzbench code, than JSON indeed sounds good.

However, for _common stats_ (that are relevant to all engines) we should carefully choose the ones that we want to allow  engines to provide (and have a schema for them) to ensure uniformity and normalized data, so we can compare them along those metrics (eg. in the reports). To me, these would be basically
- number of execs ,
- cpu usage,
- mem usage

(Is there anything else we were planning to add?)

> arbitrary key/value as stats.

Why key/value? What key would you use other than `time` or `number of execs` as key?

lszekeres: > I think in-person you preferred using number of execs rather than avg execs since the latter can be derived from the former (which conveys more info). I think it's probably better to store stats in the format they will be consumed. So assuming avg execs is more useful to users than total execs (and maybe I'm wrong) it's better to store avg rather than store total and compute average before graphing. What do you think?

Data should be stored in its most "normal form", if possible, to keep all queries simple. E.g., based on previous discussions, it would be nice if we could not only show (1) [executions / sec](https://github.com/google/fuzzbench/issues/41), but also e.g., (2) [coverage / executions](https://github.com/google/fuzzbench/issues/654#issuecomment-672496181). If we have `time, number_of_execs, edges_covered` in the [table](https://github.com/google/fuzzbench/blob/master/database/models.py#L61-L64), then both (1) and (2) can be easily queried. If we have `execs_per_sec` instead, then (2) would be much more difficult to query.

jonathanmetzman: > > It means that stats can be added and removed without changing the SQL schema.
> 
> That is what I think we should avoid. Would we allow arbitrary JSON, or enforce a schema? If the former, how are we going to use that? The PR suggests the latter, as it implements manual schema checking in `common/fuzzer_stats.py` (`SCHEMA = {'execs_per_sec': float}`). In this case what's not clear is: why would we implement this manually, adding all this code and complexity, while this is what the DB schema gives us out of the box?
> 
> > > IMO having JSONs here will be a large source of complexity (e.g. in queries) and bugs.
> > 
> > 
> > I suspect this won't actually be the case since we tend to most analysis on dataframes, so once we clean things up so that JSON fields work the same as other fields in dataframes, we wont have a problem.
> 
> Yes, this is what I was referring to: using the DB schema, we get the dataframe automatically, no additional code needed. With JSON we need to write custom code validating, and converting arbitrary schemaless JSONs to dataframe columns manually, handling all possible errors and so on, which I think will be a massive source of complexity and bugs.
> 
> > I think I want to do something similar in the next iteration. There I envision allowing fuzzers to store arbitrary key/value as stats.
> 
> What do you mean by arbitrary? Does arbitrary mean that it's relevant only to a particular fuzzer? If yes, how would we use that? Or is it just for the developer of a particular engine? If it's a fuzzer engine specific stat and only for the developer of that engine to be used by their private scripts, and not by fuzzbench code, than JSON indeed sounds good.
> 
> However, for _common stats_ (that are relevant to all engines) we should carefully choose the ones that we want to allow engines to provide (and have a schema for them) to ensure uniformity and normalized data, so we can compare them along those metrics (eg. in the reports). To me, these would be basically
> 
> * number of execs ,
> * cpu usage,
> * mem usage
> 
> (Is there anything else?)
> 
> > arbitrary key/value as stats.
> 
> Why key/value? What key would you use other than `time` or `number_of_execs` as key?



> > It means that stats can be added and removed without changing the SQL schema.
> 
> That is what I think we should avoid. Would we allow arbitrary JSON, or enforce a schema? If the former, how are we going to use that? The PR suggests the latter, as it implements manual schema checking in `common/fuzzer_stats.py` (`SCHEMA = {'execs_per_sec': float}`). In this case what's not clear is: why would we implement this manually, adding all this code and complexity, while this is what the DB schema gives us out of the box?
> 
> > > IMO having JSONs here will be a large source of complexity (e.g. in queries) and bugs.
> > 
> > 
> > I suspect this won't actually be the case since we tend to most analysis on dataframes, so once we clean things up so that JSON fields work the same as other fields in dataframes, we wont have a problem.
> 
> Yes, this is what I was referring to: using the DB schema, we get the dataframe automatically, no additional code needed. With JSON we need to write custom code validating, and converting arbitrary schemaless JSONs to dataframe columns manually, handling all possible errors and so on, which I think will be a massive source of complexity and bugs.
> 
> > I think I want to do something similar in the next iteration. There I envision allowing fuzzers to store arbitrary key/value as stats.
> 
> What do you mean by arbitrary? Does arbitrary mean that it's relevant only to a particular fuzzer?

Yeah, or group of fuzzers.

>If yes, how would we use that? Or is it just for the developer of a particular engine? If it's a fuzzer engine specific stat and only for the developer of that engine to be used by their private scripts, and not by fuzzbench code, than JSON indeed sounds good.

I think the generic analysis code probably wouldn't consume this data.
Maybe this data should be stored as JSON and the common stuff stored as columns?
(I realize this is sort of violating YAGNI)

> 
> However, for _common stats_ (that are relevant to all engines) we should carefully choose the ones that we want to allow engines to provide (and have a schema for them) to ensure uniformity and normalized data, so we can compare them along those metrics (eg. in the reports). To me, these would be basically
> 
> * number of execs ,
> * cpu usage,
> * mem usage
> 
> (Is there anything else?)

Can't think of anything else but I would be surprised if we don't add more at some point.

> 
> > arbitrary key/value as stats.
> 
> Why key/value? What key would you use other than `time` or `number_of_execs` as key?

Sorry, by key I meant name of the stat.

lszekeres: > Maybe this data should be stored as JSON and the common stuff stored as columns?

Yes, imo common ones (such as number of execs, cpu usage, mem usage) should be columns (so our analysis code can easily use them), and engine specific ones should be JSON (as fuzzbench code would not use them). What do you have in mind for the engine specific ones? Also, who would use them and how? Ie are there any requests from users to support them?

jonathanmetzman: >What do you have in mind for the engine specific ones?

Stability for example.

>Also, who would use them and how? Ie are there any requests from users to support them?

For example, vanhauser-thc could use this in analyzing which targets are stable.

lszekeres: > > What do you have in mind for the engine specific ones?
> 
> Stability for example.
> 
> > Also, who would use them and how? Ie are there any requests from users to support them?
> 
> For example, vanhauser-thc could use this in analyzing which targets are stable.

Makes sense. We might also consider asking first if people would use it, but something like this would be nice:
```
time = Column(Integer, nullable=False, primary_key=True)
edges_covered = Column(Integer, nullable=False)
num_execs = Column(Integer, nullable=True)
cpu_usage = Column(Float, nullable=True)
mem_usage = Column(Integer, nullable=True)
fuzzer_specific = Column(JSON, nullable=True)
```
where `fuzzer_specific` is whatever the fuzzer developer wants, without a schema

inferno-chromium: specific in column name is weird, i think fuzzer_stats would imply fuzzer specific stats. and, cpu and memory columns will probably come later as we add capturing hooks ? for now, i think just json is fine, since are adding fuzzer stats.

lszekeres: > specific in column name is weird, i think fuzzer_stats would imply fuzzer specific stats. and, cpu and memory columns will probably come later as we add capturing hooks ? 

sgtm. we might still consider some alternative name that differentiates a bit better between "common" and "specific" stats. eg. `edges_covered` is also "fuzzer statistic", but it's a "common" one, as opposed to say the afl specific stability. maybe something like `extra_stats` or `user_defined_stats` would capture this a bit better.

> for now, i think just json is fine, since are adding fuzzer stats.

which stats are we adding first?

inferno-chromium: i mean we will have json stats for now, since more common stats like cpu and mem are not hooked and will require custom capturing code. num_execs, can come as common one, but would need code across all fuzzers, i think this one is just doing afl as a start. once we add to all fuzzers, they can add code to set this column "num_execs"

jonathanmetzman: Yeah, I'm of two minds about this question since you (@lszekeres) raised valid points.
But I think it would be simpler to handle all of the stats which use custom capturing in the same way (JSON).
I also realized there is a pretty good reason we don't want to rely on the schema for type validation.
We wont save stats to the db in the runner, we will do this in the measurer. So I don't think we can wait until saving to the db for validation because it happens so far from the error.

fuzzers/test_fuzzers.py
```diff
@@ -76,3 +79,14 @@ def test_fuzz_function_errors():
         # so fail if we see one. If that is not the case than this assert
         # should be removed.
         assert not isinstance(error.value, TypeError)
+
+
+def test_afl_get_stats(tmp_path):
```
jonathanmetzman, line 84: What do we think about putting the test for each fuzzer's get_stats here?
The more intuitive alternative is to create fuzzers/afl/test_fuzzer.py and put it there.
But I don't really love the idea of polluting fuzzer directories with this (and with test files as well).

database/models.py
lszekeres: Do we agree to separate common metrics from fuzzer-specific ones?

If yes, then re. type/schema validation:
- Fuzzer specific metrics wouldn't have a schema, users would be allowed to provide arbitrary JSONs, right? I.e., if json.dumps() succeeds, it's valid and good to go. (We could also statically pytype-check that the fuzzer.py's get_stats() returns a dict.)
- For common metrics, let's look at them case by case:
  - Stuff like `cpu_usage` and `mem_usage` wouldn't be provided by the `fuzzer.py`, but by fuzzbench itself. No special type/schema validation is necessary.
  - For `num_execs` we need to rely on `fuzzer.py`. I'd collect this data by letting fuzzers define a `def get_num_execs() -> int` function. Early type/schema validation is done by pytype (`-> int`) at linting time.

PR #649, opened 2020-08-10
[docs] Update gem dependency kramdown

jonathanmetzman:


jonathanmetzman:
This is in response to a Github security advisory. To be honest, the advisory seems like nothing to worry about since it warns >The kramdown gem before 2.3.0 for Ruby processes the template option inside Kramdown documents by default, which allows unintended read access (such as template="/etc/passwd") or unintended embedded Ruby code execution (such as a string that begins with template="string://<%= `). NOTE: kramdown is used in Jekyll, GitLab Pages, GitHub Pages, and Thredded Forum.
However, all of the templates are committed code and are trusted and we don't really run this code in production.
But good to be safe and get rid of the warning when pushing.

PR #650, opened 2020-08-10
separate job

jonathanmetzman:


common/benchmark_config.py
```diff
@@ -30,3 +30,5 @@ def get_config_file(benchmark):
 def get_config(benchmark):
     """Returns a dictionary containing the config for a benchmark."""
     return yaml_utils.read(get_config_file(benchmark))
+
```
jonathanmetzman, line 33: I don't know why this isn't triggering presubmit to run in CI.

inferno-chromium:
Causing sporadic failures on end-to-end test for me as well [https://github.com/google/fuzzbench/runs/971664724]. I dont care, but for others, it leads to failure emails, so we need to mark this test as failing or skip somehow. @lszekeres

lszekeres:
> Causing sporadic failures on end-to-end test for me as well [https://github.com/google/fuzzbench/runs/971664724]. I dont care, but for others, it leads to failure emails, so we need to mark this test as failing or skip somehow. @lszekeres

Looking at the linked log, I think there's a good chance that this started failing due to https://github.com/google/fuzzbench/pull/653, which reverted some images_types.yaml changes. I can look into that and fix it later.

PR #653, opened 2020-08-10
Revert "Use single source of truth for building images for Local and GCB build systems"

inferno-chromium:


inferno-chromium:
TBR

jonathanmetzman:
This doesn't revert makefile generation. So makefiles are using buildkit and may or may not have this issue still while GCB does not. Also, this reverts tanishq's fix for buildkit on CI. Since CI still uses make it still uses buildkit and will again have the issue he fixed.

inferno-chromium:
> This doesn't revert makefile generation. So makefiles are using buildkit and may or may not have this issue still while GCB does not. Also, this reverts tanishq's fix for buildkit on CI. Since CI still uses make it still uses buildkit and will again have the issue he fixed.

That makefile generation was a massive CL from @lszekeres that i didnt want to revert. I wont have cycles for this. we can leave things for 2-3 weeks, but need to remember to come back to it.

jonathanmetzman:
This broke in CI before it was merged and will probably break all over the place now as well: https://github.com/google/fuzzbench/runs/967556385

Issue #647, opened 2020-08-09
experiment merging logic is broken


inferno-chromium:
add_nonprivate_experiments_for_merge_with_clobber is too buggy.

1. Merges stuff with non-completed experiments. E.g. 2020-08-03 result was merged with an incomplete 2020-08-07.
2. Merges stuff with newer experiments. E.g. 2020-08-03 was merged with a 2020-08-07.
3. And merges stuff with unneeded fuzzers, like nothing should be merged with routine variants test. @lszekeres brought this up to put a static list of interesting fuzzers to merge against.

There is also the issue with enough samples, somehow samples in 2020-08-03 are not enough.

PR #658, opened 2020-08-11
Fix report merging logic.

inferno-chromium:
- Don't merge with newer experiments.
- Fix experiment data clobbering not working with
merge_with_clobber_nonprivate.
- Choose only important fuzzers from other experiments,
  not variants.
- Choose only ended experiments.

Fixes https://github.com/google/fuzzbench/issues/647


database/models.py
```diff
@@ -25,6 +25,7 @@ class Experiment(Base):
 
     name = Column(String, nullable=False, primary_key=True)
     time_created = Column(DateTime(), server_default=sqlalchemy.func.now())
+    time_ended = Column(DateTime(), nullable=True)
```
jonathanmetzman, line 28: Given how often we shut things down early I think this is going to be pretty fragile.

inferno-chromium: We shouldn't. And there should be a field to indicate experiment finished, otherwise there isn't one. I feel service is working fine, what is the issue ? Also, this just means that incomplete experiment won't be used for merging (like if you shut it off), which also makes sense.

common/experiment_constants.py
```diff
@@ -0,0 +1,34 @@
...
+    'aflfast',
+    'aflplusplus',
+    'aflplusplus_optimal',
+    'aflsmart',
+    'eclipser',
+    'entropic',
+    'fairfuzz',
+    'fastcgs_lm',
+    'honggfuzz',
+    'honggfuzz_qemu',
```
jonathanmetzman, line 29: I think aflplusplus_qemu is more essential.

common/experiment_constants.py
```diff
@@ -0,0 +1,34 @@
...
+    'afl',
+    'afl_qemu',
+    'aflfast',
+    'aflplusplus',
+    'aflplusplus_optimal',
+    'aflsmart',
+    'eclipser',
+    'entropic',
+    'fairfuzz',
+    'fastcgs_lm',
```
jonathanmetzman, line 27: `fastcgs_lm` is still in development so I think we can probably leave it out.

database/models.py
jonathanmetzman: Large scale preemptible experiments seem to be broken and so they need to be shutdown manually or else they never finish.
Obviously that issue should be fixed and I think we should also add some sanity checking to measuring so that we dont retry forever.

experiment/reporter.py
```diff
@@ -42,6 +43,10 @@ def output_report(experiment_config: dict,
 
     reports_dir = get_reports_dir()
 
+    fuzzers = sorted(
+        set(experiment_config['fuzzers'].split(',')).union(
```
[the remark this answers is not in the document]
inferno-chromium: filing a bug instead, this is in too many places, for both fuzzers and benchmarks.

PR #660, opened 2020-08-11
Fix issue with multiple mounts and don't pull images in CI anymore

jonathanmetzman:


.github/workflows/build_and_test_run_fuzzer_benchmarks.py
```diff
@@ -81,18 +80,14 @@ def delete_docker_images():
...
-    make_targets = get_make_targets(benchmarks, fuzzer)
-    for pull_target, build_target in make_targets:
-        # Pull target first.
-        subprocess.run(['make', '-j', pull_target], check=False)
-
-        # Then build.
-        build_command = ['make', 'RUNNING_ON_CI=yes', '-j', build_target]
-        print('Running command:', ' '.join(build_command))
-        result = subprocess.run(build_command, check=False)
+    test_targets = get_make_targets(benchmarks, fuzzer)
```
inferno-chromium, line 86: did you mean make_targets here.

lszekeres:
Will this allow re-enabling buildkit, or that was an orthogonal issue?

jonathanmetzman:
It is necessary to reenable buildkit on GCB but it does not do that. It simply enables its use on GitHub actions since the revert included a fix that was needed.

PR #661, opened 2020-08-11
Temporarily remove end-to-end queue test for queue from CI.

jonathanmetzman:
Test is flaky. It can be readded when it is not flaky or if it always
returns 0.


lszekeres:
> Test is flaky. It can be readded when it is not flaky or if it always
> returns 0.

LGTM, but didn't you have another PR that made it a separate job and experimental? Did that not work?

inferno-chromium:
Merging as CI keeps failing.

PR #662, opened 2020-08-11
Add filestore column

Shadoom7:
This patch add a `experiment_filestore_name` column in the `models.Experiment`.


lszekeres:
> This patch add a `experiment_filestore_name` column in the `models.Experiment`.

Why not the store the full path to the experiment's output directory? More explicit data is usually better because it's more clear what it is and why it is there, and also we wouldn't need to construct the full path in the code.

database/alembic/versions/c83ac04855b4_add_filestore_name.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = 'c83ac04855b4'
+down_revision = '77022369cea4'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('experiment', sa.Column('experiment_filestore_name', sa.String(), nullable=False))
```
inferno-chromium, line 21: indent this file.

experiment/dispatcher.py
```diff
@@ -61,10 +61,12 @@ def _initialize_experiment_in_db(experiment_config: dict):
...
-        db_utils.get_or_create(models.Experiment,
-                               name=experiment_config['experiment'],
-                               git_hash=experiment_config['git_hash'],
-                               private=experiment_config.get('private', True))
+        db_utils.get_or_create(
+            models.Experiment,
+            name=experiment_config['experiment'],
+            git_hash=experiment_config['git_hash'],
+            private=experiment_config.get('private', True),
+            experiment_filestore_name=experiment_config['experiment_filestore'])
```
inferno-chromium, line 69: experiment_ is not needed, why not just simple "filestore"

database/alembic/versions/c83ac04855b4_add_filestore_name.py
inferno-chromium: You also need to alter current rows to have some values, otherwise this fails.

experiment/dispatcher.py
jonathanmetzman: Let's keep experiment_filestore since this is the same name as the config param. I think having multiple names is confusing.

Shadoom7:
Thank you all for the suggestions! @lszekeres @inferno-chromium @jonathanmetzman 
I still got a few questions though.
1. Should I store `gs://fuzzbench-data/[EXPERIMENT NAME]` or just `gs://fuzzbench-data`?
2. If I set it to `nullable=False` as @jonathanmetzman suggested, should I still update the current row?

inferno-chromium:
> Thank you all for the suggestions! @lszekeres @inferno-chromium @jonathanmetzman
> I still got a few questions though.
> 
> 1. Should I store `gs://fuzzbench-data/[EXPERIMENT NAME]` or just `gs://fuzzbench-data`?

Just gs://fuzzbench-data
experiment name is already available in other field.

> 2. If I set it to `nullable=True` as @jonathanmetzman suggested, should I still update the current row?

you probably need it to update current rows to have some value in that upgrade() function

> 3. If I need to update the current row, how should I get the correct filestore name (to check if it is `fuzzbench` experiment or a `fuzzbench-test` one)? Is it possible?

just update it to fuzzbench. add a comment that change it to project id.

jonathanmetzman:
> Thank you all for the suggestions! @lszekeres @inferno-chromium @jonathanmetzman
> I still got a few questions though.
> 
> 1. Should I store `gs://fuzzbench-data/[EXPERIMENT NAME]` or just `gs://fuzzbench-data`?

Heh, based on laszlo's comment (which I agree with) that it's better to store the full path rather than construct, `gs://fuzzbench-data/[EXPERIMENT NAME]` is better. But then the column doesn't match what `experiment_filestore` means elsewhere. I think we call `gs://fuzzbench-data` the `experiment filestore` while `gs://fuzzbench-data/[EXPERIMENT NAME]` is the `experiment_filestore_path`. I would just store `gs://fuzzbench-data/[EXPERIMENT NAME]` and name the column `experiment_filestore_path`. I'll probably refactor things later so that the distinction is clearer

> 2. If I set it to `nullable=True` as @jonathanmetzman suggested, should I still update the current row?

When we did the git_hash addition we faced a similar issue. We were adding a field that we could not add to old entities in the DB so we left it as `nullable=True`. I think this is probably the best solution but I really don't know what the right practice is here (maybe we should do some research)? It probably doesn't matter either way. The alternative is to put an empty string or some similar value for old entities that don't have this.

> 3. If I need to update the current row, how should I get the correct filestore name (to check if it is `fuzzbench` experiment or a `fuzzbench-test` one)? Is it possible?

Right I think this would be impossible to do correctly in a generic migration script. The problem is even worse when you consider that there are open source users that the script should work for. They are using filestores that aren't my-fuzzbench-data or fuzzbench-data.

jonathanmetzman:
Sorry I didn't see this comment before I added mine.

> > 1. Should I store `gs://fuzzbench-data/[EXPERIMENT NAME]` or just `gs://fuzzbench-data`?
> 
> Just gs://fuzzbench-data
> experiment name is already available in other field.

This is correct as well. Although I had a different opinion. I say just do what abhishek says here since doing something fast is more important than bikeshedding.

Shadoom7:
Have updated the db for `fuzzbench`.

PR #666, opened 2020-08-12
Integration of the matio project

Vaush:
Integrating the OSSFuzz benchmark for the project "matio", with the fuzz target "matio_fuzzer", dated 2019-09-04 03:05.


benchmarks/matio_matio_fuzzer_201909040305/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Copyright 2019 Google Inc.
```
jonathanmetzman, line 1: let's just call this benchmark "matio_matio_fuzzer"

jonathanmetzman:
If you pull from master you won't get the annoying presubmit failures.

jonathanmetzman:
You also need to add this benchmark to https://github.com/google/fuzzbench/blob/master/.github/workflows/build_and_test_run_fuzzer_benchmarks.py#L37 (please keep list in alpha order) to get this tested in CI.

benchmarks/matio_matio_fuzzer/Dockerfile
```diff
@@ -21,3 +21,5 @@ RUN git clone --depth 1 git://git.code.sf.net/p/matio/matio matio
 WORKDIR matio
 COPY build.sh $SRC/
 COPY *.cc *.h $SRC/
+COPY testcases/ $SRC/
+COPY testcases/ $SRC/
```
jonathanmetzman, line 25: Is this second copy needed?

Vaush: It shouldn't even be there yet, it's part of the testing I was doing on another pull request I am still working on. I will remove both of them since they are still not needed at the moment.
Thanks for noticing that

Vaush:
@jonathanmetzman Do you know why the checks are failing, but only on oss-fuzz fuzzers (whatever that means)? The previous version didn't fail any checks, and I haven't changed that much. If it's something to be expected, that's good, otherwise I will need to look into why it won't compile with those ones.
If it's not straightforward, could someone tell me what should be the difference between oss-fuzz and standard checks? Because regularly the standard check passes, and the oss-fuzz one doesn't.

jonathanmetzman:
Does make RUNNING_ON_CI=yes -j test-run-aflcc-matio_matio_fuzzer work locally?
You can inspect the output by clicking on "Details" on the failing tests.

jonathanmetzman:
Actually I suspect I know the issue. Can you pull please?

Vaush:
I run the command and it doesn't work even locally.
Also, I can't pull anything from master, I am up-to-date already, maybe you meant some other branch?

jonathanmetzman:
> I run the command and it doesn't work even locally.
> Also, I can't pull anything from master, I am up-to-date already, maybe you meant some other branch?

`git pull origin master` doesn't work?  I'll try to debug this today.

jonathanmetzman:
Sorry I don't think the issue has to do with pulling from master.
I think these fuzzers genuinely can't build this benchmark.
Can you blocklist these as is done here: https://github.com/google/fuzzbench/blob/master/benchmarks/bloaty_fuzz_target/benchmark.yaml#L19

Vaush:
Sure I can. Is there a particular way of only blocking the fuzzer for its "oss-fuzz" version (the ones that fail in testing), or failing one of the two versions is enough for us to establish the fuzzer should be excluded altogether?

jonathanmetzman:
OSS-Fuzz vs standard is a distinction we have between benchmarks purely to divide up benchmarks for building in parallel in CI.
Your benchmark is an OSS-Fuzz benchmark.
I only see failures when building OSS-Fuzz benchmarks.
This says to me that the fuzzers are failing to build your benchmark.

Vaush:
I see, it makes sense. Ok, tomorrow morning I will blocklist the ones that don't work, and update the PR.

jonathanmetzman:
> I see, it makes sense. Ok, tomorrow morning I will blocklist the ones that don't work, and update the PR.

Looking forward to it. I recommend doing this ASAP as the list of fuzzers changes nearly daily so each day this is delayed is another change that needs to be made.

Vaush:
I added the list, let's see if there are any new fuzzers that I might have missed. In general, 25 out of 28 unsupported fuzzers are either a version of aflplusplus or a qemu fuzzer, with the other 3 being aflcc, klee and libaflfuzzer.

jonathanmetzman (approved):

inferno-chromium:
Looks like you forgot to link with zlib, see https://github.com/google/fuzzbench/pull/797#issuecomment-705590737

vanhauser-thc:
> Looks like you forgot to link with zlib, see [#797 (comment)](https://github.com/google/fuzzbench/pull/797#issuecomment-705590737)

To be more precise, the mate fuzz target compile command is missing -lz

inferno-chromium:
> > Looks like you forgot to link with zlib, see [#797 (comment)](https://github.com/google/fuzzbench/pull/797#issuecomment-705590737)
> 
> To be more precise, the mate fuzz target compile command is missing -lz

yes already uploading fix, thanks for pointing the issue.

PR #671, opened 2020-08-13
Sorry one more fix necessary

vanhauser-thc:
the taint variant implementation is quite complex and I found another crash scenario :(

I also merged the libfuzzer/entropic experiment into this one.
If you could stop the experiment - if this PR comes too late for the new start - and restart it ... thank you!


inferno-chromium:
Yes, the experiments are already started, need to add this in new one.

vanhauser-thc:
I closed it because there is one more issue .... which I am so far unable to track down.
seems to be that taint afl-fuzz runs out of stack, which makes no sense from both the call stack/backtrace and the code ...

PR #674, opened 2020-08-13
[scheduler] Use instance api instead of operations to find preempted instances accurately

jonathanmetzman:
The previous query for preemption operations seems to return incomplete
results in many-trial experiments. Hopefully this one returns correct results thought it isn't
the cleanest solution.


jonathanmetzman:
Some tests still fail because I didn't update mocks.

common/gce.py
```diff
@@ -31,60 +29,14 @@ def initialize():
...
-                project=experiment_config['cloud_project'],
-                zone=experiment_config['cloud_compute_zone'])
-
-
-def get_instance_from_preempted_operation(operation, base_target_link) -> str:
-    """Returns the instance name from a preempted |operation|."""
-    return operation['targetLink'][len(base_target_link):]
+        for instance in response['items']:
+            if instance['status'] == 'TERMINATED':
+                yield instance['name']
```
inferno-chromium, line 40: nit: no reason to yield since you are converting to list anyway in one call site.

inferno-chromium:
@googlebot I consent.

inferno-chromium:
Testing in 2020-08-16.

experiment/scheduler.py
```diff
@@ -106,8 +106,12 @@ def all_trials_ended(experiment: str) -> bool:
 
 def delete_instances(instances, experiment_config):
     """Deletes |instances|."""
-    running_instances = gcloud.list_instances()
-    instances_to_delete = [i for i in instances if i in running_instances]
+    cloud_project = experiment_config['cloud_project']
```
jonathanmetzman, line 109: I'm not sure this works 100% properly.
I think the old code also got terminated instances (aka instances that aren't running but were stopped), despite the variable running_instances implying otherwise.
I'm not sure the bug here is a big deal though, worst case scenario we might not delete some preempted instances (in which case we might have to pay for their disk space).

jonathanmetzman: Actually nevermind, this calls into an equivalent function.

jonathanmetzman:
Thanks for working on this!

inferno-chromium:
Seems to be working and creating new instances, can reopen if experiment misses trials significantly, merging for now.

PR #678, opened 2020-08-14
Modify the patch for the libfuzzer/entropic's keepseed variants

dokyungs:
This patch modifies the patch for the libfuzzer/entropic's keepseed variants such that (i) seed inputs are not replaced even if -reduce_inputs=1, and (ii) the crossover mutator uses a uniform probability distribution when selecting inputs to cross over with, while using the original probability distribution computed by either Vanilla or Entropic schedule when selecting an input to mutate.


dokyungs:
> Can you add an experiment request as well (last one had to be removed for some other testing). I will merge this on weekend after current one is finished.

Added. Thanks!

PR #679, opened 2020-08-14
LTO reevaluation

vanhauser-thc:
LTO mode performs the instrumentation at link time allowing for collision free coverage.
However also optimization is done at link time.

Previously laf-intel and cmplog instrumentations were performed at compile time, and then the link time optimization re-optimized the instrumentation where it could ...

so in this new reeval laf and cmplog have been moved to link time.
unsurprising now things are as they should.

a bit surprising is however the edge explosion. e.g. bloaty with laf-intel we now get 750k edges! Not to mention the looooong time it takes to link the binary.

if this is still featisble to fuzz efficiently, or not and which other benchmarks are affected and what to learn from that - this is the purpose of this PR.

for _laf and _cmplog not all benchmark will most likely build. that is to be expected and ok, as some build times will be too long.

it will take some time however to to figure out which benchmarks e.g. with laf-intel possible to perform and which dont.
and to fix those which can.

I will @ mention once it is mergable :)


vanhauser-thc:
@inferno-chromium @jonathanmetzman 
why are the CI not starting? did I do something wrong?

vanhauser-thc:
@jonathanmetzman @lszekeres @inferno-chromium  it is ready to land. 
thank you!

vanhauser-thc:
@jonathanmetzman thank you! if you want to can put the start of the experiment a day earlier, I wanted enough buffer just in case ... I know you still have performance week ;)

jonathanmetzman:
The name doesn't affect when it gets run. It will run tonight at 6 PM PT.
The real reason why I make people add the names is so that experiment-requests.yaml has a nice mapping of experiments to reports.
I might merge this experiment with the libFuzzer experiment running today.

vanhauser-thc:
will that have the new report format? :)
Ah maybe also add weizz, andrea updated it in seperate PR.

PR #680, opened 2020-08-14
Fix coverage reports missing file issue

Shadoom7:
Some source file html are broken in the current coverage report because those files are missing.
A deeper reason is that some benchmarks only extract or build their complete source files after running `build.sh`.

This patch should fix this issue by copying the source files in `benchmark-builder` instead of in the coverage dockerfile.

### DIFF
--- a/docker/benchmark-builder/Dockerfile
+++ b/docker/benchmark-builder/Dockerfile
@@ -51,5 +51,7 @@ RUN export CHECKOUT_COMMIT=$(cat /benchmark.yaml | tr -d ' ' | grep 'commit:' | 
 RUN PYTHONPATH=$SRC python3 -u -c "from fuzzers import utils; utils.initialize_flags(); import fuzzer; fuzzer.build()"
 
 # Copy source code files of benchmark to $OUT for the report generation.
+# TODO(ruikangshi): Implement it in makefile generator and in the yaml file
+# so we could save the space in docker image.
 RUN mkdir $OUT/src && \
     cp -rL --parent $SRC $WORK $OUT/src
--- a/docker/benchmark-builder/Dockerfile
+++ b/docker/benchmark-builder/Dockerfile
@@ -49,3 +49,7 @@ RUN export CHECKOUT_COMMIT=$(cat /benchmark.yaml | tr -d ' ' | grep 'commit:' | 
     python3 -u /opt/fuzzbench/checkout_commit.py $CHECKOUT_COMMIT $SRC
 
 RUN PYTHONPATH=$SRC python3 -u -c "from fuzzers import utils; utils.initialize_flags(); import fuzzer; fuzzer.build()"
+
+# Copy source code files of benchmark to $OUT for the report generation.
+RUN mkdir $OUT/src && \
+    cp -rL --parent $SRC $WORK $OUT/src
--- a/docker/benchmark-builder/Dockerfile
+++ b/docker/benchmark-builder/Dockerfile
@@ -49,9 +49,3 @@ RUN export CHECKOUT_COMMIT=$(cat /benchmark.yaml | tr -d ' ' | grep 'commit:' | 
     python3 -u /opt/fuzzbench/checkout_commit.py $CHECKOUT_COMMIT $SRC
 
 RUN PYTHONPATH=$SRC python3 -u -c "from fuzzers import utils; utils.initialize_flags(); import fuzzer; fuzzer.build()"
-
-# Copy source code files of benchmark to $OUT for the report generation.
-# TODO(ruikangshi): Implement it in makefile generator and in the yaml file
-# so we could save the space in docker image.
-RUN mkdir $OUT/src && \
-    cp -rL --parent $SRC $WORK $OUT/src
--- a/docker/benchmark-runner/Dockerfile
+++ b/docker/benchmark-runner/Dockerfile
@@ -60,8 +60,6 @@ ARG fuzzer
 # Copy over all the build artifacts (without * to preserve directory structure).
 # This also copies seed and dictionary files if they are available.
 COPY --from=builder /out/ ./
-# Delete useless source files for coverage build.
-RUN rm -rf ./src
 # Copy the fuzzer.py file.
 COPY fuzzers/$fuzzer/fuzzer.py .
 # Copy the fuzzers directory.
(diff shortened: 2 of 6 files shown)


jonathanmetzman:
> Some source file html are broken in the current coverage report because those files are missing.
> A deeper reason is that some benchmarks only extract or build their complete source files after running `build.sh`.

Which benchmarks are these?

> 
> This patch should fix this issue by copying the source files in `benchmark-builder` instead of in the coverage dockerfile.

I don't love this idea because I don't want to increase the sizes of our runners by including stuff from SRC and WORK.
Would it be possible to copy everything in $OUT but $OUT/src?

Shadoom7:
> > Some source file html are broken in the current coverage report because those files are missing.
> > A deeper reason is that some benchmarks only extract or build their complete source files after running `build.sh`.
> 
> Which benchmarks are these?

From what I see, some of bloaty's files are built in build.sh. And libpng extract all of its source file in build.sh. But I've only checked those two, I believe other benchmarks may have similar issue.

> 
> > This patch should fix this issue by copying the source files in `benchmark-builder` instead of in the coverage dockerfile.
> 
> I don't love this idea because I don't want to increase the sizes of our runners by including stuff from SRC and WORK.
> Would it be possible to copy everything in $OUT but $OUT/src in the runner's Dockerfile?

I'm not sure I understand you. But I think coverage build won't involve any runner's Dockerfile.

Shadoom7:
I think another way is to copy it in `docker/gcb/coverage.yaml`. This won't increase the size of any runner. But it assumes that $SRC = src/, $WORK=work and $OUT=out for all dockers. Should I do this? @jonathanmetzman

jonathanmetzman:
> > > Some source file html are broken in the current coverage report because those files are missing.
> > > A deeper reason is that some benchmarks only extract or build their complete source files after running `build.sh`.
> > 
> > 
> > Which benchmarks are these?
> 
> From what I see, some of bloaty's files are built in build.sh. And libpng extract all of its source file in build.sh. But I've only checked those two, I believe other benchmarks may have similar issue.
> 
> > > This patch should fix this issue by copying the source files in `benchmark-builder` instead of in the coverage dockerfile.
> > 
> > 
> > I don't love this idea because I don't want to increase the sizes of our runners by including stuff from SRC and WORK.
> > Would it be possible to copy everything in $OUT but $OUT/src in the runner's Dockerfile?
> 
> I'm not sure I understand you. But I think coverage build won't involve any runner's Dockerfile.

benchmark-builder is the last builder image that gets built for a fuzzer-benchmark pair. When it is done building $OUT is supposed to contain the fuzzer and the benchmark/target. When the runners get built, one of the images copies everything in $OUT. Your change means the runners will now include $WORK and $SRC. I was suggesting that when this copy happens, you ensure that $OUT/src isn't copied into the runner.

>I think another way is to copy it in docker/gcb/coverage.yaml. This won't increase the size of any runner. But it assumes that $SRC = src/, $WORK=work and $OUT=out for all dockers.

I think that works too but you will have to implement in the code that generates make as well. Let's try implementing my suggestion first. Let me know if I still haven't explained it well enough.

Shadoom7:
I was going to use `.dockerignore` but the building context is the whole project which means it may affect other docker build.
So I used the stupidest way to do it in the end. (To delete it after copying it)
Please tell me if there's another way around. @jonathanmetzman

jonathanmetzman:
> I was going to use `.dockerignore` but the building context is the whole project which means it may affect other docker build.
> So I used the stupidest way to do it in the end. (To delete it after copying it)
> Please tell me if there's another way around. @jonathanmetzman

Unfortunately I think because of the way docker works the space taken up by those files is still included in the previous layer of the final docker image. Maybe land for now and assign a TODO to do it in make/gcb-yaml if we get a chance?

Shadoom7:
> > I was going to use `.dockerignore` but the building context is the whole project which means it may affect other docker build.
> > So I used the stupidest way to do it in the end. (To delete it after copying it)
> > Please tell me if there's another way around. @jonathanmetzman
> 
> Unfortunately I think because of the way docker works the space taken up by those files is still included in the previous layer of the final docker image. Maybe land for now and assign a TODO to do it in make/gcb-yaml if we get a chance?

Ahh right, the layer. Yeah I'll try to fix it if I get time. Have updated the TODO.

jonathanmetzman:
I guess why are we putting int $OUT to begin with? Is it so that the copy we do in coverage.yaml and make includes them in the tarball?
If so I think this is not really the right solution (the right one is to edit that taring so that it includes those directories).

Shadoom7:
> I guess why are we putting int $OUT to begin with? Is it so that the copy we do in coverage.yaml and make includes them in the tarball?
> If so I think this is not really the right solution (the right one is to edit that taring so that it includes those directories).

It's true. So we should only edit the taring command in `coverage.yaml` and the one in `local_build` right?

jonathanmetzman:
> > I guess why are we putting int $OUT to begin with? Is it so that the copy we do in coverage.yaml and make includes them in the tarball?
> > If so I think this is not really the right solution (the right one is to edit that taring so that it includes those directories).
> 
> It's true. So we should only edit the taring command in `coverage.yaml` and the one in `local_build` right?

Yes. I don't want this to block you but if you think that solution would be easier I'd prefer it. Sorry for suggesting the wrong path initially.

Shadoom7:
> > > I guess why are we putting int $OUT to begin with? Is it so that the copy we do in coverage.yaml and make includes them in the tarball?
> > > If so I think this is not really the right solution (the right one is to edit that taring so that it includes those directories).
> > 
> > 
> > It's true. So we should only edit the taring command in `coverage.yaml` and the one in `local_build` right?
> 
> Yes. I don't want this to block you but if you think that solution wouldn't be too hard I'd prefer it. Sorry for suggesting the wrong path initially.

No worries! It's some tiny changes, I'll finish and test it this evening.

jonathanmetzman:
It looks like the approach this CL takes breaks in CI anyway: https://github.com/google/fuzzbench/pull/680/checks?check_run_id=986869872

inferno-chromium:
How big of a issue is this, is this needed before the next big experiment ?

Shadoom7:
> How big of a issue is this, is this needed before the next big experiment ?

I think it's kinda important. Many coverage reports are crashed right now.
I've ran a test today. This patch is ready to go. Better be merged before next experiment.

PR #681, opened 2020-08-16
Fixes for weizz

andreafioraldi:
The weizz script that Marc pushed here it is for fork-based qemu mode, let's specify it.
I updated also the commit id, there are some fixes.

The previous run with weizz failed on most targets problably due to the lack of -m none, but I didn't tested weizz on docker and the fuzzer can crash if the memory is too low. weizz is an academic poc, it can allocates even some gigabytes of memory as hashmap in the worst case on large targets to approximate taint tracking.

Interestingly but not surprisingly, it was great on chunk based formats (the ones that it is designed for) like tls and png and bad on others like sql.


andreafioraldi:
I'm interested to see also aflsmart in the next run in which there is weizz because weizz is basically aflsmart with format reverse engineering instead of with an human written model.

PR #683, opened 2020-08-17
add afltriple

datadancer:
I made changes to afl and want to test it's performance. I added the fuzzer according to the docs and test a litttle using my local machine. Now I want to test and compare it with other fuzzers. Thanks a lot.


datadancer:
@googlebot I fixed it.

PR #685, opened 2020-08-17
Some fix and improvements on differential plots.

Shadoom7:
This patch adds a new ranking graph to summarize the rank of each fuzzer based on their unique coverage on each benchmark.
Screenshot 2020-08-17 at 5 45 15 PM

This patch also adds space between lines of fuzzer buttons in coverage reports collapsible box.

### DIFF
--- a/analysis/experiment_results.py
+++ b/analysis/experiment_results.py
@@ -105,17 +105,18 @@ class ExperimentResults:  # pylint: disable=too-many-instance-attributes
         """A pivot table of medians for each fuzzer on each benchmark."""
         return data_utils.experiment_pivot_table(
             self._experiment_snapshots_df, data_utils.benchmark_rank_by_median)
-    
+
     @property
-    def benchmark_unique_coverage_list(self):
+    def _benchmarks_unique_coverage_list(self):
         """A list containing unique coverage dataframe for each benchmark."""
-        return [benchmark._unique_region_cov_df for benchmark in self.benchmarks]
-    
+        return [benchmark.unique_region_cov_df for benchmark in self.benchmarks]
+
     @property
     def rank_by_unique_coverage_average_normalized_score(self):
         """Rank fuzzers using average normalized score on unique coverage across
         benchmarks."""
-        return coverage_data_utils.rank_by_unique_coverage_average_normalized_score(self.benchmark_unique_coverage_list)
+        return coverage_data_utils.rank_by_average_normalized_score(
+            self._benchmarks_unique_coverage_list)
 
     @property
     def rank_by_average_rank_and_average_rank(self):
--- a/analysis/experiment_results.py
+++ b/analysis/experiment_results.py
@@ -17,6 +17,7 @@ import functools
 import os
 
 from analysis import benchmark_results
+from analysis import coverage_data_utils
 from analysis import data_utils
 from analysis import stat_tests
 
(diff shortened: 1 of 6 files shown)


analysis/report_templates/default.html
```diff
@@ -96,6 +96,13 @@ <h5 class="center-align">By avg. rank</h5>
                     {{ experiment.rank_by_median_and_average_rank.round(2).to_frame().to_html() }}
                 </div>
             </div>
+            
+            <div class="row">
+                <div class="col s5 offset-s1">
+                    <h5 class="center-align">By avg. score on unique coverage</h5>
+                    {{ experiment.rank_by_unique_coverage_average_normalized_score.round(2).to_frame().to_html() }}
+                </div>
+            </div>
```
lszekeres, line 105: Considering adding this to `analysis/report_templates/experimental.html` instead.

I don't think ranking by unique coverage can be considered a typical / usual way of ranking fuzzers. I wouldn't include it by default in the main report.

analysis/experiment_results.py
```diff
@@ -105,6 +106,18 @@ def summary_table(self):
         return data_utils.experiment_pivot_table(
             self._experiment_snapshots_df, data_utils.benchmark_rank_by_median)
 
+    @property
+    def _benchmarks_unique_coverage_list(self):
+        """A list containing unique coverage dataframe for each benchmark."""
+        return [benchmark.unique_region_cov_df for benchmark in self.benchmarks]
```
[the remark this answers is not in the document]
Shadoom7: Yeah good point! I'll remove this one.

analysis/report_templates/default.html
inferno-chromium: +++++++1, still very early and we need to verify results and don't think it should be used for ranking.

analysis/report_templates/default.html
```diff
@@ -274,7 +281,9 @@ <h5 class="center-align">Pairwise unique coverage</h4>
                         <div class="row">
                             {% for fuzzer in benchmark.fuzzer_names %}
                             <div class="col">
-                                <a class="waves-effect waves-light btn" href="{{ benchmark.get_filestore_name(fuzzer) }}/coverage/reports/{{ benchmark.name }}/{{ fuzzer }}/index.html">{{ fuzzer }}</a>
+                                <p>
+                                    <a class="waves-effect waves-light btn" href="{{ benchmark.get_filestore_name(fuzzer) }}/coverage/reports/{{ benchmark.name }}/{{ fuzzer }}/index.html">{{ fuzzer }}</a>
+                                </p>
```
lszekeres, line 286: Buttons are most typically used for interacting with the current page, and rarely for links.
Consider using a collection of links instead: https://materializecss.com/collections.html#links
(Perhaps even with an "external link" icon: https://fontawesome.com/icons/external-link-alt)

analysis/report_templates/default.html
Shadoom7: Got it!

analysis/report_templates/default.html
Shadoom7: The collection list would be too long in a full experiment. Does it matter?

lszekeres: Since we put them in a collapsibles, i think it's fine, no?

Shadoom7: Yeah that makes sense. I'll change it.

Shadoom7:
Now the `Pairwise unique coverage` plots have the fuzzers in order of their unique coverage.

PR #687, opened 2020-08-18
Update llvm/compiler-rt upstream commit number for libfuzzer and entropic

dokyungs:
This patch updates llvm/compiler-rt upstream commit number for libfuzzer and entropic, and removes now-stale variants, fixcrossover and interceptors.


dokyungs:
@morehouse FYI, this patch bumps up the LLVM/compiler-rt version for both libfuzzer and entropic.

jonathanmetzman:
Ughh...why so many queued jobs in CI.

PR #688, opened 2020-08-18
Add a temporary entropic variant for testing scheduling policy that uses exectime

dokyungs:
This patch adds an entropic variant to test a scheduling policy that prioritizes inputs with lower execution time.

In my local environment, this gives a throughput increase as well as a coverage increase on the SQLITE3 benchmark. An experiment is requested to see its effect on a larger scale.

### DIFF
--- a/.github/workflows/fuzzers.yml
+++ b/.github/workflows/fuzzers.yml
@@ -37,6 +37,7 @@ jobs:
           # Temporary variants.
           - libfuzzer_keepseed
           - entropic_keepseed
+          - entropic_exectime
           - aflplusplus_classic
           - aflplusplus_lto_pcguard
           - aflplusplus_lto
--- a/fuzzers/entropic_exectime/builder.Dockerfile
+++ b/fuzzers/entropic_exectime/builder.Dockerfile
@@ -1,0 +1,28 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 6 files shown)


jonathanmetzman:
> This patch adds an entropic variant to test a scheduling policy that prioritizes inputs with lower execution time.
> 

Interesting, AFL uses this approach. I think (could very well be wrong) AFL executes new interesting inputs multiple times so the time it uses is an average and thus not so prone to weird variance. Might be something to consider.
Also, if this gets upstreamed you'll probably have to disable the feature when a random seed is specified


> In my local environment, this gives a throughput increase as well as a coverage increase on the SQLITE3 benchmark. An experiment is requested to see its effect on a larger scale.

Nice!

dokyungs:
@jonathanmetzman I just rebased the patch and wonder whether my experiment request would still be okay. Could you please review?

mboehme:
Awesome @dokyungs! There are some really nice improvements!

I actually tried to accomodate prioritizing by execution time in the Entropic review for LibFuzzer, but it didn't land.
KCC had some concerns about execution time as source of non-determinism between campaigns that are started with the same random seed: 
https://reviews.llvm.org/D73776?id=257332#1997664
https://reviews.llvm.org/D73776?id=257332#1998190
You can get a significant performance boost, but at the cost of deterministic crash reproduction on stateful programs.

It actually lived on in FuzzBench for a while: https://github.com/google/fuzzbench/commit/de1b86acab087c26a5828cb7b817735abc089af0#diff-7521e1961f04599fc99be889564117c6R332

Maybe it could be an experimental option?

mboehme:
@dokyungs The only other tweak that I did over upstream Entropic was to disable 5min restarts for FuzzBench:
https://github.com/google/fuzzbench/pull/467

My thinking was that LibFuzzer in Fuzzbench is run with `-ignore_crashes` which restarts LF every 5 minutes. Incidence counts will get lost. The patch still restarts LF after a crash, but not after every 5 minutes.

mboehme:
Oh, it landed 3 days ago 👍. That was quick!

https://github.com/llvm/llvm-project/commit/5cda4dc7b4d28fcd11307d4234c513ff779a1c6f#diff-c3f8db3928d5d4ba1151657fa434f679

PR #689, opened 2020-08-19
Build all benchmarks with latest clang.

inferno-chromium:
This will help us the use the same latest clang from latest OSS-Fuzz images across all benchmarks. This is important as sometimes they can be new tool additions or image changes over time, e.g. https://github.com/google/oss-fuzz/pull/4332


inferno-chromium:
Checking to see if this builds properly for all old benchmarks.

jonathanmetzman:
This PR is obsolete.

PR #690, opened 2020-08-19
Base entropic_keepseed variant on entropic_exectime variant

dokyungs:
This patch rebases the entropic_keepseed variant on entropic_exectime. In the initial 9 hours of the 2020-08-20 FuzzBench run, the entropic_exectime variant achieved an order-of-magnitudes higher throughput than entropic_fixcrossover 2020-08-15 (from tens of exec/s to thousands of exec/s). This patch requests experimenting with the entropic_keepseed variant again, to test whether or not the effect of keeping seed inputs is more pronounced with the throughput increase.

### DIFF
--- a/service/experiment-requests.yaml
+++ b/service/experiment-requests.yaml
@@ -19,6 +19,10 @@
 # are still testing this feature. You should request an experiment by contacting
 # us as you normally do.
 
+- experiment: 2020-08-21
+  fuzzers:
+    - entropic_keepseed
+
 - experiment: 2020-08-20
   fuzzers:
     - aflplusplus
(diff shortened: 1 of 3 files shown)


service/experiment-requests.yaml
```diff
@@ -19,6 +19,10 @@
 # are still testing this feature. You should request an experiment by contacting
 # us as you normally do.
 
+- experiment: 2020-08-20
```
jonathanmetzman, line 22: Duplicate experiment name.

dokyungs: Fixed!

PR #692, opened 2020-08-21
Generate reports directory

Shadoom7:
This patch will
-generate hierarchical directory for clang coverage reports
-simplify the logic and structure of `coverage_utils.py`
-fix the edge mismatch issue by using the list


inferno-chromium:
This is superceded by https://github.com/google/fuzzbench/pull/714, uptaking the fix there.

PR #696, opened 2020-08-27
Fix for out-of-memory issues after clang cov switch.

inferno-chromium:
Switch to serialized code coverage report generation. Each llvm-cov
invocation can take several gigabytes of memory, so parallelizing
it on one dispatcher does not work. Also, disable php benchmark
until we get a chance to analyze large memory footprint (>100Gb).


experiment/stop_experiment.py
```diff
@@ -38,17 +38,18 @@ def stop_experiment(experiment_name, experiment_config_filename):
...
-    ]
+    experiment_instances = []
     dispatcher_instance = experiment_utils.get_dispatcher_instance_name(
         experiment_name)
     if dispatcher_instance not in instances:
         logger.warning('Dispatcher instance not running, skip.')
     else:
         experiment_instances.append(dispatcher_instance)
 
+    trial_prefix = 'r-' + experiment_name
```
inferno-chromium, line 49: This is important since dispatcher needs to be killed first to avoid race conditions with instance deletes.

jonathanmetzman: ACK

service/automatic_run_experiment.py
```diff
@@ -54,7 +54,9 @@
     'libpcap_fuzz_both',
     'mbedtls_fuzz_dtlsclient',
     'openssl_x509',
-    'php_php-fuzz-parser',
+
+    # Temporarily disabled due to out-of-memory issues with clang coverage.
```
jonathanmetzman, line 58: Definitely shouldn't come in this PR, but I think we should switch to searching for the benchmarks instead of hardcoding a list. We only hardcoded when we had benchmarks like perl that didn't work. Instead we can have a list of benchmarks we ignore.

experiment/coverage_utils.py
```diff
@@ -49,40 +48,42 @@ def upload_coverage_info_to_bucket():
...
     logs.initialize()
-    logger.info('Generating coverage report for benchmark: {benchmark} \
-                fuzzer: {fuzzer}.'.format(benchmark=benchmark, fuzzer=fuzzer))
-    generator = CoverageReporter(fuzzer, benchmark, experiment)
+    logger.info(
+        ('Generating coverage report for '
+         'benchmark: {benchmark} fuzzer: {fuzzer}.').format(benchmark=benchmark,
+                                                            fuzzer=fuzzer))
+
+    generator = CoverageReporter(experiment, fuzzer, benchmark)
```
jonathanmetzman, line 72: should probably be called `coverage_reporter`.

jonathanmetzman:
THanks!

experiment/coverage_utils.py
inferno-chromium: Done!

service/automatic_run_experiment.py
inferno-chromium: Yes right, will do it next week.

PR #697, opened 2020-08-27
Add (libfuzzer|entropic)_(before|after) variants

dokyungs:
`libfuzzer_after` is the same as `libfuzzer_keepseed`.
`entropic_after` is the same as `entropic_keepseed` (which in turn is based on `entropic_exectime`).

The above two variants are all based on the following llvm commit, which contains interceptor/fixcrossover patch.
https://github.com/llvm/llvm-project/commits/bb54bcf84970c04c9748004f3a4cf59b0c1832a7

`libfuzzer_before` and `entropic_before` are based on the following llvm commit, before interceptor/fixcrossover appeared.
https://github.com/llvm/llvm-project/commits/8ef9e2bf355d05bc81d8b0fe1e5333eec59a0a91

### DIFF
--- a/.github/workflows/fuzzers.yml
+++ b/.github/workflows/fuzzers.yml
@@ -35,8 +35,12 @@ jobs:
           - honggfuzz_qemu
           - weizz
           # Temporary variants.
+          - libfuzzer_before
+          - libfuzzer_after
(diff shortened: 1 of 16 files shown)


dokyungs:
@morehouse @hctim FYI

PR #698, opened 2020-08-27
Improve fuzzer and benchmark validation

jonathanmetzman:
Don't allow capital letters as GCB doesn't like them in docker image names.
Add tests.
Do more to validate benchmark.yaml.

### DIFF
--- a/common/benchmark_utils.py
+++ b/common/benchmark_utils.py
@@ -15,6 +15,8 @@
 import os
 import re
 
+import yaml
+
 from common import environment
 from common import logs
 from common import benchmark_config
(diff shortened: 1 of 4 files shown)


common/benchmark_utils.py
```diff
@@ -53,16 +56,34 @@ def get_builder_image_url(benchmark, fuzzer, docker_registry):
...
-    logs.error('%s must have a benchmark.yaml.', benchmark)
-    return False
+    return True
+
+
+def validate(benchmark):
+    """Returns True if |benchmark| is a valid fuzzbench benchmark."""
+    if not validate_name(benchmark):
+        return False
+    if benchmark not in get_all_benchmarks():
```
inferno-chromium, line 72: nit: newline before and after block for readability.

common/benchmark_utils.py
```diff
@@ -53,16 +56,34 @@ def get_builder_image_url(benchmark, fuzzer, docker_registry):
...
+        logs.error('%s must have a benchmark.yaml.', benchmark)
+        return False
+    try:
+        get_fuzz_target(benchmark)
```
inferno-chromium, line 76: maybe add get_project as well.

jonathanmetzman: I don't think project is actually mandatory. The function is actually dead code. I'm removing this function.

PR #704, opened 2020-08-31
Don't clear existing coverage jsons in report generation.

inferno-chromium:
Switch report directory rsync to not delete existing reports
directory on GCS as it will contain clang coverage reports.
This is a regression from the recent clang coverage switch
where we keep coverage_reports.json in reports dir.


inferno-chromium:
TBR

PR #714, opened 2020-09-07
Clang code coverage fixes

inferno-chromium:
- In preperation for breaking up measurer into individual bots,
  sync report and data files individually for each
  (fuzzer, benchmark) combination. This ensures that we get
  data from non-failing runs, while failing ones will be logged.
- Try-catch global on coverage run for each pair, otherwise
  measurer process will silently exit and it won't be possible
  to debug root cause.
- Fix bug where profdata might not exist for some trial pairs,
  so ignore those. Otherwise, merging of profdata files will
  fail.
- Fix llvm-cov new_process call to not exception when failed
  to generate html files. See expect_zero=False with logging
  of error.
- Fix region calculation issue pointed in
  https://github.com/google/fuzzbench/pull/692. Uptake fix
  after Ruikang internship.
- Fixes #714, use regions instead of edges after clang code
  coverage switch.

### DIFF
--- a/experiment/measurer.py
+++ b/experiment/measurer.py
@@ -15,6 +15,7 @@
 """Module for measuring snapshots from trial runners."""
 
 import collections
+import gc
 import multiprocessing
 import os
 import pathlib
@@ -69,10 +70,11 @@ def measure_main(experiment_config):
     max_total_time = experiment_config['max_total_time']
     measure_loop(experiment, max_total_time)
 
+    # Clean up resources.
+    gc.collect()
+
     # Do the final measuring and store the coverage data.
-    coverage_utils.store_coverage_data(experiment_config)
     coverage_utils.generate_coverage_reports(experiment_config)
-    coverage_utils.upload_coverage_info_to_bucket()
 
     logger.info('Finished measuring.')
 
@@ -319,7 +321,8 @@ class SnapshotMeasurer(coverage_utils.TrialCoverage):  # pylint: disable=too-man
 
     def __init__(self, fuzzer: str, benchmark: str, trial_num: int,
                  trial_logger: logs.Logger):
-        super().__init__(fuzzer, benchmark, trial_num, trial_logger)
+        super().__init__(fuzzer, benchmark, trial_num)
+        self.logger = trial_logger
         self.corpus_dir = os.path.join(self.measurement_dir, 'corpus')
 
         self.crashes_dir = os.path.join(self.measurement_dir, 'crashes')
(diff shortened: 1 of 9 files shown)


inferno-chromium:
Full report works - https://www.fuzzbench.com/reports/2020-09-07/index.html with differential coverage now.

experiment/measurer.py
```diff
@@ -364,6 +367,26 @@ def run_cov_new_units(self):
...
+                                                      summary_only=summary_only)
+
+        if result.retcode != 0:
+            self.logger.error(
+                'Coverage summary json file generation failed for '
+                'cycle: %d.', cycle)
+            if cycle != 0:
+                self.logger.error(
+                    'Coverage summary json file generation failed for '
+                    'cycle: %d.', cycle)
```
lszekeres, line 385: Do we need to log this error twice?

lszekeres (approved):

lszekeres:
> Full report works - https://www.fuzzbench.com/reports/2020-09-07/index.html with differential coverage now.

Awesome!! Thanks!

inferno-chromium:
@jonathanmetzman fyi.

experiment/measurer.py
```diff
@@ -69,10 +70,11 @@ def measure_main(experiment_config):
     max_total_time = experiment_config['max_total_time']
     measure_loop(experiment, max_total_time)
 
+    # Clean up resources.
```
jonathanmetzman, line 73: How sure are we this is needed?

experiment/measurer.py
```diff
@@ -364,6 +367,22 @@ def run_cov_new_units(self):
...
             self.UNIT_BLACKLIST[self.benchmark].union(set(crashing_units)))
 
+    def generate_summary(self, cycle: int, summary_only=True):
+        """Transforms the .profdata file into json form."""
+        coverage_binary = coverage_utils.get_coverage_binary(self.benchmark)
+        result = coverage_utils.generate_json_summary(coverage_binary,
+                                                      self.profdata_file,
+                                                      self.cov_summary_file,
+                                                      summary_only=summary_only)
+        if result.retcode != 0:
```
jonathanmetzman, line 377: Personally would avoid the extra indent and do `if result.retcode == 0: return`

jonathanmetzman:
Thanks!

experiment/measurer.py
inferno-chromium: Can fix in next CL.

experiment/measurer.py
inferno-chromium: I was seeing some ooms before that don't happen anymore. We won't need this after measurer is broken into individual tasks.

PR #721, opened 2020-09-08
Retry variance test, new variants, description.md

vanhauser-thc:
This PR:
  * adds a description.md file to all afl++ instances
  * retries the aflplusplus_same1-3 run, as a misplaced larger than instead of lesser than introduced variance
  * the misplaced comparison could have an impact on the overall scheduling and different schedulung so this sadly needs a lot of tests variants :-(

plus request a new experiment.


vanhauser-thc:
I think this is the first time ever everyihing built, no formatting issue and no bug in the service request :)
@jonathanmetzman  lets get that lets get that test rolling to hopefully finally see what the CPU variance is... fingers crossed

jonathanmetzman (approved):

jonathanmetzman:
15 fuzzers is quite a lot. I'm hoping we don't run into some kind of scaling issue because of this.

vanhauser-thc:
at least then this is a test case to check for scaling issues ;)
yeah I would have loved avoiding this, but it is a central functionality that I had to fix and checking out the impact is important :(

PR #726, opened 2020-09-11
Add libfuzzer and entropic norestart variants

mboehme:
This is to evaluate disabling libfuzzer and entropic restart every 5 minutes when -fork=1

Currently, in fork mode, LibFuzzer runs a new instance (job) every five minutes and then merges the generate corpus into the main corpus. The new LF instance (job) starts with a very small subset of the main corpus.

In this commit, in fork mode, LibFuzzer runs a new instance only when a crash is encountered and merges the current corpus into the main corpus every five minutes. The new LF instance (job) always starts with the entire main corpus. 

Some seed files might have been deleted before the merge. So, there is a thread-unsafe quick check whether the file still exists before it is merged.

### DIFF
--- a/.github/workflows/fuzzers.yml
+++ b/.github/workflows/fuzzers.yml
@@ -35,8 +35,11 @@ jobs:
           - honggfuzz_qemu
           - weizz
           # Temporary variants.
+          - libfuzzer_norestart
           - libfuzzer_magicbytes
           - entropic_magicbytes
+          - entropic_norestart_noexectime
+          - entropic_norestart_yesexectime
           - aflplusplus_qemu5
           - aflplusplus_pre3
           - aflplusplus_pre3_fix
--- a/fuzzers/entropic/builder.Dockerfile
+++ b/fuzzers/entropic/builder.Dockerfile
@@ -18,7 +18,7 @@ FROM $parent_image
 RUN git clone https://github.com/llvm/llvm-project.git /llvm-project && \
     cd /llvm-project && \
     git checkout 5cda4dc7b4d28fcd11307d4234c513ff779a1c6f && \
-    cd /llvm-project/compiler-rt/lib/fuzzer && \
+    cd compiler-rt/lib/fuzzer && \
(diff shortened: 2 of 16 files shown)


mboehme:
presubmit says "Code is not formatted correctly, please run 'make format'". When I run `make format` locally, it only reports "Success!", without changes to the code. What do I do to pass presubmit?

inferno-chromium:
> presubmit says "Code is not formatted correctly, please run 'make format'". When I run `make format` locally, it only reports "Success!", without changes to the code. What do I do to pass presubmit?

make format is like a best attempt, and many lint failures require manual minor fixing, e.g. in CI, this one needs a new line.

--- /home/runner/work/fuzzbench/fuzzbench/fuzzers/entropic_norestart_noexectime/fuzzer.py	(original)
+++ /home/runner/work/fuzzbench/fuzzbench/fuzzers/entropic_norestart_noexectime/fuzzer.py	(reformatted)
@@ -17,6 +17,7 @@
 
 from fuzzers import utils
 from fuzzers.libfuzzer import fuzzer as libfuzzer_fuzzer
+
 
 def build():
     """Build benchmark."""

------------------------------------
Your code has been rated at 10.00/10

jonathanmetzman:
> presubmit says "Code is not formatted correctly, please run 'make format'". When I run `make format` locally, it only reports "Success!", without changes to the code. What do I do to pass presubmit?

I'm surprised by this.
In theory that message should only show up if `make format` would result in changing the code. So it's bizarre to me that you would see that message but `make format` doesn't actually fix it.
Maybe the presubmit error was caused by something else like a linting failure (which `make format` won't fix).


(probably irrelevant) The only weirdness I've seen here is with respect to pylint directives for specific function arguments. When you do this and then run `make format` the linter will complain but once you fix that issue both should be happy.

mboehme:
I checked whether anything changed using `git diff`. Maybe it doesn't show changes to whitespace?

PR #728, opened 2020-09-13
New schedule test

vanhauser-thc:
A fix in a central scheduling setup results that the scheduling efficient seems to be degraded - or at least the order of effectiveness has changed.
this compares  the original best schedule with all the fixed ones, plus tests a change in the havoc routine.


inferno-chromium (approved):

inferno-chromium:
I am curious if 

make test-run-aflplusplus_no_mopt-openthread-2019-12-23
succeeds locally for you. all these openthread failure seem to show up on openthread, so just curious, maybe try it and see if there is an easy fix. Who does not love green CI :), we can try to look more in next week or so.

vanhauser-thc:
@inferno-chromium yeah I am curious too, as I wrote in the previous PR - why does afl fail but aflfast pass? they are both using the same build.
That aflplusplus_optimal passes can be explained by that I am using a dedicated llvm 11 install.

So my guess is it is a clang issue. Maybe it was updated for llvm 12 for openthread and that is why it fails?

vanhauser-thc:
I think the easiest fix would be to get rid of -Werror in the openthread checkout directory by

sed -i 's/Wshadow -Werror /Wshadow/g' config*

not sure that is the solution you'd like though ;)

inferno-chromium:
> I think the easiest fix would be to get rid of -Werror in the openthread checkout directory by
> 
> sed -i 's/Wshadow -Werror /Wshadow/g' config*
> 
> not sure that is the solution you'd like though ;)

Error disappear if i bump up to latest revision, doing that.

jonathanmetzman:
lgtm

PR #730, opened 2020-09-14
Don't print decimals for unique edges on ranking plot.

### LINKED ISSUE #723: Unique edges does not need two numbers after decimal place
These will never be anything other than 0. 
They also make numbers more likely to bleed into other bars.

lszekeres:
Fixes #723.


inferno-chromium (approved):

jonathanmetzman:
lgtm

PR #734, opened 2020-09-14
Add config attribute to allow skipping fuzzers for a benchmark.

inferno-chromium:
This is needed for KLEE and AFLCC which do not support all
benchmarks, but we still need to support building the rest
and running an experiment without constant build retries.


laurentsimon (approved):

inferno-chromium:
Since fuzzer does not config file, i thought to just use benchmark config here. it feels weird to add fuzzer config just for this one attribute, plus might be better to think in terms of benchmarks, that which fuzzer don't work with it

experiment/build/builder.py
```diff
@@ -50,6 +51,27 @@
 logger = logs.Logger('builder')  # pylint: disable=invalid-name
 
 
+def get_fuzzer_benchmark_pairs(fuzzers, benchmarks):
+    """Return a tuple of (fuzzer, benchmark) pairs to build. Excludes
+    blacklisted fuzzers from each benchmark.
```
lszekeres, line 56: nit: consider using "skip list" instead here and everywhere else.

lszekeres (approved):

experiment/build/builder.py
inferno-chromium: Done!

lszekeres: thanks!

jonathanmetzman:
Not sure i like this and it feels pretty rushed.

inferno-chromium:
> Not sure i like this and it feels pretty rushed.

Ok, i can either revert or take your time to review. I will hold off on https://github.com/google/fuzzbench/pull/737 as well.

PR #736, opened 2020-09-14
Fix bibtex

lszekeres:


inferno-chromium (approved):

jonathanmetzman:
Why not leave this and get rid of the blogpost?

inferno-chromium:
> Why not leave this and get rid of the blogpost?

This is what Laszlo said on chat - "using the blog post might be slightly better, as it's more customary to cite "articles" rather than code repos". I am fine with either.

jonathanmetzman (approved):
lgtm

PR #738, opened 2020-09-15
Fix weizz_qemu in CI.

inferno-chromium:


lszekeres (approved):

service/core-fuzzers.yaml
```diff
@@ -16,3 +16,4 @@ fuzzers:
     - libfuzzer
```
jonathanmetzman, line 16: I'm getting confused what the purpose of this file is.
Is it supposed to include every fuzzer that's going to be included in fuzzbench on a permanent basis?

fuzzers/weizz_qemu/fuzzer.py
```diff
@@ -40,6 +40,13 @@ def fuzz(input_corpus, output_corpus, target_binary):
     os.environ['WEIZZ_NO_AFFINITY'] = '1'
     os.environ['WEIZZ_I_DONT_CARE_ABOUT_MISSING_CRASHES'] = '1'
     os.environ['WEIZZ_CTX_SENSITIVE'] = '1'
+
+    # Weizz needs at least one non-empty seed to start.
+    if len(os.listdir(input_corpus)) == 0:
+        with open(os.path.join(input_corpus, 'default_seed'),
```
jonathanmetzman, line 46: I'd do the os.path.join on another line to avoid the ugly line break.

service/core-fuzzers.yaml
inferno-chromium: This is used for deciding which benchmarks are the standard ones to be included/added/merged with a new experiment.

fuzzers/weizz_qemu/fuzzer.py
inferno-chromium: Done! Also, using a helper now for both afl and weizz.

service/core-fuzzers.yaml
inferno-chromium: We will also use this to decide which fuzzers are variants and which ones are not. This will be used to seperate official experiment reports in root dir and variant experiment reports in experimental sub-dir. @lszekeres is working on this.

docs/_data/fuzzers.yaml
```diff
@@ -18,8 +18,8 @@ Fuzzers:
   year: 2020
   url: https://arxiv.org/pdf/1911.00621.pdf
   dblp: journals/corr/abs-1911-00621
-  repo:
-  dir:
+  repo: https://github.com/andreafioraldi/weizz-fuzzer
+  dir: weizz_qemu
```
jonathanmetzman, line 22: It's odd to me this directory is weizz_qemu and not weizz.

jonathanmetzman (commented):
lgtm

docs/_data/fuzzers.yaml
inferno-chromium: Since it is a binary only fuzzer i think - https://github.com/andreafioraldi/weizz-fuzzer , needs qemu. also @vanhauser-thc did it explicitly in https://github.com/google/fuzzbench/pull/681/files

vanhauser-thc: the reason is that weizz can work with llvm and with qemu, and this is the qemu version. so the naming is correct. (the llvm has less features as far as I understood andrea)

PR #743, opened 2020-09-17
Request a new experiment for aflsmart

thuanpv:
Hi,

This is a request for a new experiment to evaluate the recently added feature of AFLSmart (https://github.com/google/fuzzbench/commit/30540d1f5bfd756abb65577e5d57b89909e60493).

Thanks,

Thuan

### DIFF
--- a/service/experiment-requests.yaml
+++ b/service/experiment-requests.yaml
@@ -16,6 +16,10 @@
 # You can run "make presubmit" to do basic validation on this file.
 # Please add new experiment requests towards the top of this file.
 
+- experiment: 2020-09-17
+  fuzzers:
+    - aflsmart
+
 - experiment: 2020-09-16
   fuzzers:
     - aflplusplus_coe_12
--- a/service/experiment-requests.yaml
+++ b/service/experiment-requests.yaml
@@ -16,7 +16,7 @@
 # You can run "make presubmit" to do basic validation on this file.
 # Please add new experiment requests towards the top of this file.
 
-- experiment: 2020-09-17
+- experiment: 2020-09-18
   fuzzers:
     - aflsmart
 


service/experiment-requests.yaml
```diff
@@ -16,6 +16,10 @@
 # You can run "make presubmit" to do basic validation on this file.
 # Please add new experiment requests towards the top of this file.
 
+- experiment: 2020-09-18
```
inferno-chromium, line 19: Just fyi, had to change the date since 2020-09-17 is a taken name in db (we have a broken test experiment). Experiment will start soon regardless (i mean in a day since currently 2020-09-16 is using up all the cores).

inferno-chromium (approved):
Do you want to bump up the revision for aflsmart here - https://github.com/google/fuzzbench/blob/master/fuzzers/aflsmart/builder.Dockerfile#L41 ?
Otherwise results should already be available at https://www.fuzzbench.com/reports/2020-08-23/index.html

service/experiment-requests.yaml
thuanpv: Yes, I would like to evaluate the newest version of aflsmart (a9d60257a6b5a7df2e177bddc6982376723bfd90) which support the new feature. Thanks.

thuanpv: https://github.com/google/fuzzbench/blob/master/fuzzers/aflsmart/builder.Dockerfile#L41 has already been updated in my previous PR (#712)

inferno-chromium (approved):

inferno-chromium:
Just fyi, once experiment finishes, it auto-merges results with other fuzzers as well (e.g. like honggfuzz, aflplusplus, etc)

PR #745, opened 2020-09-17
Fix exception in local experiment, allow session across threads.

inferno-chromium:


inferno-chromium:
Verified that sqlite exception is gone.

jonathanmetzman (approved):
lgtm

jonathanmetzman:
LGTM

PR #746, opened 2020-09-17
Fix docs to better seperate local and cloud workflows.

inferno-chromium:


inferno-chromium:
TBR minor doc fix. (esp navigation).

PR #751, opened 2020-09-21
Simplify pre and non-pre vms handling decisions

inferno-chromium:
Currently, the logic is too complicated that decides whether to
spin a preemptible or non-preemptible vm once a current vm preempts.
Simplify to start a non-preemptible vm always after experiment
expected end time. This also takes care of a case where vms are never
started if can_start_nonpreemptible is false.

### DIFF
--- a/experiment/scheduler.py
+++ b/experiment/scheduler.py
@@ -187,9 +187,6 @@ class TrialInstanceManager:  # pylint: disable=too-many-instance-attributes
     handle_preempted_trials method should be called in the scheduling loop.
     See the docstring for handle_preempted_trials for how it works.
     """
-    # How many times the number of trials we need preemptibles can we launch.
-    MAX_PREEMPTIBLES_MULTIPLIER = 4
-
     # How long can we keep trying preemptibles before we have to switch to a
     # nonpreemptibles or stopping the experiment.
     PREEMPTIBLE_WINDOW_MULTIPLIER = 1
@@ -197,12 +194,6 @@ class TrialInstanceManager:  # pylint: disable=too-many-instance-attributes
     def __init__(self, num_trials, experiment_config):
         self.experiment_config = experiment_config
         self.num_trials = num_trials
-
-        # Bound for the number of preemptibles we can start if the experiment
-        # specified preemptible_runners.
-        self.max_preemptibles = (self.num_trials *
-                                 self.MAX_PREEMPTIBLES_MULTIPLIER)
-        logger.info('Max preemptibles: %d.', self.max_preemptibles)
 
         # Attributes for preemptible retry window. The preemptible retry window
         # is a time period that starts when the last initial trial is started.
@@ -260,20 +251,12 @@ class TrialInstanceManager:  # pylint: disable=too-many-instance-attributes
 
         return datetime_now() > preemptible_window_end_time
 
-    def can_start_preemptible(self, preemptible_starts: int) -> bool:
-        """Returns True if we can start a preemptible trial.
-        |preemptible_starts| is the number of preemptibles we've already
-        started."""
+    def can_start_preemptible(self) -> bool:
+        """Returns True if we can start a preemptible trial."""
         if not self.experiment_config.get('preemptible_runners'):
             # This code shouldn't be executed in a non preemptible experiment.
             # But just in case it is, it's not OK to create a preemptible trial
             # in a non-preemptible experiment.
-            return False
-
-        if preemptible_starts > self.max_preemptibles:
-            # Don't create more than the maximum number of preemptibles or else
-            # costs can be infinite in the (highly unlikely) worst case
-            # scenario.
             return False
 
         if self.preemptible_window_passed():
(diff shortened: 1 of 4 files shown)


experiment/scheduler.py
```diff
@@ -188,34 +187,17 @@ class TrialInstanceManager:  # pylint: disable=too-many-instance-attributes
...
-
-    # The maximum fraction of total trials in the experiment that can be done
-    # using nonpreemptibles. This helps bound the cost in unexpected situations.
-    NONPREEMPTIBLES_FRACTION = 1 / 20
-
-    MAX_FRACTION_FOR_NONPREEMPTIBLES = 1 / 4
-
     # How many times the number of trials we need preemptibles can we launch.
-    MAX_PREEMPTIBLES_MULTIPLIER = 2
+    MAX_PREEMPTIBLES_MULTIPLIER = 4
```
inferno-chromium, line 191: Since pre n1-standard-1 is 0.01 vs $0.04749975, we can easily cap it to 4x. I dont think eviction rate is anywhere near this, so we will never reach near it.

experiment/scheduler.py
```diff
@@ -188,34 +187,17 @@ class TrialInstanceManager:  # pylint: disable=too-many-instance-attributes
...
 
     # How long can we keep trying preemptibles before we have to switch to a
     # nonpreemptibles or stopping the experiment.
-    PREEMPTIBLE_WINDOW_MULTIPLIER = 3
+    PREEMPTIBLE_WINDOW_MULTIPLIER = 1
```
inferno-chromium, line 195: This is to avoid 4-5 day long experiments that is currently happening.

experiment/scheduler.py
jonathanmetzman: I frankly don't know if a multiplier is needed anymore. It was very specific to the previous algorithm where we had a guaranteed bounded cost in the worst case scenario.

experiment/scheduler.py
```diff
@@ -188,34 +187,17 @@ class TrialInstanceManager:  # pylint: disable=too-many-instance-attributes
     handle_preempted_trials method should be called in the scheduling loop.
     See the docstring for handle_preempted_trials for how it works.
     """
-    # Hard limit on the number of nonpreemptibles we will use. This bounds
```
jonathanmetzman, line 191: If switching to preemptibles earlier should be as cost effective as I'm led to beleive why do we need to remove these bounds on nonpreemptibles fraction and MAX_NONPREEMPTIBLES.

service/experiment-config.yaml
```diff
@@ -2,7 +2,7 @@
 # Unless you are a fuzzbench maintainer running this service, this
 # will not work with your setup.
 
-trials: 20
+trials: 15
```
jonathanmetzman, line 5: Why change this?

experiment/scheduler.py
inferno-chromium: I think this seems overly cautious to keep this, i think we can remove it as if there are too many preemptions in the first day, it is fine as well, we can continue to keep recreating bots, no need to bound it (as costs will still be low even if we keep respinning vms).

service/experiment-config.yaml
inferno-chromium: I thought 15 was the agreed upon number of trials needed, why do 20. i think 20 was needed when we fail on some trials with these 4-5 days long experiments, now it should 2 days and non-preemptibles will be created after first day.

experiment/scheduler.py
inferno-chromium: So, these non premptibles bounds are bad, as it would mean we don't create trial at all if can_start_nonpreemptible is false. Keeping a simple logic feels better here, only create preemptibles the first day, and on second day, create nonpreemptibles whatever cost it is. It is always just a hundred or so, that might be needed on second day, so delaying the experiments for 3-4 days does not feel right here.

service/experiment-config.yaml
jonathanmetzman: I don't think we agreed 15 was needed. I think we decided 15 is sufficient for non-preemptible experiments since it balances costs and accuracy. We need 20 for statistical power.

experiment/scheduler.py
jonathanmetzman: I don't agree that they are bad. And we can simplify having them by just preventing creation of more than 10% nonpreemptibles.
But it's good to have this limit so that we don't end up doing a fully nonpreemptible experiment in rare cases

service/experiment-config.yaml
inferno-chromium: Done reverted.

experiment/scheduler.py
inferno-chromium: It is complicating the code for unseen rare circumstance. I have never seen preemptible vms not available. 
As per old code, if 10% non-preemptible limit is reached, we don't create any more vms, that does not seem right either. We dont fall back to create preemptibles in that case.

jonathanmetzman: It seems right to me since this is such a rare event it's better to fail and let developer know than to try to handle it and waste a ton of money. I understand the event is rare but I think it's a good idea to defend against it.

inferno-chromium:
Closing so that i can redo this.

PR #753, opened 2020-09-21
Add nginx benchmark, add support for hg in checkout_commit.py

inferno-chromium:


jonathanmetzman (approved):
lgtm

inferno-chromium:
@vanhauser-thc - i really want to get nginx benchmark, was wondering if you can check what is going on with the compile failures with aflplusplus here. Use this patch on trunk, and see failure with test-aflplusplus-nginx_http_request_fuzzer. Rest all fuzzers are all ok, so i am kind of blocked here. I can blacklist it, but at this point, too many afl++ variants and blacklisting will be pain.

vanhauser-thc:
> @vanhauser-thc - i really want to get nginx benchmark, was wondering if you can check what is going on with the compile failures with aflplusplus here. Use this patch on trunk, and see failure with test-aflplusplus-nginx_http_request_fuzzer. Rest all fuzzers are all ok, so i am kind of blocked here. I can blacklist it, but at this point, too many afl++ variants and blacklisting will be pain.

uh nice find. dict2file llvm module bug. will add a fix once I have figured out where it breaks.

vanhauser-thc:
why the hell is nginx using 96 bit integers ... should work now

inferno-chromium:
> why the hell is nginx using 96 bit integers ... should work now

Thanks a lot, retriggered CI.

vanhauser-thc:
Coe2 fast2 etc will fail because they use an old commit that is required for @mboehme ‘s patch

inferno-chromium:
> Coe2 fast2 etc will fail because they use an old commit that is required for @mboehme ‘s patch

I wont commit this then for the current 09-22, 09-23 experiments.

inferno-chromium:
@vanhauser-thc - any idea why aflplusplus_optimal still fails.

vanhauser-thc:
of course it does ... the feature is implemented twice and I forgot to fix it in LTO. my bad. it was 2:30 in the morning ...

PR #760, opened 2020-09-22
Add support for using latest OSS-Fuzz corpora, add libxslt benchmark.

inferno-chromium:


inferno-chromium:
Ready for review! Unsupported fuzzers tracked in https://github.com/google/fuzzbench/issues/762. 
Adding a benchmark is pain these days, some fuzzers always break.

.github/workflows/build_and_test_run_fuzzer_benchmarks.py
```diff
@@ -88,9 +100,9 @@ def make_builds(benchmarks, fuzzer):
         make_target = get_make_target(fuzzer, benchmark)
         make_command = ['make', 'RUNNING_ON_CI=yes', '-j', make_target]
         print('Running command:', ' '.join(make_command))
-        result = subprocess.run(make_command, check=False)
-        if not result.returncode == 0:
+        if not run_with_retries(make_command):
```
jonathanmetzman, line 103: Maybe reuse: https://github.com/google/fuzzbench/blob/master/common/retry.py

.github/workflows/fuzzers.yml
```diff
@@ -37,7 +37,6 @@ jobs:
           - honggfuzz_qemu
           - weizz_qemu
           # Temporary variants.
-          - aflcc_no_orig_dict
```
jonathanmetzman, line 40: Don't understand what this removal is doing here.

benchmarks/libxslt_xpath/build.sh
```diff
@@ -0,0 +1,74 @@
...
+        -o $OUT/$fuzzer \
+        $LIB_FUZZING_ENGINE \
+        libexslt/.libs/libexslt.a libxslt/.libs/libxslt.a \
+        ../libxml2/.libs/libxml2.a \
+        $CRYPTO_LIBS
+
+    zip -j $OUT/${fuzzer}_seed_corpus.zip tests/fuzz/seed/$fuzzer/*
+done
+
+cp tests/fuzz/xpath.dict tests/fuzz/xpath.xml $OUT/
```
jonathanmetzman, line 73: Don't know if I want zips commited to this repo or if we should put them in GCB bucket. Need to think about this.

fuzzers/eclipser/fuzzer.py
```diff
@@ -65,7 +65,7 @@ def fuzz(input_corpus, output_corpus, target_binary):
         # Default is low (0.5 sec), recommended to use higher:
         # https://github.com/google/fuzzbench/issues/70#issuecomment-596060572
         '--exectimeout',
-        '2000',
+        '5000',
```
jonathanmetzman, line 68: Is this change necessary because eclipser breaks otherwise?

.github/workflows/fuzzers.yml
inferno-chromium: aflcc does not work across a lot of benchmarks [atleast 4-5]. While aflcc is explicitly excluded, this variant is not. I prefer to keep this temporary variant out of CI, rather than adding more blacklist for this temp variant.

benchmarks/libxslt_xpath/build.sh
inferno-chromium: I prefer zip here as it is more clear for people looking at benchmark directory. Going to GCS folder and gsutil might appear obvious to us, but not for the common crowd.

fuzzers/eclipser/fuzzer.py
inferno-chromium: Yes right, this timeout hits with seed corpus now. Also, it will hit as we add more seed corpus, i remember seeing with libaom that i was planning to add too.

.github/workflows/build_and_test_run_fuzzer_benchmarks.py
inferno-chromium: Done!

inferno-chromium:
Addressed comments, PTAL.

common/retry.py
```diff
@@ -41,6 +39,8 @@ def wrap(  # pylint: disable=too-many-arguments
         exception_type=Exception,
         retry_on_false=False):
     """Retry decorator for a function."""
+    # To avoid circular dependency.
```
jonathanmetzman, line 42: This seems to be hiding the dependency rather than avoiding it.
It's fine if this is needed because this CL didn't introduce this problem but comment is misleading IMO.

benchmarks/oss_fuzz_benchmark_integration.py
```diff
@@ -96,6 +99,34 @@ def copy_oss_fuzz_files(project, commit_date, benchmark_dir):
...
+        full_fuzz_target = fuzz_target
+
+    corpus_backup_url = CORPUS_BACKUP_URL_FORMAT.format(
+        project=project, fuzz_target=full_fuzz_target)
+    command = ['gsutil', 'ls', corpus_backup_url]
+    result = subprocess.run(command, stdout=subprocess.PIPE)
+    if result.returncode:
+        raise Exception(f'No corpus found for project {project}.')
+
+    archive_filename = f'{fuzz_target}_seed_corpus.zip'
```
jonathanmetzman, line 116: Do you think we should call this seed corpus or something else?

common/retry.py
inferno-chromium: Yes otherwise it breaks CI. Since we need to use logs in retry, i dont see any other way to break the circular deps. This is hacky, but no other way.

benchmarks/oss_fuzz_benchmark_integration.py
inferno-chromium: Done, made it oss_fuzz_corpus.zip

inferno-chromium:
Done, addressed comments. PTAL.

benchmarks/libxslt_xpath/build.sh
inferno-chromium: I will switch it to GCS if this explodes things, right now we are using simple benchmarks with 10-20mb benchmarks.

jonathanmetzman: Yes but the annoying thing is that unless we clean up the history corpora will take up space forever.

jonathanmetzman (approved):
LGTM for now.

PR #763, opened 2020-09-23
update libaflfuzzer, remove variants, change variants, new experiment

vanhauser-thc:
This removes all my unneeded variants. Please note that others are by Marcel and I dont want to interfere with his experiments.

When running just afl++ variants, the variance of the same 15-20 execution runs on all benchmarks can result to a variance of up to 0.75 in average coverage. so the last run was not conclusive and showed a result that was different to a similar one. So I am rerunning a few some of the previous variants.

libaflfuzzer should no be fixed and able to fuzz all benchmarks. only php still makes trouble it seems.


vanhauser-thc:
btw another experiment request will be coming from Marcel.
I am fine if you combine this one with the Klee/aflcc one - that would altogeter make 14 which would fit one experiment.
that way there would be no long queue of experiments :)

jonathanmetzman (approved):
LGTM

PR #765, opened 2020-09-23
New AFL++ power schedules (variants of FAST and COE)

mboehme:
* Applied fastcount patch to COE2, COE3, and FAST2. Fastcount is the correct fix for #438. The performance hit doesn't come from hashing. It comes from looping over all seeds to find one that has the same checksum every time a new input is generated.
* COE3 (which prunes more aggressively than COE2) was quite successful. Added more aggressive pruning in COE4.
* Added FAST variants FAST3 (equivalent to COE3 but w/o cut-off) and FAST4 (equivalent to COE4 but w/o cut-off).
* Created new shorthandicap where lateboost ios removed and fastcount is added.


inferno-chromium:
atleast one variant fuzzers/aflplusplus_fast3 is missing a runner.dockerfile (may forgot to git add). can test these locally with make test-run-<fuzzer_name>-<variant name>.

mboehme:
Github actions + Fuzzbench CI = ❤️

inferno-chromium (approved):

Issue #770, opened 2020-09-25
Presubmit failure: AssertionError - "experiment/test_run_coverage.py"


BharathMonash:
I get this error every-time I  ```make presubmit``` 

```experiment/test_run_coverage.py::TestIntegrationRunCoverage::test_integration_do_coverage_run_crash FAILED```
```experiment/test_run_coverage.py::TestIntegrationRunCoverage::test_integration_do_coverage_run_no_crash FAILED```

```ERROR    root:logs.py:161 Coverage run failed. Extras: {'coverage_binary': '/home/bk**/fuzzbench/experiment/test_data/test_run_coverage/fuzz-target-clang-cov', 'output': "/home/bk**/fuzzbench/experiment/test_data/test_run_coverage/fuzz-target-clang-cov: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.22' not found```

Maybe this particular test should be skipped for now?

inferno-chromium:
Can you try https://github.com/google/fuzzbench/pull/771 to see if this fixes it for you ? Do you run into any other test failures ?

BharathMonash:
#772 - This fixes it, Thanks Abhishek. 
Sometimes I do run into rsync tests failures (running on gcloud).

PR #771, opened 2020-09-25
Add libstdc++.so.6 dep explicity for coverage binaries.

inferno-chromium:
Should fix
https://github.com/google/fuzzbench/issues/770


inferno-chromium:
@jonathanmetzman - thoughts here, i added libstdc++ explicitly and changed rpath so hopefully #770 works. Or do you want to rebuild these binaries statically ?

jonathanmetzman:
Not sure. This isn't the first test that uses binaries we've gotten complaints about compatibility for.
Maybe it's better just to hide these tests so that they aren't run on most users' machines.

PR #780, opened 2020-09-29
test new changes

vanhauser-thc:
This tests Marcel's updated schedules and also code changes.

PR #781, opened 2020-09-29
Cleanup temporary AFL++ variants

mboehme:


inferno-chromium:
Can you please remove these from https://github.com/google/fuzzbench/blob/master/.github/workflows/fuzzers.yml#L39 list as well.

inferno-chromium:
Don't worry, i will remove it in https://github.com/google/fuzzbench/pull/780, since there will be conflict.

PR #782, opened 2020-09-29
Fix local experiment docs to use gcr.io/fuzzbench registry.

inferno-chromium:
Right now, custom registry e.g. local-server:5000 don't work
due to missing dispatcher-image instructions and missing support
in make. For now, just recommend using gcr.io/fuzzbench, this
still keeps images locally.


jonathanmetzman (approved):
LGTM

Issue #162, opened 2020-03-26
Support downloading of OSS-Fuzz corpora for use in benchmarks


jonathanmetzman:
This should be easy to implement but maybe a bit tough to manage in terms of running this service, should we put these benchmarks in the same report?

This feature would be quite nice as it would allow us to determine if a fuzzer can break through a coverage wall that ClusterFuzz encounters from fuzzing (using LibFuzzer, HonggFuzz and AFL).

jonathanmetzman:
This would be even more useful in conjunction with #77. It is needed according to #156.

inferno-chromium:
This is done, although some folks prefer to do this dynamically. I am not sure of the value and it creates more confusion to put oss_fuzz_corpus in some bucket.

inferno-chromium:
Experiment is now running https://www.fuzzbench.com/reports/experimental/2020-10-07-ossfuzzc/index.html with latest corpus (from backup archive). corpora used = https://storage.cloud.google.com/storage/browser/fuzzbench-data/2020-10-07-ossfuzzc/oss_fuzz_corpora

PR #783, opened 2020-09-30
Archive OSS-Fuzz corpus in GCS, use at runtime based on flag.

inferno-chromium:
- Add support in oss_fuzz_benchmark_integration.py to copy
  OSS-Fuzz corpus to GCS bucket gs://fuzzbench-oss-fuzz-corpora
  instead of benchmark directory. This saves space and allows
  easy update and large corpus support.
- Add flag --oss-fuzz-corpus in run_experiment.py to allow
  running benchmarks that have OSS-Fuzz corpus to use it
  as seed corpus.
- Add support in runner to download OSS-Fuzz corpus and
  use as seed corpus if flag is set.
- Migrage libxslt_xpath benchmark to the new system.

Fixes https://github.com/google/fuzzbench/issues/162

### DIFF
--- a/benchmarks/libxslt_xpath/Dockerfile
+++ b/benchmarks/libxslt_xpath/Dockerfile
@@ -25,4 +25,3 @@ RUN git clone --depth 1 https://gitlab.gnome.org/GNOME/libxml2.git
 RUN git clone --depth 1 https://gitlab.gnome.org/GNOME/libxslt.git
 WORKDIR libxslt
 COPY build.sh $SRC/
-COPY oss_fuzz_corpus.zip $SRC/xpath_seed_corpus.zip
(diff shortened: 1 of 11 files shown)


jonathanmetzman:
I think this isn't quite what we agreed on in the meeting. Let me check my notes and explain more here.

inferno-chromium:
> I think this isn't quite what we agreed on in the meeting. Let me check my notes and explain more here.

These are the things i remember.
- We want to turn this feature on and off for any experiment run. By default it should be false. We want to run some experiment with all oss-fuzz corpora enabled to see where coverage wall is broken.
- Corpora should be on GCS
- I think you said that it should be dynamically pulled when available [gs://{some_bucket}/some_project/some_fuzzer like]. I instead added the oss_fuzz_corpus attribute, as it helps people to know where this corpus resides. We can remove the attribute, but then it wont let people know from where it is used.

Feel free to share what do you think, it should be easy enough to change. Just want to wrap this up. Once we have this, we can add corpus for remaining oss-fuzz benchmark..

PR #786, opened 2020-10-01
Add aflplusplus_lightweizz

andreafioraldi:
As the results of weizz on the targets that it supports are encouraging (it is not general-purpose, it targets only chunk based binary formats and it is expected that it sucks on php and shines on libpng) I wanted to implement a lightweight version of the weizz mutator on top of cmplog and afl++.

The current main limitations of weizz are:

+ QEMU and fork(), it does not support persistent mode and LLVM mode, so it is slow
+ crashes on programs that I never tested, it's my first fuzzer and it's pure academic code lol
+ need to bitflip the entire input to approximate taint, that is expensive

In this afl++ patch, I simplified the weizz mutator and tried to use redqueen to extract structural information without bitflipping (likely will not work, let's see). It uses ofc persistent shmem in llvm mode and, hopefully, it is stable.


andreafioraldi:
There are some commits from Marc in addition, I didn't notice them

inferno-chromium:
experiment-requests.yaml needs update since 10-01 is already requested by Marcel.

andreafioraldi:
fixed, i named it -lightweizz

inferno-chromium:
> fixed, i named it -lightweizz

thanks, need to do one last "make format" for minor presubmit failure fix.

andreafioraldi:
It's not the first time that presubmit on my laptop does not report errors :(

inferno-chromium:
> It's not the first time that presubmit on my laptop does not report errors :(

What is the issue, does it succeed for you when it shouldnt ? Sorry for hassle.

inferno-chromium:
> > What is the issue, does it succeed for you when it shouldnt ?
> 
> Yes, dunno why, I do it before committing (I guess it uses something similar to git diff) and after to double check,

I would say do 'make presubmit' after you do the local commit. without git commit, maybe git diff or something fails.

vanhauser-thc:
@googlebot i consent

jonathanmetzman:
> > > What is the issue, does it succeed for you when it shouldnt ?
> > 
> > 
> > Yes, dunno why, I do it before committing (I guess it uses something similar to git diff) and after to double check,
> 
> I would say do 'make presubmit' after you do the local commit. without git commit, maybe git diff or something fails.

It shouldn't fail. I find this issue very puzzling. It's never happened to me.

PR #794, opened 2020-10-05
KLEE fixes.

inferno-chromium:
- Add max memory limit.
- Remove hack for fuzz target, pass it using env var.

### DIFF
--- a/docker/benchmark-builder/Dockerfile
+++ b/docker/benchmark-builder/Dockerfile
@@ -48,4 +48,5 @@ COPY benchmarks/$benchmark/benchmark.yaml /
 RUN export CHECKOUT_COMMIT=$(cat /benchmark.yaml | tr -d ' ' | grep 'commit:' | cut -d ':' -f2) && \
     python3 -u /opt/fuzzbench/checkout_commit.py $CHECKOUT_COMMIT $SRC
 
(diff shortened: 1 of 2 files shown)


inferno-chromium:
Still testing in a live experiment.

jonathanmetzman (approved):
lgtm

inferno-chromium:
It does not stop OOMs (obviously :), but good change anyway to remove fuzz target hack, so committing.

lszekeres:
> It does not stop OOMs (obviously :), but good change anyway to remove fuzz target hack, so committing.

The -max-memory value is more like a soft limit than a hard one, it's used to prevent adding more paths/states if we're over it:
https://github.com/klee/klee/blob/3983b23eac93b0e6f28ffba4b626401c5280c10f/lib/Core/Executor.cpp#L2965

Consider setting something like 1gb, and perhaps that way we won't go over 3gb :)

inferno-chromium:
> > It does not stop OOMs (obviously :), but good change anyway to remove fuzz target hack, so committing.
> 
> The -max-memory value is more like a soft limit than a hard one, it's just used to prevent adding more paths/states if we're over it:
> https://github.com/klee/klee/blob/3983b23eac93b0e6f28ffba4b626401c5280c10f/lib/Core/Executor.cpp#L2965
> 
> Consider setting something like 1gb, and perhaps that way we won't go over 3gb :)

Ok started 2020-10-07-klee experiment with 1 gb, lets see how it goes.

PR #796, opened 2020-10-07
Use OSS-Fuzz target name for non-OSS-Fuzz benchmarks.

inferno-chromium:
This will be useful to do some experiments where we want to start
from saturated OSS-Fuzz corpora and this fix helps to make sure we
use that corpora from OSS-Fuzz (History: non OSS-Fuzz benchmarks
are actually FTS benchmarks made from old OSS-Fuzz project commits,
so we can use same target names).


lszekeres (approved):

jonathanmetzman (commented):
LGTM

jonathanmetzman:
It's fine if we want to rename these targets so that we can get the corpora.
But the code won't depend on using OSS-Fuzz projects right?

inferno-chromium:
> It's fine if we want to rename these targets so that we can get the corpora.
> But the code won't depend on using OSS-Fuzz projects right?

yes code is not dependent in any way, just to get corpus, e.g. https://pantheon.corp.google.com/storage/browser/fuzzbench-data/2020-10-07-ossfuzzc/oss_fuzz_corpora

PR #797, opened 2020-10-07
remove variants, add variants, request experiment

vanhauser-thc:
tries two things:
1. cache all queue entries in memory (faster splicing etc.)
2. instead of cycles: fuzz entries depending on their performance score


inferno-chromium (approved):

inferno-chromium:
aflplusplus_memcache_inf seems to fail with matio which is a recently added benchmark for bug based benchmarking, could this be an easy fix, otherwise we can merge and you can look later ?

inferno-chromium:
> I am afk, previously all new variants failed because the Fuzzer was not linked with zlib, so that was a benchmark bug. Dunno if that is the issue you mean? If it crashes during the test run then it runs out of memory which is fine and within the test parameters :)

yes right benchmark issue and looks even known (we shouldnt have allowed that many blacklists - https://github.com/google/fuzzbench/pull/666/files). merging this, this benchmark is only in CI and not in automated service, so not impacting prod.

PR #800, opened 2020-10-08
Fix matio benchmark, add missing zlib dependency.

inferno-chromium:
Also, fix the unsupported fuzzers list.

### DIFF
--- a/benchmarks/matio_matio_fuzzer/benchmark.yaml
+++ b/benchmarks/matio_matio_fuzzer/benchmark.yaml
@@ -3,38 +3,5 @@ commit_date: 2019-09-04 02:05:00+00:00
 fuzz_target: matio_fuzzer
 project: matio
 unsupported_fuzzers:
-  - aflcc
-  - aflplusplus
-  - aflplusplus_coe
-  - aflplusplus_coe_maxfactor
-  - aflplusplus_coe_nofav
-  - aflplusplus_coe_rareoften
-  - aflplusplus_coe_skip
-  - aflplusplus_coe_24
-  - aflplusplus_coe_48
-  - aflplusplus_coe_64
-  - aflplusplus_exploit
-  - aflplusplus_exploit_48
-  - aflplusplus_exploit_64
-  - aflplusplus_explore
-  - aflplusplus_explore_min8
-  - aflplusplus_explore_min16
-  - aflplusplus_explore_pow3
-  - aflplusplus_explore_pow5
-  - aflplusplus_fast
-  - aflplusplus_fast_maxfactor
-  - aflplusplus_fast_nofav
-  - aflplusplus_fast_rareoften
-  - aflplusplus_fast_skip
-  - aflplusplus_fast_24
-  - aflplusplus_fast_48
-  - aflplusplus_fast_64
-  - aflplusplus_lightweizz
-  - aflplusplus_qemu
-  - aflplusplus_rare
-  - afl_qemu
-  - honggfuzz_qemu
   - klee
-  - libaflfuzzer
-  - weizz_qemu
 
--- a/benchmarks/matio_matio_fuzzer/build.sh
+++ b/benchmarks/matio_matio_fuzzer/build.sh
@@ -26,5 +26,5 @@ for fuzzers in $(find $SRC -name '*_fuzzer.cc'); do
   base=$(basename -s .cc $fuzzers)
   $CXX $CXXFLAGS -std=c++11 -Iinclude \
   $fuzzers ./getopt/.libs/libgetopt.a \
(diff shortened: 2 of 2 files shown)


inferno-chromium:
@Vaush as fyi

inferno-chromium:
Merging to unblock CI.

jonathanmetzman:
Very surprising to me that this mistake was made. @Vaush didn't we just copy this directly from the Dockerfile?

inferno-chromium:
And https://github.com/google/fuzzbench/commit/9560b28e2d7c2dc1bf615eec047f34a04eda09c2 as well. i just fixed by checking OSS-Fuzz project integration.

Vaush:
> 
> 
> Very surprising to me that this mistake was made. @Vaush didn't we just copy this directly from the Dockerfile?

Yes, the Dockerfile, as of now, hasn't been modified. It's exactly the same as the one generated by the script to integrate OSS-Fuzz benchmarks.
I am not entirely sure why the problem didn't show up with afl, libfuzzer and others, but my guess is that whoever integrated matio in OSS-Fuzz made the mistake and it was never caught because the fuzzer used on OSS-Fuzz doesn't trigger the issue?
I am don't know exactly how the integration from OSS-Fuzz to FuzzBench works internally, but I am guessing the compile command is taken from the OSS-Fuzz project, so it would at least make sense if this was the case.

Vaush:
Just to be 100% sure, I see that the fuzzer list was fixed, and zlib included and linked to, so there is nothing I need to do on my part for this, right?

inferno-chromium:
> > Very surprising to me that this mistake was made. @Vaush didn't we just copy this directly from the Dockerfile?
> 
> Yes, the Dockerfile, as of now, hasn't been modified. It's exactly the same as the one generated by the script to integrate OSS-Fuzz benchmarks.
> I am not entirely sure why the problem didn't show up with afl, libfuzzer and others, but my guess is that whoever integrated matio in OSS-Fuzz made the mistake and it was never caught because the fuzzer used on OSS-Fuzz doesn't trigger the issue?
> I am don't know exactly how the integration from OSS-Fuzz to FuzzBench works internally, but I am guessing the compile command is taken from the OSS-Fuzz project, so it would at least make sense if this was the case.

Just fyi, problem did show for afl, libfuzzer and honggfuzz fuzzers, CI didnt break as you added those in unsupported_fuzzers.

Vaush:
> 
> 
> > > Very surprising to me that this mistake was made. @Vaush didn't we just copy this directly from the Dockerfile?
> > 
> > 
> > Yes, the Dockerfile, as of now, hasn't been modified. It's exactly the same as the one generated by the script to integrate OSS-Fuzz benchmarks.
> > I am not entirely sure why the problem didn't show up with afl, libfuzzer and others, but my guess is that whoever integrated matio in OSS-Fuzz made the mistake and it was never caught because the fuzzer used on OSS-Fuzz doesn't trigger the issue?
> > I am don't know exactly how the integration from OSS-Fuzz to FuzzBench works internally, but I am guessing the compile command is taken from the OSS-Fuzz project, so it would at least make sense if this was the case.
> 
> Just fyi, problem did show for afl, libfuzzer and honggfuzz fuzzers, CI didnt break as you added those in unsupported_fuzzers.

Only for their qemu version though, which shouldn't be what's used by OSS-Fuzz. Afl and libfuzzer built normally, I checked locally with those two before even sending the PR.

inferno-chromium:
> > > > Very surprising to me that this mistake was made. @Vaush didn't we just copy this directly from the Dockerfile?
> > > 
> > > 
> > > Yes, the Dockerfile, as of now, hasn't been modified. It's exactly the same as the one generated by the script to integrate OSS-Fuzz benchmarks.
> > > I am not entirely sure why the problem didn't show up with afl, libfuzzer and others, but my guess is that whoever integrated matio in OSS-Fuzz made the mistake and it was never caught because the fuzzer used on OSS-Fuzz doesn't trigger the issue?
> > > I am don't know exactly how the integration from OSS-Fuzz to FuzzBench works internally, but I am guessing the compile command is taken from the OSS-Fuzz project, so it would at least make sense if this was the case.
> > 
> > 
> > Just fyi, problem did show for afl, libfuzzer and honggfuzz fuzzers, CI didnt break as you added those in unsupported_fuzzers.
> 
> Only for their qemu version though, which shouldn't be what's used by OSS-Fuzz. Afl and libfuzzer built normally, I checked locally with those two before even sending the PR.

it was breaking compile for me for libfuzzer, but whatever, it is fixed now. no worries, this was more of a fyi.

Vaush:
I see, that's interesting. Thanks for fixing this, I am just trying to understand the situation completely so I don't repeat the same mistake on the next benchmarks.

PR #803, opened 2020-10-08
Don't stop on crashes for afl based fuzzers.

inferno-chromium:
This is needed for benchmarking OSS-Fuzz corpora which can
have corpus items with crashes.


laurentsimon (approved):

jonathanmetzman (commented):
LGTM

Issue #801, opened 2020-10-08
libFuzzer, entropic patch.diff is causing crashes on "No such file or directory" when provided with seed inputs


inferno-chromium:
E.g. run
`make run-libfuzzer-matio_matio_fuzzer`

You will see a crash

```
8851: cov: 137 ft: 190 corp: 38 exec/s 110 oom/timeout/crash: 7/0/0 time: 36s job: 8 dft_time: 0
No such file or directory: ��hg; exiting
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/out/fuzzer.py", line 44, in fuzz
    extra_flags=['-keep_seed=1', '-cross_over_uniform_dist=1'])
  File "/out/fuzzer.py", line 76, in run_fuzzer
    subprocess.check_call(command)
  File "/usr/local/lib/python3.7/subprocess.py", line 363, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/out/matio_fuzzer', '-print_final_stats=1', '-close_fd_mask=3', '-fork=1', '-ignore_ooms=1', '-ignore_timeouts=1', '-ignore_crashes=1', '-detect_leaks=0', '-keep_seed=1', '-cross_over_uniform_dist=1', '/out/corpus', '/out/seeds']' returned non-zero exit status 1.
ERROR:root:Executed command: "nice -n 5 python3 -u -c import fuzzer; fuzzer.fuzz('/out/seeds', '/out/corpus', '/out/matio_fuzzer')" returned: 1.
ERROR:root:Fuzz process returned nonzero. Extras: {'traceback': 'Traceback (most recent call last):\n  File "/src/experiment/runner.py", line 224, in run_fuzzer\n    env=fuzzer_environment)\n  File "/src/common/new_process.py", line 124, in execute\n    raise subprocess.CalledProcessError(retcode, command)\nsubprocess.CalledProcessError: Command \'[\'nice\', \'-n\', \'5\', \'python3\', \'-u\', \'-c\', "import fuzzer; fuzzer.fuzz(\'/out/seeds\', \'/out/corpus\', \'/out/matio_fuzzer\')"]\' returned non-zero exit status 1.\n'}
```

This is blocking the work on trying to run an experiment with OSS-Fuzz corpora on all benchmarks to see if they can break the coverage wall.

inferno-chromium:
[...]

PR #806, opened 2020-10-09
Workaround for #801

mboehme:


inferno-chromium:
Can you please add entropic as well. Wish these could come back, since it affects rankings significantly.

mboehme:
If this isn't very urgent, I can look at it properly within a week?

inferno-chromium:
> If this isn't very urgent, I can look at it properly within a week?

Yes not urgent at all, whenever you get time. Ideally we want that patch to stablize so that we can get in LLVM and benefit all consumers in industry (similar to how entropic is default now).

inferno-chromium:
> If this isn't very urgent, I can look at it properly within a week?

Yes not urgent at all, whenever you get time. Ideally we want that patch to stablize so that we can get in LLVM and benefit all consumers in industry (similar to how entropic is default now).

mboehme:
Sure. Closing this then.

jonathanmetzman:
> > If this isn't very urgent, I can look at it properly within a week?
> 
> Yes not urgent at all, whenever you get time. Ideally we want that patch to stablize so that we can get in LLVM and benefit all consumers in industry (similar to how entropic is default now).

Youre not saying we need to upstream everything before testing it right? I think it's fine (even better) to do the testing here first.

inferno-chromium:
> > > If this isn't very urgent, I can look at it properly within a week?
> > 
> > 
> > Yes not urgent at all, whenever you get time. Ideally we want that patch to stablize so that we can get in LLVM and benefit all consumers in industry (similar to how entropic is default now).
> 
> Youre not saying we need to upstream everything before testing it right? I think it's fine (even better) to do the testing here first.

Yes +1. i mean we can test things here and let it stabilize here, before submitting an upstream patch to llvm. We have seen that this breaks [i had to do this disabling before https://www.fuzzbench.com/reports/experimental/2020-10-11-ossfuzzc/index.html] and causes libfuzzer/entropic to exit/crash when a number of seeds are used. So, in a way it is important to fix this, but it can wait a week or two, otherwise, we can just commit this one to temporarily disable it.

Issue #812, opened 2020-10-11
Determining which experiment data was used in a report.


wideglide:
I didn't see this covered in the documentation or another issue, but please point me to a reference if it exists.
 I don't see how one could correlate which experiment data is used to generate a specific report.

For example with the 2020-09-24 report, the linked raw [data.csv.gz](https://www.fuzzbench.com/reports/2020-09-24/data.csv.gz) has lots of additional data that was filtered in order to generate the report.
Determining which fuzzers were filtered is pretty easy, but determining which experiments were used doesn't seem so straightforward.  There are 6 six different experiments where `aflplusplus` was run.  Is the data shown in the report only from the most recent one (2020-09-19)?  Is this always the case?

![image](https://user-images.githubusercontent.com/558033/95686422-c10c2f80-0bcb-11eb-8baa-033e86ab09fc.png)

inferno-chromium:
ah right, we should be only saving filtered data, that is a bug.

See code
```
    if benchmarks is not None:
        experiment_df = data_utils.filter_benchmarks(experiment_df, benchmarks)

    if fuzzers is not None:
        experiment_df = data_utils.filter_fuzzers(experiment_df, fuzzers)

    if label_by_experiment:
        experiment_df = data_utils.label_fuzzers_by_experiment(experiment_df)

    if end_time is not None:
        experiment_df = data_utils.filter_max_time(experiment_df, end_time)

    if merge_with_clobber or merge_with_clobber_nonprivate:
        experiment_df = data_utils.clobber_experiments_data(
            experiment_df, experiment_names)
```

Basically main thing is clobber_experiments_data, only fuzzers run in that experiment are used and then merged with core-fuzzers [https://github.com/google/fuzzbench/blob/master/service/core-fuzzers.yaml] from earlier experiments.

jonathanmetzman:
I'm frankly not so sure we should be only saving filtered data. Probably better would be to store all of it but provide the command used for filtering etc.

jonathanmetzman:
I'm not sure there are downsides to the new approach so I guess I'm fine with it.

inferno-chromium:
> I'm frankly not so sure we should be only saving filtered data. Probably better would be to store all of it but provide the command used for filtering etc.

The problem was it was saving all the other experiment data from merge experiment list. So, basically data.csv.gz would have data from like 10 older experiments, even when it might be using data from last one [like current one + core-fuzzers from lets say last one].

PR #813, opened 2020-10-12
Store filtered data in csv.

inferno-chromium:
Since we merge data across earlier experiments and do lot of filters,
it can be very confusing to add data from all experiments. Only keep
the filtered data in csv that was used to generate the report.

Fixes https://github.com/google/fuzzbench/issues/812

### DIFF
--- a/analysis/generate_report.py
+++ b/analysis/generate_report.py
@@ -156,8 +156,6 @@ def generate_report(experiment_names,
         experiment_df = pd.read_csv(data_path)
     else:
         experiment_df = queries.get_experiment_data(experiment_names)
-        # Save the raw data along with the report.
-        experiment_df.to_csv(data_path)
 
     data_utils.validate_data(experiment_df)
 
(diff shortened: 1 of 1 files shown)


lszekeres (approved):

analysis/generate_report.py
```diff
@@ -177,6 +175,11 @@ def generate_report(experiment_names,
         experiment_df = data_utils.clobber_experiments_data(
             experiment_df, experiment_names)
 
+    # Save the filtered raw data along with the report if not using cached data
+    # or if the data does not exist.
+    if not from_cached_data or not os.path.exists(data_path):
```
jonathanmetzman, line 180: The `or not os.path.exists(data_path)` results in wrong behavior.
What if I regenerate a report with different flags?

inferno-chromium: If you are regenerating a report, you probably would be using cached data flag to reuse data. and then it wont overwrite.

This is important if you are reusing report directory. lets say you generated report with experiment 1 and then try to generate report with experiment 2, you would want to make sure that you have data.csv from experiment 2. wdyt ?

PR #815, opened 2020-10-12
fix seed selection bug and retest

vanhauser-thc:
its crazy.
I totally fucked up the seed selection algorithm - and it outperformed by a margin :D 
Now this fixes that bug - and lets see if it still performs well ...
also new test of the testcache implementation.


inferno-chromium (approved):

jonathanmetzman (commented):
LGTM

PR #822, opened 2020-10-13
Add optional experiment description

### LINKED ISSUE #805: Suggestion: Add PR description to generated reports for experimentals
Adding the intention for an experiment to the PR description seems to work well for my recent PRs. If these descriptions seem useful, maybe they could be added to the generated reports to provide the context of the results?

jonathanmetzman:
This can be specified using the service or manually.
When specified, Fuzzbench will include the description in the experiment report.
Fixes #805.

### DIFF
--- a/analysis/generate_report.py
+++ b/analysis/generate_report.py
@@ -157,6 +157,8 @@ def generate_report(experiment_names,
     else:
         experiment_df = queries.get_experiment_data(experiment_names)
 
+    description = queries.get_experiment_description(experiment_names[0])
+
     data_utils.validate_data(experiment_df)
 
     if benchmarks is not None:
@@ -197,7 +199,8 @@ def generate_report(experiment_names,
 
     template = report_type + '.html'
     detailed_report = rendering.render_report(experiment_ctx, template,
-                                              in_progress, coverage_report)
+                                              in_progress, coverage_report,
+                                              description)
 
     filesystem.write(os.path.join(report_directory, 'index.html'),
                      detailed_report)
--- a/analysis/generate_report.py
+++ b/analysis/generate_report.py
@@ -147,7 +147,8 @@ def generate_report(experiment_names,
             queries.add_nonprivate_experiments_for_merge_with_clobber(
                 experiment_names))
 
-    report_name = report_name or experiment_names[0]
+    main_experiment_name = experiment_names[0]
+    report_name = report_name or main_experiment_name
 
     filesystem.create_directory(report_directory)
 
@@ -157,7 +158,7 @@ def generate_report(experiment_names,
     else:
         experiment_df = queries.get_experiment_data(experiment_names)
 
-    description = queries.get_experiment_description(experiment_names[0])
+    description = queries.get_experiment_description(main_experiment_name)
 
     data_utils.validate_data(experiment_df)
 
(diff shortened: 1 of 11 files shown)


analysis/generate_report.py
```diff
@@ -157,6 +157,8 @@ def generate_report(experiment_names,
     else:
         experiment_df = queries.get_experiment_data(experiment_names)
 
+    description = queries.get_experiment_description(experiment_names[0])
```
inferno-chromium, line 160: nit: move experiment_names[0] into a local main_experiment and use both at line 150 and 160.

analysis/queries.py
```diff
@@ -36,6 +36,15 @@ def get_experiment_data(experiment_names):
     return pd.read_sql_query(snapshots_query.statement, db_utils.engine)
 
 
+def get_experiment_description(experiment_name):
+    """Get the description of the experiment named by |experiment_name|."""
+    # Do another query for the description so we don't explode the size of the
+    # results from get_experiment_data.
+    return db_utils.query(Experiment.description)\
+            .select_from(Experiment)\
+            .filter(Experiment.name.in_(experiment_names)).one()
```
inferno-chromium, line 45: typo experiment_names ? without s ?

analysis/report_templates/default.html
```diff
@@ -303,6 +303,12 @@ <h2 class="green-text text-darken-3">experiment data</h2>
             The experiment was conducted using this FuzzBench commit:
             <a href="https://github.com/google/fuzzbench/commits/{{ experiment.git_hash }}">{{ experiment.git_hash }}</a>
             {% endif %}
+
+            {% if description %}
```
inferno-chromium, line 307: I dont think anyone will find this at bottom, shouldnt it be at the top of page ? if you still prefer to keep in this last section, it should be the starting part in this section ?

service/test_automatic_run_experiment.py
```diff
@@ -94,7 +94,7 @@ def test_run_requested_experiment(mocked_get_requested_experiments,
     ]
     expected_calls = [
         mock.call(expected_experiment_name, expected_config_file,
-                  expected_benchmarks, expected_fuzzers)
+                  expected_benchmarks, expected_fuzzers, None)
```
inferno-chromium, line 97: probably add a description in EXPERIMENT_REQUESTS so that we can see it works.

inferno-chromium (approved):

analysis/generate_report.py
jonathanmetzman: Done.

analysis/report_templates/default.html
jonathanmetzman: >I dont think anyone will find this at bottom, shouldnt it be at the top of page ?

Probably not but I think it should go in this section. Maybe the section should be moved up?

I don't know. I think putting it last makes it easy to tell when the description ends. Maybe if we had some way of description from other text it could be moved up in this section. 
I'm totally fine if someone wants to move this later, but I don't really want to get into heavy frontend work in this PR.

service/test_automatic_run_experiment.py
jonathanmetzman: Done.

PR #828, opened 2020-10-14
Use aflplusplus and aflplusplus_optimal in core fuzzers list.

inferno-chromium:


inferno-chromium:
@vanhauser-thc - just confirming that you are recommending this for all cases right ?

vanhauser-thc:
honestly both should be in there.
aflplusplus_optimal is an optimal setting carefully crafted for each target.
aflplusplus is a generic setting, same for all.
this basically shows the difference you can achieve when analyzing the target - or not.
most other fuzzers (I think aflcc is the only exception that comes to my mind) do not even have configurable options that can have such a different impact.

vanhauser-thc:
if you plan again for a restart experiment (like you just did) I would add a specific fuzzer setup for this as the two existing setups are not effective in that circumstance.

andreafioraldi:
> honestly both should be in there.
> aflplusplus_optimal is an optimal setting carefully crafted for each target.
> aflplusplus is a generic setting, same for all.
> this basically shows the difference you can achieve when analyzing the target - or not.
> most other fuzzers (I think aflcc is the only exception that comes to my mind) do not even have configurable options that can have such a different impact.

I agree for the experimental reports, optimal is useful, but not for the generic ones. As optimal is an overfit, if we compare fuzzers we are cherry picking with optimal, and this is scientifically unfair.

inferno-chromium:
> > honestly both should be in there.
> > aflplusplus_optimal is an optimal setting carefully crafted for each target.
> > aflplusplus is a generic setting, same for all.
> > this basically shows the difference you can achieve when analyzing the target - or not.
> > most other fuzzers (I think aflcc is the only exception that comes to my mind) do not even have configurable options that can have such a different impact.
> 
> I agree for the experimental reports, optimal is useful, but not for the generic ones. As optimal is an overfit, if we compare fuzzers we are cherry picking with optimal, and this is scientifically unfair.

Ok, @vanhauser-thc @andreafioraldi - are you both on concensus now to have just regular aflplusplus in official experiments.

andreafioraldi:
I understand that the ability to configure afl++ to perform better is an engineering plus, but fuzzbench mostly evaluates algorithms.

vanhauser-thc:
I disagree with Andrea. optimal should be the core Fuzzer. Optimal is not the only one having target specific settings, although it uses it the most

inferno-chromium:
closing until we get concensus.

vanhauser-thc:
closing it basically means the end of discussion.
if there is consensus - would the issue be still open until we dont have a consense anymore? ;)

what is your opinion on aflplusplus vs aflplusplus_optimal @inferno-chromium ?

IMHO until there is consensus either both or none of the two variants should be in core_fuzzers. otherwise it is already decided.

inferno-chromium:
i am fine with keeping both. @jonathanmetzman any thoughts ?

jonathanmetzman:
sgtm

inferno-chromium:
> closing it basically means the end of discussion.
> if there is consensus - would the issue be still open until we dont have a consense anymore? ;)
> 
> what is your opinion on aflplusplus vs aflplusplus_optimal @inferno-chromium ?
> 
> IMHO until there is consensus either both or none of the two variants should be in core_fuzzers. otherwise it is already decided.

i am fine with both and looks like Jonathan is fine too, so committing.

PR #835, opened 2020-10-15
Set sanitizer flags when benchmark is bug based benchmark.

inferno-chromium:


inferno-chromium:
> I think this might require a little more thought.
> Initially I wanted each benchmark to include sanitizer flags in its build.sh.
> This would have made it very easy for one benchmark to use ASAN, another UBSAN another ASAN+UBSAN etc.
> But I realize that may cause issues with fuzzers that need an instrumented binary.
> 
> So we have two options I think:
> 
> 1. Go with the above approach and don't support those fuzzers for crash benchmarks.
> 2. Have some way of mapping sanitizer flags to benchmarks and add those flags during a build (similar to what you did above except not hardcoded).
> 
> I don't think the approach you did here is good because it assumes we want ASAN for all crash benchmarks.

We can easily extend this if we needed a different sanitizer by adding a "sanitizer: " in benchmark.yaml and then we add the flag value based on it. we can even do "sanitizer: address, undefined" to support both.

fuzzers/utils.py
```diff
@@ -55,31 +74,23 @@ def append_flags(env_var, additional_flags, env=None):
...
-    can be compiled without a sanitizer. If |env| is not provided, the program's
-    environment will be used."""
-    if env is None:
-        env = os.environ
-    env['CFLAGS'] = ' '.join(NO_SANITIZER_COMPAT_CFLAGS)
-    env['CXXFLAGS'] = ' '.join(NO_SANITIZER_COMPAT_CXXFLAGS)
+def get_config_value(attribute):
+    """Gets config attribute value from benchmark config yaml file."""
+    with open(BENCHMARK_CONFIG_YAML_PATH) as file_handle:
+        # Hacky yaml parsing without PyYAML dependency.
```
jonathanmetzman, line 87: Why are we doing this? Why can't we use pyyaml. I think we should never ever do this.

fuzzers/utils.py
```diff
@@ -150,19 +161,44 @@ def get_dictionary_path(target_binary):
     return None
 
 
-def set_default_optimization_flag(env=None):
-    """Set default optimization flag if none is already set."""
-    if not env:
+def set_fuzz_target(env=None):
+    """Set compilation flags."""
```
jonathanmetzman, line 165: Comment is wrong.

fuzzers/utils.py
```diff
@@ -150,19 +161,44 @@ def get_dictionary_path(target_binary):
     return None
 
 
-def set_default_optimization_flag(env=None):
-    """Set default optimization flag if none is already set."""
-    if not env:
+def set_fuzz_target(env=None):
```
jonathanmetzman, line 164: I'm confused why this function is necessary now?

fuzzers/utils.py
inferno-chromium: The reason being at fuzzer runtime, we use runner.Dockerfile from fuzzer directory and we can't install pyyaml dependency there.

jonathanmetzman: Are you talking about runtime or buildtime (double checking).
I didn't fuzzers/utils.py used to only be a buildtime thing.
In any case, I'm pretty sure we have pyyaml at runtime because fuzzer.py at runtime is invoked by experiment/runner.py which I believe has these dependencies.

fuzzers/utils.py
inferno-chromium: Used by some fuzzers

aflcc/fuzzer.py:    fuzz_target = os.getenv('FUZZ_TARGET')
aflplusplus/fuzzer.py:        # For CmpLog build, set the OUT and FUZZ_TARGET environment
aflplusplus/fuzzer.py:        fuzz_target = os.getenv('FUZZ_TARGET')
aflplusplus/fuzzer.py:            new_env['FUZZ_TARGET'] = os.path.join(cmplog_build_directory,
klee/fuzzer.py:    fuzz_target = os.getenv('FUZZ_TARGET')

i removed the hack we had in dockerfile to parse it since we would be parsing type attribute anyway here.

fuzzers/utils.py
inferno-chromium: Adding at buildtime is okish, i can add requirements.txt for benchmark-builder.
but runtime, it was not working, let me check more.

fuzzers/utils.py
inferno-chromium: fixed

fuzzers/utils.py
inferno-chromium: Fixed as per offline discussion.

inferno-chromium:
PTAL.

inferno-chromium:
Recap on new patch

Use one requirements.txt for builder and runner.
    
    - Use main requirements.txt in base-image.
    - In benchmark-builder, copy python3 packages from base-images as well.
    - In fuzzer runner, it inherits from base-image, so it gets the python deps automatically.

requirements.txt
```diff
@@ -12,6 +12,7 @@ pyfakefs==3.7.1
 pytest==5.3.5
 python-dateutil==2.8.1
 pytz==2019.3
+psutil==5.7.2
```
jonathanmetzman, line 15: I'm opposed to adding packages to requirements.txt because a fuzzer needs it. I don't think this is the right place.

fuzzers/klee/runner.Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM gcr.io/fuzzbench/base-runner
+FROM gcr.io/fuzzbench/base-image
```
jonathanmetzman, line 15: Does base-runner no longer exist?

inferno-chromium: yes right, it does not since July, this is all old code that was working due to cache, i deleted it now.

requirements.txt
inferno-chromium: Should i add another requirements.txt inside base-image directory with just those 2-3 deps, i thought you preferred one requirements.txt.

jonathanmetzman: I prefer 1 requirements.txt for all of FuzzBench (not including fuzzers/benchmarks).
However, if a fuzzer or a benchmark needs Python packages, they need to install it in builder/runner.
Just like it would be wrong to add the apt package ninja to base-image because some random benchmark needs it, it is also wrong to add a python package because one benchmark needs it.

inferno-chromium: psutil is a general utility, that is the reason i put it in requirements.txt [for other fuzzer use it is nice to print things with it]. currently people needs to add pylint hacks and move the import if some thing is installed in their runner docker. pick one option after this comment and i will either keep it as-is or move it back.

jonathanmetzman: > psutil is a general utility, that is the reason i put it in requirements.txt [for other fuzzer use it is nice to print things with it]. currently people needs to add pylint hacks and move the import if some thing is installed in their runner docker. pick one option after this comment and i will either keep it as-is or move it back.

I think it should not be installed anywhere general I think it should be installed only by the fuzzers that use it.
I thought the purpose of your change was to eliminate the second requirements.txt we had for the runner images right?

fuzzers/klee/runner.Dockerfile
jonathanmetzman: ACK. Thanks!

requirements.txt
inferno-chromium: yes that is the purpose, so reverting the psutil and keeping it to klee (as originally).

fuzzers/utils.py
```diff
@@ -17,15 +17,35 @@
 import configparser
 import contextlib
 import os
+import re
 import shutil
 import subprocess
 import tempfile
+import yaml
```
jonathanmetzman, line 24: Needs to be seperated from other imports as others are of stdlib packages and yaml is third_party.

fuzzers/utils.py
```diff
@@ -17,15 +17,35 @@
...
+# libraries linked into the sanitizer runtime. These flags link against those
+# libraries.
+NO_SANITIZER_COMPAT_CFLAGS = [
+    '-pthread', '-Wl,--no-as-needed', '-Wl,-ldl', '-Wl,-lm',
+    '-Wno-unused-command-line-argument'
+]
+
 OSS_FUZZ_LIB_FUZZING_ENGINE_PATH = '/usr/lib/libFuzzingEngine.a'
+BENCHMARK_CONFIG_YAML_PATH = '/benchmark.yaml'
+CONFIG_ATTR_VAL_REGEX = re.compile(
```
jonathanmetzman, line 47: Can remove.

fuzzers/utils.py
```diff
@@ -150,21 +157,46 @@ def get_dictionary_path(target_binary):
...
+        append_flags('CFLAGS',
+                     NO_SANITIZER_COMPAT_CFLAGS + [DEFAULT_OPTIMIZATION_LEVEL],
+                     env=env)
+        append_flags('CXXFLAGS',
+                     NO_SANITIZER_COMPAT_CFLAGS +
+                     [LIBCPLUSPLUS_FLAG, DEFAULT_OPTIMIZATION_LEVEL],
+                     env=env)
 
 
 def initialize_flags(env=None):
```
jonathanmetzman, line 194: Maybe change this function name to `initialize_env` since it is initializing more than flags.

fuzzers/utils.py
```diff
@@ -150,21 +157,46 @@ def get_dictionary_path(target_binary):
...
     """Set initial flags before fuzzer.build() is called."""
-    set_no_sanitizer_compilation_flags(env)
-    set_default_optimization_flag(env)
+    set_fuzz_target(env)
+    set_compilation_flags(env)
 
-    for flag_var in ['CFLAGS', 'CXXFLAGS']:
+    for flag_var in ['FUZZ_TARGET', 'CFLAGS', 'CXXFLAGS']:
```
jonathanmetzman, line 199: `flag_var` should be renamed if we want to keep this loop, but frankly maybe we should just delete it?

jonathanmetzman (approved):

fuzzers/utils.py
inferno-chromium: useful for debugging, renaming and keeping

jonathanmetzman:
Did you intend to merge this even though CI is failing?

inferno-chromium:
> Did you intend to merge this even though CI is failing?

Ci failure was expected since aflcc_no_orig_dict was intended to be removed. aflcc_no_orig_dict was a temporary variant addition that was not properly added in CI when it originally landed. also, it missed the benchmark blacklists (similar to aflcc).

aflcc_no_orig_dict is not included in fuzzer list in /home/runner/work/fuzzbench/fuzzbench/.github/workflows/fuzzers.yml.
Failed.
Makefile:60: recipe for target 'presubmit' failed

Regarding qemu benchmark failures, they are not related to this. Something with matio benchmark does not work with qemu fuzzers, leads to 
```
2020-10-22T02:25:36.7057986Z [1;91m[-] [0mWhoops, the target binary crashed suddenly, before receiving any input
2020-10-22T02:25:36.7058551Z     from the fuzzer! There are several probable explanations:
2020-10-22T02:25:36.7059015Z 
2020-10-22T02:25:36.7060030Z     - The binary is just buggy and explodes entirely on its own. If so, you
2020-10-22T02:25:36.7060551Z       need to fix the underlying problem or find a better replacement.
2020-10-22T02:25:36.7061178Z 
2020-10-22T02:25:36.7061880Z     - Less likely, there is a horrible bug in the fuzzer. If other options
2020-10-22T02:25:36.7062409Z       fail, poke <lcamtuf@coredump.cx> for troubleshooting tips.
2020-10-22T02:25:36.7063061Z )B[?25h[0m[1;91m
2020-10-22T02:25:36.7063523Z [-] PROGRAM ABORT : [1;97mFork server crashed with signal 9[1;91m
2020-10-22T02:25:36.7064082Z          Location : [0minit_forkserver(), afl-fuzz.c:2212
```

PR #836, opened 2020-10-16
Store crashes for libFuzzer, honggfuzzer in corpus sub-dir.

inferno-chromium:
This is needed for bug based benchmarking, as we archive corpus
dir and re-run it in measurer.

### DIFF
--- a/fuzzers/afl/fuzzer.py
+++ b/fuzzers/afl/fuzzer.py
@@ -69,9 +69,6 @@ def run_afl_fuzz(input_corpus,
                  hide_output=False):
     """Run afl-fuzz."""
     # Spawn the afl fuzzing process.
-    # FIXME: Currently AFL will exit if it encounters a crashing input in seed
-    # corpus (usually timeouts). Add a way to skip/delete such inputs and
-    # re-run AFL.
     print('[run_afl_fuzz] Running target with afl-fuzz')
     command = [
         './afl-fuzz',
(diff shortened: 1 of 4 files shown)


fuzzers/afl/fuzzer.py
```diff
@@ -69,9 +69,6 @@ def run_afl_fuzz(input_corpus,
                  hide_output=False):
     """Run afl-fuzz."""
     # Spawn the afl fuzzing process.
-    # FIXME: Currently AFL will exit if it encounters a crashing input in seed
```
inferno-chromium, line 72: Stale comment, already fixed by AFL_SKIP_CRASHES above.

fuzzers/libfuzzer/fuzzer.py
```diff
@@ -63,6 +63,9 @@ def run_fuzzer(input_corpus, output_corpus, target_binary, extra_flags=None):
         # Don't use LSAN's leak detection. Other fuzzers won't be using it and
         # using it will cause libFuzzer to find "crashes" no one cares about.
         '-detect_leaks=0',
+
+        # Store crashes along with corpus for bug based benchmarking.
+        f'-artifact_prefix={output_corpus}/',
```
jonathanmetzman, line 68: This might have some weird effects because libFuzzer reloads its corpus directories (maybe this one wont be reloaded because it is an output corpus).
Maybe the right approach is to put output corpora dirs in a subdirectory of output_corpus? WDYT?

inferno-chromium: Fixed.

inferno-chromium:
@jonathanmetzman - fixed as per discussion. PTAL.

jonathanmetzman (approved):
lgtm

PR #837, opened 2020-10-16
Requesting new experiments for several AFL schedules

mboehme:
Evaluating several AFL schedules. There are two edge-count-based boosting schedules (afl_branches_u8, afl_branches_16). For each edge, AFL already maintains favourite seeds (the fastest and smallest seed). We choose identify rarely executed edges and mark those favourites as 'champions', and skip most non-champions as long as some champions exist. There are two FAST schedules that also consider the number of times a seed is fuzzed (afl_old_fast_1, afl_old_fast_025). There are two schedules that further de-prioritize slow seeds (afl_quick, afl_fast_quick). In previous experiments, we observed that AFL doesn't get through the entire queue in 23 hours. There are five baselines (afl, afl_shuffled, afl_nofast, afl_fast, and afl_coe) where afl_shuffled starts with a shuffled queue, and afl_nofast uses xxh3 for hashing.


fuzzers/afl_nofast/fuzzer.py
```diff
@@ -0,0 +1,109 @@
...
+        # Pass INT_MAX to afl the maximize the number of persistent loops it
+        # performs.
+        '2147483647'
+    ]
+    print('[run_afl_fuzz] Running command: ' + ' '.join(command))
+    output_stream = subprocess.DEVNULL if hide_output else None
+    subprocess.check_call(command, stdout=output_stream, stderr=output_stream)
+
+
+def fuzz(input_corpus, output_corpus, target_binary):
```
inferno-chromium, line 105: Instead of copying afl's fuzzer.py, please call into afl's fuzzer helper functions. This helps to make all afl changes in one place rather than all variants.

fuzzers/afl_shuffled/fuzzer.py
```diff
@@ -0,0 +1,111 @@
...
+    # OSS-Fuzz is used.
+    os.environ['AFL_SKIP_CRASHES'] = '1'
+    # Enable shuffle queue
+    os.environ['AFL_SHUFFLE_QUEUE'] = '1'
+
+    # AFL needs at least one non-empty seed to start.
+    utils.create_seed_file_for_empty_corpus(input_corpus)
+
+
+def run_afl_fuzz(input_corpus,
```
inferno-chromium, line 67: Same as comment above.

inferno-chromium (changes requested):
LGTM with minor fix needed, see comments.

mboehme:
Done. Thanks Fuzzbench team and have a great weekend!

inferno-chromium:
@mboehme - funny, i was going to write a comment on CI, but you are just too fast :)

PR #838, opened 2020-10-16
Wrap LLVMFuzzerTestOneInput

jonathanmetzman:
Wrap LLVMFuzzerTestOneInput.

This can be used as a proof of concept for sampling.
It may not work with all benchmarks such as bloaty.
That is because some benchmarks depend on FUZZER_LIB being
an actual library rather than CFLAGS.

An alternative approach is to use sed and awk to rewrite source
files containing LLVMFuzzerTestOneInput.


jonathanmetzman:
Wow looks like every standard benchmark builds with the wrapped LLVMFuzzerTestOneInput.

PR #848, opened 2020-10-18
Requesting new AFL experiment

mboehme:
Experiments from 2020-10-17 showed quite some overhead [afl vs afl_nofast]. Checking overhead of xxh3 (afl_xxh3). Reduced overhead of addressing 8MB of memory for hit counts (afl_fast, etc. u8* instead of u32*). After some debugging, it seems that schedules that speed up cycles are penalized, because queue_cycle determines a seed's handicap (fuzzed 4x longer) and the size of the mutated blocks in the seed (larger blocks for later cycles). For schedules afl_fast_no_cycles, afl_oldfast_1, afl_fast_branches_u8/16, the current cycle is computed as "fuzzed_paths / queued_paths".

Some insights from the experiments on 2020-10-17,
* [afl vs afl_shuffled] A shuffled input queue doesn't make too much of a difference. At least it doesn't explain the difference on SQLite3 and Proj4.
* [afl vs afl_quick] Fuzzing less very slow inputs  doesn't make too much of a difference, either. In terms of average ranking, they are about the same. In terms of average score, afl_quick is slightly better.
* [afl vs afl_nofast] There is non-negligible overhead in computing the XXH3 chksum and addressing 8MB of memory, particularly for SQLite3 and Proj4.


inferno-chromium (approved):

mboehme:
@inferno-chromium Thanks for merging! Can't seem to find the report, though. The corresponding folder in the `fuzzbench-data` bucket only contains an input folder, but no build-logs or coverage-binaries. Maybe ran out of authorization keys, so the Cloud Build didn't run?

PR #849, opened 2020-10-19
Use sanitizers when compiling matio

Vaush:
In PR #666 I forgot to add the flags to compile matio with ASAN and UBSAN, which are needed since the bugs we have for it have been found by OSS-Fuzz using those sanitizers.

### DIFF
--- a/benchmarks/matio_matio_fuzzer/build.sh
+++ b/benchmarks/matio_matio_fuzzer/build.sh
@@ -20,7 +20,7 @@
 ./configure
 make -j$(nproc)
 make install
-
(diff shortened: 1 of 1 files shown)


inferno-chromium:
This is not needed after https://github.com/google/fuzzbench/pull/835

jonathanmetzman:
Agr

> This is not needed after #835

Agreed. Sorry for not explaining this properly, but we will do something different than what I told you in our meetings.
We aren't going to set sanitizer flags in build.sh because some fuzzers will probably need to modify with these and putting them in build.sh leaves fuzzer.py with no way to modify them.
Instead, we will set them in the environment that fuzzer.py can modify before calling utils.build_benchmark to actually build them.

Issue #821, opened 2020-10-13
Add method for verifying bugs in benchmarks


jonathanmetzman:
We are going to add benchmarks in which OSS-Fuzz found many bugs as part of bug-based benchmarking.
We need a method of verifying these crashes reproduce in the fuzzbench builds of these benchmarks.

Related to #165.

PR #850, opened 2020-10-19
Added reproducibility check functionality

Vaush:
Added the possibility of testing the bugs available for a benchmark (stored in a "testcases" directory in the benchmark's folder) for reproducibility (see Issue #821 ).
This is done simply by running compiling and running the benchmark with libfuzzer, using its capability of running reproducibility tests when one or more files are passed as arguments instead of folders.
This PR adds a new make rule, "make repro-check-$FUZZER-$BENCHMARK" in order to run the check. The make rule is only available for benchmarks whose folder contains a "testcases" directory.


Vaush:
CC @jonathanmetzman , please tell me if you already receive notifications that mention your issues, I won't CC you anymore in case

docker/generate_makefile.py
```diff
@@ -56,16 +56,26 @@ def _print_makefile_run_template(image):
...
             print('\t-e MAX_TOTAL_TIME=20 \\\n\t-e SNAPSHOT_PERIOD=10 \\')
         if run_type == 'debug':
             print('\t--entrypoint "/bin/bash" \\\n\t-it ', end='')
+        elif run_type == 'repro-check':
+            if os.path.isdir(os.path.join(BENCHMARK_DIR, benchmark,
+                                          'testcases')):
+                print('\t--entrypoint /bin/bash ', end='')
+                params = " -c \"for f in \\$$OUT/testcases/*; do\
+                        \\$$OUT/\\$$FUZZ_TARGET -timeout=25\
+                        -rss_limit_mb=2560 \\$$f; done;\""
```
jonathanmetzman, line 70: Where does `2560` come from? Is it because some of the testcases from Matio cause an OOM with 2048?

docker/benchmark-builder/Dockerfile
```diff
@@ -45,6 +45,12 @@ ENV ASAN_OPTIONS="detect_leaks=0"
 RUN mkdir /opt/fuzzbench/
 COPY docker/benchmark-builder/checkout_commit.py /opt/fuzzbench/
 COPY benchmarks/$benchmark/benchmark.yaml /
+# Use multiple copy on benchmark.yaml and the testcases folder, so as to not throw an error in case the folder doesn't exist
```
jonathanmetzman, line 48: If we check that testcases exists in generate_makefile, I'm not sure why this clever hack is needed.

docker/generate_makefile.py
Vaush: No, it's from OSS-Fuzz, docs say that the memory limit while fuzzing is 2.5GB

docker/benchmark-builder/Dockerfile
Vaush: Because the benchmark-builder is run on all the benchmarks, even those without a testcases folder. To avoid generating a whole different image just to check reproducibility, I modified the normal builder to copy the folder if present, and then the folder will be used when the repro-check make rule is called.

Vaush: But please, do tell me if there is a better way to copy the testcases folder into the final image, because I don't like this hack either, I just didn't find a way to avoid using it.

jonathanmetzman: Yes there is a better way. 
Please get rid of these changes entirely.
Then when you start the container add something like ` -v /testcases:$PATH_TO_TESTCASES_ON_HOST` this tells docker to mount `$PATH_TO_TESTCASES_ON_HOST` to `testcases` as a volume, so the testcases can be accessed.

docker/generate_makefile.py
```diff
@@ -37,12 +37,15 @@ def _print_makefile_run_template(image):
     fuzzer = image['fuzzer']
     benchmark = image['benchmark']
 
-    for run_type in ('run', 'debug', 'test-run'):
+    for run_type in ('run', 'debug', 'test-run', 'repro-check'):
```
inferno-chromium, line 40: s/repro-check/repro-crash or repro-bug ?

docker/generate_makefile.py
```diff
@@ -37,12 +37,15 @@ def _print_makefile_run_template(image):
...
+        if run_type == 'repro-check':
+            bench_path = os.path.join(BENCHMARK_DIR, benchmark, 'testcases')
+            if not os.path.isdir(bench_path):
+                continue
```
inferno-chromium, line 44: i think we need to exception here saying, "This is not a bug-based benchmark, testcases dir missing!"

docker/generate_makefile.py
```diff
@@ -56,16 +59,28 @@ def _print_makefile_run_template(image):
...
+        params = ""
         if run_type == 'test-run':
             print('\t-e MAX_TOTAL_TIME=20 \\\n\t-e SNAPSHOT_PERIOD=10 \\')
         if run_type == 'debug':
             print('\t--entrypoint "/bin/bash" \\\n\t-it ', end='')
+        elif run_type == 'repro-check':
+            print('\t-v {path}:/testcases \\\n\
+\t--entrypoint /bin/bash '.format(path=bench_path),
+                  end='')
+            params = " -c \"for f in /testcases/*; do\
```
inferno-chromium, line 71: Can this be not printed at the same time ?

inferno-chromium (changes requested):

docker/generate_makefile.py
Vaush: But isn't this run for all benchmarks? If it's not a bug based benchmark we simply don't create the make rule to run the reproducibility check, otherwise we will get exceptions every time the makefile is generated

docker/generate_makefile.py
Vaush: No, it needs to be printed after the image name, which is done after the ifs for every rule type

docker/generate_makefile.py
inferno-chromium: ah ok, sg then, no change needed.

docker/generate_makefile.py
```diff
@@ -56,16 +59,28 @@ def _print_makefile_run_template(image):
...
+                    -rss_limit_mb=2560 \\$$f; done;\""
+
         elif run_type == 'run':
             print('\t-it ', end='')
         else:
             print('\t', end='')
 
-        print(os.path.join(BASE_TAG, image['tag']))
+        print(os.path.join(BASE_TAG, image['tag']), end='')
+        print(params)
```
inferno-chromium, line 83: can you do a 
if (params) and only print when it is non-empty ? that would simplify tests.

inferno-chromium (approved):
LGTM with minor fix needed.

docker/generate_makefile.py
```diff
@@ -37,12 +37,15 @@ def _print_makefile_run_template(image):
     fuzzer = image['fuzzer']
     benchmark = image['benchmark']
 
-    for run_type in ('run', 'debug', 'test-run'):
+    for run_type in ('run', 'debug', 'test-run', 'repro-bugs'):
+        if run_type == 'repro-bugs':
+            bench_path = os.path.join(BENCHMARK_DIR, benchmark, 'testcases')
```
inferno-chromium, line 42: probably rename bench_path to something more clear like
crash_testcases_dir ?

PR #854, opened 2020-10-21
[GCB] Fix builds in other cloud projects and fix caching

jonathanmetzman:
Fix builds: Make sure we tag every image using the
gcr.io/fuzzbench style to ensure subsequent build steps can
use them.

Caching:
Previously when GCB builds tried to build a specific image they
used the last build of that image pushed to gcr.io/fuzzbench as
a cache. This broke caching when doing a build in another project
of an image that was never pushed to gcr.io/fuzzbench.

Also fix tests for both features and add a test to ensure other
docker registries are handled properly.

### DIFF
--- a/experiment/build/generate_cloudbuild.py
+++ b/experiment/build/generate_cloudbuild.py
@@ -23,14 +23,62 @@ from common.utils import ROOT_DIR
 from experiment.build import build_utils
 
 DOCKER_IMAGE = 'docker:19.03.12'
+STANDARD_DOCKER_REGISTRY = 'gcr.io/fuzzbench'
 
 
-def get_experiment_tag_for_image(image_specs, tag_by_experiment=True):
-    """Returns the registry with the experiment tag for given image."""
-    tag = posixpath.join(get_docker_registry(), image_specs['tag'])
+def _get_image_tag(image_specs,
+                   tag_by_experiment=False,
+                   use_standard_registry=False):
+    """Returns an image tag for |image_specs|. Uses an experiment-specific tag
+    if |tag_by_experiment|. The registry is determined by the env var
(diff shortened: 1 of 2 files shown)


experiment/build/generate_cloudbuild.py
```diff
@@ -98,10 +150,11 @@ def create_cloudbuild_spec(image_templates,
         }
         step['args'] = [
             'build', '--tag',
-            posixpath.join(docker_registry, image_specs['tag']), '--tag',
-            get_experiment_tag_for_image(image_specs), '--cache-from',
-            get_experiment_tag_for_image(image_specs, tag_by_experiment=False),
-            '--build-arg', 'BUILDKIT_INLINE_CACHE=1'
+            _get_experiment_image_tag(image_specs), '--tag',
+            _get_gcb_image_tag(image_specs), '--tag',
```
inferno-chromium, line 154: Should gcb break if you provide the same tag twice ?

Like _get_experiment_image_tag(image_specs) can be == _get_gcb_image_tag(image_specs) (can be seen in some tests). it probably makes sense to check if and not add a duplicate if it is the same tag.

inferno-chromium (approved):

experiment/build/generate_cloudbuild.py
jonathanmetzman: I'm pretty sure "it" (it would probably be docker that breaks) does not break.

inferno-chromium: fine either way, nice to fix with a 2 line fix.

jonathanmetzman: I tested that it does not break. I don't see why we should add this logic. It just complicates things without providing any benefits.

PR #856, opened 2020-10-21
Hack around container-optimized-os issues

jonathanmetzman:
Container-optimized-os doesn't allow writing to HOME.
Allowing docker to authenticate to the artifact registry requires
writing to the HOME directory of root.
To support this, use the HOME directory of another user on the
instance: chronos.

Also, create /home/chronos if it doesn't already exist. This will
allow use of these scripts on instances that aren't using COS.

### DIFF
--- a/experiment/resources/dispatcher-startup-script-template.sh
+++ b/experiment/resources/dispatcher-startup-script-template.sh
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Hack because container-optmized-os doesn't support writing to /home/root
+export HOME=/home/chronos
+mkdir HOME || true
 docker-credential-gcr configure-docker -include-artifact-registry
 echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope
 docker run --rm \
--- a/experiment/resources/dispatcher-startup-script-template.sh
+++ b/experiment/resources/dispatcher-startup-script-template.sh
@@ -15,7 +15,7 @@
 
 # Hack because container-optmized-os doesn't support writing to /home/root
 export HOME=/home/chronos
-mkdir HOME || true
+mkdir -p $HOME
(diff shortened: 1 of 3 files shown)


jonathanmetzman:
PTAL.
This change is necessary to use the artifact registry.
Really surprising that Google cloud's docs don't explain how to do this, but I was able to find out about this from stackoverflow: https://stackoverflow.com/questions/38520729/docker-login-with-root-user-on-container-vm-image

experiment/resources/dispatcher-startup-script-template.sh
```diff
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Hack because container-optmized-os doesn't support writing to /home/root
```
inferno-chromium, line 16: Maybe add the thing about you are storing docker credentials to make it clear what is the intent.

experiment/resources/dispatcher-startup-script-template.sh
```diff
@@ -13,6 +13,9 @@
...
+export HOME=/home/chronos
+mkdir HOME || true
```
inferno-chromium, line 18: s/HOME/$HOME
instead of || true, can just do "mkdir -p $HOME"

inferno-chromium (changes requested):

experiment/resources/runner-startup-script-template.sh
```diff
@@ -23,9 +23,10 @@ echo core >/proc/sys/kernel/core_pattern
 
 # Start docker.
 {% if not local_experiment %}
-# Hack because container-optmized-os doesn't support writing to /home/root
+# Hack because container-optmized-os doesn't support writing to /home/root.
+# docker-credential-gcr needs to write to a dotfile in $HOME.
```
inferno-chromium, line 27: nit: add this same line to other file as well.

inferno-chromium (approved):

PR #865, opened 2020-10-23
Write the makefile instead of printing it

jonathanmetzman:
This will make testing nicer and makes the makefile generation more readable.


docker/generate_makefile.py
```diff
@@ -58,116 +62,120 @@ def _print_makefile_run_template(image):
...
-            params = " -c \"for f in /testcases/*; do\
-                    echo _________________________________________;\
-                    echo \\$$f:;\
-                    \\$$OUT/\\$$FUZZ_TARGET -timeout=25\
-                    -rss_limit_mb=2560 \\$$f; done;\""
-
+            section += f'\t-v {testcases_dir}:/testcases \\\n\t'
+            section += '--entrypoint /bin/bash '
+            section += os.path.join(BASE_TAG, image['tag'])
+            section += ' -c "for f in /testcases/*; do '
```
inferno-chromium, line 75: Did you verify this works, i wonder why it was added after the set of ifs before.

inferno-chromium (approved):
this is amazing, those 'end' hacks were really crazy

docker/generate_makefile.py
jonathanmetzman: Yep I verified this.

PR #868, opened 2020-10-26
Update Eclipser to v2.0

jchoi2022:
Eclipser previously had its own simplified random fuzzing module that mimics AFL. This was to support fuzzing multiple input sources (stdin, file, etc.) at once. We recently decided to focus on fuzzing a single input source, as most fuzzers do. This allows us to directly integrate Eclipser with AFL. From v2.0, Eclipser only performs grey-box concolic testing, and relies on AFL for random fuzzing. For the details and context of the update, please refer to https://github.com/SoftSec-KAIST/Eclipser#eclipser-v20.

This PR updates Eclipser to run v2.0 code in parallel with AFL.

### DIFF
--- a/fuzzers/eclipser/fuzzer.py
+++ b/fuzzers/eclipser/fuzzer.py
@@ -50,10 +50,9 @@ def build():
     new_env['CC'] = 'clang'
     new_env['CXX'] = 'clang++'
     new_env['FUZZER_LIB'] = '/libStandaloneFuzzTarget.a'
-    # Reset compilation flags only using NO_SANITIZER_COMPAT_* flags. Eclipser
-    # module prefers unoptimized binaries, as hacky binary code can impede its
-    # grey-box concolic technique. Note that we can still use optimized binaries
-    # for AFL module.
+    # Ensure to compile with NO_SANITIZER_COMPAT* flags even for bug benchmarks,
+    # as QEMU is incompatible with sanitizers. Also, Eclipser prefers clean and
+    # unoptimized binaries. We leave fast random fuzzing as AFL's job.
     new_env['CFLAGS'] = ' '.join(utils.NO_SANITIZER_COMPAT_CFLAGS)
     cxxflags = [utils.LIBCPLUSPLUS_FLAG] + utils.NO_SANITIZER_COMPAT_CFLAGS
     new_env['CXXFLAGS'] = ' '.join(cxxflags)
@@ -107,7 +106,7 @@ def afl_master(input_corpus, output_corpus, target_binary):
 
 def afl_worker(input_corpus, output_corpus, target_binary):
     """Run AFL worker instance."""
-    print('[afl_slave] Run AFL worker')
+    print('[afl_worker] Run AFL worker')
     afl_fuzzer.run_afl_fuzz(input_corpus, output_corpus, target_binary,
                             ['-S', 'afl-worker'], True)
 
--- a/fuzzers/eclipser/fuzzer.py
+++ b/fuzzers/eclipser/fuzzer.py
@@ -100,66 +100,16 @@ def eclipser(input_corpus, output_corpus, target_binary):
 
 def afl_master(input_corpus, output_corpus, target_binary):
     """Run AFL master instance."""
-    command = [
-        './afl-fuzz',
-        '-i',
-        input_corpus,
-        '-o',
-        output_corpus,
-        # Use no memory limit as ASAN doesn't play nicely with one.
-        '-m',
-        'none',
-        # Use same default 1 sec timeout, but add '+' to skip hangs.
-        '-t',
-        '1000+',
-        '-M',
-        'afl-master',
-    ]
-    dictionary_path = utils.get_dictionary_path(target_binary)
-    if dictionary_path:
-        command.extend(['-x', dictionary_path])
-    command += [
-        '--',
-        target_binary,
-        # Pass INT_MAX to afl the maximize the number of persistent loops it
-        # performs.
-        '2147483647'
-    ]
-    print('[afl_master] Run AFL master with command: ' + ' '.join(command))
-    output_stream = subprocess.DEVNULL
-    subprocess.check_call(command, stdout=output_stream, stderr=output_stream)
+    print('[afl_master] Run AFL master')
+    afl_fuzzer.run_afl_fuzz(input_corpus, output_corpus, target_binary,
+                            ['-M', 'afl-master'], True)
 
 
 def afl_worker(input_corpus, output_corpus, target_binary):
     """Run AFL worker instance."""
-    command = [
-        './afl-fuzz',
-        '-i',
-        input_corpus,
-        '-o',
-        output_corpus,
-        # Use no memory limit as ASAN doesn't play nicely with one.
-        '-m',
-        'none',
-        # Use same default 1 sec timeout, but add '+' to skip hangs.
-        '-t',
-        '1000+',
-        '-S',
-        'afl-worker',
-    ]
-    dictionary_path = utils.get_dictionary_path(target_binary)
-    if dictionary_path:
-        command.extend(['-x', dictionary_path])
-    command += [
-        '--',
-        target_binary,
-        # Pass INT_MAX to afl the maximize the number of persistent loops it
-        # performs.
-        '2147483647'
-    ]
-    print('[afl_worker] Run AFL worker with command: ' + ' '.join(command))
-    output_stream = subprocess.DEVNULL
-    subprocess.check_call(command, stdout=output_stream, stderr=output_stream)
+    print('[afl_slave] Run AFL worker')
+    afl_fuzzer.run_afl_fuzz(input_corpus, output_corpus, target_binary,
+                            ['-S', 'afl-worker'], True)
 
 
 def fuzz(input_corpus, output_corpus, target_binary):
--- a/fuzzers/eclipser/fuzzer.py
+++ b/fuzzers/eclipser/fuzzer.py
@@ -11,88 +11,181 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Integration code for Eclipser fuzzer."""
+"""Integration code for Eclipser fuzzer. Note that starting from v2.0, Eclipser
+relies on AFL to perform random-based fuzzing."""
 
-import os
+import shutil
 import subprocess
-import time
-from multiprocessing import Process
+import os
+import threading
 
 from fuzzers import utils
+from fuzzers.afl import fuzzer as afl_fuzzer
+
+
+def get_uninstrumented_outdir(target_directory):
+    """Return path to uninstrumented target directory."""
+    return os.path.join(target_directory, 'uninstrumented')
 
 
 def build():
     """Build benchmark."""
-    # QEMU does not work with sanitizers.
-    # See https://github.com/SoftSec-KAIST/Eclipser/issues/5
 
-    os.environ['CC'] = 'clang'
-    os.environ['CXX'] = 'clang++'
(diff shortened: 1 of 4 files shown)


jchoi2022:
The two checks seem to have failed due to an out-of-memory error of the docker container (exit code 137). In local tests, Eclipser seems to run OK on the failed targets. Is there any method to test whether it will run OK in the Google Cloud (i.e. FuzzBench service)?

vanhauser-thc:
@jchoi2022 the instances have 3.5GB memory each - if this info helps

fuzzers/eclipser/fuzzer.py
```diff
@@ -11,88 +11,181 @@
...
-            'dotnet',
-            '/Eclipser/build/Eclipser.dll',
-            'decode',
-            '-i',
-            os.path.join(encoded_temp_corpus, 'testcase'),
-            '-o',
-            output_corpus,
-        ])
+
+def afl_master(input_corpus, output_corpus, target_binary):
```
inferno-chromium, line 101: Can you please use afl_fuzzer.run_afl_fuzz to avoid code duplication with additional_flags option (that is also important as we sometimes update the helper for corner cases).

fuzzers/eclipser/fuzzer.py
```diff
@@ -11,88 +11,181 @@
...
+        # Pass INT_MAX to afl the maximize the number of persistent loops it
+        # performs.
+        '2147483647'
+    ]
+    print('[afl_master] Run AFL master with command: ' + ' '.join(command))
+    output_stream = subprocess.DEVNULL
+    subprocess.check_call(command, stdout=output_stream, stderr=output_stream)
+
+
+def afl_worker(input_corpus, output_corpus, target_binary):
```
inferno-chromium, line 133: Can you please use afl_fuzzer.run_afl_fuzz to avoid code duplication with additional_flags option (that is also important as we sometimes update the helper for corner cases).

fuzzers/eclipser/fuzzer.py
```diff
@@ -11,88 +11,181 @@
...
-        '1048576',
-        # Default is low (0.5 sec), recommended to use higher:
-        # https://github.com/google/fuzzbench/issues/70#issuecomment-596060572
+        '-v',  # Controls the verbosity.
+        '2',
         '--exectimeout',
         '5000',
     ]
-    if os.listdir(input_corpus):  # Important, otherwise Eclipser crashes.
+    if os.listdir(input_corpus):  # Specify inputs only if any seed exists.
```
inferno-chromium, line 95: Can you please add support for crash dir, similar to this
https://github.com/google/fuzzbench/blob/master/fuzzers/libfuzzer/fuzzer.py#L52
we are working on crash based benchmarking, so this will be great to add as you are modifying this.

inferno-chromium (changes requested):

inferno-chromium:
> The two checks seem to have failed due to an out-of-memory error of the docker container (exit code 137). In local tests, Eclipser seems to run OK on the failed targets. Is there any method to test whether it will run OK in the Google Cloud (i.e. FuzzBench service)?

Where do you see these failures ? Can you paste the log link ?

fuzzers/eclipser/fuzzer.py
jonathanmetzman: In case it's not clear what this comment means, we want any crashes eclipser finds to be located in `output_corpus`.

fuzzers/eclipser/fuzzer.py
jchoi2022: I initially tried to reuse afl_fuzzer.run_afl_fuzz(), but it turned out that '-d' option in run_afl_fuzz() is incompatible with '-M' or '-S' options. When such options are given together, AFL refuses to run (cf. https://github.com/google/AFL/blob/fab1ca5ed7e3552833a18fc2116d33a9241699bc/afl-fuzz.c#L7510). 

So at the moment, it seems impossible to reuse run_afl_fuzz() here. Removing '-d' option from run_afl_fuzz() is one possible option, but it will affect many other fuzzers relying on this function. So I thought it would be better to address it in a separate issue or PR.

fuzzers/eclipser/fuzzer.py
jchoi2022: (Same as above)

fuzzers/eclipser/fuzzer.py
jchoi2022: Is there any specific requirement (or protocol) for the directory path for crashes?

Currently, Eclipser uses a similar path convention with AFL. As a result, eclipser/fuzzer.py will store the found crashes in
`<output_corpus>/afl-master/crashes`, `<output_corpus>/afl-worker/crashes`, and `<output_corpus>/eclipser-output/crashes`. 

Is this behavior okay?

inferno-chromium: Ah perfect, in afl's case, nothing is needed to be changed, it should just work.

fuzzers/eclipser/fuzzer.py
inferno-chromium: Basically, add a condition in run_afl_fuzzer, like:

# '-d' flag is incompatible with '-M' and '-S' options.
if not additional_flags or ('-M' not in additional_flags and '-S' not in additional_flags):
  flags.append('-d') 

We shouldn't be duplicating code, esp afl one we keep improving from time to time.

fuzzers/eclipser/fuzzer.py
inferno-chromium: See comment above, run_afl_fuzz can be fixed.

jchoi2022:
The two checks (build (eclipser, oss-fuzz) and build (eclipser, standara)) have failed when I pushed the first commit (d0ff08a7), but the failure had disappeared when I pushed the second commit (9aa5c2bb). It does not seem the second commit actually resolved the problem. Rather, I think the OOM error of docker just occurs non-deterministically.

jchoi2022:
@vanhauser-thc Thank you for the information. Then I may have to consider reducing the number of AFL instances to save memory, if OOM issue persists in the Google Cloud.

fuzzers/eclipser/fuzzer.py
jchoi2022: I see, that seems to be a nicer solution. I will add a new commit to resolve this.

jchoi2022: @jonathanmetzman While refactoring fuzzers/afl/fuzzer.py, I found a confusing comment.  In commit 60b67bf7, the comment says `Use deterministic mode as it does best when we don't have seeds which is often the case`, but isn't `-d` option intended to "skip" deterministic mode?

fuzzers/eclipser/fuzzer.py
jonathanmetzman: Yes. It is fine.

inferno-chromium:
One question: do you need both afl master and slave thread, usually across fuzzers, we use just one afl slave thread.

jchoi2022:
Ideally, it would be good to have a master AFL instance dedicated for deterministic mode. However, I was also thinking that using only one AFL slave instance would be more consistent with other AFL* fuzzers. Also, removing master instance may help reducing memory usage, as discussed above.

I will add a commit to remove master instance shortly.

inferno-chromium (approved):

inferno-chromium:
I think Eclipser wont work in CI here due to short timeout and it is not a blocker. Merging for now.

```
[eclipser] Run Eclipser with command: dotnet /Eclipser/build/Eclipser.dll -p /out/uninstrumented/cms_transform_fuzzer -s /out/corpus -o /out/corpus/eclipser_output --arg foo -f foo -v 2 --exectimeout 5000 -i /out/seeds
[00:00:00:00] [*] Fuzz target : /out/uninstrumented/cms_transform_fuzzer
[00:00:00:00] [*] Total 1 initial seeds
[00:00:00:00] [*] No time limit given, run infinitely
[00:00:00:00] [*] Start fuzzing
[00:00:00:00] Fuzzing with:  ...800bytes... (0) (Right)
[00:00:00:05] Fuzzing with:  ...800bytes... (1) (Right)
/home/runner/work/_temp/4659f939-8d85-4bc9-9fe6-9469fe1608b4.sh: line 1:  2766 Killed                  PATH=.venv/bin/:$PATH PYTHONPATH=. python3 .github/workflows/build_and_test_run_fuzzer_benchmarks.py standard eclipser
docker/generated.mk:21682: recipe for target 'test-run-eclipser-lcms-2017-03-21' failed
make: *** [test-run-eclipser-lcms-2017-03-21] Killed
Error: Process completed with exit code 137.
```

jonathanmetzman:
> I think Eclipser wont work in CI here due to short timeout and it is not a blocker. Merging for now.
> 
> ```
> [eclipser] Run Eclipser with command: dotnet /Eclipser/build/Eclipser.dll -p /out/uninstrumented/cms_transform_fuzzer -s /out/corpus -o /out/corpus/eclipser_output --arg foo -f foo -v 2 --exectimeout 5000 -i /out/seeds
> [00:00:00:00] [*] Fuzz target : /out/uninstrumented/cms_transform_fuzzer
> [00:00:00:00] [*] Total 1 initial seeds
> [00:00:00:00] [*] No time limit given, run infinitely
> [00:00:00:00] [*] Start fuzzing
> [00:00:00:00] Fuzzing with:  ...800bytes... (0) (Right)
> [00:00:00:05] Fuzzing with:  ...800bytes... (1) (Right)
> /home/runner/work/_temp/4659f939-8d85-4bc9-9fe6-9469fe1608b4.sh: line 1:  2766 Killed                  PATH=.venv/bin/:$PATH PYTHONPATH=. python3 .github/workflows/build_and_test_run_fuzzer_benchmarks.py standard eclipser
> docker/generated.mk:21682: recipe for target 'test-run-eclipser-lcms-2017-03-21' failed
> make: *** [test-run-eclipser-lcms-2017-03-21] Killed
> Error: Process completed with exit code 137.
> ```

Oh so test-run-eclipser-$BENCHMARK fails everywhere?

inferno-chromium:
> > I think Eclipser wont work in CI here due to short timeout and it is not a blocker. Merging for now.
> > ```
> > [eclipser] Run Eclipser with command: dotnet /Eclipser/build/Eclipser.dll -p /out/uninstrumented/cms_transform_fuzzer -s /out/corpus -o /out/corpus/eclipser_output --arg foo -f foo -v 2 --exectimeout 5000 -i /out/seeds
> > [00:00:00:00] [*] Fuzz target : /out/uninstrumented/cms_transform_fuzzer
> > [00:00:00:00] [*] Total 1 initial seeds
> > [00:00:00:00] [*] No time limit given, run infinitely
> > [00:00:00:00] [*] Start fuzzing
> > [00:00:00:00] Fuzzing with:  ...800bytes... (0) (Right)
> > [00:00:00:05] Fuzzing with:  ...800bytes... (1) (Right)
> > /home/runner/work/_temp/4659f939-8d85-4bc9-9fe6-9469fe1608b4.sh: line 1:  2766 Killed                  PATH=.venv/bin/:$PATH PYTHONPATH=. python3 .github/workflows/build_and_test_run_fuzzer_benchmarks.py standard eclipser
> > docker/generated.mk:21682: recipe for target 'test-run-eclipser-lcms-2017-03-21' failed
> > make: *** [test-run-eclipser-lcms-2017-03-21] Killed
> > Error: Process completed with exit code 137.
> > ```
> 
> Oh so test-run-eclipser-$BENCHMARK fails everywhere?
> I think that's not great because it's a good first line of defense against fuzzer-benchmark pairs that are broken.
> It's harder to figure these out otherwise.

not everywhere, let me investigate more. trying in https://github.com/google/fuzzbench/pull/890.
prod is all running fine - https://www.fuzzbench.com/reports/2020-10-27/index.html
but you are right, CI should work too or we should look at fixing things. Just too much stuff to handle.

PR #869, opened 2020-10-26
Request new afl++ testcache experiment

vanhauser-thc:
New - and hopefully final testcache experiment.
20mb and 200mb variants are misleading, it is 50mb and 500mb but did not want to rename :)

note that I also removed @Microsvuln two afl++ variants as he likely isnt aware that variants should be cleaned when they are not needed anymore. if you still need them just raise you hand and I will restore them. :)


jonathanmetzman (approved):

vanhauser-thc:
clarified with @Microsvuln that they can be removed. he still needs them but he needs actually a different template, not the _optimal one :)

Microsvuln:
True, I should submit two new variants for my tests but from _explore and not _optimal.

vanhauser-thc:
@inferno-chromium @jonathanmetzman 
that one built failure is a fluke and not part of the experiment anyway :)

PR #870, opened 2020-10-26
Integrating opensc project with target fuzz_pkcs15_reader

Vaush:
It does not build with libfuzzer, does build with aflplusplus. Needs to be fixed before merging.


jonathanmetzman:
Don't upload these without updating the list of benchmarks please. It just results in a waste of cycles by using AFL to rebuild every other benchmark.

Vaush:
> 
> 
> Don't upload these without updating the list of benchmarks please. It just results in a waste of cycles by using AFL to rebuild every other benchmark.

Yes, this was the first one, I discovered AFL would rebuild everything right after I uploaded this one. I am not doing it with the others.

benchmarks/opensc_fuzz_pkcs15_reader/build.sh
```diff
@@ -0,0 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+./bootstrap
+# FIXME FUZZING_LIBS="$LIB_FUZZING_ENGINE" fails with some missing C++ library, I don't know how to fix this
```
inferno-chromium, line 19: Looks like you are still figuring this ?

benchmarks/opensc_fuzz_pkcs15_reader/build.sh
```diff
@@ -0,0 +1,28 @@
...
+./configure --disable-shared --disable-pcsc --enable-ctapi --enable-fuzzing FUZZING_LIBS="$LIB_FUZZING_ENGINE"
+make -j4
+
+cp src/tests/fuzzing/fuzz_asn1_print $OUT
```
inferno-chromium, line 23: Should only add the fuzzer that is tested, ignore/dont build those.

inferno-chromium (changes requested):
Needs CI addition.

benchmarks/opensc_fuzz_pkcs15_reader/build.sh
Vaush: Actually, this is not my comment, it's the script's original author's, but as I mentioned it has got some weird build behaviour that I need to check before finalizing the PR

inferno-chromium:
Not fixable.

PR #871, opened 2020-10-26
Integrate project harfbuzz with target hb-subset-fuzzer

Vaush:


benchmarks/harfbuzz_hb-subset-fuzzer/hb-draw-fuzzer.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
+timeout = 5
```
inferno-chromium, line 2: Remove all other option files other than subset fuzzer.

benchmarks/harfbuzz_hb-subset-fuzzer/build.sh
```diff
@@ -0,0 +1,56 @@
...
+# Build the library.
+meson --default-library=static --wrap-mode=nodownload \
+      -Dexperimental_api=true \
+      -Dfuzzer_ldflags="$(echo $LIB_FUZZING_ENGINE)" \
+      $build \
+   || (cat build/meson-logs/meson-log.txt && false)
+
+# Build the fuzzers.
+ninja -v -j$(nproc) -C $build test/fuzzing/hb-{shape,draw,subset,set}-fuzzer
+mv $build/test/fuzzing/hb-{shape,draw,subset,set}-fuzzer $OUT/
```
inferno-chromium, line 39: remove everything other than subset fuzzer in this build.sh files.

inferno-chromium (changes requested):

jonathanmetzman:
I think you need to blocklist klee.

Vaush:
> 
> 
> I think you need to blocklist klee.

I will blocklist it, but apparently the problem is that klee doesn't support emtpy seeds (at least, that's what the error says). Still, I don't think deleting part of the corpus would be a good idea.

inferno-chromium (approved):

PR #873, opened 2020-10-26
Integrate project ndpi with target fuzz_ndpi_reader

Vaush:


inferno-chromium:
this PR needs a rebase, also aflplusplus_optimal and klee are failing due missing libpcap something

```
#19 45.52 checking pkg-config is at least version 0.9.0... yes
#19 45.53 checking for pcap_open_live in -lpcap... no
#19 45.63 
#19 45.63 ERROR: Missing libpcap(-dev) library required to compile the example application
#19 45.63 ERROR: Please install it and try again
#19 45.67 + make
#19 45.67 make: *** No targets specified and no makefile found.  Stop.
#19 45.67 Traceback (most recent call last):
#19 45.67   File "<string>", line 1, in <module>
#19 45.67   File "/src/fuzzers/klee/fuzzer.py", line 198, in build
#19 45.67     utils.build_benchmark()
#19 45.67   File "/src/fuzzers/utils.py", line 68, in build_benchmark
#19 45.67     subprocess.check_call(['/bin/bash', '-ex', build_script], env=env)
#19 45.67   File "/usr/local/lib/python3.7/subprocess.py", line 363, in check_call
#19 45.67     raise CalledProcessError(retcode, cmd)
#19 45.67 subprocess.CalledProcessError: Command '['/bin/bash', '-ex', '/src/build.sh']' returned non-zero exit status 2.
#19 ERROR: executor failed running [/bin/sh -c PYTHONPATH=$SRC python3 -u -c "from fuzzers import utils; utils.initialize_env(); from fuzzers.$FUZZER import fuzzer; fuzzer.build()"]: runc did not terminate sucessfully
------
 > importing cache manifest from gcr.io/fuzzbench/builders/klee/ndpi_fuzz_ndpi_reader:
------
```

inferno-chromium (changes requested):
Needs changes as per comment above.

Vaush:
> 
> 
> Needs changes as per comment above.

Yes, I will address the fuzzers that are not working. I am not sure why it needs a rebase though, could you clarify?

inferno-chromium:
> > Needs changes as per comment above.
> 
> Yes, I will address the fuzzers that are not working. I am not sure why it needs a rebase though, could you clarify?

i thought i saw some minor conflict in .github/workflows/build_and_test_run_fuzzer_benchmarks.py, basically as your other prs get committed, might be minor annoyance to pull and update this file.

Vaush:
> 
> 
> > > Needs changes as per comment above.
> > 
> > 
> > Yes, I will address the fuzzers that are not working. I am not sure why it needs a rebase though, could you clarify?
> 
> i thought i saw some minor conflict in .github/workflows/build_and_test_run_fuzzer_benchmarks.py, basically as your other prs get committed, might be minor annoyance to pull and update this file.

Ah, I see, I thought about that while I was uploading the PRs. I'll see to it

Vaush:
Ok, I added libpcap-dev to the apt install list, checked locally and aflplusplus_optimal built and opened a debug session without issues.
Now, of the fuzzers are throwing an error, as it seems the benchmark used a specific version of libpcap downloaded by itself. I don't know why aflplusplus_optimal and klee can't see that version though. I'll revert the changes

inferno-chromium:
> Ok, I added libpcap-dev to the apt install list, checked locally and aflplusplus_optimal built and opened a debug session without issues.
> Now, of the fuzzers are throwing an error, as it seems the benchmark used a specific version of libpcap downloaded by itself. I don't know why aflplusplus_optimal and klee can't see that version though. I'll revert the changes

@vanhauser-thc - any thoughts on why this libpcap thing only impacts aflplusplus_optimal and not regular aflplusplus one. these benchmarks are planned for bug/crash-based benchmarking work.

vanhauser-thc:
aflplusplus is "tracepc, cmplog, dict2file", whereas aflplusplus_optimal for libpcap is "tracepc, dict2file". so this is not the issue. so it must rather be the different builder.Dockerfile.

I would love to get rid of the specific llvm download I have to do for _optimal to be able to use LTO instrumentation, but this requires llvm11+ to be available for all benchmarks. is this the case now?

but even that does not answer why configure does not find libpcap-dev. I would assume - it is simply not installed by the benchmark as it should (for whatever reason).

Vaush:
@vanhauser-thc I did a couple of checks, if run with the standard parameters "tracepc, cmplog, dict2file", the benchmark wors. I suspect that's because in the fuzzer.py, aflplusplus_optimal checks the name of the benchmark, and while libpcap is indeed "tracepc, dict2file", we are actually fuzzing ndpi, a new benchmark we are integrating, which is not on the list.
In aflplusplus_optimal, those that are not on the list are called with "lto, cmplog, fixed". Now, I am not sure what the optimal parameters for ndpi should be, nor why the default parameters for aflplusplus_optimal are different from those used by aflplusplus itself, but these parameters are the reason why the benchmark is not building.
I think the modification should be in the list of benchmarks in aflplusplus_optimal, adding ndpi with the right set of parameters.
@inferno-chromium @jonathanmetzman  see if this makes sense to you, and @vanhauser-thc do you have any idea about the parameters for ndpi? The aflplusplus default ones work, but I am guessing they are not optimal by any means

vanhauser-thc:
> @vanhauser-thc I did a couple of checks, if run with the standard parameters "tracepc, cmplog, dict2file", the benchmark works. I suspect that's because in the fuzzer.py, aflplusplus_optimal checks the name of the benchmark, and while libpcap is indeed "tracepc, dict2file", we are actually fuzzing ndpi

ah I did not know that.
easy fix: copy the libpcap  "else if "+ following line for from aflplusplus_optimal and set in the name of the new benchmark.

LTO is great, but there are benchmarks that dont work with it (too long compile times, too early constructors, ...)

inferno-chromium:
Great, thanks @vanhauser-thc 

@Vaush - can you add that condition, then aflplusplus_optimal should get green and ready to land.
klee probably needs a blocklist right, add that as well.

PR #879, opened 2020-10-27
Integrate project envoy with target filter_fuzz_test

Vaush:


Vaush:
@inferno-chromium @jonathanmetzman Do you know what happened here? It seems the build_fuzzers checks were not added for some reason. Have I missed something?

benchmarks/envoy_filter_fuzz_test/build.sh
```diff
@@ -0,0 +1,136 @@
...
+#
+declare -r EXTRA_BAZEL_FLAGS="$(
+for f in ${CFLAGS}; do
+  echo "--conlyopt=${f}" "--linkopt=${f}"
+done
+for f in ${CXXFLAGS}; do
+  echo "--cxxopt=${f}" "--linkopt=${f}"
+done
+
+if [ "$SANITIZER" = "undefined" ]
```
inferno-chromium, line 42: We don't set SANITIZER, i think you should check using undefined in CFLAGS or set this always.

benchmarks/envoy_filter_fuzz_test/build.sh
```diff
@@ -0,0 +1,136 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+export CFLAGS="$CFLAGS"
+export CXXFLAGS="$CXXFLAGS"
+
+declare -r FUZZER_TARGETS_CC=$(find . -name *_fuzz_test.cc)
```
inferno-chromium, line 21: Envoy is a crazy slow build, so lets optimize here by just specifying the filter_fuzz_test fuzzer file explicitly.

inferno-chromium (changes requested):
Once you repush CL with these fixes, hope the benchmark starts building. If not, we can investigate what is going wrong with CI. I dont see anything wrong from this CL itself.

inferno-chromium:
Please fix syntax error in build.sh

```
#18 0.404 /src/build.sh: command substitution: line 43: syntax error near unexpected token `('
#18 0.404 Traceback (most recent call last):
#18 0.404   File "<string>", line 1, in <module>
#18 0.404   File "/src/fuzzers/afl/fuzzer.py", line 40, in build
#18 0.405     utils.build_benchmark()
#18 0.405   File "/src/fuzzers/utils.py", line 68, in build_benchmark
#18 0.405     subprocess.check_call(['/bin/bash', '-ex', build_script], env=env)
#18 0.405   File "/usr/local/lib/python3.7/subprocess.py", line 363, in check_call
#18 0.405     raise CalledProcessError(retcode, cmd)
#18 0.405 subprocess.CalledProcessError: Command '['/bin/bash', '-ex', '/src/build.sh']' returned non-zero exit status 2.
#18 ERROR: executor failed running [/bin/sh -c PYTHONPATH=$SRC python3 -u -c "from fuzzers import utils; utils.initialize_env(); from fuzzers.$FUZZER import fuzzer; fuzzer.build()"]: runc did not terminate sucessfully
```

Vaush:
I will check why so many fuzzers are failing, envoy really is a though benchmark.
As a reminder, I am going through all the benchmarks to update their base-image SHA, but envoy has a bug (mentioned in the Dockerfile) that will require me not to specify the base-image to be used.

benchmarks/envoy_filter_fuzz_test/build.sh
jonathanmetzman: We probably should then right?

benchmarks/envoy_filter_fuzz_test/build.sh
jonathanmetzman: +1

jonathanmetzman:
> I will check why so many fuzzers are failing, envoy really is a though benchmark.

Worth noting that some envoy fuzz targets use libprotobuf-mutator. In OSS-Fuzz, I think these fuzzers build with AFL but running them with any fuzzer other than libFuzzer is useless at this point. So we should double check that the target doesn't use libprotobuf-mutator before spending effort on it.

> As a reminder, I am going through all the benchmarks to update their base-image SHA, but envoy has a bug (mentioned in the Dockerfile) that will require me not to specify the base-image to be used.

I'm a bit confused by this sentence. How are you planning on updating most of the dockerfiles? And yes, envoy was broken for a while in OSS-Fuzz and needed a specific version of base-builder because it needed a specific version of clang.

benchmarks/envoy_filter_fuzz_test/build.sh
Vaush: I did modify it in order to recognize if undefined is used in -fsanitize, but I guess setting it would be a good idea. Still, what if 2 sanitizers are used, such as ASAN and UBSAN? $SANITIZER would be "address, undefined"

Vaush:
> 
> 
> > I will check why so many fuzzers are failing, envoy really is a though benchmark.
> 
> Worth noting that some envoy fuzz targets use libprotobuf-mutator. In OSS-Fuzz, I think these fuzzers build with AFL but running them with any fuzzer other than libFuzzer is useless at this point. So we should double check that the target doesn't use libprotobuf-mutator before spending effort on it.
> 
> > As a reminder, I am going through all the benchmarks to update their base-image SHA, but envoy has a bug (mentioned in the Dockerfile) that will require me not to specify the base-image to be used.
> 
> I'm a bit confused by this sentence. How are you planning on updating most of the dockerfiles? And yes, envoy was broken for a while in OSS-Fuzz and needed a specific version of base-builder because it needed a specific version of clang.

I already have updated them locally, I will send another PR updating the Dockerfiles that we have merged already, and for the ones we haven't, I will just update their respective PRs.
I don't know exactly what kind of bug envoy had, I just know that in its Dockerfile there's a warning not to specify a SHA for the base-image, and in fact when I did so locally by mistake, it didn't build anymore.
I will check this libprotobuf-mutator thing you mentioned but do you have maybe a link to a specific issue or something like that?

benchmarks/envoy_filter_fuzz_test/build.sh
jonathanmetzman: If we set it at all then maybe it would be `address, undefined`. We don't combine sanitizers in OSS-Fuzz.

Another issue we need to figure out is how we will use UBSAN.
In OSS-Fuzz, we don't actually do -fsanitize=undefined. We specify specific ubsan checks so the flags look more like this:
`-fsanitize=unsigned-overflow,signed-overflow,bool`

jonathanmetzman:
> > > I will check why so many fuzzers are failing, envoy really is a though benchmark.
> > 
> > 
> > Worth noting that some envoy fuzz targets use libprotobuf-mutator. In OSS-Fuzz, I think these fuzzers build with AFL but running them with any fuzzer other than libFuzzer is useless at this point. So we should double check that the target doesn't use libprotobuf-mutator before spending effort on it.
> > > As a reminder, I am going through all the benchmarks to update their base-image SHA, but envoy has a bug (mentioned in the Dockerfile) that will require me not to specify the base-image to be used.
> > 
> > 
> > I'm a bit confused by this sentence. How are you planning on updating most of the dockerfiles? And yes, envoy was broken for a while in OSS-Fuzz and needed a specific version of base-builder because it needed a specific version of clang.
> 
> I already have updated them locally, I will send another PR updating the Dockerfiles that we have merged already, and for the ones we haven't, I will just update their respective PRs.
> I don't know exactly what kind of bug envoy had, I just know that in its Dockerfile there's a warning not to specify a SHA for the base-image, and in fact when I did so locally by mistake, it didn't build anymore.
> I will check this libprotobuf-mutator thing you mentioned but do you have maybe a link to a specific issue or something like that?

There's no issue. But if you can find a function `LLVMFuzzerTestOneInput` for the fuzz target then that means it is probably fine. libprotobuf-mutator based fuzzers will use something like DEFINE_PROTO_FUZZER instead.

benchmarks/envoy_filter_fuzz_test/build.sh
Vaush: I see...so we use not sanitizers at all in fuzzbench, or are there some default ones? If we use UBSAN with the flags instead of undefined, that might also be a reason why envoy is not building, since I am looking for the "undefined" sanitizer. If we don't do that in FuzzBench, though, I guess it means our only clue right now is the libprotobuf-mutator thing

jonathanmetzman: >I see...so we use not sanitizers at all in fuzzbench, or are there some default ones

I don't exactly understand this.
We don't use all of the default UBSAN checks in OSS-Fuzz.

>if we use UBSAN with the flags instead of undefined

In Fuzzbench, I don't think we are using *any* ubsan flags yet.

>If we don't do that in FuzzBench, though, I guess it means our only clue right now is the libprotobuf-mutator thing

Actually I don't really consider this a clue. AFL builds fine with libprotobuf-mutator targets. Running them together also doesn't result in an error. But it's useless to run AFL with them. It's a runtime issue and shouldn't affect builds.

I think the best clue we have is simply that envoy is a very finnicky project. When I think of projects with build problems in OSS-Fuzz envoy is probably the first that comes to mind. It's constantly having them.

Vaush:
I can confirm that the target does indeed use DEFINE_PROTO_FUZZER, and the errors are all over the place, there doesn't seem to be a clear, single reason for them. Should we just discard this one?

inferno-chromium:
> I can confirm that the target does indeed use DEFINE_PROTO_FUZZER, and the errors are all over the place, there doesn't seem to be a clear, single reason for them. Should we just discard this one?

yes ignore this one then.

PR #882, opened 2020-10-27
Integrate project php with target php-fuzz-execute

Vaush:


Vaush:
@inferno-chromium @jonathanmetzman I just noticed there's a warning above the CI list not to build php benchmarks as it fills up disk, should I delete the line? But how would we test it then? Maybe I should delete it just before merging?

benchmarks/php_php-fuzz-execute/build.sh
```diff
@@ -0,0 +1,68 @@
...
+# Generate corpuses and dictionaries.
+sapi/cli/php sapi/fuzzer/generate_all.php
+
+# Copy dictionaries to expected locations.
+cp sapi/fuzzer/dict/unserialize $OUT/php-fuzz-unserialize.dict
+cp sapi/fuzzer/dict/parser $OUT/php-fuzz-parser.dict
+cp sapi/fuzzer/json.dict $OUT/php-fuzz-json.dict
+
+FUZZERS="php-fuzz-json
+php-fuzz-exif
```
jonathanmetzman, line 55: Only copy the fuzzer we are using please.

jonathanmetzman:
> @inferno-chromium @jonathanmetzman I just noticed there's a warning above the CI list not to build php benchmarks as it fills up disk, should I delete the line? But how would we test it then? Maybe I should delete it just before merging?

Interesting, if this works, then I suppose we will have to delete this line.
We may need to get rid of this benchmark after we try it out though. We had to get rid of the last PHP benchmark because of issues measuring. I think there's a good chance that will happen again here.

inferno-chromium:
> > > @inferno-chromium @jonathanmetzman I just noticed there's a warning above the CI list not to build php benchmarks as it fills up disk, should I delete the line? But how would we test it then? Maybe I should delete it just before merging?
> > 
> > 
> > Interesting, if this works, then I suppose we will have to delete this line.
> > We may need to get rid of this benchmark after we try it out though. We had to get rid of the last PHP benchmark because of issues measuring. I think there's a good chance that will happen again here.
> 
> I see, but doesn't fuzzbench still have a php benchmark? I have another one to integrate, and it ended up having the same name as the existing one

php benchmark is commented out in service atm. it ooms out on clang code coverage, so skipped for now until more analysis can be done.

Vaush:
@vanhauser-thc do you happen to know if the php_fuzz_execute target needs to have different parameters than php_fuzz_parser? I copied them, but as you can see it run out of time (took more than 6 hours). I'm asking because I would like to avoid occupying resources for another 6 hours just to try different parameters combinations, but I will go for the default ones if you don't know the good ones already.
@jonathanmetzman @inferno-chromium I have no clue as to why weizz_qemu is failing, apparently before even starting its test-run. I will have a look around, but if it happened before already maybe you know something about how to fix it?

vanhauser-thc:
@Vaush the same one should work - it is still php I guess? but I do not understand why it runs out of time. the "cmplog" parameter doubles the build time (because a second target binary is built) but that is it, and looking at the other fuzzer the build time should be 17 minutes.
A long build time rather indicates that LTO is used, which is the default, and that would be selected if 'php_php-fuzz-execute' is not the exact benchmark name.
Alternatively you can try `'tracepc', 'cmplog', 'dict2file'` as build parameters.

jonathanmetzman:
> > > > @inferno-chromium @jonathanmetzman I just noticed there's a warning above the CI list not to build php benchmarks as it fills up disk, should I delete the line? But how would we test it then? Maybe I should delete it just before merging?
> > > 
> > > 
> > > Interesting, if this works, then I suppose we will have to delete this line.
> > > We may need to get rid of this benchmark after we try it out though. We had to get rid of the last PHP benchmark because of issues measuring. I think there's a good chance that will happen again here.
> > 
> > 
> > I see, but doesn't fuzzbench still have a php benchmark? I have another one to integrate, and it ended up having the same name as the existing one
> 
> php benchmark is commented out in service atm. it ooms out on clang code coverage, so skipped for now until more analysis can be done.

It runs out of disk space right? It's confusing to say it OOMs here when the issue is not that it runs out of RAM.

inferno-chromium:
> > > > > @inferno-chromium @jonathanmetzman I just noticed there's a warning above the CI list not to build php benchmarks as it fills up disk, should I delete the line? But how would we test it then? Maybe I should delete it just before merging?
> > > > 
> > > > 
> > > > Interesting, if this works, then I suppose we will have to delete this line.
> > > > We may need to get rid of this benchmark after we try it out though. We had to get rid of the last PHP benchmark because of issues measuring. I think there's a good chance that will happen again here.
> > > 
> > > 
> > > I see, but doesn't fuzzbench still have a php benchmark? I have another one to integrate, and it ended up having the same name as the existing one
> > 
> > 
> > php benchmark is commented out in service atm. it ooms out on clang code coverage, so skipped for now until more analysis can be done.
> 
> It runs out of disk space right? It's confusing to say it OOMs here when the issue is not that it runs out of RAM.

We are probably talking about different things. I was saying about why that php benchmark was disabled in prod service (not CI), there is some sort of OOM or hang. If this was just a disk size issue, could have knocked it out way earlier.

jonathanmetzman:
> > > > > > @inferno-chromium @jonathanmetzman I just noticed there's a warning above the CI list not to build php benchmarks as it fills up disk, should I delete the line? But how would we test it then? Maybe I should delete it just before merging?
> > > > > 
> > > > > 
> > > > > Interesting, if this works, then I suppose we will have to delete this line.
> > > > > We may need to get rid of this benchmark after we try it out though. We had to get rid of the last PHP benchmark because of issues measuring. I think there's a good chance that will happen again here.
> > > > 
> > > > 
> > > > I see, but doesn't fuzzbench still have a php benchmark? I have another one to integrate, and it ended up having the same name as the existing one
> > > 
> > > 
> > > php benchmark is commented out in service atm. it ooms out on clang code coverage, so skipped for now until more analysis can be done.
> > 
> > 
> > It runs out of disk space right? It's confusing to say it OOMs here when the issue is not that it runs out of RAM.
> 
> We are probably talking about different things. I was saying about why that php benchmark was disabled in prod service (not CI), there is some sort of OOM or hang. If this was just a disk size issue, could have knocked it out way earlier.

Sorry my bad, I was talking about in CI.

Vaush:
Again, did a test-run locally and aflplusplus_optimal didn't timeout. I need to upload the sha256 and template fix, so we'll see if the error repeats itself

Vaush:
@inferno-chromium @jonathanmetzman there is something a bit weird going on here. I know that afl rebuilds all the benchmarks, so it doesn't surprise me to see 6 different targets in the building logs, but if you check the logs for aflplusplus_optimal, the check that is running out of time, you'll see that the task gets canceled while trying to build matio for some reason?
Might it be because I had to change aflplusplus_optimal's fuzzer.py in order to set the correct parameters for php_php-fuzz-execute?
Locally the test-run completes successfully, even if it takes a substantial amount of time (still less than 360 minutes though).
My only way of dealing with this would be to ignore the cancelled check and just proceed with the merge, but I'd be glad if you had a better solution

inferno-chromium:
> @inferno-chromium @jonathanmetzman there is something a bit weird going on here. I know that afl rebuilds all the benchmarks, so it doesn't surprise me to see 6 different targets in the building logs, but if you check the logs for aflplusplus_optimal, the check that is running out of time, you'll see that the task gets canceled while trying to build matio for some reason?
> Might it be because I had to change aflplusplus_optimal's fuzzer.py in order to set the correct parameters for php_php-fuzz-execute?
> Locally the test-run completes successfully, even if it takes a substantial amount of time (still less than 360 minutes though).
> My only way of dealing with this would be to ignore the cancelled check and just proceed with the merge, but I'd be glad if you had a better solution

Yes ignore that one.

inferno-chromium (changes requested):
Can you update the fix to remove unneeded fuzzers, dict from build.sh and then we can merge this.

Vaush:
@inferno-chromium the reason I didn't upload the commit is that locally the benchmark started throwing an error when executed with afl, and I don't know why. I am not sure if it was a local thing or if it has the same problem now that it has been merged

PR #886, opened 2020-10-28
Remove variants, update variants

vanhauser-thc:
This removes the unneeded temporary variants and updates the core variants for the upcoming experiment.
no commit id was changed, so basically this could be accepted when a single benchmark successfully built.


vanhauser-thc:
aflpp qemu fails with a new benchmark, investigating ...

inferno-chromium:
> This is not an afl++ issue. seems to be something in matio or qemu. afl_qemu fails too.
> 
> I installed qemu-user in the container and using qemu-x86_x64 on the target fails as well:
> 
> ```
> root@352895f85bd0:/out# apt install qemu-user
> Reading package lists... Done
> Building dependency tree       
> Reading state information... Done
> ....
> root@352895f85bd0:/out# qemu-x86_64 ./matio_fuzzer 
> warning: TCG doesn't support requested feature: CPUID.01H:ECX.vmx [bit 5]
> Killed
> ```
> 
> increasing the qemu stack size to 60mb doesnt help either ...
> I think this benchmark should be disabled for all qemu fuzzers

sure filed https://github.com/google/fuzzbench/issues/894

PR #892, opened 2020-10-28
[benchmark integration] Check third_party/oss-fuzz is a git repo.

jonathanmetzman:
If third_party/oss-fuzz isn't a repo, then exit and explain to
users that they need to checkout submodules.
Previously, if it was not a repo, `git reset --hard` would be done
on the fuzzbench repo, which could unexpectedly cause users to lose
work.


inferno-chromium (approved):
Why need logs.error, can just add message in RuntimeError ?

jonathanmetzman:
> Why need logs.error, can just add message in RuntimeError ?

I don't do it ther ebecause it feels weird putting instructions inside of an exception.
Frankly I don't even want the runtimeerror, I just use it to bail out of the program

Issue #895, opened 2020-10-28
Switch to queue based measurer


jonathanmetzman:
This issue tracks switching to the queue based measurer.
This will make measuring horizontally scalable, making experiments shorter, cheaper and results more realtime. It is also necessary for the new architecture we are switching to.
This is essentially the second time I am trying to do this. The first time I did it on a second branch called queue. queue was pretty much ready to merge in #570.
But #570 is basically impossible to land now because we completely changed measuring thanks to clang coverage.
Instead of trying this same failed approach again, I am going to merge in pieces of #570 incrementally, before switching over to the queue system.

PR #896, opened 2020-10-28
[queue]Use one source of truth for cycle filenames.

jonathanmetzman:
Use one source of truth for naming of cycle files (e.g. corpus-archive-0001).

Part of queue based measurer (#895).


jonathanmetzman:
TBR

inferno-chromium (commented):
LGTM, extension could be argument to function too, but not a big deal.

jonathanmetzman:
I think it was, but a reviewer (maybe @mbarbella-chromium?) asked that I remove this?
Not sure, this review probably happened months ago.

PR #901, opened 2020-10-28
Blacklist not working fuzzers in project matio

Vaush:
This is a temporary fix for the issue of fuzzers not being able to build with matio


Vaush:
@inferno-chromium I have reintroduced the old blacklist, and reverted the zlib change for now (but I will put it back if needed). The only doubt is that some new fuzzers might have been introduced that don't build against matio, so we might have to blacklist those too

inferno-chromium:
this is not needed i think now.

PR #903, opened 2020-10-28
Fix matio not building

Vaush:
@inferno-chromium In my local tests this version compiled with afl, aflcc and aflplusplus, so it should build everywhere. If it is so, we don't need to blocklist any fuzzer. We just installed the wrong library for zlib.


benchmarks/matio_matio_fuzzer/benchmark.yaml
```diff
@@ -2,7 +2,3 @@ commit: 7a9958485036ef37b4e88df5cbe32bd7ec4c0d25
 commit_date: 2019-09-04 02:05:00+00:00
 fuzz_target: matio_fuzzer
 project: matio
-type: bug
```
inferno-chromium, line 5: can you bring back "type: bug", that adds the sanitizer flags "-fsanitize=address,undefined".

inferno-chromium (changes requested):

inferno-chromium:
@Vaush - aflcc is failing and *_qemu seems to fail as well, i think you might be missing that "type: bug" in your local testing.

Vaush:
Yep, I was just testing that locally, aflcc and qemus build without type: bug. Also, I found online that afl_qemu is not compatible with asan

Vaush:
I am not sure how we should proceed, apparently we need to disable asan, or maybe both asan and ubsan, but only for some fuzzers

Vaush:
> 
> 
> I am not sure how we should proceed, apparently we need to disable asan, or maybe both asan and ubsan, but only for some fuzzers

Which would be a needed thing, by the way, since asan and ubsan are needed to detect some of our crashes, but at the same time not all fuzzers can support them (and as such won't be able to detect some bugs), and that should be reflected in our analysis results

inferno-chromium:
> > I am not sure how we should proceed, apparently we need to disable asan, or maybe both asan and ubsan, but only for some fuzzers
> 
> Which would be a needed thing, by the way, since asan and ubsan are needed to detect some of our crashes, but at the same time not all fuzzers can support them (and as such won't be able to detect some bugs), and that should be reflected in our analysis results

yes just blocklist them since this benchmark is for crash based benchmarking and for now,if asan does not work with qemu* fuzzers and aflcc, we cant do much here. can check with @jonathanmetzman later, but for now this seems ok.

Vaush:
> 
> 
> > > I am not sure how we should proceed, apparently we need to disable asan, or maybe both asan and ubsan, but only for some fuzzers
> > 
> > 
> > Which would be a needed thing, by the way, since asan and ubsan are needed to detect some of our crashes, but at the same time not all fuzzers can support them (and as such won't be able to detect some bugs), and that should be reflected in our analysis results
> 
> yes just blocklist them since this benchmark is for crash based benchmarking and for now,if asan does not work with qemu* fuzzers and aflcc, we cant do much here. can check with @jonathanmetzman later, but for now this seems ok.

We will probably have the same problem with all the others we already merged though, once I had type:bug to them

Vaush:
I'm working on a kind of definitive fix for this. Meanwhile, here's some info on afl in qemu mode and sanitizers: https://afl-1.readthedocs.io/en/latest/notes_for_asan.html
Apparently they can work, in some cases, with some options set. I don't know if it's worth the effort instead of just excluding sanitizers

inferno-chromium:
> I'm working on a kind of definitive fix for this. Meanwhile, here's some info on afl in qemu mode and sanitizers: https://afl-1.readthedocs.io/en/latest/notes_for_asan.html
> Apparently they can work, in some cases, with some options set. I don't know if it's worth the effort instead of just excluding sanitizers

We can't exclude asan since we need it for reliable crashes and stacktraces [for bug based benchmarking]. what are those options ? i see https://afl-1.readthedocs.io/en/latest/notes_for_asan.html#interactions-with-the-qemu-mode which makes it say it never works. @vanhauser-thc as well for thoughts ?

Vaush:
> 
> 
> > I'm working on a kind of definitive fix for this. Meanwhile, here's some info on afl in qemu mode and sanitizers: https://afl-1.readthedocs.io/en/latest/notes_for_asan.html
> > Apparently they can work, in some cases, with some options set. I don't know if it's worth the effort instead of just excluding sanitizers
> 
> We can't exclude asan since we need it for reliable crashes and stacktraces [for bug based benchmarking]. what are those options ? i see https://afl-1.readthedocs.io/en/latest/notes_for_asan.html#interactions-with-the-qemu-mode which makes it say it never works. @vanhauser-thc as well for thoughts ?

I meant excluding them were they are not supported. If afl_qemu can't work with sanitizers, for instance, there's really no other options. I am almost done writing the functionality for fuzzers to "ban" certain sanitizers for themselves, I will create a PR soon and you can tell me what you think about it.
Also, you are right about the content of the link, I read the section after the QEMU one as a continuation, but it's actually talking about normal AFL.

inferno-chromium:
Should this be closed, do you still want this to update dockerfile package name ?

Vaush:
> 
> 
> Should this be closed, do you still want this to update dockerfile package name ?

I think it's still needed, since otherwise afl didn't work by what I remember. Also, there are 2 new testcases. I will update the PR to solve the conflicts

Vaush:
@inferno-chromium I think this is ready for merging

inferno-chromium (approved):

PR #907, opened 2020-10-28
[queue][GCE] Add instance group code to gce module.

jonathanmetzman:
Also add tests.
See #895.

### DIFF
--- a/common/gce.py
+++ b/common/gce.py
@@ -54,3 +54,55 @@ def get_preempted_instances(project, zone):
         if (instance['scheduling']['preemptible'] and
                 instance['status'] == 'TERMINATED'):
             yield instance['name']
+
+
+def get_instance_group_managers():
+    """Returns the instance group managers resource."""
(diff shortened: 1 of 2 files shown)


common/gce.py
```diff
@@ -54,3 +54,55 @@ def get_preempted_instances(project, zone):
...
+                              project=project)
+    return request.execute()
+
+
+def create_instance_group(name: str, instance_template_url: str,
+                          base_instance_name: str, project: str, zone: str):
+    """Creates an instance group named |name| from the template specified by
+    |instance_template_url|."""
+    managers = get_instance_group_managers()
+    target_size = 1
```
inferno-chromium, line 99: probably should add size as an argument to this function and can default=1
size=1

inferno-chromium (approved):

common/gce.py
jonathanmetzman: Don't think this is necessary.

PR #909, opened 2020-10-28
[queue][gcloud] Add instance template code to gcloud

jonathanmetzman:
See #895

### DIFF
--- a/common/gcloud.py
+++ b/common/gcloud.py
@@ -14,6 +14,7 @@
 """Google cloud related code."""
 
 import enum
+import posixpath
 import subprocess
 from typing import List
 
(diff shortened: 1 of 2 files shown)


common/gcloud.py
```diff
@@ -112,3 +113,33 @@ def run_local_instance(startup_script: str = None) -> bool:
...
+        'gcloud', 'compute', '--project', project, 'instance-templates',
+        'create-with-container', template_name, '--no-address',
+        '--image-family=cos-stable', '--image-project=cos-cloud',
+        '--region=%s' % zone, '--scopes=cloud-platform',
+        '--machine-type=n1-standard-1', '--boot-disk-size=50GB',
+        '--preemptible', '--container-image', docker_image
+    ]
+    for item in env.items():
+        command.extend(['--container-env', '%s=%s' % item])
+    new_process.execute(command)
```
inferno-chromium, line 134: Should we check if it succeeded, otherwise exception out ? can also fix test to account for this.

inferno-chromium (approved):

common/gcloud.py
jonathanmetzman: If it fails, it exceptions. I think that's fine for now.

PR #910, opened 2020-10-28
[queue][gcloud] Add module for scheduling measure workers

jonathanmetzman:
See #895

### DIFF
--- a/experiment/schedule_measure_workers.py
+++ b/experiment/schedule_measure_workers.py
@@ -1,0 +1,155 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Module for starting instances to run measure workers."""
+import collections
+import os
+import posixpath
+import sys
+import time
+
+from common import experiment_utils
+from common import gce
+from common import gcloud
+from common import logs
+from common import queue_utils
+from common import yaml_utils
+
+logger = logs.Logger('schedule_measure_workers')  # pylint: disable=invalid-name
+
+# This is the default quota on GCE.
+# TODO(metzman): Use the GCE API to determine this quota.
+MAX_INSTANCES_PER_GROUP = 1000
+
+
+def get_instance_group_name(experiment: str):
+    """Returns the name of the instance group of measure workers for
+    |experiment|."""
+    # "worker-" needs to come first because name cannot start with number.
+    return 'worker-' + experiment
+
+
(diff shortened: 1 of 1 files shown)


experiment/schedule_measure_workers.py
```diff
@@ -0,0 +1,155 @@
...
+        experiment_config['experiment'])
+    project = experiment_config['cloud_project']
+    zone = experiment_config['cloud_compute_zone']
+    num_instances = gce.get_instance_group_size(instance_group_name, project,
+                                                zone)
+
+    # TODO(metzman): Use autoscaling as it probably can deal with quotas more
+    # easily.
+    if not num_instances_needed:
+        # Can't go below 1 instance per group.
```
inferno-chromium, line 125: i remember 0 was allowed for instance group, are you sure about this ?

experiment/schedule_measure_workers.py
```diff
@@ -0,0 +1,155 @@
...
+
+    zone = experiment_config['cloud_compute_zone']
+    instance_template_url = gcloud.create_instance_template(
+        instance_template_name, docker_image, env, project, zone)
+
+    instance_group_name = get_instance_group_name(experiment)
+
+    # GCE will create instances for this group in the format
+    # "m-$experiment-$UNIQUE_ID". Use 'm' is short for "measurer".
+    base_instance_name = 'm-' + experiment
```
inferno-chromium, line 79: nit: can create a helper function for this, similar to others.

experiment/schedule_measure_workers.py
```diff
@@ -0,0 +1,155 @@
...
+from common import gce
+from common import gcloud
+from common import logs
+from common import queue_utils
+from common import yaml_utils
+
+logger = logs.Logger('schedule_measure_workers')  # pylint: disable=invalid-name
+
+# This is the default quota on GCE.
+# TODO(metzman): Use the GCE API to determine this quota.
```
inferno-chromium, line 31: Is this todo really needed ? gce instance manager limit is 4000 and can be hardcoded.

experiment/schedule_measure_workers.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # TODO(metzman): Use autoscaling as it probably can deal with quotas more
+    # easily.
+    if not num_instances_needed:
+        # Can't go below 1 instance per group.
+        logs.info('num_instances_needed = 0, resizing to 1.')
+        num_instances_needed = 1
+
+    if num_instances_needed != num_instances:
+        # TODO(metzman): Add some limits so always have some measurers but not
```
inferno-chromium, line 130: Can just do this, by adding 
num_instances_needed = max(num_instances_needed, MIN_IDLE_INSTANCES) ?

inferno-chromium (approved):

experiment/schedule_measure_workers.py
jonathanmetzman: Pretty sure, but last time I checked was a few months ago.
I don't think this is worth investigating now.

experiment/schedule_measure_workers.py
jonathanmetzman: I think our quota is 1000, and we should have this TODO because hardcoding will break open source users right?

experiment/schedule_measure_workers.py
jonathanmetzman: I like having the logging that the if condition lets me do.

PR #911, opened 2020-10-28
[presubmit] Don't complain that coverage is not tested in CI.

jonathanmetzman:


jonathanmetzman:
Maybe we should be testing coverage builds in CI? This change makes sense if we don't want to.

inferno-chromium (approved):

inferno-chromium:
> Maybe we should be testing coverage builds in CI? This change makes sense if we don't want to.

i think yes, can you add it.

inferno-chromium (changes requested):
requesting changes to actually test coverage config.

inferno-chromium:
We are now testing coverage builds in CI, closing. https://github.com/google/fuzzbench/blob/master/.github/workflows/fuzzers.yml#L20

jonathanmetzman:
> We are now testing coverage builds in CI, closing. https://github.com/google/fuzzbench/blob/master/.github/workflows/fuzzers.yml#L20

We actually are not testing coverage in CI. That change introduced a bug. See complaint about coverage build failure here: https://github.com/google/fuzzbench/pull/921#issuecomment-744596051

Thats because CI uses `test-run-` make targets. But these don't exist for coverage builds because they don't have a `fuzz` function.

PR #912, opened 2020-10-28
[CI] Seperate buggy benchmarks into their own build jobs

### LINKED ISSUE #891: CI times out as number of benchmarks scale, happening with Eclipser.
Filed from https://github.com/google/fuzzbench/pull/868#issuecomment-717917112

jonathanmetzman:
Do this to avoid timing out.
Fixes #891

### DIFF
--- a/.github/workflows/build_and_test_run_fuzzer_benchmarks.py
+++ b/.github/workflows/build_and_test_run_fuzzer_benchmarks.py
@@ -25,7 +25,6 @@ ALWAYS_BUILD_FUZZER = 'afl'
 NUM_RETRIES = 2
 RETRY_DELAY = 60
 
-# TODO(tanq16): Get list of Benchmarks automatically.
 
 # Don't build php benchmark since it fills up disk in GH actions.
 OSS_FUZZ_BENCHMARKS = {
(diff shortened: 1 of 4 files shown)


jonathanmetzman:
Supposedly we get up to 256 jobs per workflow run.
From: https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration
```
Job matrix - A job matrix can generate a maximum of 256 jobs per workflow run. This limit also applies to self-hosted runners.
```
But it seems like we aren't getting that. I've experienced this before when we set up CI.
Maybe we should file an issue against GitHub?
Super annoying there's no way to debug this.

jonathanmetzman:
> Supposedly we get up to 256 jobs per workflow run.
> From: https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration
> 
> ```
> Job matrix - A job matrix can generate a maximum of 256 jobs per workflow run. This limit also applies to self-hosted runners.
> ```
> 
> But it seems like we aren't getting that many.
> Look here (https://github.com/google/fuzzbench/runs/1323852957) only presubmit is run. Builds aren't run.
> I've experienced this before when we set up CI.
> I count 37 fuzzers. 37*3 = 111.
> Maybe we should file an issue against GitHub?
> Super annoying there's no way to debug this.

Nevermind, this has to do with the fact that these builds only happen when certain files are edited.

jonathanmetzman:
In my latest one patch I modified eclipser to see if it can build on time.

inferno-chromium (approved):

jonathanmetzman:
Eclipser still timing out.

jonathanmetzman:
Maybe we should still land this anyway?

inferno-chromium (approved):

inferno-chromium:
> Maybe we should still land this anyway?

Yes, we should. funny part, eclipser didnt timeout in that other pr i tried - https://github.com/google/fuzzbench/pull/890

Issue #894, opened 2020-10-28
matio benchmark fails with qemu fuzzers


inferno-chromium:
See https://github.com/google/fuzzbench/pull/886#issuecomment-717992496

@Vaush fyi, any thoughts here ?

inferno-chromium:
> matio is a very problematic benchmark. Originally, in #666, it didn't work with aflplusplus and its variants, nor with qemu fuzzers, any of them (see [#666 (comment)](https://github.com/google/fuzzbench/pull/666#issuecomment-704359550)). After your fix, #800, by the checks it works on aflplusplus but doesn't work in almost any other fuzzer, such as afl or libfuzzer. I am honestly not sure about what's happening here, but I will look into it. I will open a draft PR to fix matio.

Can you disable/blocklist it on all the qemu benchmarks where it fails. this avoid dev confusion on CI.

Vaush:
We did originally, I will blocklist them again while I try to fix it

vanhauser-thc:
> matio is a very problematic benchmark. Originally, in **#666**, ...

very prophetic :D

inferno-chromium:
fuzzers are blacklisted.

PR #914, opened 2020-10-29
Allow fuzzers to ban sanitizers

Vaush:
In reference to #666 #903 and #894.
It was discovered that some fuzzers cannot work together with certain sanitizers. While we require sanitizers for our bug based benchmarks, if we cannot use them we simply won't be able to record data for a certain fuzzer and will have to discard it; considering that it would instead refuse to build at all if we forced it to use a sanitizer, I think it could be useful to be able for fuzzers to never build their benchmarks with certain sanitizers, no matter what the benchmark wants.

Another option would be to create a separate category altogether for bug based benchmarks, but if we didn't test this new group too by default, we could risk getting a lot of new fuzzers that don't check compatibility with these benchmarks.
Also, if we ever needed to use sanitizers for normal benchmarks, we would run into this problem all over again.

I have also a test commit with a modified afl_qemu that can build matio by banning ASAN and UBSAN, please tell me if you want me to upload it to try the functionality out.


Vaush:
@inferno-chromium @jonathanmetzman This is the functionality I was talking about. Tell me if you think it could be useful, right now it seems the best way to solve the problem with afl_qemu and our bug based benchmarks.

inferno-chromium:
For now, we are blocklisting them as these fuzzers wont be usuable for bug based benchmarking for now, like we need to catch bugs reliably with stacktraces. Doing this in https://github.com/google/fuzzbench/pull/917. for pending benchmark CLs, can you add similar template in benchmark.yaml.

Vaush:
> 
> 
> For now, we are blocklisting them as these fuzzers wont be usuable for bug based benchmarking for now, like we need to catch bugs reliably with stacktraces. Doing this in #917. for pending benchmark CLs, can you add similar template in benchmark.yaml.

Sure

PR #929, opened 2020-11-02
Upgrade python requirement from 3.7 to 3.8, add clusterfuzz pkg

inferno-chromium:
- Upgrade python 3.7 to 3.8 in images.
- Upgrade pylint, pytype and yapf packages, fix lint failures.
- Add libClusterFuzz for stacktrace parsing.

### DIFF
--- a/docker/dispatcher-image/Dockerfile
+++ b/docker/dispatcher-image/Dockerfile
@@ -21,7 +21,6 @@ WORKDIR $WORK
 
 # Install runtime dependencies for benchmarks, easy json parsing.
 RUN apt-get update -y && apt-get install -y \
-    git \
     jq \
     libglib2.0-0 \
     libxml2 \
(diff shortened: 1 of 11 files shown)


lszekeres (approved):

inferno-chromium:
Removed libClusterFuzz code since Oliver is putting a pip package, so that we can then just do pip install clusterfuzz. Also, this was kind of unrelated change to this 3.8 upgrade, so nice to go in another cl.

docker/dispatcher-image/Dockerfile
```diff
@@ -41,6 +39,9 @@ RUN curl https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64 > \
     /usr/local/bin/cloud_sql_proxy
 RUN chmod +x /usr/local/bin/cloud_sql_proxy
 
+# Add libClusterFuzz for stacktrace parsing.
+RUN pip3 install clusterfuzz
```
jonathanmetzman, line 43: Why can't we put this in requirements.txt?
Linting will explode if we start using this package without doing that.

jonathanmetzman (commented):
LGTM
Thanks!

docker/dispatcher-image/Dockerfile
inferno-chromium: Sure, will do.

PR #930, opened 2020-11-03
Remove virtualenv use from local experiment

inferno-chromium:
This was missed in last cl when we removed its use in dispatcher-image.


jonathanmetzman (approved):
LGTM

PR #935, opened 2020-11-04
adds tabbed coverage growth plots (linear and logscale)

wideglide:
- changes the coverage growth plot in the default report to
  generate both linear and logscale plots and a tabbed interface
  to switch between the two.  This will take much longer to generate,
  but adds a requested feature to all reports.

### DIFF
--- a/analysis/benchmark_results.py
+++ b/analysis/benchmark_results.py
@@ -239,25 +239,26 @@ class BenchmarkResults:
             self._get_full_path(plot_filename))
         return plot_filename
 
-    @property
-    def coverage_growth_plot(self):
-        """Coverage growth plot."""
-        plot_filename = self._prefix_with_benchmark('coverage_growth.svg')
-        self._plotter.write_coverage_growth_plot(
-            self._benchmark_df, self._get_full_path(plot_filename), wide=True)
-        return plot_filename
-
-    @property
-    def coverage_growth_plot_logscale(self):
-        """Coverage growth plot."""
-        plot_filename = self._prefix_with_benchmark(
-            'coverage_growth_logscale.svg')
+    def _coverage_growth_plot(self, filename, logscale=False):
+        """Coverage growth plot helper function"""
+        plot_filename = self._prefix_with_benchmark(filename)
         self._plotter.write_coverage_growth_plot(
             self._benchmark_df,
             self._get_full_path(plot_filename),
             wide=True,
-            logscale=True)
+            logscale=logscale)
         return plot_filename
+
+    @property
+    def coverage_growth_plot(self):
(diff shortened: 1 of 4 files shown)


jonathanmetzman (approved):
Thanks a lot.
WOuld you mind uploading a zip containing this so I can see what it looks like?

analysis/plotting.py
```diff
@@ -179,7 +179,8 @@ def write_coverage_growth_plot(self, benchmark_df, image_path, wide=False):
         self._write_plot_to_image(self.coverage_growth_plot,
                                   benchmark_df,
                                   image_path,
-                                  wide=wide)
+                                  wide=wide,
+                                  logscale=False)
```
inferno-chromium, line 183: This seems wrong to me, i think you need to add logscale argument and then do logscale=logscale here.

analysis/benchmark_results.py
```diff
@@ -247,6 +247,18 @@ def coverage_growth_plot(self):
             self._benchmark_df, self._get_full_path(plot_filename), wide=True)
         return plot_filename
 
+    @property
+    def coverage_growth_plot_logscale(self):
+        """Coverage growth plot."""
+        plot_filename = self._prefix_with_benchmark(
```
inferno-chromium, line 253: since coverage_growth_plot_logscale and coverage_growth_plot share all the code except logscale arg, can you create a common helper and use in both places.

inferno-chromium (changes requested):
Minor fixup needed.

analysis/plotting.py
wideglide: yep, fixed this another related issue

analysis/benchmark_results.py
wideglide: just fixed this as well

wideglide: Here's an example chart:

image

wideglide: I can't upload the zip  for some reason, but here's the [example report](https://tr0gd0r.com/download/example_dual_2020-11-01).

analysis/generate_report.py
```diff
@@ -155,10 +155,10 @@ def generate_report(experiment_names,
     data_path = os.path.join(report_directory, 'data.csv.gz')
     if from_cached_data and os.path.exists(data_path):
         experiment_df = pd.read_csv(data_path)
+        description = "from cached data"
```
inferno-chromium, line 158: at some point, i will fix this, not needed for this cl. i think description is there in experiment_df.

inferno-chromium (approved):
Thank you so much!

lszekeres (approved):
Very nice! Thanks so much!!

Issue #938, opened 2020-11-06
Pytype only runs with python3.7 in presubmit


BharathMonash:
I see that the CI has been upgraded to `python3.8` but `pytype` still looks for `python3.7` in presubmit, totally skipping the pytpe checks.  

Not sure if `pytype` supports all the needed features in python 3.8. Is that why it hasn't been changed?

PR #939, opened 2020-11-06
Enable pytype that now works on python 3.8, also upgrade pytest.

inferno-chromium:
Fixes https://github.com/google/fuzzbench/issues/938


jonathanmetzman (approved):
LGTM
Thanks!

PR #951, opened 2020-11-13
Integrate project php with target php-fuzz-parser, commit date 2020-07-25

Vaush:


inferno-chromium (changes requested):
for aflplusplus_optimal failure, i think you need to add it here as well

```
    elif benchmark_name == 'php_php-fuzz-execute':
        aflplusplus_fuzzer.build("classic", "ctx", "cmplog")
    elif benchmark_name == 'php_php-fuzz-parser':
        aflplusplus_fuzzer.build("classic", "ctx", "cmplog")
```

Vaush:
@inferno-chromium I had to change aflplusplus_optimal's fuzzer.py and the CI is acting up again. I tested in local and test-run goes well, should we just ignore the checks again?

PR #954, opened 2020-11-14
Test if php-fuzz-execute actually works

Vaush:
Dummy PR, will be deleted


inferno-chromium:
yes it does seem to work, closing.

PR #958, opened 2020-11-30
Remove unneeded timeout afl hack with 1000+.

inferno-chromium:
This is not needed after AFL_SKIP_CRASHES=1 which skips crashes
in seed corpus. Remove this hack as it break afl based fuzzers
on some benchmarks, e.g. https://www.fuzzbench.com/reports/2020-11-23/index.html#libpcap_fuzz_both

This regressed in https://github.com/google/fuzzbench/commit/c7e7bda1f373b406e37e27efd324c42fd388b945


inferno-chromium:
does not work for timeouts, looks like we need to still keep it.

PR #959, opened 2020-12-01
Save and plot crashes for bug based benchmarking.

inferno-chromium:


experiment/measurer/measure_manager.py
```diff
@@ -150,7 +150,7 @@ def save_snapshots():
         if not snapshots:
             return
 
-        db_utils.bulk_save(snapshots)
+        db_utils.add_all(snapshots)
```
inferno-chromium, line 153: This is needed since Snapshot has relationship to crashes and we need to add them both. bulk_save does not do relationships.

experiment/measurer/measure_manager.py
```diff
@@ -617,7 +628,8 @@ def measure_snapshot_coverage(fuzzer: str, benchmark: str, trial_num: int,
         return models.Snapshot(time=this_time,
                                trial_id=trial_num,
                                edges_covered=regions_covered,
-                               fuzzer_stats=fuzzer_stats_data)
+                               fuzzer_stats=fuzzer_stats_data,
+                               crashes=[])
```
inferno-chromium, line 632: no files changed in corpus archive which means no new crash files as well, use []

analysis/queries.py
```diff
@@ -15,21 +15,27 @@
...
+        Trial.experiment, Trial.fuzzer, Trial.benchmark,
+        Trial.time_started, Trial.time_ended,
+        Snapshot.trial_id, Snapshot.time, Snapshot.edges_covered,
+        Snapshot.fuzzer_stats, Crash.crash_key)\
         .select_from(Experiment)\
         .join(Trial)\
         .join(Snapshot)\
+        .join(Crash,
+              and_(Snapshot.time==Crash.time,
+                   Snapshot.trial_id == Crash.trial_id), isouter=True)\
```
inferno-chromium, line 38: is_outer as we need to keep rows where there are no crashes, this is needed for regular coverage benchmarking

experiment/measurer/measure_manager.py
jonathanmetzman: ACK

experiment/measurer/measure_manager.py
jonathanmetzman: ACK

analysis/data_utils.py
```diff
@@ -26,8 +26,15 @@ def validate_data(experiment_df):
...
-        'time_ended', 'time', 'edges_covered'
+        'experiment',
+        'benchmark',
+        'fuzzer',
+        'trial_id',
+        'time_started',
+        'time_ended',
+        'time',
+        'edges_covered',
+        'crash_key',
```
jonathanmetzman, line 37: I'm somewhat surprised this isn't multiple keys, how can this be?

analysis/data_utils.py
```diff
@@ -312,3 +319,13 @@ def experiment_level_ranking(experiment_snapshots_df,
     pivot_table = experiment_pivot_table(experiment_snapshots_df,
                                          benchmark_level_ranking_function)
     return experiment_level_ranking_function(pivot_table)
+
+
+def get_bugs_snaphot(benchmark_df):
```
jonathanmetzman, line 324: I'd probably try to be consistent and only use the word "crash" or "bug" and not both, at least not internally.

jonathanmetzman (approved):
LGTM

analysis/data_utils.py
inferno-chromium: Done!

analysis/data_utils.py
inferno-chromium: This is a different attribute crash_key which we join with Crash table, see analysis/queries.py. We can't use crashes since it is just a relationship, also makes it hard to work as a dataframe.

PR #962, opened 2020-12-02
Collapse simple template to improve deduplication.

inferno-chromium:


experiment/measurer/test_run_crashes.py
```diff
@@ -61,3 +61,15 @@ def test_filter_crash_type():
...
+# pylint: disable=protected-access
+def test_filter_crash_state():
+    """Tests _filter_crash_state."""
+    assert (run_crashes._filter_crash_state(
+        'Segv on unknown address:int* std::__1::fill_n<int*, long, int>\n'
+        'int arrow::util::RleDecoder::GetBatchWithDict<int>\n'
+        'parquet::DictDecoderImpl<parquet::PhysicalType<\n') ==
+            'Segv on unknown address:int* std::__1::fill_n\n'
+            'int arrow::util::RleDecoder::GetBatchWithDict\n'
+            'parquet::DictDecoderImpl<parquet::PhysicalType<\n')
```
mbarbella-chromium, line 75: Should we try to handle this case as well? We're in a bit of a weird state either way with the truncated lines, so I don't think too much could go wrong by removing anything after < to the end either.

mbarbella-chromium (approved):

experiment/measurer/run_crashes.py
```diff
@@ -30,14 +30,21 @@
     'crash_testcase', 'crash_type', 'crash_address', 'crash_state',
     'crash_stacktrace'
 ])
+
 SIZE_REGEX = re.compile(r'\s([0-9]+|{\*})$', re.DOTALL)
+TEMPLATE_REGEX = re.compile(r'<[^>]+>')
```
jonathanmetzman, line 35: Could you make it explicit that this is C++ templates?
I initial thought this had something to do with HTML templates.

experiment/measurer/test_run_crashes.py
inferno-chromium: Done, added to regex for truncated stuff.

experiment/measurer/run_crashes.py
inferno-chromium: Done!

jonathanmetzman (approved):
LGTM, Thanks!

PR #963, opened 2020-12-02
Fix bugs graph to account for different trials.

inferno-chromium:


analysis/plotting.py
```diff
@@ -445,6 +445,7 @@ def crash_plot(self, benchmark_df, axes=None):
                             x='time',
                             hue='fuzzer',
                             data=crash_df[crash_df.time <= snapshot_time],
+                            ci=None if self._quick else 95,
```
jonathanmetzman, line 448: What's 95?

inferno-chromium: Some confidence interval, we use the same in coverage graph too

```
ci
Size of the confidence interval to draw when aggregating with an estimator. “sd” means to draw the standard deviation of the data. Setting to None will skip bootstrapping.
```

lszekeres (approved):
LGTM, thanks!

PR #966, opened 2020-12-03
Use parameterized coverage growth plot function for bugs.

lszekeres:
- Remove redundant `crash_plot` function.
- Refactor function for computing `bugs_covered` column.

### DIFF
--- a/analysis/benchmark_results.py
+++ b/analysis/benchmark_results.py
@@ -327,10 +327,12 @@ class BenchmarkResults:
     def _crash_plot(self, filename, logscale=False):
         """Crash plot."""
         plot_filename = self._prefix_with_benchmark(filename)
-        self._plotter.write_crash_plot(self._benchmark_df,
-                                       self._get_full_path(plot_filename),
-                                       wide=True,
-                                       logscale=logscale)
+        self._plotter.write_coverage_growth_plot(
+            data_utils.add_bugs_covered_column(self._benchmark_df),
+            self._get_full_path(plot_filename),
+            wide=True,
+            logscale=logscale,
+            bugs=True)
         return plot_filename
 
     @property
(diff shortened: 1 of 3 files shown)


analysis/benchmark_results.py
```diff
@@ -327,10 +327,12 @@ def pairwise_unique_coverage_plot(self):
     def _crash_plot(self, filename, logscale=False):
         """Crash plot."""
         plot_filename = self._prefix_with_benchmark(filename)
-        self._plotter.write_crash_plot(self._benchmark_df,
-                                       self._get_full_path(plot_filename),
-                                       wide=True,
-                                       logscale=logscale)
+        self._plotter.write_coverage_growth_plot(
+            data_utils.add_bugs_covered_column(self._benchmark_df),
```
inferno-chromium, line 331: nit: Jonathan said in my cl to use either crash or bug in code, as mixing can be confusing. What do you think. "Bugs coverage" seems more clear than "Crashes coverage". Can you rename the tab name "Crashes" in report html template as well and use bug then.

inferno-chromium (approved):

analysis/benchmark_results.py
jonathanmetzman: +1 for bugs instead of crashes.

jonathanmetzman (approved):
LGTM

analysis/benchmark_results.py
lszekeres: Let me do the renames everywhere (including the parts outside this pr) in another pr later today.

inferno-chromium: SGTM, this rename is only needed for analysis/ code, dont worry about it being called crash in experiment/measurer, that part is all ok. just need this in report template and that analysis/*.py

PR #968, opened 2020-12-03
Add markers to make overlapping lines somewhat more visible.

lszekeres:
Not sure it helps too much:
Before:
image
After:
image
Should we keep them?


inferno-chromium (approved):
this is better i think.

jonathanmetzman (approved):
LGTM

PR #971, opened 2020-12-04
Disable libFuzzer, entropic patches

inferno-chromium:
These affect bug based benchmarking and also cause issues with
seed inputs (https://github.com/google/fuzzbench/issues/801)


jonathanmetzman (approved):
LGTM

PR #978, opened 2020-12-05
Requesting new AFL++ Experiments

mboehme:
In the previous experiment, I implemented the schedules into an older, intermediate version of AFL++ and compared to a new version of AFL++. So, the comparison to baseline AFL++ is not valid. There was also a divide-by-zero in most of the schedules, when dividing by log(hits). Repeating the previous experiment, and trying a few new schedules.
* `nolog` removes the logarithm from the bitmapsize when weighting. In my test runs, bitmap_size is nicely distributed..
* `noperf` removes the perf_score based boosting.
* `branches` revives the idea of champions. For each branch, there is a "favourite" seed, i.e., the fastest and smallest seed. We extend this idea and consider a favourite as a champion if the corresponding branch was executed by less than 255 generated inputs.
* `cutoff` assigns zero weight to seeds with above average hit counts.

Thanks Fuzzbench team!


mboehme:
Fir the bug-based benchmark, it says `50637 Error: Operation was canceled.`

Seems unrelated?

PR #979, opened 2020-12-06
Match sanitizer build flags and _OPTIONS same as OSS-Fuzz/ClusterFuzz

inferno-chromium:
- This was causing unintended crash blockers in several benchmarks, e.g. float-cast-overflow. This matches the config used in OSS-Fuzz.
- Match sanitizer options, this is needed, e.g. fixes missing stacktrace since UBSan requires print_stacktrace=1


jonathanmetzman (approved):
LGTM

PR #980, opened 2020-12-07
Modify bugs_covered_column behavior and usage

wideglide:
Adds the `bugs_covered` column to the experiment dataframe instead of the benchmark dataframe.  Also, modifies the behavior so that the `bugs_covered` column is added before the csv is exported.  This will allow for easier analysis during report generation and from the published data.

### DIFF
--- a/analysis/benchmark_results.py
+++ b/analysis/benchmark_results.py
@@ -328,7 +328,7 @@ class BenchmarkResults:
         """Bug coverage growth plot."""
         plot_filename = self._prefix_with_benchmark(filename)
         self._plotter.write_coverage_growth_plot(
-            data_utils.add_bugs_covered_column(self._benchmark_df),
+            self._benchmark_df,
             self._get_full_path(plot_filename),
             wide=True,
             logscale=logscale,
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -112,7 +112,7 @@ def add_bugs_covered_column(experiment_df):
     grouping1 = ['fuzzer', 'benchmark', 'trial_id', 'crash_key']
     grouping2 = ['fuzzer', 'benchmark', 'trial_id']
     grouping3 = ['fuzzer', 'benchmark', 'trial_id', 'time']
-    df = experiment_df
+    df = experiment_df.sort_values(grouping3)
     df['firsts'] = ~df.duplicated(subset=grouping1) & ~df.crash_key.isna()
     df['bugs_cumsum'] = df.groupby(grouping2)['firsts'].transform('cumsum')
     df['bugs_covered'] = df.groupby(grouping3)['bugs_cumsum'].transform('max')
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -106,15 +106,16 @@ def filter_max_time(experiment_df, max_time):
     return experiment_df[experiment_df['time'] <= max_time]
 
 
-def add_bugs_covered_column(benchmark_df):
-    """Return a modified benchmark df in which the |crash_key| column is
-    replaced with a |bugs_covered| column."""
-    new_df = benchmark_df.drop(columns='crash_key').drop_duplicates()
-    new_df['bugs_covered'] = new_df.apply(lambda x: benchmark_df[
-        (benchmark_df['trial_id'] == x['trial_id']) &
-        (benchmark_df['fuzzer'] == x['fuzzer']) &
-        (benchmark_df['time'] <= x['time'])]['crash_key'].nunique(),
-                                          axis=1)
+def add_bugs_covered_column(experiment_df):
+    """Return a modified experiment df in which adds a |bugs_covered| column,
+    a cumulative count of bugs covered over time."""
+    grouping1 = ['fuzzer', 'benchmark', 'trial_id', 'crash_key']
+    grouping2 = ['fuzzer', 'benchmark', 'trial_id']
+    grouping3 = ['fuzzer', 'benchmark', 'trial_id', 'time']
+    df['firsts'] = ~df.duplicated(subset=grouping1) & ~df.crash_key.isna()
+    df['bugs_cumsum'] = df.groupby(grouping2)['firsts'].transform('cumsum')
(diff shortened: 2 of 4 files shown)


analysis/data_utils.py
```diff
@@ -106,15 +106,17 @@ def filter_max_time(experiment_df, max_time):
...
+def add_bugs_covered_column(experiment_df):
+    """Return a modified experiment df in which adds a |bugs_covered| column,
+    a cumulative count of bugs covered over time."""
+    grouping1 = ['fuzzer', 'benchmark', 'trial_id', 'crash_key']
+    grouping2 = ['fuzzer', 'benchmark', 'trial_id']
+    grouping3 = ['fuzzer', 'benchmark', 'trial_id', 'time']
+    df = experiment_df.sort_values(grouping3)
+    df['firsts'] = ~df.duplicated(subset=grouping1) & ~df.crash_key.isna()
+    df['bugs_cumsum'] = df.groupby(grouping2)['firsts'].transform('cumsum')
+    df['bugs_covered'] = df.groupby(grouping3)['bugs_cumsum'].transform('max')
```
lszekeres, line 118: Wondering why we need to `max` and why can't we just `cumsum` the `firsts`:
```
df['bugs_covered'] = df.groupby(grouping3)['firsts'].transform('cumsum')
```

analysis/data_utils.py
```diff
@@ -106,15 +106,17 @@ def filter_max_time(experiment_df, max_time):
...
-def add_bugs_covered_column(benchmark_df):
-    """Return a modified benchmark df in which the |crash_key| column is
-    replaced with a |bugs_covered| column."""
-    new_df = benchmark_df.drop(columns='crash_key').drop_duplicates()
-    new_df['bugs_covered'] = new_df.apply(lambda x: benchmark_df[
-        (benchmark_df['trial_id'] == x['trial_id']) &
-        (benchmark_df['fuzzer'] == x['fuzzer']) &
-        (benchmark_df['time'] <= x['time'])]['crash_key'].nunique(),
-                                          axis=1)
+def add_bugs_covered_column(experiment_df):
```
lszekeres, line 109: Beautiful! Thanks so much!

Please consider adding a unit test for this if you happen to have time (otherwise no worries).

lszekeres (approved):

analysis/data_utils.py
wideglide: `cumsum` gives the current count, incrementing for every first over the current group.  I couldn't figure out another way to cumulatively sum, but just report the max for each time period.  It seemed less error prone to have `bugs_covered` reported the same way `edges_covered` does in the dataframe so that they can be used interchangeably.

analysis/benchmark_results.py
```diff
@@ -328,7 +328,7 @@ def _bug_coverage_growth_plot(self, filename, logscale=False):
         """Bug coverage growth plot."""
         plot_filename = self._prefix_with_benchmark(filename)
         self._plotter.write_coverage_growth_plot(
-            data_utils.add_bugs_covered_column(self._benchmark_df),
+            self._benchmark_df,
```
inferno-chromium, line 331: How does this work keeping the crash_key column as-is ? Wont this have duplicate rows for particular (trial_id, fuzzer, time) combination due to different crash keys.

wideglide: yes this is a still(?) a problem.  I think you currently have those duplicates in the data for every other chart as the duplicates were only filtered for the bug_coverage_growth plot?   However, I think the data (identifiable bugs) could still be useful to keep for unique bug analysis.  Similar to the unique coverage charts that are generated now.  In keeping the duplicate rows with different `crash_key` values, the `edges_covered` and `bugs_covered` numbers are all the same for a unique (benchmark, fuzzer, trial_id, time) combination.

This is probably why you have a really high count (45) for libfuzzer samples on arrow_parquet-arrow-fuzz summary?

Possible fixes:
1. Add a filter to `get_benchmark_snapshot` to drop the `crash_key` column and remove duplicates.
2. Add a filter for both edge and bug coverage growth plots to remove duplicates.

inferno-chromium (changes requested):
LGTM, need small changes.
1)
To fix the crash_key duplicate row, can you add

```
experiment_df = experiment_df.drop(columns='crash_key').drop_duplicates()
```
with a comment after it is converted to csv [after the ```experiment_df.to_csv(data_path)``` line]

2) Can you make bugs_covered an integer instead of float.

analysis/data_utils.py
wideglide: added a simple unit test!

analysis/benchmark_results.py
wideglide: Added the filter after the data is exported to csv.

analysis/test_data_utils.py
```diff
@@ -129,6 +139,13 @@ def test_filter_benchmarks():
     assert filtered_df.benchmark.unique() == benchmarks_to_keep
 
 
+def test_add_bugs_covered_columns():
+    experiment_df = create_experiment_data()
+    experiment_df = data_utils.add_bugs_covered_column(experiment_df)
+
+    assert 'bugs_covered' in experiment_df.columns
```
inferno-chromium, line 146: Usual test always validate the value, but i can add something on this later.

inferno-chromium (approved):

wideglide:
ok, I still missed something here, because `crash_key` is in the list of columns removed in `drop_uninteresting_columns()` when the Experiment results are initialized.
So now there's a problem because it tries to remove the column a second time.
I'm not sure how it was able to remove the column previously in `add_bugs_covered_column()`?

inferno-chromium:
> ok, I still missed something here, because `crash_key` is in the list of columns removed in `drop_uninteresting_columns()` when the Experiment results are initialized.
> So now there's a problem because it tries to remove the column a second time.
> I'm not sure how it was able to remove the column previously in `add_bugs_covered_column()`?

Fixing these new exceptions in https://github.com/google/fuzzbench/pull/982

PR #981, opened 2020-12-08
new benchmark request

vanhauser-thc:
This tests the new aflfast changes by @mboehme vs the original implementation and also how they impact explore/exploit.

this is the final benchmark before the 3.0 release :)
(there will be one last benchmark for the release itself that will compare 3.0 vs 2.68c)


vanhauser-thc:
All failures are the bug ones which is not surprising, plus aflplusplus_268c_qemu fails - dunno why, because I didnt touch that in this PR. but it is php exe that fails, so likely this was not blacklisted?

so please accept and start the PR :)
@inferno-chromium @lszekeres

vanhauser-thc:
btw I have very little time currently but I plan to check the issues with the bug targets over xmas/newyear.

jonathanmetzman (approved):
LGTM

PR #982, opened 2020-12-08
Fix exceptions from #980

inferno-chromium:
Also, remove random incorrect tests. bugs_covered will always
exist after #980, so need to keep that column. Also, don't remove
crash_key as it is already removed with drop_duplicates.

### DIFF
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -45,8 +45,8 @@ def validate_data(experiment_df):
 def drop_uninteresting_columns(experiment_df):
     """Returns table with only interesting columns."""
     return experiment_df[[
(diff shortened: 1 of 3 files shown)


lszekeres (approved):
LGTM. Thanks!

wideglide:
one more thing.
if you add 
```python
if 'bugs_covered' not in experiment_df.columns:
```
 as a condition before calling `add_bugs_covered`, then it will work from cached data that already has the column added.

inferno-chromium:
> one more thing.
> if you add
> 
> ```python
> if 'bugs_covered' not in experiment_df.columns:
> ```
> 
> as a condition before calling `add_bugs_covered`, then it will work from cached data that already has the column added.

I mean it can recalculate the column, not a big deal right ?

PR #986, opened 2020-12-14
Add llvm-cov options for coverage summary generation

BharathMonash:
This PR is about adding a few more options to the `llvm-cov` call in `coverage_utils.py`.

The options added on top of the existing ones are as follows:

- `-num-thread=1` :  llvm-cov auto-detects an appropriate number of threads to use. Here we are forcing usage of a single thread to export coverage data.

- `-region-coverage-gt=0` : As the current `covered_regions.json` only accounts for regions with hits greater than 0, we can simply add this option rather than filtering it out later in (`extract_covered_regions_from_summary_json`).

- `-skip-expansions` : This option enables us to skip exporting of macro expansion coverage data.

All these changes are made so that during measuring we can generate a full fledged summary to export segment and function coverage. As suggested, the changes to code to record segment and function coverage will follow in a new PR.

### DIFF
--- a/experiment/measurer/coverage_utils.py
+++ b/experiment/measurer/coverage_utils.py
@@ -251,7 +251,7 @@ class TrialCoverage:  # pylint: disable=too-many-instance-attributes
 def generate_json_summary(coverage_binary,
                           profdata_file,
                           output_file,
-                          summary_only=True):
+                          summary_only=False):
     """Generates the json summary file from |coverage_binary|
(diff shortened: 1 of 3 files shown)


inferno-chromium (approved):

inferno-chromium:
Running in https://www.fuzzbench.com/reports/2020-12-15/index.html, does seem to be going ok, but need to wait on experiment finishing fine (another dayish).

PR #987, opened 2020-12-15
Generate segment and function coverage data over time for an experiment

BharathMonash:
As suggested, opening a new PR for easier review. I have maintained a seperate file for the code added for generating function and segment coverage and same goes for the unit tests as well.

Linking the previous PR to retain all our conversations :)
Previous PR: #832 

PR is about generating 3 compressed CSV files recording the segment and function coverage over time. The CSV files (headers mentioned alongside) being generated are: (NOTE: These CSV would be available in a compressed (.gz) format)

 ```
1. segments.csv (only covered segments) - (benchmark_id, fuzzer_id, trial_id, time_stamp, file_id, line, col)
2. functions.csv - (benchmark_id, fuzzer_id, trial_id, time_stamp, function_id, hits)
3. names.csv  - (id, name, type)
```
Filestore path for these files:  **`experiment_data/$(EXPERIMENT_NAME)/coverage/data/{files above}.csv.gz`** 

The segment and function coverage is being recorded while measuring snapshots (every 900 secs), hence the time_stamp column. The generated data captures the concrete code elements that are covered over time. For functions, we also maintain hit counts. For segments, we only record those that have been covered and add the newly covered segments for later time stamps. This makes it space efficient alongside compression.

This data is generated over the entire campaign at every snapshot point when coverage is measured but the CSV files are only available after the experiment ends.

```
Experiment config (local):
        benchmark: libjpeg-turbo-07-2017, freetype2-2017
        fuzzers: libfuzzer, afl
        trials: 2 (each)
        max_total_time: 1860 (2 cycles)

Output files size for compressed CSVs:
        segment.csv.gz:  230.5 kB
        functions.csv.gz: 63.6 kB
        names.csv.gz:  32.0 kB
```

Based on the experiment results shown above, if we have to estimate the sizes on a full experiment, it can be calculated as:
```((size_of_segment.csv.gz + (size_of_function.csv.gz * measure_cycles * trials)) * fuzzers + size_of_name.csv.gz) * benchmarks ```

So for a 24-hour experiment (96 cycles), 20 benchmark, 15 fuzzers and 10 trials the estimated size is ~ 740 mB. Due to certain repetitive patterns in the data, the final compressed sized may even result to lower than 740 mB as compression engines give a better compression rate if the records are similar in a way.

### CODE REVIEW

**Comment by mboehme** [comment]
@googlebot I consent.

File: experiment/measurer/test_experiment_coverage_utils.py [hunk 1]
```diff
@@ -0,0 +1,44 @@
...
+
+def get_test_data_path(*subpaths):
+    """Returns the path of |subpaths| relative to TEST_DATA_PATH."""
+    return os.path.join(TEST_DATA_PATH, *subpaths)
+
+
+def test_extract_segments_and_functions_from_summary_json(fs):
+    """Tests that extract_covered_regions_from_summary_json returns the covered
+    segments and functions from summary json file."""
+    num_functions_in_cov_summary = 3  # For testing.
```
jonathanmetzman, line 30: Remove these "For testing" comments please.

experiment/measurer/test_measure_manager.py
```diff
@@ -25,7 +25,8 @@
 from database import models
 from database import utils as db_utils
 from experiment.build import build_utils
-from experiment.measurer import measure_manager
+from experiment.measurer import measure_manager, experiment_coverage_utils as \
```
jonathanmetzman, line 28: Please don't import two modules on one line (think it's a style-guide violation).
Also, please just import experiment_coverage_utils as experiment_coverage_utils (not exp_cov_utils please).

experiment/measurer/measure_manager.py
```diff
@@ -130,8 +143,12 @@ def measure_all_trials(experiment: str, max_total_time: int, pool, q) -> bool:
     if not unmeasured_snapshots:
         return False
 
+    process_specific_df_containers = \
```
jonathanmetzman, line 146: Try to break this up on multiple lines without a backslash please (I think `presubmit.py format` will help here).

experiment/measurer/experiment_coverage_utils.py
```diff
@@ -0,0 +1,189 @@
...
+        csv_filestore_helper('segments.csv.gz', self.segment_df)
+        csv_filestore_helper('names.csv.gz', self.name_df)
+
+
+def extract_segments_and_functions_from_summary_json(  # pylint: disable=too-many-locals
+        summary_json_file, benchmark, fuzzer, trial_id, time):
+    """Return a trial-specific data frame container with segment and function
+     coverage information given a trial-specific coverage summary json file."""
+
+    process_specific_df_container = DataFrameContainer()
```
jonathanmetzman, line 159: Are you calling this "process_specific" because there will be one for each item we map using the multiprocessing pool?
Because I think that's a weird detail to name this var after (I think it's also wrong, the processes in the pool are reused). I believe we are mapping over trials, can you name at after that (e.g. `trial_df_container`).

experiment/measurer/experiment_coverage_utils.py
```diff
@@ -0,0 +1,189 @@
+# Copyright 2020 Google LLC
```
jonathanmetzman, line 1: Really awesome this is in its own module.

Could you please think of a better name? `experiment_coverage_utils.py` doesn't really say much to me other than this is not `coverage_utils.py`.

experiment/measurer/experiment_coverage_utils.py
```diff
@@ -0,0 +1,189 @@
...
+
+class DataFrameContainer:  # pylint: disable=too-many-instance-attributes
+    """Maintains segment and function coverage information, and writes this
+    information to CSV files."""
+
+    def __init__(self):
+        """Constructor"""
+        self.segment_entries = []
+        self.function_entries = []
+        self.fuzzer_names = dict()
```
jonathanmetzman, line 35: nit: just do `{}` instead of `dict`

experiment/measurer/experiment_coverage_utils.py
```diff
@@ -0,0 +1,189 @@
...
+        self.file_names = dict()
+
+        # Will be initialized upon done_adding_entries().
+        self.segment_df = None
+        self.function_df = None
+        self.name_df = None
+
+    def add_function_entry(self, benchmark, fuzzer, trial_id, function,
+                           function_hits, time):
+        # pylint: disable=too-many-arguments
```
jonathanmetzman, line 47: I think you want to put this on the line with the arguments (you may need to run lint and format a few times to get it right, lint and format have some weird disagreement about this but eventually get to the point where both agree).

experiment/measurer/experiment_coverage_utils.py
```diff
@@ -0,0 +1,189 @@
...
+        for name in self.fuzzer_names:
+            name_entries.append([self.fuzzer_names[name], name, "fuzzer"])
+        for name in self.function_names:
+            name_entries.append([self.function_names[name], name, "function"])
+        for name in self.file_names:
+            name_entries.append([self.file_names[name], name, "file"])
+        self.name_df = pd.DataFrame(name_entries,
+                                    columns=["id", "name", "type"])
+
+    def name_to_id(self, name):  # pylint: disable=no-self-use
```
jonathanmetzman, line 108: Is this really necessary? How much diskspace are we saving here?

jonathanmetzman (commented):
I did a first pass. I'm going to try to finish reviewing this before I go on vacation next Tuesday.

experiment/measurer/test_experiment_coverage_utils.py
```diff
@@ -0,0 +1,44 @@
+# Copyright 2020 Google LLC
```
jonathanmetzman, line 1: I think you should probably write more tests. I'll try to think of some concrete suggestions when I understand this PR better. But on its surface this seems like it won't be easy to maintain without more tests.

experiment/measurer/experiment_coverage_utils.py
BharathMonash: According to my readings, this step reduces the disk space consumption by about 5-6 folds. We record file and function names and they are pretty long in general (generally 30+ characters since file names are relative paths and function names in benchmarks are long as well). Hence reducing it to a 7 character unique id helps reduce diskspace consumed considerably.

Also, for segments, even-though we only record newly discovered segments from the second cycle onwards, there can be up to 25k+ entries in the first cycle itself. For functions, we record the function names, and we record in every cycle irrespective of wether the function was previously covered or not.

BharathMonash:
Thanks @jonathanmetzman. I have have made a few changes as requested.  Let me come up with a better name for the module and come up with concrete test cases as well. Any suggestions from your side for the test cases are welcome :)

experiment/measurer/experiment_coverage_utils.py
mboehme: Not very good at naming, either. My suggestions:
* `coverage_printer.py`
* `coverage_data_printer.py`
* `data_to_csv_utils.py`

experiment/measurer/test_experiment_coverage_utils.py
mboehme: @BharathMonash We could write tests to check for the following:

- [ ] Check whether functions.csv.gz, segments.csv.gz, and names.csv.gz are empty.
- [ ] Check whether names and IDs are properly resolved (benchmark, fuzzer, function, file, type).
- [ ] Check whether you find the expected values in the right columns for all columns (e.g., afl under fuzzers, file.c:12 file, etc).
- [ ] Check whether there are no duplicate entries in segment_df and function_df after remove_redundant_duplicates is called.
- [ ] Check whether `experiment_specific_df_container` is correctly filled in `measure_all_trials` when there are multiple snapshots (5?) in the multiprocessing pool.

BharathMonash: Hi @mboehme and @jonathanmetzman, I have added few test cases and changed the name for the file as well. Although, the last suggested test case (with multiprocessing) could not be achieved as python's multiprocessing package has a few OS calls which pyfakefs cannot handle as they don't support for multiprocessing package as of now.

Please let me know if any test case can be improved further or if new test cases can be added.

BharathMonash:
Hi @jonathanmetzman, hope this note finds you well. This is just a gentle reminder to review the requested changes :)

jonathanmetzman:
> Hi @jonathanmetzman, hope this note finds you well. This is just a gentle reminder to review the requested changes :)

Sorry @BharathMonash I'll try to get to this today

BharathMonash:
> > Hi @jonathanmetzman, hope this note finds you well. This is just a gentle reminder to review the requested changes :)
> 
> Sorry @BharathMonash I'll try to get to this today

Thanks @jonathanmetzman, completely at your convenience. Sorry, I didn't mean to rush you.

experiment/measurer/coverage_printer_utils.py
```diff
@@ -0,0 +1,187 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Utility DataFrameContainer class and functions for report generation on
+segment and function coverage for the entire experiment
+(all Fuzzer-benchmark-trial combinations)."""
+
+import os
+import hashlib
+import pandas as pd
```
jonathanmetzman, line 20: nit: put a space before this import to seperate stdlib imports from third-party.

experiment/measurer/measure_manager.py
```diff
@@ -27,6 +27,7 @@
 import time
 from typing import List, Set
 import queue
+import pandas as pd
```
jonathanmetzman, line 30: Put a space before this import please.

experiment/measurer/experiment_coverage_utils.py
jonathanmetzman: Maybe coverage_over_time_utils?

experiment/measurer/experiment_coverage_utils.py
jonathanmetzman: > According to my readings, this step reduces the disk space consumption by about 5-6 folds. 

Right, but how big is a file without this compression vs with it?
If we are going from 80 MB in size for a full experiment to 20 MB, I don't think it is worth it.

I'm also concerned with this method you've used. It isn't sound. There can be collisions here if two things happen to have the same 8 char prefix. Why can't we simply use a numerical ID?

jonathanmetzman: Also have you tried using bz2 for compression instead of gz? It generally compresses to smaller sizes but is also slower.

experiment/measurer/experiment_coverage_utils.py
```diff
@@ -0,0 +1,189 @@
...
+                        trial_id,
+                        file['filename'],
+                        segment[0],  # Segment line.
+                        segment[1],  # Segment column.
+                        time)
+
+    except (ValueError, KeyError, IndexError):
+        coverage_utils.logger.error(
+            'Failed when extracting trial-specific segment and function '
+            'information from coverage summary')
```
jonathanmetzman: nit: end with period.

experiment/measurer/measure_manager.py
```diff
@@ -42,6 +43,7 @@
 from database import utils as db_utils
 from database import models
 from experiment.build import build_utils
+from experiment.measurer import experiment_coverage_utils as exp_cov_utils
```
jonathanmetzman: don't do `import ... as ...` please use the real name.

experiment/measurer/coverage_printer_utils.py
```diff
@@ -0,0 +1,187 @@
...
+
+from common import experiment_path as exp_path
+from common import filestore_utils
+from experiment.measurer import coverage_utils
+
+
+class DataFrameContainer:  # pylint: disable=too-many-instance-attributes
```
jonathanmetzman, line 27: I think this name is very vague and would like a better one.
Maybe CoverageOverTimeData or DetailedCoverageData?

experiment/measurer/experiment_coverage_utils.py
jonathanmetzman: Or maybe detailed_coveage_utils?

experiment/measurer/measure_manager.py
```diff
@@ -70,21 +72,29 @@ def measure_main(experiment_config):
     initialize_logs()
     logger.info('Start measuring.')
 
+    # Create data frame container for segment and function coverage info.
+    experiment_specific_df_container = (
```
jonathanmetzman, line 76: This name is super vague. Let's rename it once we've renamed DataFrameContainer.

experiment/measurer/measure_manager.py
```diff
@@ -180,6 +198,18 @@ def save_snapshots():
     # If we have any snapshots left save them now.
     save_snapshots()
 
+    # Concatenate all process-specific data frames and remove duplicates.
```
jonathanmetzman, line 201: What does "process" mean here?

experiment/measurer/measure_manager.py
```diff
@@ -574,9 +615,9 @@ def get_fuzzer_stats(stats_filestore_path):
     return json.loads(stats_str)
 
 
-def measure_trial_coverage(  # pylint: disable=invalid-name
-        measure_req, max_cycle: int,
-        q: multiprocessing.Queue) -> models.Snapshot:
+def measure_trial_coverage(  # pylint: disable=invalid-name,too-many-arguments
+        measure_req, max_cycle: int, q: multiprocessing.Queue,
+        process_specific_df_containers) -> models.Snapshot:
```
jonathanmetzman, line 620: Don't really know what "process" means here.

jonathanmetzman:
@inferno-chromium could you also please take a look and share your thoughts on this?

experiment/measurer/experiment_coverage_utils.py
BharathMonash: Since we call this method in each thread, we use hashing to resolve names to ids as we want to have consistent IDs for the same names in different threads and hashing is a good way to do it. 

Moreover, we call `name_to_id` and `remove_redundant_duplicates` in each thread as this helps us reduce the memory consumption (as it reduces the df size frequently). We can use numerical ID but then we would have to resolve names to IDs after we collect all the data for the experiment which would increase the memory consumption considerably. While hashing there is possibility of collision of IDs, no doubt, but the probability is quite low. So, currently I'm truncating the computed hash at 7 characters but we can increase this to say 10 or 12 characters which would drop the probability of collision further.

Also for my local experiments (not full-scale), using `name_to_id` helps reduce disk space consumed by the `*.csv.gz` by 5-6 folds. Our estimate for the disk space consumption for the entire experiment (full-scale) is 800-850 MB currently. Without this step the estimate would shoot up to around 6-7 GB for a full-scale experiment.

experiment/measurer/measure_manager.py
BharathMonash: Process was to indicate that it's specific to a single thread. Will change it to trial, as we are mapping over trials.

experiment/measurer/measure_manager.py
BharathMonash: Will changed it to "trial". "process" here was to indicate that it's specific to a single thread.

experiment/measurer/coverage_printer_utils.py
```diff
@@ -0,0 +1,187 @@
...
+        self.fuzzer_names[fuzzer] = fuzzer_id
+        self.benchmark_names[benchmark] = benchmark_id
+        self.file_names[file_name] = file_id
+
+        segment_entry = [
+            benchmark_id, fuzzer_id, trial_id, time, file_id, line, column
+        ]
+        self.segment_entries.append(segment_entry)
+
+    def done_adding_entries(self):
```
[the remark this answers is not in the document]
BharathMonash: Changed `done_adding_entries` to `generate_data_frames_after_adding_all_entries`.

experiment/measurer/measure_manager.py
BharathMonash: Change `experiment_specific_df_container` to `experiment_coverage_data_container`

BharathMonash:
Thanks, @jonathanmetzman. I have incorporated the requested changes. 
I have left a few comments on `name_to_id` stuff to explain why we are using it in detail. Let me know what you think about it or if you have a better idea :)

experiment/measurer/experiment_coverage_utils.py
BharathMonash: > > According to my readings, this step reduces the disk space consumption by about 5-6 folds.
> 
> Right, but how big is a file without this compression vs with it?
> If we are going from 80 MB in size for a full experiment to 20 MB, I don't think it is worth it.
> 
> I'm also concerned with this method you've used. It isn't sound. There can be collisions here if two things happen to have the same 8 char prefix. Why can't we simply use a numerical ID?

Since we call this method in each thread (multiprocessing), we use hashing to resolve names to ids as we want to have consistent IDs for the same names in different threads and hashing is a good way to do it.

Moreover, we call `name_to_id` and `remove_redundant_duplicates` in each thread as this helps us reduce the memory consumption (as it reduces the df size frequently). We can use numerical ID but then we would have to resolve names to IDs after we collect all the data for the experiment which would increase the memory consumption considerably. While hashing there is possibility of collision of IDs, no doubt, but the probability is quite low. So, currently I'm truncating the computed hash at 7 characters but we can increase this to say 10 or 12 characters which would drop the probability of collision further.

Also for my local experiments (not full-scale), using `name_to_id` helps reduce disk space consumed by the *.csv.gz by 5-6 folds. Our estimate for the disk space consumption for the entire experiment (full-scale) is 800-850 MB currently. Without this step the estimate would shoot up to around 6-7 GB for a full-scale experiment.

experiment/measurer/coverage_over_time_utils.py
```diff
@@ -0,0 +1,189 @@
...
+    def __init__(self):
+        """Constructor"""
+        self.segment_entries = []
+        self.function_entries = []
+        self.fuzzer_names = {}
+        self.benchmark_names = {}
+        self.function_names = {}
+        self.file_names = {}
+
+        # Will be initialized upon
```
jonathanmetzman, line 41: Change comment to: "Set by generate_data_frames_after_adding_all_entries()" please.

experiment/measurer/measure_manager.py
```diff
@@ -70,21 +73,29 @@ def measure_main(experiment_config):
     initialize_logs()
     logger.info('Start measuring.')
 
+    # Create data frame container for segment and function coverage info.
+    experiment_coverage_data_container = (
```
jonathanmetzman, line 77: Please rename this to `detailed_coverage_data`.

experiment/measurer/coverage_over_time_utils.py
```diff
@@ -0,0 +1,189 @@
+# Copyright 2020 Google LLC
```
jonathanmetzman, line 1: Err...I appreciate the rename but the two suggestions I made were alternatives. So please just rename this module detailed_coverage_data_utils.py. Thanks!

BharathMonash: Oops, sorry for that :)

BharathMonash: Done. Thanks!

experiment/measurer/measure_manager.py
BharathMonash: Renamed it.

experiment/measurer/coverage_over_time_utils.py
BharathMonash: Changed it.

jonathanmetzman (approved):
This LGTM.
I'll wait for Abhishek to review but once he approves I'll merge on Tuesday (Monday is a holiday here in the states)

mboehme:
Thanks for the green light @jonathanmetzman! We haven't been able to trail this PR at Fuzzbench scale. Would suggest to run a Fuzzbench-scale experiment on this branch before merging.

inferno-chromium:
> Thanks for the green light @jonathanmetzman! We haven't been able to trial this PR at Fuzzbench scale. Would suggest to run a Fuzzbench-scale experiment on this branch before merging.

I will run a trial official experiment using this PR over the weekend. Also, will try to review soonish.

BharathMonash:
Thanks @jonathanmetzman and @inferno-chromium. I hope that the trial run is successful 🤞.

inferno-chromium:
> Thanks @jonathanmetzman and @inferno-chromium. I hope that the trial run is successful 🤞.

Experiment failing on
Error occurred during measuring.
https://www.fuzzbench.com/reports/2021-01-15/index.html
```
2021-01-16 14:11:33.837 PST
Error occurred during measuring.
Expand all | Collapse all{
 insertId: "1ggan0tfu029ej"  
 jsonPayload: {
  component: "dispatcher"   
  experiment: "2021-01-15"   
  instance_name: "d-2021-01-15"   
  message: "Error occurred during measuring."   
  subcomponent: "measurer"   
  traceback: "Traceback (most recent call last):
  File "/work/src/experiment/measurer/measure_manager.py", line 112, in measure_loop
    if not measure_all_trials(experiment, max_total_time, pool,
  File "/work/src/experiment/measurer/measure_manager.py", line 148, in measure_all_trials
    manager.list(  # pytype:disable=attribute-error
  File "/usr/local/lib/python3.8/multiprocessing/managers.py", line 740, in temp
    token, exp = self._create(typeid, *args, **kwds)
  File "/usr/local/lib/python3.8/multiprocessing/managers.py", line 623, in _create
    conn = self._Client(self._address, authkey=self._authkey)
  File "/usr/local/lib/python3.8/multiprocessing/connection.py", line 502, in Client
    c = SocketClient(address)
  File "/usr/local/lib/python3.8/multiprocessing/connection.py", line 630, in SocketClient
    s.connect(address)
ConnectionRefusedError: [Errno 111] Connection refused
```

mboehme:
Sorry about that. We'll investigate. Worst case, we'll trade memory for disk storage and write the trial specific coverage data to the (trial specific) subject-fuzzer/trial folder to handle concurrency in the file system and post-process sequentially. This would also be less invasive for the measurer and simplify unit testing.

inferno-chromium:
> Sorry about that. We'll investigate. Worst case, we'll trade memory for disk storage and write the trial specific coverage data to the (trial specific) subject-fuzzer/trial folder to handle concurrency in the file system and post-process sequentially. This would also be less invasive for the measurer and simplify unit testing.

No worries at all, it is completely hard to predict this fat measurer. Sometime in Q2, we will split it, right now focusing on bug benchmarks. For the CL, using disk sounds totally good, ton of disk space and we can even add more as needed.

PR #990, opened 2020-12-16
Bug based top level ranking for bug based reports.

lszekeres:
For bug based experiments, the top level ranking and critical difference diagram
will be computed based on the number of bugs found, not by the number of code
regions covered.


inferno-chromium (approved):

jonathanmetzman (commented):
LGTM

Issue #994, opened 2020-12-16
Make sure build step and fuzzer run have right sanitizer options for bug based benchmarks


inferno-chromium:
Right now, we set these only in measurer, need to fix this.

Also, dont degrade fuzzing performance with https://github.com/google/fuzzbench/pull/988#issuecomment-746009404

PR #988, opened 2020-12-15
afl++ support for bug benchmarks

vanhauser-thc:
when UBSAN is activated this results in a huge increase in compile time.
I have a theory why this could be, but no clue how to resolve this.
So for bug benchmarks we switch back to lame clang -fsanitize=coverage... :(
but this will just degrade exec/s and none of the features.

PR #995, opened 2020-12-17
Bug benchmarking fixes

inferno-chromium:
- Set sanitizer options in runner, to match measurer.
- Remove unsigned-integer-overflow, it breaks builds and is also
  unused in OSS-Fuzz, we only see stacks and then silence them.
  Best to just remove it altogether.
- Move sanitizer.py to common/ since runner cannot access files
  in experiment/measurer and also this is now for common given
  the usecase.
- Optimize sanitizer flags for fuzzing, see
  https://github.com/google/AFL/blob/master/afl-fuzz.c#L7558
  and https://github.com/google/fuzzbench/pull/988#issuecomment-746009404
Fixes https://github.com/google/fuzzbench/issues/994

### DIFF
--- a/docker/benchmark-runner/Dockerfile
+++ b/docker/benchmark-runner/Dockerfile
@@ -59,6 +59,8 @@ ENV ROOT_DIR=/src
 # Copy over all the build artifacts (without * to preserve directory structure).
 # This also copies seed and dictionary files if they are available.
 COPY --from=builder /out/ ./
+# Copy the benchmarks directory.
+COPY benchmarks $ROOT_DIR/benchmarks
 # Copy the fuzzers directory.
 COPY fuzzers $ROOT_DIR/fuzzers
 # Create empty __init__.py to allow python deps to work.
--- a/fuzzers/test_utils.py
+++ b/fuzzers/test_utils.py
@@ -99,12 +99,12 @@ def test_initialize_env_in_environment_with_sanitizer(fs, environ):
         '-fsanitize=address -fsanitize=array-bounds,bool,builtin,enum,'
         'float-divide-by-zero,function,integer-divide-by-zero,null,object-size,'
         'return,returns-nonnull-attribute,shift,signed-integer-overflow,'
-        'unsigned-integer-overflow,unreachable,vla-bound,vptr -O3')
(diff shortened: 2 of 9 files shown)


docker/benchmark-runner/Dockerfile
```diff
@@ -59,6 +59,8 @@ ENV ROOT_DIR=/src
 # Copy over all the build artifacts (without * to preserve directory structure).
 # This also copies seed and dictionary files if they are available.
 COPY --from=builder /out/ ./
+# Copy the benchmarks directory.
```
inferno-chromium, line 62: This is needed for getting benchmark config in runner.

jonathanmetzman (approved):
LGTM

fuzzers/test_utils.py
```diff
@@ -131,10 +131,10 @@ def test_initialize_env_in_var_with_sanitizer(fs):
...
         'return,returns-nonnull-attribute,shift,signed-integer-overflow,'
-        'unsigned-integer-overflow,unreachable,vla-bound,vptr -O3')
+        'unreachable,vla-bound,vptr -O3')
     assert env.get('CXXFLAGS') == (
         '-fsanitize=address -fsanitize=array-bounds,bool,builtin,enum,'
         'float-divide-by-zero,function,integer-divide-by-zero,null,object-size,'
         'return,returns-nonnull-attribute,shift,signed-integer-overflow,'
-        'unsigned-integer-overflow,unreachable,vla-bound,vptr -stdlib=libc++ '
+        'unreachable,vla-bound,vptr -stdlib=libc++ '
         '-O3')
```
jonathanmetzman, line 140: I wonder if `-O3` is why some bugs we find on OSS-Fuzz don't repro here. We don't use O3 in OSS-Fuzz.

inferno-chromium: Should i switch to -O1 ?

inferno-chromium: Done, for bug based benchmarking, switched to -O1.

jonathanmetzman: Not sure, probably the right call.
Thanks!

PR #996, opened 2020-12-18
New experiment

vanhauser-thc:
this tests two variants of romu (new RNG) and skim (fast? coverage analysis).
I hope the UBSAN/ASAN issues are fixed soon so we can see how afl++ performs on the bug benchmarks :)


inferno-chromium:
> this tests two variants of romu (new RNG) and skim (fast? coverage analysis).
> I hope the UBSAN/ASAN issues are fixed soon so we can see how afl++ performs on the bug benchmarks :)

that is fixed and running in https://www.fuzzbench.com/reports/2020-12-19-bug/index.html

inferno-chromium (approved):

PR #1002, opened 2020-12-22
Allow bug based benchmarks to be requested via type attribute.

inferno-chromium:


inferno-chromium:
TBR minor addition to make service work with bug benchmarks. that benchmark hardcode will be removed in new year, merging for now.

service/automatic_run_experiment.py
```diff
@@ -47,7 +47,7 @@
 
 # TODO(metzman): Stop hardcoding benchmarks and support marking benchmarks as
 # disabled using a config file in each benchmark.
-BENCHMARKS = [
+CODE_BENCHMARKS = [
```
jonathanmetzman, line 50: nit: probably would have called "coverage"

service/experiment-requests.yaml
```diff
@@ -18,6 +19,15 @@
 # You can run "make presubmit" to do basic validation on this file.
 # Please add new experiment requests towards the top of this file.
 
+- experiment: 2020-12-22-bug
+  description: "afl++ bugs benchmarking"
+  type: bug
```
jonathanmetzman, line 24: Probably would have called this "benchmarks_type"

PR #1004, opened 2020-12-23
remove temporary variants, update commit id, request new experiment

vanhauser-thc:


service/experiment-requests.yaml
```diff
@@ -19,6 +19,16 @@
 # You can run "make presubmit" to do basic validation on this file.
 # Please add new experiment requests towards the top of this file.
 
+- experiment: 2020-12-23
+  description: "best of current state"
+  fuzzers:
```
inferno-chromium, line 24: Do you want a regular coverage experiment or "type: bug" for bug based benchmarking.

vanhauser-thc: a regular with fries on the side please! :)

vanhauser-thc:
@inferno-chromium  the experiment is not starting - could it be that due to the new type: bug feature there is a bug that only bug benchmarks can be started?

inferno-chromium:
> @inferno-chromium the experiment is not starting - could it be that due to the new type: bug feature there is a bug that only bug benchmarks can be started?

> a regular with fries on the side please! :)

While you were enjoying the fries, fuzzbench was working hard :) fyi, if all the fuzzers requested are in the core-fuzzers.yaml list, then they go in the main directory (and not experimental), so your experiment results are at https://www.fuzzbench.com/reports/2020-12-23/index.html

vanhauser-thc:
ah I thought all requested ones would be in experimental. good to know, thanks!
enjoy the holidays!

jonathanmetzman:
> ah I thought all requested ones would be in experimental. good to know, thanks!
> enjoy the holidays!

I think this is probably unintuitive behavior we should probably change.

inferno-chromium:
> > ah I thought all requested ones would be in experimental. good to know, thanks!
> > enjoy the holidays!
> 
> I think this is probably unintuitive behavior we should probably change.

yes this was due to missing aflplusplus_optimal in core-fuzzers.yaml, i fixed it already by adding it.

PR #1006, opened 2020-12-26
compare afl++ and eclipser to 2x afl++

vanhauser-thc:
afl++ combined with eclipser performs really, really well.
However we have two processes working at the same time, which is not really a fair comparison to other fuzzers.
So to see what the value of eclipser is, this compares two parallel running afl++ in sync compare to afl++ and eclipser.
this could be further enhanced by not running the double afl++ both with cmplog, but this way this does not require a rewrite of our main fuzzer.py


vanhauser-thc:
fyi aflplusplus_optimal and aflplusplus_eclipser both are running on all benchmarks in prod, so it is just in the CI that they are failing. I would propose fixing the CI so that it mirrors prod in the limits etc.

inferno-chromium (approved):

PR #1015, opened 2021-01-04
Don't process crashes for non-bug benchmarks.

inferno-chromium:
Also, optimize by not processing ooms and timeout by looking
at filenames.

Basically, in regular non-bug benchmark experiments, woff benchmark generates a lot of ooms, so we shouldn't be processing them.


jonathanmetzman (approved):
LGTM

PR #1025, opened 2021-01-11
Log build failures

jonathanmetzman:
Put logs of build failures in GCS

### DIFF
--- a/experiment/build/gcb_build.py
+++ b/experiment/build/gcb_build.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Module for building things on Google Cloud Build for use in trials."""
 
+import subprocess
 import tempfile
 from typing import Dict
 
@@ -100,8 +101,12 @@ def _build(
         result = new_process.execute(command,
                                      write_to_stdout=False,
                                      kill_children=True,
-                                     timeout=timeout_seconds)
+                                     timeout=timeout_seconds,
+                                     expect_zero=False)
+        # TODO(metzman): Refactor code so that local_build stores logs as well.
         build_utils.store_build_logs(config_name, result)
+        if result.retcode != 0:
+            raise subprocess.CalledProcessError(result.retcode, command)
     return result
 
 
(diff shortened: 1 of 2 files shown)


experiment/build/gcb_build.py
```diff
@@ -100,8 +101,12 @@ def _build(
         result = new_process.execute(command,
                                      write_to_stdout=False,
                                      kill_children=True,
-                                     timeout=timeout_seconds)
+                                     timeout=timeout_seconds,
+                                     expect_zero=False)
+        # TODO(metzman): Refactor code so that local_build stores logs as well.
         build_utils.store_build_logs(config_name, result)
+        if result.retcode != 0:
+            raise subprocess.CalledProcessError(result.retcode, command)
```
inferno-chromium, line 109: I don't understand this. How will raising an error capture this in logs ? Won't it just cause uncaught exceptions. I don't see any exceptions caught and then retried in https://github.com/google/fuzzbench/blob/4eec33d326b73a46f1e1875bb64edc267eb2ec8a/experiment/build/builder.py#L129

jonathanmetzman: >How will raising an error capture this in logs ?

The capture is done in the line before.
The difference is that [we no longer expect_zero](https://github.com/google/fuzzbench/pull/1025/files/f8c5dd112f3e5c4dd9f4d9fed0df3c3f010fde9b#diff-d1b96055e2c59a647f342428dc8962f1dda8ac13b339e51685c89772d8279aa8R105) so instead of exceptioning on build fail (like it used to) new_process.execute will return.

>Won't it just cause uncaught exceptions.

No. Without this patch, the code throws exceptions as well under the same circumstances. The only difference is now they are thrown after we have a chance to save the logs.
The exceptions are caught in https://github.com/google/fuzzbench/blob/master/experiment/build/builder.py

inferno-chromium (approved):
Thanks for explaining!

PR #1032, opened 2021-01-13
Disable lafintel for bug based benchmarks

inferno-chromium:
lafintel seems to not work with sanitizer flags, crashes on
startup.

```
[*] Spinning up the fork server...

[-] Whoops, the target binary crashed suddenly, before receiving any input
    from the fuzzer! There are several probable explanations:
```


jonathanmetzman (approved):
LGTM

PR #1038, opened 2021-01-13
Migrate from deprecated n1-standard-1 to e2-standard-2

inferno-chromium:
- There is no e2-standard-1. We just rely on --cpus=1 flag, and we can decide later if we want to allow all fuzzers to use 2 cpus.
- See go/n1-to-e2-guide

### DIFF
--- a/common/gcloud.py
+++ b/common/gcloud.py
@@ -27,8 +27,12 @@ DISPATCHER_BOOT_DISK_SIZE = '4TB'
 DISPATCHER_BOOT_DISK_TYPE = 'pd-ssd'
 
 # Constants for runner specs.
-RUNNER_MACHINE_TYPE = 'n1-standard-1'
+RUNNER_MACHINE_TYPE = 'e2-standard-2'
 RUNNER_BOOT_DISK_SIZE = '30GB'
+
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
Why are we doing this?
Will this be cheaper even with 2 cores vs 1?
Will we make up for this in quota? Right now we are running an experiment on 6k instances, that won't work with a 10k core quota.
I wonder if we can rely on docker to enforce 1 core, or if having 2 cores on the instance can have some unwanted effects.

inferno-chromium:
> Why are we doing this?

n1-standard-1 is going away. forwarded an email.

> Will this be cheaper even with 2 cores vs 1?

yes e2-standard-2 premptible is almost coming inline with n1-standard-1 (soonish). costs taken care of, this reco is coming from cloud team.

> Will we make up for this in quota? Right now we are running an experiment on 6k instances, that won't work with a 10k core quota.

Yes this is already, see updated quota :)

> I wonder if we can rely on docker to enforce 1 core, or if having 2 cores on the instance can have some unwanted effects.

yes, but no 1-core config anymore.

jonathanmetzman (approved):
LGTM

PR #1039, opened 2021-01-14
Use secret manager to reuse service account keys

jonathanmetzman:
Do this instead of recreating them every experiment.
This should replace a common source of silent issues in FuzzBench.

### DIFF
--- a/docker/dispatcher-image/startup-dispatcher.sh
+++ b/docker/dispatcher-image/startup-dispatcher.sh
@@ -28,15 +28,9 @@ tar -xvzf ${WORK}/src.tar.gz -C ${WORK}/src
 
 # Set up credentials locally as cloud metadata service does not scale.
 credentials_file=${WORK}/creds.json
-iam_account=`gcloud config get-value account`
-gcloud iam service-accounts keys create ${credentials_file} \
-    --iam-account=${iam_account}
+PYTHONPATH=${WORK}/src python3 \
+  ${WORK}/src/experiment/cloud/service_account_key.py $credentials_file $CLOUD_PROJECT
 
 # Start dispatcher.
 PYTHONPATH=${WORK}/src GOOGLE_APPLICATION_CREDENTIALS=${credentials_file} \
     python3 "${WORK}/src/experiment/dispatcher.py"
-
-# Revoke created credentials.
-key_id=$(cat "${credentials_file}" | jq -r ".private_key_id")
-gcloud iam service-accounts keys delete ${key_id} -q \
-    --iam-account=${iam_account}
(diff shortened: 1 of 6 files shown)


experiment/cloud/secret_manager.py
```diff
@@ -0,0 +1,68 @@
+# Copyright 2021 Google LLC
```
jonathanmetzman, line 1: @inferno-chromium I want to start putting stuff in more specific modules than common.
Since there's no way this needs to be used at build time, it only needs to be used during an experiment (as with most cloud-related code, which I will move soon).

inferno-chromium: Sounds good!

docker/dispatcher-image/startup-dispatcher.sh
```diff
@@ -28,15 +28,9 @@ tar -xvzf ${WORK}/src.tar.gz -C ${WORK}/src
 
 # Set up credentials locally as cloud metadata service does not scale.
 credentials_file=${WORK}/creds.json
-iam_account=`gcloud config get-value account`
-gcloud iam service-accounts keys create ${credentials_file} \
-    --iam-account=${iam_account}
+PYTHONPATH=${WORK}/src python3 \
```
jonathanmetzman, line 31: I don't know if we can set GOOGLE_APPLICATION_CREDENTIALS once we've already imported and started using google cloud libraries. So to be safe, I get this key using a script and then start the dispatcher.

inferno-chromium: i think it works, i remember it checks for auth after the first call. but it is fine if you want to do it this way.

jonathanmetzman: First call to an API library? That's problematic because we want to use logging in the key module.
THe only way it would work is if it didn't do any caching of the key at all, so after using the library in the default manner, we could switch to the safe way by setting the env var within the process (this is particularly dangerous if we create any multiprocessing pools on the file level, since those processes wont have any env vars we set after).

inferno-chromium: i mean it happens on google.auth.default() call, not on init.

jonathanmetzman:
Ping

experiment/cloud/secret_manager.py
```diff
@@ -0,0 +1,66 @@
...
+from google.cloud import secretmanager
+
+
+def get_secret_manager_client():
+    """Returns the secretmanager client."""
+    return secretmanager.SecretManagerServiceClient()
+
+
+def get_parent(project):
+    """Creates the parent resource."""
```
inferno-chromium, line 26: Return parent resource ?
maybe get_parent_resource to be more clear.

inferno-chromium (approved):
Thanks a lot!

PR #1045, opened 2021-01-20
Update honggfuzz to the newest version (1ec27b2e3652e6c0c94c28f547226a2dc6009f04)

robertswiecki:


inferno-chromium (approved):

inferno-chromium:
Results should show up in 2021-01-20 experiment which is just started

PR #1051, opened 2021-01-25
Replace terms with more inclusive ones.

jonathanmetzman:


jonathanmetzman:
@laurentsimon is the change to aflcc/fuzzer.py non-breaking?

inferno-chromium (approved):
Thanks!

PR #1055, opened 2021-01-26
[run_experiment] Log more when we fail to create dispatcher.

jonathanmetzman:


experiment/run_experiment.py
```diff
@@ -406,6 +405,7 @@ def start(self):
             '${WORK}/src/experiment/dispatcher.py || '
             '/bin/bash'  # Open shell if experiment fails.
         ]
+        logs.info('Starting dispatcher with container name: %s', container_name)
```
inferno-chromium, line 408: nit: "with instance name" to match other message

common/gcloud.py
```diff
@@ -86,7 +87,13 @@ def create_instance(instance_name: str,
         command.extend(
             ['--metadata-from-file', 'startup-script=' + startup_script])
 
-    return new_process.execute(command, expect_zero=False, **kwargs)[0] == 0
+    result = new_process.execute(command, expect_zero=False, **kwargs)
+    if result.retcode == 0:
+        return True
+
+    logs.info('Failed to create instance. Command: %s failed. Output: %s',
```
inferno-chromium, line 94: logs.error ?

inferno-chromium (approved):

experiment/run_experiment.py
jonathanmetzman: It's in the command.

common/gcloud.py
jonathanmetzman: It's not worth reporting as an error when we are creating runners and there is no quota. So I leave it up to the caller.

PR #1062, opened 2021-01-28
[SymQEMU] Fuzzer Integration

jonathanmetzman:
CC @vanhauser-thc @laurentsimon @sebastianpoeplau

### DIFF
--- a/fuzzers/symqemu/builder.Dockerfile
+++ b/fuzzers/symqemu/builder.Dockerfile
@@ -1,0 +1,104 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+ARG parent_image
+FROM $parent_image
+
+RUN mkdir /symqemu /symcc
+RUN git clone https://github.com/eurecom-s3/symcc --depth 1 /symcc/src
+RUN git clone https://github.com/eurecom-s3/symcc --depth 1 /symqemu/src/
+
+# Download and compile AFL v2.56b.
+# Set AFL_NO_X86 to skip flaky tests.
+RUN git clone https://github.com/google/AFL.git /afl && \
+    cd /afl && \
+    git checkout 8da80951dd7eeeb3e3b5a3bcd36c485045f40274 && \
+    AFL_NO_X86=1 make
+
+# Use afl_driver.cpp from LLVM as our fuzzing library.
+RUN apt-get update && \
+    apt-get install wget -y && \
+    wget https://raw.githubusercontent.com/llvm/llvm-project/5feb80e748924606531ba28c97fe65145c65372e/compiler-rt/lib/fuzzer/afl/afl_driver.cpp -O /afl/afl_driver.cpp && \
+    clang -Wno-pointer-sign -c /afl/llvm_mode/afl-llvm-rt.o.c -I/afl && \
+    clang++ -stdlib=libc++ -std=c++11 -O2 -c /afl/afl_driver.cpp && \
+    ar r /libAFL.a *.o
+
+RUN sed -i -- 's/# deb-src/deb-src/g' /etc/apt/sources.list
+RUN apt-get update && \
+    apt-get install -y cmake cargo ninja-build flex bison && \
+    apt-get build-dep -y --no-install-recommends qemu
+
+ENV CC clang
+ENV CXX clang++
+
+RUN git clone -b z3-4.8.7 https://github.com/Z3Prover/z3.git /z3_src && \
+    cd /z3_src && \
+    mkdir build && \
+    cd build && \
+    cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/z3 .. && \
+    ninja && \
+    ninja install && \
+    cd / && \
+    rm -rf /z3_src
+
+# We need libLLVMSupport, but the version in /usr/local/lib was built against
+# libstdc++ and thus won't work with our libc++ build. We compile a new version
+# that uses libc++.
+RUN mkdir /llvm && \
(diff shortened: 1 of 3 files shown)


vanhauser-thc:
I recommend to switch symcc to https://github.com/vanhauser-thc/symcc because the original fuzzing helper tends to panic fast - and then symcc/symqemu is dead.

jonathanmetzman:
> I recommend to switch symcc to https://github.com/vanhauser-thc/symcc because the original fuzzing helper tends to panic fast - and then symcc/symqemu is dead.

I'll probably do this as I'm not an expert in symqemu. But I'm surprised given that they got some good results in their paper which used their version of sysmcc

fuzzers/symqemu/builder.Dockerfile
```diff
@@ -0,0 +1,105 @@
...
+    ninja && \
+    ninja install && \
+    cd / && \
+    rm -rf /z3_src
+
+# We need libLLVMSupport, but the version in /usr/local/lib was built against
+# libstdc++ and thus won't work with our libc++ build. We compile a new version
+# that uses libc++.
+RUN mkdir /llvm && \
+  git clone -b llvmorg-10.0.0 --depth 1 https://github.com/llvm/llvm-project.git /llvm/src && \
```
inferno-chromium, line 55: nit: indent more by 2 space this block.

vanhauser-thc:
> > I recommend to switch symcc to https://github.com/vanhauser-thc/symcc because the original fuzzing helper tends to panic fast - and then symcc/symqemu is dead.
> 
> I'll probably do this as I'm not an expert in symqemu. But I'm surprised given that they got some good results in their paper which used their version of sysmcc

I dont understand that either.
the problem is the call to afl-showmap. if that times out or something happens than rust issues a panic.

happened to me (https://github.com/eurecom-s3/symcc/issues/44) and richard (https://github.com/eurecom-s3/symcc/issues/45) when we individually tried it out fuzzing campaigns

fuzzers/symqemu/fuzzer.py
```diff
@@ -0,0 +1,159 @@
...
+        '/symcc/build/SymRuntime-prefix/src/SymRuntime-build/libSymRuntime.so',
+        build_directory)
+    print('[post_build] Copying SymQEMU to $OUT directory')
+    shutil.copy('/symqemu/build/x86_64-linux-user/symqemu-x86_64',
+                build_directory)
+    print('[post_build] Copying fuzzing helper to $OUT directory')
+    shutil.copy('/root/.cargo/bin/symcc_fuzzing_helper', build_directory)
+
+
+def run_afl_fuzz(input_corpus, output_corpus, target_binary):
```
inferno-chromium, line 92: Use afl.run_afl_fuzz

inferno-chromium (approved):
Does SymQEMU work on bug based benchmarks, if not, please add to unsupported_fuzzers attribute of bug benchmarks (i see a lot of other qemu fuzzers there).

vanhauser-thc:
also I do not think this implementation works.
the fuzzing helper calls afl-showmap and will use the coverage map for the decision if a new path has been generated or not.
the coverage map from sancov is different to the one qemu_mode/symqemu generates.
so the afl-fuzz instance needs also to be a `-Q` instance.

@laurentsimon @sebastianpoeplau correct?

jonathanmetzman:
> Does SymQEMU work on bug based benchmarks, if not, please add to unsupported_fuzzers attribute of bug benchmarks (i see a lot of other qemu fuzzers there).

I don't think so. Will do.

fuzzers/symqemu/builder.Dockerfile
jonathanmetzman: Done.

vanhauser-thc:
for me building with this branch fails:

```
$ make test-run-symqemu-libjpeg-turbo-07-2017
[...]
#19 72.36 + clang++ -pthread -Wl,--no-as-needed -Wl,-ldl -Wl,-lm -Wno-unused-command-line-argument -stdlib=libc++ -O3 -fsanitize-coverage=trace-pc-guard -std=c++11 /src/libjpeg_turbo_fuzzer.cc -I . .libs/libturbojpeg.a /libAFL.a -o /out/uninstrumented/libjpeg_turbo_fuzzer
#19 73.16 + cp -r /opt/seeds /out/uninstrumented/
#19 73.16 [post_build] Copying afl-fuzz to $OUT directory
#19 73.17 [post_build] Copying afl-showmap to $OUT directory
#19 73.17 [post_build] Copying libz3 to $OUT directory
#19 73.30 [post_build] Copying libSymRuntime to $OUT directory
#19 73.31 [post_build] Copying SymQEMU to $OUT directory
#19 73.32 [post_build] Copying fuzzing helper to $OUT directory
#19 73.32 Traceback (most recent call last):
#19 73.32   File "<string>", line 1, in <module>
#19 73.32   File "/src/fuzzers/symqemu/fuzzer.py", line 89, in build
#19 73.32     shutil.copy('/root/.cargo/bin/symcc_fuzzing_helper', build_directory)
#19 73.32   File "/usr/local/lib/python3.8/shutil.py", line 415, in copy
#19 73.32     copyfile(src, dst, follow_symlinks=follow_symlinks)
#19 73.32   File "/usr/local/lib/python3.8/shutil.py", line 261, in copyfile
#19 73.32     with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
#19 73.32 FileNotFoundError: [Errno 2] No such file or directory: '/root/.cargo/bin/symcc_fuzzing_helper'
------
failed to solve with frontend dockerfile.v0: failed to build LLB: executor failed running [/bin/sh -c PYTHONPATH=$SRC python3 -u -c "from fuzzers import utils; utils.initialize_env(); from fuzzers.$FUZZER import fuzzer; fuzzer.build()"]: runc did not terminate sucessfully
make: *** [docker/generated.mk:35924: .symqemu-libjpeg-turbo-07-2017-builder] Error 1
```

plus - it looks wrong. it is compiling the uninstrumented binary, however the `-fsanitize-coverage=trace-pc-guard` option is set together with /libAFL.a

sebastianpoeplau:
@jonathanmetzman, sorry for the delay! Looks good to me. This is very close to what we used for the paper; I'm not sure what has changed in FuzzBench since, especially regarding compilation flags, but the approach generally looks right to me, and I see you've made some adjustments in the flag setup.

@vanhauser-thc, I've replied in more detail to your qustion about coverage data in your [SymCC issue](https://github.com/eurecom-s3/symcc/issues/49) on the topic. In short, I think there is no problem as long as `afl-showmap` and the AFL process are consistent (i.e., both in QEMU mode or both native, which the fuzzing helper makes sure of); SymQEMU uses its own independent coverage mechanism for branch decisions ([inherited from QSYM](https://github.com/sslab-gatech/qsym/blob/master/qsym/pintool/afl_trace_map.cpp)).

PR #1069, opened 2021-02-02
Add file bug benchmark

wideglide:
Adding file v5.35 as a bug benchmark.  Granted, file may not be a critical piece of open source software, but these four bugs should be the appropriate level of difficulty for Fuzzbench 24hr experiments.  Unfortunately, I needed to modify the fuzzing harness to expose the problematic functions which also makes the benchmark execute more slowly since it needs to write the buffer to a file each iteration.  
I'm intentionally disabling the provided seed corpus as fuzzers will make progress with an empty seed.


inferno-chromium (approved):
Need to add this here as well - https://github.com/google/fuzzbench/blob/master/service/automatic_run_experiment.py#L80, can do in a new pr.

PR #1074, opened 2021-02-04
fixes afl++ optimal and v300c

vanhauser-thc:


inferno-chromium (approved):

vanhauser-thc:
@inferno-chromium I see this coming up in the CI logs:
```
No space left on device | build (aflplusplus_eclipser, oss-fuzz):                     .github#L1
```
could that be (one of) the issue(s)?

inferno-chromium:
> @inferno-chromium I see this coming up in the CI logs:
> 
> ```
> No space left on device | build (aflplusplus_eclipser, oss-fuzz):                     .github#L1
> ```
> 
> could that be (one of) the issue(s)?

Where do you see this.
aflplusplus_eclipser is too slow to complete in 6 hrs, so i am close to giving up and even thinking to remove it from CI.

vanhauser-thc:
@inferno-chromium 
look here, scroll down: https://github.com/google/fuzzbench/actions/runs/538024385

vanhauser-thc:
btw I am fine with removing aflplusplus_eclipser.
it has top results, but the results are not fair :) as it is afl-fuzz plus another running process (eclipser). if I do the same setup with afl-fuzz cmplog + afl-fuzz plain in fuzzbench the results are better, so eclipser is actually not that good, just exploiting spare CPU cycles.

I am also fine with disabling _optimal. it needs re-adjusting to the changes to the targets so its setup is not "optimal" anymore.

PR #1075, opened 2021-02-05
DO NOT MERGE - CI test

vanhauser-thc:


vanhauser-thc:
@inferno-chromium  afl is breaking here ;)

inferno-chromium:
So, aflplusplus_asan is working fine for me locally (so weird, no segfault). I wonder if this clang-12 and gcc use is causing some issues, i never liked this in first place. If we start using latest clang in all benchmarks, then we just use that, instead of installing our own -clang-12, gcc?

vanhauser-thc:
the crash is in afl-fuzz which is compiled solely with gcc. I can add another asan variant and use clang for that. it is "clang" everywhere - right? not clang-12

vanhauser-thc:
ah no I am wrong. only aflplusplus_optimal uses gcc. all other use clang for everything:
```
export CC=clang
```
so that is not it ... but I will add a gcc asan variant now ;)

inferno-chromium:
i tested with 19d8f00963a2eb3526fc442c0b1685c166b68410 locally in aflplusplus_optimal and still see segfault. i hope gcc is the fault and gcc asan shows up issue. 

i was saying about clang-12 as i see all these package installs here, when clang in those benchmarks should be latest (or almost latest since some hardcode base-builder hashes that i want to remove soonish)
```
RUN apt-get update && \
    apt-get install -y wget libstdc++-5-dev libexpat1-dev && \
    apt-get install -y apt-utils apt-transport-https ca-certificates && \
    echo deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-12 main >> /etc/apt/sources.list && \
    echo deb http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu xenial main >> /etc/apt/sources.list && \
    wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
    apt-key adv --recv-keys --keyserver keyserver.ubuntu.com 1E9377A2BA9EF27F && \
    apt-get update && \
    apt-get install -y clang-12 clangd-12 clang-tools-12 libc++1-12 libc++-12-dev \
      libc++abi1-12 libc++abi-12-dev libclang1-12 libclang-12-dev libclang-common-12-dev \
      libclang-cpp12 libclang-cpp12-dev liblld-12 liblld-12-dev liblldb-12 \
      liblldb-12-dev libllvm12 libomp-12-dev libomp5-12 lld-12 lldb-12 \
      llvm-12 llvm-12-dev llvm-12-runtime llvm-12-tools && \
    apt-get install -y gcc-9 g++-9
```

vanhauser-thc:
It cannot have anything to do with gcc. the issues also come up for other variants that are clang only (e.g. dict2file, introspection).

optimal needs an overhaul, not only remove the extra clang installation, but also assessing what is optimal. that is why optimal can be disabled for the time being. it will take a few weeks to bring it to an "optmal" state again, and a) finding out what the issue is with the segfault and b) oss-fuzz integration are more important.

I dont get it who asan is not breaking :) kafkaesque ...

inferno-chromium:
> It cannot have anything to do with gcc. the issues also come up for other variants that are clang only (e.g. dict2file, introspection).
> 
> optimal needs an overhaul, not only remove the extra clang installation, but also assessing what is optimal. that is why optimal can be disabled for the time being. it will take a few weeks to bring it to an "optmal" state again, and a) finding out what the issue is with the segfault and b) oss-fuzz integration are more important.
> 
> I dont get it who asan is not breaking :) kafkaesque ...

Yes will disable these 2 in CI [will add to https://github.com/google/fuzzbench/pull/1076]. Totally agree, OSS-Fuzz integration+optimization is important to show AFL++ impact.

PR #1076, opened 2021-02-06
Spring fuzzer variant cleanup, better docker cleanup.

inferno-chromium:
- Remove 5 old unused temp variants.
- Remove manul from CI.

Needed before we can add more benchmarks to the matrix.


fuzzbench/test_e2e/test_e2e_run.py
```diff
@@ -58,10 +58,8 @@ def test_jobs_dependency(self, experiment_config, redis_connection):  # pylint:
                 for dep in image['depends_on']:
                     assert jobs[dep].ended_at <= jobs[name].started_at
 
-    def test_all_jobs_finished_successfully(
-        self,
-        experiment_config,  # pylint: disable=redefined-outer-name
-        redis_connection):  # pylint: disable=redefined-outer-name
+    def test_all_jobs_finished_successfully(self, experiment_config,
```
inferno-chromium, line 61: crazy lint indent failure, this is only workaround.

.github/workflows/fuzzers.yml
```diff
@@ -21,17 +21,17 @@ jobs:
           - aflcc
           - aflfast
           - aflplusplus
-          - aflplusplus_optimal
-          - aflplusplus_eclipser
+          # Disabled due to segfaults and timeouts.
+          # See https://github.com/google/fuzzbench/issues/1073.
+          # - aflplusplus_optimal
```
vanhauser-thc, line 26: needs also removal in core fuzzers perhaps?

inferno-chromium: Yes done.

service/core-fuzzers.yaml
```diff
@@ -13,4 +12,3 @@ fuzzers:
     - lafintel
     - libfuzzer
     - mopt
-    - neuzz
```
inferno-chromium, line 16: latest neuzz does not work on all benchmarks, see https://www.fuzzbench.com/reports/experimental/2021-02-08-paper/index.html, so removing from core-fuzzers for now.

.github/workflows/fuzzers.yml
```diff
@@ -21,17 +21,17 @@ jobs:
...
+          # - aflplusplus_eclipser
           - aflsmart
           - entropic
           - fairfuzz
-          - fastcgs_lm
           - honggfuzz
           - lafintel
           - klee
           - libfuzzer
-          - manul
```
jonathanmetzman, line 34: Why remove manul? Because it isn't so actively developed?

jonathanmetzman (approved):
LGTM

.github/workflows/fuzzers.yml
inferno-chromium: Yes, and dont remember seeing interesting results. We need to scale benchmarks, so thinking to keep CI lean on what we periodically benchmark against.

Issue #1073, opened 2021-02-04
aflplusplus fuzzer failures


inferno-chromium:
Failure like 

1. "make test-run-aflplusplus_optimal-libhevc_hevc_dec_fuzzer", 
2. "make test-run-aflplusplus_optimal-freetype2-2017" 

are legit and show locally as well [@vanhauser-thc fyi]
There are others at https://github.com/google/fuzzbench/actions/runs/536892982, please do take time to reproduce locally. I am guessing all of this came after using clang-12 package. Can we try building with clang available in base-builder, that should be 12 as well.

inferno-chromium:
Locally this one
3. make test-run-aflplusplus_v300c-freetype2-2017 fails with
```
#19 2.381 configure: error: C compiler cannot create executables
#19 2.381 See `config.log' for more details
#19 2.513 Traceback (most recent call last):
#19 2.513   File "<string>", line 1, in <module>
#19 2.513   File "/src/fuzzers/aflplusplus_v300c/fuzzer.py", line 30, in build
#19 2.513     aflplusplus_fuzzer.build("tracepc", "cmplog", "dict2file")
#19 2.513   File "/src/fuzzers/aflplusplus/fuzzer.py", line 147, in build
#19 2.513     utils.build_benchmark()
#19 2.513   File "/src/fuzzers/utils.py", line 76, in build_benchmark
#19 2.513     subprocess.check_call(['/bin/bash', '-ex', build_script], env=env)
#19 2.513   File "/usr/local/lib/python3.8/subprocess.py", line 364, in check_call
#19 2.513     raise CalledProcessError(retcode, cmd)
#19 2.513 subprocess.CalledProcessError: Command '['/bin/bash', '-ex', '/src/build.sh']' returned non-zero exit status 77.
```

inferno-chromium:
Locally this one reproduces as well
4. make test-run-aflplusplus_cmplog_transform-libhevc_hevc_dec_fuzzer
```
[*] Fuzzing test case #0 (1 total, 0 uniq crashes found, perf_score=100, exec_us=5028, hits=0, map=1490)...
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/src/fuzzers/aflplusplus_cmplog_transform/fuzzer.py", line 35, in fuzz
    aflplusplus_fuzzer.fuzz(input_corpus,
  File "/src/fuzzers/aflplusplus/fuzzer.py", line 208, in fuzz
[...]

PR #1083, opened 2021-02-14
ci test  - do not merge

vanhauser-thc:


vanhauser-thc:
I think I found the issue. `-march=native` seems to generate mis-optimized code for some specific intel CPUs that violate memory and result in a segfault. What is weird is that it happens with both clang and gcc compiler. But removing this compile flag seems to fix it at least for the CI.

I noted another issue btw - only outdated python3 is available in some build images, e.g. for bloaty. qemu does not build with that python3.5. hence aflplusplus_qemu fails for this and other benchmark targets.

inferno-chromium:
> I think I found the issue. `-march=native` seems to generate mis-optimized code for some specific intel CPUs that violate memory and result in a segfault. What is weird is that it happens with both clang and gcc compiler. But removing this compile flag seems to fix it at least for the CI.
> 
> I noted another issue btw - only outdated python3 is available in some build images, e.g. for bloaty. qemu does not build with that python3.5. hence aflplusplus_qemu fails for this and other benchmark targets.

You rock Sir!!! yes fixed my local too.
Can you reenable 
aflplusplus_optimal is not included in fuzzer list in /home/runner/work/fuzzbench/fuzzbench/.github/workflows/fuzzers.yml.
aflplusplus_eclipser is not included in fuzzer list in /home/runner/work/fuzzbench/fuzzbench/.github/workflows/fuzzers.yml.
so that presubmit is fixed. then can merge this (i will approve this).

for those old build images [old python], i do plan to remove build hash hardcode from their dockerfile. i think all build benchmark should use the same oss-fuzz latest base-builder.

inferno-chromium (approved):
Just uncomment this

          # Disabled due to segfaults and timeouts.
          # See https://github.com/google/fuzzbench/issues/1073.	          # See https://github.com/google/fuzzbench/issues/1073.
          # - aflplusplus_optimal	          # - aflplusplus_optimal
          # - aflplusplus_eclipser	          # - aflplusplus_eclipser

vanhauser-thc:
so weird. I have seen march=native breaking cross-system usage, obivously, but same-system ... I suspected that when I saw that your system is a xeon, so it had to be something about the CPU, because it could not be the compiler or kernel or docker version...

vanhauser-thc:
optimal fails, will fix that the next days

PR #1084, opened 2021-02-15
fix afl++ optimal

vanhauser-thc:


inferno-chromium (approved):

vanhauser-thc:
that was too early, there is still one that did not work - systemd ... it too has only clang-10

PR #1087, opened 2021-02-15
final workaround for optimal

vanhauser-thc:
systemd only has llvm 10


vanhauser-thc:
aflplusplus_optimal fails for bug php_execute because: `#19 492.0 /usr/bin/ld: final link failed: No space left on device` :)
it worked yesterday so that is fine :)  
waiting for optimal oss-fuzz to complete ...

jonathanmetzman (approved):
lgtm

PR #1088, opened 2021-02-15
Remove shallow bug benchmarks, remove hardcoding of benchmark lists.

inferno-chromium:
- Remove hardcoding of benchmark lists in experiment service, ci.
  php benchmark works in prod now.
- Remove super shallow bug benchmarks, where bugs usually hit in
  a few seconds. E.g. see https://www.fuzzbench.com/reports/2020-12-19-bug/index.html


inferno-chromium:
This is needed to add new bug based benchmark suggestions easily from Marcel.

.github/workflows/build_and_test_run_fuzzer_benchmarks.py
```diff
@@ -170,8 +129,8 @@ def do_build(build_type, fuzzer, always_build):
         return make_builds(benchmarks, fuzzer)
 
     # Otherwise, only build benchmarks that have changed.
-    changed_benchmarks = set(change_utils.get_changed_benchmarks(changed_files))
-    benchmarks = benchmarks.intersection(changed_benchmarks)
+    changed_benchmarks = change_utils.get_changed_benchmarks(changed_files)
+    benchmarks = set(benchmarks).intersection(set(changed_benchmarks))
```
jonathanmetzman, line 133: `set(changed_benchmarks)` isn't needed, `changed_benchmarks` is sufficient. You can pass a list to intersection.

service/automatic_run_experiment.py
```diff
@@ -46,53 +46,6 @@
...
-    'bloaty_fuzz_target',
-    'curl_curl_fuzzer_http',
-    'jsoncpp_jsoncpp_fuzzer',
-    'libpcap_fuzz_both',
-    'libxslt_xpath',
-    'mbedtls_fuzz_dtlsclient',
-    'openssl_x509',
-
-    # Temporarily disabled due to out-of-memory issues with clang coverage.
-    # 'php_php-fuzz-parser',
```
jonathanmetzman, line 62: Your new change means we are going to use `php_php-fuzz-parser` again in automatic experiments.
Do we have reason to think it will work now?

.github/workflows/build_and_test_run_fuzzer_benchmarks.py
inferno-chromium: Done!

service/automatic_run_experiment.py
inferno-chromium: Yes it works fine for some time now
https://www.fuzzbench.com/reports/2021-02-13-paper/index.html#php_php-fuzz-parser
I think it was that num-threads=1 change that Bharat did some time back.

jonathanmetzman (approved):
LGTM

Issue #1086, opened 2021-02-15
Documentation issue: docs/developing-fuzzbench/custom_analysis_and_reports.md


temp1337temp1337:
I was trying to generate a simple report, following the command in the related documented. However, upon executing the command: 
```
mkdir ~/my-report; cd ~/my-report
wget https://www.fuzzbench.com/reports/sample/data.csv.gz
PYTHONPATH=<fuzzbench_root> python3 analysis/generate_report.py \
  [experiment_name] \
  --report-dir ~/my-report \
  --from-cached-data
```

I encountered the following issues:
- The postgresql service needs to be up and running and the postgresql password needs to be set. I am not sure if this is mentioned elsewhere though.
- The file provided for the `data.csv.gz` is missing the `crash_key` column, which is one of the expected ones; at the `data_utils.validate_data(experiment_df)` of the `generate_report.py`.

Eventually I executed the command as:
```
POSTGRES_PASSWORD=<password> PYTHONPATH=fuzzbench python3 analysis/generate_report.py \
  test-experiment \
  --report-dir ~/my-report \
  --from-cached-data
```

Upon adding a dummy column on the `data.csv.gz` and setting the password for the `postgres` user. Could you please update that documentation file?

Finally, there is a small mistake in the same document. The line `--report_type experimental` should be `--report-type experimental`.

Thank you.

Issue #1078, opened 2021-02-07
Feature wish: time graph over all benchmarks


vanhauser-thc:
Every individual benchmark has a graph  that shows the coverage from 0:15 until 23:00, e.g.
![graph](https://www.fuzzbench.com/reports/2021-01-31-jigsaw/curl_curl_fuzzer_http_coverage_growth.svg)

It would be nice to have a the same kind of graph for all targets combined.
this would show fuzzers who are good at startup with not many seeds, those who who have a strong mid/end game strategy.

I dont think this is that helpful in the summary section so either in the summary to fold out or at the bottom?

jonathanmetzman:
> Every individual benchmark has a graph that shows the coverage from 0:15 until 23:00, e.g.
> ![graph](https://camo.githubusercontent.com/1d56cb1c31adf4cede5d8f432f7342a4cb8097bdbbb9b37a0927af9d654da6cc/68747470733a2f2f7777772e66757a7a62656e63682e636f6d2f7265706f7274732f323032312d30312d33312d6a69677361772f6375726c5f6375726c5f66757a7a65725f687474705f636f7665726167655f67726f7774682e737667)
> 
> It would be nice to have a the same kind of graph for all targets combined.
> this would show fuzzers who are good at startup with not many seeds, those who who have a strong mid/end game strategy.
> 
> I dont think this is that helpful in the summary section so either in the summary to fold out or at the bottom?

I like the idea but it might require some thought how to do this.
Right now the graph is a median/mean of trials. 
The naive approach would be to just combine trials from each benchmark, but that will make the graph hard to understand/interpret.

vanhauser-thc:
I agree that this graph might not be easy to understand.

Thats why I propose to make it available but not prominently - either folded away, or at the bottom.

how to do the graph is I think easy - the average of the % coverage per benchmark target for that specific time
 the issue is rather that the benchmarks do not start at the time, so intermediate results are a bit weird - but that is true for the ranking as well, it fluctuates a lot.
at the end of the experiment its fine though.

PR #1090, opened 2021-02-16
Fuzzer report and new summary table

wideglide:
-see commit notes

two examples attached:
- [report-2021-01-30.tar.gz](https://github.com/google/fuzzbench/files/5987574/report-2021-01-30.tar.gz)
- [report-2021-01-31-fuzzer.tar.gz](https://github.com/google/fuzzbench/files/5987584/report-2021-01-31-fuzzer.tar.gz)

### DIFF
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Utility functions for data (frame) transformations."""
+import pandas as pd
 from analysis import stat_tests
 from common import environment
 
@@ -34,7 +35,6 @@ def validate_data(experiment_df):
         'time_ended',
         'time',
         'edges_covered',
-        'crash_key',
     }
     missing_columns = expected_columns.difference(experiment_df.columns)
     if missing_columns:
@@ -114,6 +114,9 @@ def filter_max_time(experiment_df, max_time):
 def add_bugs_covered_column(experiment_df):
     """Return a modified experiment df in which adds a |bugs_covered| column,
     a cumulative count of bugs covered over time."""
+    if 'crash_key' not in experiment_df:
+        experiment_df['bugs_covered'] = 0
+        return experiment_df
     grouping1 = ['fuzzer', 'benchmark', 'trial_id', 'crash_key']
     grouping2 = ['fuzzer', 'benchmark', 'trial_id']
     grouping3 = ['fuzzer', 'benchmark', 'trial_id', 'time']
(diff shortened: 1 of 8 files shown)


lszekeres:
Thanks a lot! 

It looks like this is two relatively independent changes: ie. (1) the new experiment summary table, and (2) the new fuzzer oriented template. Would it make sense to have these two in separate PRs and get them reviewed separately?

wideglide:
@lszekeres The changes are not really independent because I use the new summary table in the fuzzer template report.  However, switching to the the new summary table in the default report can easily be overridden.  I didn't actually remove the old summary table, I just added a new `percent_summary_table` method in `experiment_results.py`.  The change to the default report template is very minor: only calling the new summary table method.

Also this partially addresses #1086 so that reports can be generated from older `data.csv.gz` files where `crash_key` is not present.

lszekeres (commented):
- I like the new experiment summary table with the colors and percentages.

- I'm not sure how to use the new fuzzer-based report though. The plots seem to compare a single fuzzer's results on different benchmarks (eg. how code coverage grows on different benchmarks with afl). Even if coverage is in % of max, I'm not sure why'd one compare results between different benchmarks, given these benchmark programs can have very different characteristics, so it's like comparing apples to oranges. Could you say a bit more about the motivation and how you use this report type?

wideglide:
In reviewing results on previous experiments using the default report, I find it challenging to really get a wholistic view of how a fuzzer performs across benchmarks.  I found my self scrolling between the benchmarks trying to remember relative performance on different benchmarks.  I think the new summary table helps to address that weakness significantly, but a report that is oriented on fuzzers (instead of benchmarks) allows for a snapshot of each fuzzer across all benchmarks in pictorial and textual form.  It is easier for someone to identify where a fuzzer is strong and where it might be weak.  


Of the new charts, I find the "Trial Ranks" chart to be use most useful, but maybe not quite intuitive enough.  I might need to add some additional documentation/ text to aide comprehension. 
- The "Rank" text shows the ranking of the fuzzer on each benchmark with the average rank listed at the top. This allows me to see the correlation of why each fuzzer achieved the overall rank that it did.  
- "Rank" itself is weak measure of central tendency, so the box plots of trial ranking distribution shows a visual depiction of whether the rank achieved is significant.
    - Looking at the chart below, AFLplusplus ranked 1st in [curl, harfbuzz, libxml2, and freetype2] and the ranking for those benchmarks is significant because all of the individual trial rankings are above the 90th percentile.
    - For benchmarks like re2, you can see that AFLplusplus still ranked 1st but that ranking is not significant because the trial ranks are very spread out.
    - Likewise, on libfuzzer's chart, you can tell that it did very poorly on at least 4 benchmarks as those individual trial ranks are less than the 10th percentile.  
- The _percent of max_ and _fuzzer consistency_ are less helpful as they tend to show more about the characteristics of each benchmark more then the fuzzer, but if you were interested in how one fuzzer compared to the best performance in the experiment across all benchmarks you could see that in the _percent of max_ chart.
- The code coverage graph is probably my least favorite graph, and I'm not sure if it is worth keeping, but I saw the interest (#1078) in a comprehensive coverage graph and attempted to generate one.  I think if you added all the coverages together, the resulting graph would be dominated by one or two benchmarks, and you wouldn't really know which ones where a factor.  

image

wideglide:
One more note, because the report is organized by fuzzer in rank ascending order, you are likely to want to compare one fuzzer to the fuzzer either directly above or below it, which should mean less scrolling to find the data you are looking for if you want to compare charts.

lszekeres:
Thanks for the detailed explanation, and sorry for the slow response!

> Of the new charts, I find the "Trial Ranks" chart to be use most useful, but maybe not quite intuitive enough. I might need to add some additional documentation/ text to aide comprehension.

Yes, this plot can be useful, I agree. Yes, adding a detailed description how to read them under the plots would help a lot.

> * The _percent of max_ and _fuzzer consistency_ are less helpful as they tend to show more about the characteristics of each benchmark more then the fuzzer, but if you were interested in how one fuzzer compared to the best performance in the experiment across all benchmarks you could see that in the _percent of max_ chart.

Right, I'm worried these might give more confusion than clarity. They also seem a bit redundant with the first plot (trial ranks). I'd drop these for now.

> * The code coverage graph is probably my least favorite graph, and I'm not sure if it is worth keeping, but I saw the interest (#1078) in a comprehensive coverage graph and attempted to generate one.  I think if you added all the coverages together, the resulting graph would be dominated by one or two benchmarks, and you wouldn't really know which ones where a factor.

Yes, I find this very confusing as it compares apples to oranges (benchmarks). I'd drop this for now too.

Consider starting with the Trial Ranking Distribution chart only in this PR, and explaining in detail how to read it in the report.

wideglide:
I went ahead and split this into two separate PRs as you originally suggested.   I'll take another look at how to make this report more intuitive and useful.

PR #1097, opened 2021-02-25
Evaluation of a prototype for collision-free context sensitivity

andreafioraldi:
Hi to all!
I'm pushing 4 temp AFL++ variants and one experiment request to test a prototype for collision-free context sensitivity.
It works only with LLVM 10 atm, but if the results will be good, we will do the effort to write a real implementation and not just a prototype.


andreafioraldi:
I also remove afl_buf as the experiment to test it showed that such feedback (minimizing the distance of each mem access between the start and the end of the corresponding buffer) is decremental.

andreafioraldi:
Static analysis, as usual, is broken :-P
I'll try to figure our the set of benchmarks that fail to compile (aspell and bloaty atm for sure) and mark pcguard ctx as an incompatible fuzzer. I'm interested to see how it performs on programs that compile successfully and only then maybe spend time developing the static analysis if the results are promising.

andreafioraldi:
Super strange, on my machine `make debug-aflplusplus_classic_ctx-bloaty_fuzz_target` succeed and even fuzz.

inferno-chromium:
> Super strange, on my machine `make debug-aflplusplus_classic_ctx-bloaty_fuzz_target` succeed and even fuzz.

Click Details in the CI failures, e.g. one failure was on `test-run-aflplusplus_classic_ctx-aspell_aspell_fuzzer` (bug based benchmark)

inferno-chromium:
main thing is aflplusplus_pcguard breaking everywhere. bug based benchmark failure can be ignored if you dont want to run those, only coverage-based ones.

andreafioraldi:
Yes I'm looking at the details of the CI. The issue is that I cannot reproduce the failure on mu local fuzzbench. For instance, https://github.com/google/fuzzbench/pull/1097/checks?check_run_id=1978957602 fails with a strange apt error, while on my machine make test-run-aflplusplus_pcguard-stb_stbi_read_fuzzer runs fine

andreafioraldi:
Ok seems that the problem with ctx_pcguard is the there is no enough RAM for pointer analysis and that some configure scripts are broken when using gllvm (that, for ppl that will try to use gllvm on fuzzbench in the future, can be solved setting WLLVM_CONFIGURE_ONLY=1 while running configure, but cannot be controlled from fuzzer.py atm).

I'll patch fuzzbench and run a local experiment for this so I close this PR.
Maybe consider splitting configure and make for the benchmarks so fuzzers can set different envs for configure and build.

PR #1102, opened 2021-03-03
Switch from e2 to n1 pre-emptibles due to cloud quota issues.

inferno-chromium:
Also, this ensures that intel is used always.


inferno-chromium:
tbr

jonathanmetzman (commented):
LGTM

vanhauser-thc:
I think the current runs need to be restarted?
The tests are now up since 8+ hours, but are not progressing.

PR #1103, opened 2021-03-04
remove php_php-fuzz-parser as it is hanging measurer.

inferno-chromium:
previously this benchmark was excluded, but got reenabled
when hardcoding was removed in
https://github.com/google/fuzzbench/commit/23de3782d98ed55c0a26c5aa4bd48be597c4eab8
remove it for now to stabilize system first.

PR #1105, opened 2021-03-06
Revert "remove php_php-fuzz-parser as it is hanging measurer."

inferno-chromium:
Reverts google/fuzzbench#1103

This was not a culprit.


jonathanmetzman:
LGTM

PR #1107, opened 2021-03-08
Restore intended behavior of run_experiment.py and omitting

jonathanmetzman:
benchmarks.

### DIFF
--- a/experiment/run_experiment.py
+++ b/experiment/run_experiment.py
@@ -473,14 +473,18 @@ def main():
         'more benchmarks.')
 
     all_benchmarks = benchmark_utils.get_all_benchmarks()
-    all_fuzzers = fuzzer_utils.get_fuzzer_names()
-
+    code_coverage_benchmarks = [
+        benchmark for benchmark in all_benchmarks if benchmark_utils.get_type(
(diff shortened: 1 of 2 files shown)


experiment/run_experiment.py
```diff
@@ -473,14 +473,18 @@ def main():
         'more benchmarks.')
 
     all_benchmarks = benchmark_utils.get_all_benchmarks()
-    all_fuzzers = fuzzer_utils.get_fuzzer_names()
-
+    code_coverage_benchmarks = [
```
inferno-chromium, line 476: This should be a helper function in benchmark_utils, just like one that just adds these two - get_oss_fuzz_coverage_benchmarks(),get_standard_coverage_benchmarks().

experiment/run_experiment.py
```diff
@@ -494,6 +498,9 @@ def main():
                         '--description',
                         help='Description of the experiment.',
                         required=False)
+
+    all_fuzzers = fuzzer_utils.get_fuzzer_names()
```
inferno-chromium, line 502: nit: can remove newline after this line.

inferno-chromium (approved):

inferno-chromium (approved):
Thanks!

PR #1124, opened 2021-03-31
afl++: ctx collision free experiment

andreafioraldi:
This PR is a followup to #1097.

We fixed the problems with the builds of some benchmarks and now I request a coverage experiment.


andreafioraldi:
Seems that just bug and oss-fuzz benchmarks are failing, perhaps some because the linker was killed by the kernel. Linking can take up to 16 GB of ram and I don't think the CI has that amount of memory, but I think is fine as I requested a standard experiment no?

inferno-chromium (approved):

PR #1129, opened 2021-04-12
Fix issue with query calls not properly scoped.

inferno-chromium:
Query calls can be followed by join/filter sub-calls which were
not caught for exceptions and hence session rollback did not
occur. This caught furthur calls to exception out on
"Can't reconnect until invalid transaction is rolled back"

### DIFF
--- a/database/utils.py
+++ b/database/utils.py
@@ -15,9 +15,9 @@
 
 import os
 import threading
+from contextlib import contextmanager
 
 import sqlalchemy
-from sqlalchemy import orm
 
 # pylint: disable=invalid-name,no-member
 engine = None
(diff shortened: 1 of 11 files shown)


database/utils.py
```diff
@@ -87,48 +64,46 @@ def cleanup():
...
-@use_session
-def bulk_save(entities):
+def add_all(entities):
     """Save all |entities| to the database connected to by session."""
-    try:
-        session.bulk_save_objects(entities)
-        session.commit()
-        return
-    except Exception as e:
-        session.rollback()
```
jonathanmetzman, line 110: I'm somewhat worried we are getting rid of all of these rollbacks. Why are doing this?

jonathanmetzman (commented):
Much of this seems like a good change to me. The global session stuff was poorly designed and it's good we are getting rid of it.
But this PR doesn't add more rollbacks. In fact it seems to remove the number of SQL operations we can rollback.
Can we add more of these?

database/utils.py
inferno-chromium: Also, we are not fixing global session problem atm, right now it is trying to fix the missing rollbacks first. so session_scope gives all the missing rollbacks.

jonathanmetzman: Ah sorry misread this PR then.
`session_scope` isn't the best name then for what it currently does, but it's fine I guess.

jonathanmetzman (approved):
LGTM

database/utils.py
inferno-chromium: in next pr, i can do the actual throwing away of session in session scope, right now trying to stabilize things first, esp for 7 day exp.

PR #1132, opened 2021-04-16
Add CLI sugar to support reproducing experiments using a few commands

jonathanmetzman:
Support reproducing experiments from its config file.
Instruct users on how to do this in each report.
Also update python to 3.9 (because of a glinux issues, update packages that needed to be updated because of the python update, this included pylint, which threw new warnings that this commit also fixes).


analysis/report_templates/default.html
```diff
@@ -388,6 +388,18 @@ <h2 class="green-text text-darken-3">experiment data</h2>
             <br><br>
             The experiment was conducted using this FuzzBench commit:
             <a href="https://github.com/google/fuzzbench/commits/{{ experiment.git_hash }}">{{ experiment.git_hash }}</a>
+
+            <br><br>
+            To reproduce this experiment run the following commands in your FuzzBench repo:</a><br>
+            <code>
+            git checkout {{ experiment.git_hash }}  # Check out the right commit. <br>
+            # Download the internal config file. <br>
```
inferno-chromium, line 396: does <br> work inside <code>

analysis/report_templates/default.html
```diff
@@ -388,6 +388,18 @@ <h2 class="green-text text-darken-3">experiment data</h2>
             <br><br>
             The experiment was conducted using this FuzzBench commit:
             <a href="https://github.com/google/fuzzbench/commits/{{ experiment.git_hash }}">{{ experiment.git_hash }}</a>
+
+            <br><br>
+            To reproduce this experiment run the following commands in your FuzzBench repo:</a><br>
+            <code>
+            git checkout {{ experiment.git_hash }}  # Check out the right commit. <br>
```
inferno-chromium, line 395: move "# Check out the right commit." to previous line to match others.

experiment/dispatcher.py
```diff
@@ -195,7 +195,7 @@ def main():
     except Exception as error:
         logs.error('Error conducting experiment.')
         raise error
-    experiment_config_file_path = os.path.join(_get_config_dir(),
+    experiment_config_file_path = os.path.join(_get_config_path(),
```
inferno-chromium, line 198: this looks wrong, shouldnt it be just just _get_config_path() and not experiment.yaml

common/experiment_utils.py
```diff
@@ -23,6 +23,12 @@
 CONFIG_DIR = 'config'
 
 
+def get_internal_experiment_config_relative_path():
```
inferno-chromium, line 26: is internal important in name ?

experiment/run_experiment.py
```diff
@@ -193,8 +192,9 @@ def validate_experiment_name(experiment_name: str):
 def set_up_experiment_config_file(config):
     """Set up the config file that will actually be used in the
     experiment (not the one given to run_experiment.py)."""
-    filesystem.recreate_directory(CONFIG_DIR)
-    experiment_config_filename = os.path.join(CONFIG_DIR, 'experiment.yaml')
+    filesystem.recreate_directory(experiment_utils.CONFIG_DIR)
+    experiment_config_filename = os.path.join(experiment_utils.CONFIG_DIR,
```
inferno-chromium, line 196: why not the helper function ?

inferno-chromium (approved):

analysis/report_templates/default.html
jonathanmetzman: Done.

analysis/report_templates/default.html
jonathanmetzman: `<br>` you mean? Yes I tested it

common/experiment_utils.py
jonathanmetzman: I think so because I want to distinguish between the internal config file and the external one, which is different.

experiment/dispatcher.py
jonathanmetzman: Fixed.

experiment/run_experiment.py
jonathanmetzman: Done.

PR #1136, opened 2021-04-27
Integrating the benchmark "matio" with target "matio_fuzzer", commit date "2019-09-11 02:28:00+00:00"

Vaush:

There are 45 known bugs. We could reproduce 32 of them.


Vaush:
I didn't blacklist aflfast, hoping it's a fluke. Let's see.

PR #1137, opened 2021-04-27
Integrating the benchmark "arrow" with target "parquet-arrow-fuzz", commit date "2020-02-25 02:36:00+00:00"

Vaush:

There are 34known bugs. We could reproduce 29 of them.


jonathanmetzman (approved):
LGTM

PR #1138, opened 2021-04-29
frida + qemu experiment

vanhauser-thc:
tests the new frida_mode


vanhauser-thc:
this will need a fix in frida first... :(

PR #1145, opened 2021-05-10
Integrating the benchmark "ffmpeg" with target "ffmpeg_DEMUXER_fuzzer", commit date "2019-06-14 01:51:00+00:00

Vaush:

There are 37 known bugs. We could reproduce 35 of them.


Vaush:
@inferno-chromium you might notice I kept the SHA in the dockerfile here. When removed, the benchmark doesn't build anymore, complaining about an error with clang-12 (try it with afl if you want)

Vaush:
Sorry, I used the wrong commit message, it should be "Blacklisted aflplusplus_qemu_inmem". I am not sure if there's any way to fix it now though

PR #1148, opened 2021-05-10
Integrating the benchmark "wireshark" with target "fuzzshark_ip", commit date "2018-10-07 01:19:00+00:00"

Vaush:
Benchmark chosen based on the number of days ellapsed between each bug's introduction in the code and its discovery by OSS-Fuzz.

There are 10 known bugs. We could reproduce 8 of them.


Vaush:
@jonathanmetzman @inferno-chromium I am waiting to blacklist fuzzers in this one, since as I mentioned the issue is the aflplusplus runs out of disk space, it needs 38GBs. Is there anyway to make it work, or should I just go ahead and blacklist it?

PR #1154, opened 2021-05-21
Experiment for AFL++ cloning strategies

andreafioraldi:
I'd like to benchmark different function cloning strategies that we are developing to achieve context sensitivity at compile time


andreafioraldi:
The linker is continuosly killed, problably OOM, that's why the CI is failing

inferno-chromium (approved):

andreafioraldi:
Hey @inferno-chromium what's the status of this experiment? Estimated time of start?

PR #1157, opened 2021-05-24
[reports] Speed up report generation significantly

jonathanmetzman:
Reduce report generation time from minutes to ~20 seconds.
These measures which rely on multiprocessing won't be used on the dispatcher to avoid overloading them. But they should help local report generation a lot.

1. For differential coverage, use multiprocessing if requested  to download coverage data from GCS.
2. For report plots, use multiprocessing to generate plots and cache them in parallel so that rendering with jinja is instantaneous.


analysis/rendering.py
```diff
@@ -21,8 +21,54 @@
...
+    # Plotting each benchmark's plots in parallel seems to be faster than
+    # plotting each plot in parallel. I suspect this is because each call to
+    # _warm_benchmark_plot_cache requires copying a benchmark, which has some
+    # overhead.
+    arguments = []
+    for benchmark in experiment_results.benchmarks:
+        attrs = []
+        if benchmark.type == 'bug':
+            attrs.extend([
+                'bug_box_plot',
```
lszekeres, line 45: Could we somehow avoid having a redundant list of functions here, and figure them out automatically? One idea that comes to mind is to call the jinja render() with some fake/mock environment first which just collects the list of these functions.

jonathanmetzman: Umm... I agree having redundant functions is quite ugly. But I'm not sure this solution is better. I guess it is less likely to break. But it just feels a bit icky to use mock in production. Need to think about this more.

inferno-chromium:
@jonathanmetzman - this would be nice to get landed, fast report generation is awesome. i am fine with some repetitions.

PR #1158, opened 2021-05-25
symcc_afl: improve build and enable more benchmarks

DavidKorczynski:
This PR changes the set up for symcc_afl in the following ways:
1) Improves the build so it's compatible with more benchmarks.
2) Changes the build to use a pre-compiled version of z3 instead of compiling it from source.
3) Fixes the commit from which symcc will be cloned.

The PR also requests a new experiment.


DavidKorczynski:
This one is good to go @inferno-chromium @jonathanmetzman @laurentsimon  - the CI might fail due to out-of-disk issues, however, I have run symcc_afl on all of the benchmarks locally and verified that it runs on the enabled ones.

inferno-chromium (approved):

PR #1155, opened 2021-05-21
afl++ release test

vanhauser-thc:
new benchmark to check for regressions for a new release

PR #1160, opened 2021-05-26
Update google-* libs due to version conflicts.

inferno-chromium:
See
https://stackoverflow.com/questions/67652313/python-app-hosted-on-app-engine-is-not-able-to-connect-with-firestore-where-as-w
and issue pointed in
https://github.com/google/fuzzbench/pull/1155#issuecomment-848422635


inferno-chromium:
TBR

jonathanmetzman:
LGTM.
Thanks!

PR #1163, opened 2021-05-29
afl++ variants for saturated tests

vanhauser-thc:
variants for special run with saturated corpus.


vanhauser-thc:
still the bug benchmark is the issue so can be ignored

inferno-chromium (approved):

PR #1164, opened 2021-05-30
add experiment request

johamb:
I have not added a new fuzzer. 
I would like to request an experiment to compare afl, honggfuzz and libfuzzer.

### DIFF
--- a/service/experiment-requests.yaml
+++ b/service/experiment-requests.yaml
@@ -18,6 +18,16 @@
 #
 # You can run "make presubmit" to do basic validation on this file.
 # Please add new experiment requests towards the top of this file.
+
+- experiment: 2021-06-02
+  description: "Experiment to compare afl with honggfuzz and libfuzzer"
+  fuzzers:
+    - afl
+    - aflplusplus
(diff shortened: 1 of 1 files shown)


inferno-chromium:
> I have not added a new fuzzer.
> I would like to request an experiment to compare afl, honggfuzz and libfuzzer.

Do results from recent experiment https://www.fuzzbench.com/reports/experimental/2021-05-29-symccafl/index.html suffice?

johamb:
> Do results from recent experiment https://www.fuzzbench.com/reports/experimental/2021-05-29-symccafl/index.html suffice?

Hi, since I plan to use the results in my studies, I was hoping to schedule an experiment with exactly these fuzzers so that I can use the report and data directly. But I realize that the results will be the same as the ones from previous experiments with these fuzzers. 
So I will leave it up to you to decide if such an experiment is worth it. If you want to accept my pr to schedule the experiment, I would be delighted. But otherwise I can gather the data I need from previous reports.

inferno-chromium (approved):

PR #1182, opened 2021-06-12
Test afl++ fixes

vanhauser-thc:
This run tests three fixes:
  - crash in cmplog during runs
  - cmplog not fully working in frida mode
  - cmplog not fully working in source mode (found due klee presentation)
plus a new run of function + loop counting


vanhauser-thc:
the fails are "out of disk space" :)
even afl_frida_old fails which is the same commit from 3 weeks ago.
can be merged :)

inferno-chromium (approved):

Issue #946, opened 2020-11-09
Local experiment issue : ERRO[0612] error waiting for container: unexpected EOF - [SOLVED]


Microsvuln:
Hi.

I have a problem running local experiments, I get the following error while building benchmarks after running this :

`(.venv) arash@fuzzbench-scale1:~/new/fuzzbench$ PYTHONPATH=. python3 experiment/run_experiment.py --experiment-config experiment-config.yaml --benchmarks bloaty_fuzz_target harfbuzz-1.3.2 libjpeg-turbo-07-2017 libpcap_fuzz_both libpng-1.2.56 libxml2-v2.9.2 --experiment-name $EXPERIMENT_NAME --fuzzers afl aflql`

And the error log (summarized) :

```
INFO:root:Building using (<function build_measurer at 0x7f41267d5430>): [('bloaty_fuzz_target',), ('harfbuzz-1.3.2',), ('libjpeg-turbo-07-2017',), ('libpcap_fuzz_both',), ('libpng-1.2.56',), ('libxml2-v2.9.2',)]
INFO:root:Building measurer for benchmark: bloaty_fuzz_target.
INFO:root:Building measurer for benchmark: harfbuzz-1.3.2.
INFO:root:Building measurer for benchmark: libjpeg-turbo-07-2017.
INFO:root:Building measurer for benchmark: libpcap_fuzz_both.
INFO:root:Building measurer for benchmark: libpng-1.2.56.
INFO:root:Building measurer for benchmark: libxml2-v2.9.2.
INFO:root:Done building measurer for benchmark: libpcap_fuzz_both.
INFO:root:Done building measurer for benchmark: bloaty_fuzz_target.
INFO:root:Done building measurer for benchmark: libpng-1.2.56.
INFO:root:Done building measurer for benchmark: libjpeg-turbo-07-2017.
INFO:root:Done building measurer for benchmark: harfbuzz-1.3.2.
INFO:root:Done building measurer for benchmark: libxml2-v2.9.2.
INFO:root:Build successes: [('bloaty_fuzz_target',), ('harfbuzz-1.3.2',), ('libjpeg-turbo-07-2017',), ('libpcap_fuzz_both',), ('libpng-1.2.56',), ('libxml2-v2.9.2',)]
INFO:root:Done building measurers.
INFO:root:Building all fuzzer benchmarks.
[...]

PR #1188, opened 2021-06-24
[HOTFIX] Remove entrypoint from builder dockerfile

andreafioraldi:
My last patch conflicts with the experiments run.
Fuzzbench expects to overwrite the command of the builder when starting an experiment, and this is impossible when there is an ENTRYPOINT in the dockerfile.
This should be the reason of the failed libafl experiment.


andreafioraldi:
The debug-builder works, I have troubles with local experiments (`ERRO[0094] error waiting for container: unexpected EOF` #946  even with an experiment of 3 fuzzers and 2 benchs having a VM with 100gb of ram and 100 cpus) so I cannot confirm the fix

jonathanmetzman (approved):
LGTM

Issue #1192, opened 2021-07-03
symcc + libxml2


vanhauser-thc:
the results of last symbolic run shows that symcc has issues with various targets although they compile and start running fine.
but at some late point there is no data anymore, e.g.:

gs://fuzzbench-data/2021-06-27-symbolic/experiment-folders/libxml2-v2.9.2-symcc_aflplusplus_single/trial-1462846/corpus/corpus-archive-0089.tar.gz

the corresponding result log `gs://fuzzbench-data/2021-06-27-symbolic/experiment-folders/libxml2-v2.9.2-symcc_aflplusplus_single/trial-1462846/results/fuzzer-log.txt` does not show any issues though.

is there somewhere more data what the issue is? I would think that OOM etc. would show up in the log.

PR #1193, opened 2021-07-03
new symbolic solver benchmark

vanhauser-thc:
only run symbolic benchmarks on targets that works for all solvers.
disabled dictionaries, synced afl++ options and commit.
symcc runtime issues (OOM?) are present, but everything else will be comparable (#1192)


DavidKorczynski:
@googlebot I consent.

vanhauser-thc:
the fails are fine and not concerning the run, can be merged :)

PR #1199, opened 2021-07-08
compare two different afl++ customized versions (dd and no dd)

elManto:
This experiment compares two different instrumentations on top of afl++


elManto:
@googlebot I signed it!

inferno-chromium:
Currently fuzzbench infra is running several experiments, we will be able to do this sometime next week.

jonathanmetzman (approved):
Thanks for waiting. I'll merge this if CI passes.

elManto:
The CI killed the bench on the bug but I verified locally that all the projects build

PR #1202, opened 2021-07-09
Experiment for AFL++ cloning strategies using different pointer analyses

pietroborrello:
I'd like to request an experiment to compare different strategies to achieve context sensitivity at compile time using different pointer analyses


pietroborrello:
@googlebot I signed it!

andreafioraldi:
@googlebot I consent.

jonathanmetzman:
Hi,
Thanks for this PR.
Unfortunately, we need the cores to run our own experiments until next week, so we can't run this experiment yet. I'll merge this when we have free cores/attention to accept this.

pietroborrello:
It seems the linker is continuosly killed, problably OOM, that's why the CI is failing on https://github.com/google/fuzzbench/actions/runs/1027102495 for our fuzzers. 
If there is a way to detect when running CI, we may set a lower number of clones to avoid filling the (somehow) limited Travis memory. However, some pointer analysis may still require a lot of memory (e.g. 8/16GB, not sure what are the limits in Travis)

jonathanmetzman:
> It seems the linker is continuosly killed, problably OOM, that's why the CI is failing on https://github.com/google/fuzzbench/actions/runs/1027102495 for our fuzzers.
> If there is a way to detect when running CI, we may set a lower number of clones to avoid filling the (somehow) limited Travis memory. However, some pointer analysis may still require a lot of memory (e.g. 8/16GB, not sure what are the limits in Travis)

It might be a pain to do this because docker builds try to be reproducible so we'd need to define some dockerfile argument and set it at build time (I don't think this is worth the effort).

I'd rather if you could just build it locally, or we try building these in production and if they fail oh well. Let me know when you want this merged.

pietroborrello:
> It might be a pain to do this because docker builds try to be reproducible so we'd need to define some dockerfile argument and set it at build time (I don't think this is worth the effort).

Yes it makes sense, was just wondering if there was already a way to distinguish, but it does not seem the case

> I'd rather if you could just build it locally, or we try building these in production and if they fail oh well. Let me know when you want this merged.

We already tested the bug benchmarks locally and excluded the ones that do not build, so you can merge this. Thank you very much

PR #1205, opened 2021-07-15
[fixed] missing commas, reformatted code more easily readable code and change to fstring format

slowy07:
fixing on ```experiment/run_experiment.py``` contains:
- [x] adding missing commas
- [x] reformatted code for more easily readable
- [x] and change ```%s``` format to ```fstring``` format for more readable and less of errors but fast  
- [x] passed tested with flake8 
- [x] reformatted code without change str quote marks

### DIFF
--- a/experiment/run_experiment.py
+++ b/experiment/run_experiment.py
@@ -39,57 +39,51 @@ from common import new_process
 from common import utils
 from common import yaml_utils
 
-BENCHMARKS_DIR = os.path.join(utils.ROOT_DIR, "benchmarks")
-FUZZERS_DIR = os.path.join(utils.ROOT_DIR, "fuzzers")
-OSS_FUZZ_PROJECTS_DIR = os.path.join(
-    utils.ROOT_DIR, "third_party", "oss-fuzz", "projects"
-)
-RESOURCES_DIR = os.path.join(utils.ROOT_DIR, "experiment", "resources")
-FUZZER_NAME_REGEX = re.compile(r"^[a-z0-9_]+$")
-EXPERIMENT_CONFIG_REGEX = re.compile(r"^[a-z0-9-]{0,30}$")
-FILTER_SOURCE_REGEX = re.compile(
-    r"("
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
I appreciate this change but it changes all of our single quotes to double quotes. We don't want that.

slowy07:
> I appreciate this change but it changes all of our single quotes to double quotes. We don't want that.

okay, solved @jonathanmetzman ! 👍🏻

experiment/run_experiment.py
```diff
@@ -88,28 +88,27 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
     for param in required_params:
         if param not in config:
             valid = False
-            logs.error('Config does not contain "%s".', param)
+            logs.error(f'Config does not contain "{str(param)}".')
```
jonathanmetzman, line 91: This pattern is going to beak presubmit.
The standard pattern we adhere to is to use % format strings for logs, not f-strings.

experiment/run_experiment.py
```diff
@@ -402,15 +405,16 @@ def start(self):
...
             '/bin/bash',
             '-c',
-            'rsync -r '
-            '"${EXPERIMENT_FILESTORE}/${EXPERIMENT}/input/" ${WORK} && '
-            'mkdir ${WORK}/src && '
-            'tar -xvzf ${WORK}/src.tar.gz -C ${WORK}/src && '
-            'PYTHONPATH=${WORK}/src python3 '
-            '${WORK}/src/experiment/dispatcher.py || '
-            '/bin/bash'  # Open shell if experiment fails.
+            'rsync -r ',
```
jonathanmetzman, line 408: I don't understand this change. This is supposed to be one argument and you made it many.

jonathanmetzman (commented):
Sorry i don't think we want this. I didn't realize what it's purpose was but it mostly is bad (changing % format strings in logging to f-strings. Thank you for your help, but I don't think we will land this.

PR #1209, opened 2021-07-19
Support multiple experiments when generating coverage reports.

jonathanmetzman:
1. Handle multiple experiments when generating coverage reports.
2. Prepare for using pool to generate coverage reports in generate_report.
3. Use tempfiles for copying json coverage files from filestore so we don't need to pass temp dir everywhere.
4. Get rid of useless ls call that slows things down.
5. Move path handling logic to own functions.
6. Add functions for key handling in coverage_data_utils to functions instead of handling in ad-hoc manner.
7. Move logic regarding gcs URLs to filestore_utils.
8. Add tests.


common/filestore_utils.py
```diff
@@ -26,7 +29,23 @@ def _using_gsutil():
     except KeyError:
         return True
 
-    return experiment_filestore_path.startswith('gs://')
+    return is_filestore_path_gcs(experiment_filestore_path)
+
+
+def is_filestore_path_gcs(filestore_path):
```
inferno-chromium, line 35: usually we name like is_gcs_filestore_path ?

common/filestore_utils.py
```diff
@@ -26,7 +29,23 @@ def _using_gsutil():
...
+    """Returns True if |filestore_path is a GCS URL. Assumes that GCS paths are
+    using gs:// and not http."""
+    return filestore_path.startswith(GCS_GSUTIL_PREFIX)
+
+
+def get_user_facing_path(filestore_path):
```
inferno-chromium, line 41: name is slightly confusing, get_web_url ?

common/filestore_utils.py
```diff
@@ -26,7 +29,23 @@ def _using_gsutil():
     except KeyError:
         return True
 
-    return experiment_filestore_path.startswith('gs://')
+    return is_filestore_path_gcs(experiment_filestore_path)
+
+
+def is_filestore_path_gcs(filestore_path):
+    """Returns True if |filestore_path is a GCS URL. Assumes that GCS paths are
```
inferno-chromium, line 36: nit: missing | after filestore_path

analysis/test_coverage_data_utils.py
```diff
@@ -91,3 +107,92 @@ def test_get_pairwise_unique_coverage_table():
...
+        'experiment_filestore': 'gs://fuzzbench-data2',
+        'experiment': 'exp2',
+        'fuzzer': FUZZER,
+        'benchmark': BENCHMARK
+    }])
+    filestore_path = (
+        coverage_data_utils.get_experiment_filestore_path_for_fuzzer_benchmark(
+            FUZZER, BENCHMARK, df))
+    assert filestore_path in ('gs://fuzzbench-data/exp1'
+                              'gs://fuzzbench-data2/exp2')
```
inferno-chromium, line 173: missing "," in tuple.

inferno-chromium (approved):
Thanks, very nice set of tests as well!

common/filestore_utils.py
jonathanmetzman: Done.

common/filestore_utils.py
jonathanmetzman: Done.

analysis/test_coverage_data_utils.py
jonathanmetzman: Good catch. Done.

common/filestore_utils.py
jonathanmetzman: get_web_url is misleading because it returns a unix path if filestore_path is a unix path.

Issue #1228, opened 2021-08-20
Wrong categorization for eclipser?


Strafo:
Hi,
I was comparing the fuzzers categorized as "Binary-only" listed in the workflows / fuzzers.yml file.
From the results of my local experiment, I believe that eclipser has an excessive advantage over the other fuzzers.
I think the big advantage is that, compared to other "binary-only" fuzzers, eclipser runs alongside AFL not in qemu-mode.
So I was wondering if it was more appropriate to move eclipser under the label "concolic execution" and add a new fuzzer eclipser_afl_qemu under "binary-only".

jonathanmetzman:
Good eye. Yes this makes sense.
Previously eclipser did not use AFL and was in fact binary-only, but now it is not.
Would you like to submit a PR to fix this?

PR #1229, opened 2021-08-20
New eclipser_afl_qemu fuzzer

Strafo:
Linked to #1228 
Additions/changes:
- new eclipser_afl_qemu for binary-only fuzzing
- change of categorization for the standard eclipser fuzzer in the fuzzers.yml file


inferno-chromium (approved):

inferno-chromium:
Looks like eclipser_afl_qemu is running out of memory in CI, any ways to optimize this, as otherwise we won't want failing fuzzers in CI. We can check it in without the CI part, but ideally we want every fuzzer checked in to work in CI as well.

Strafo:
unfortunately i couldn't find ways to make buid faster :(
qemu consumes quite a bit

Issue #1237, opened 2021-08-31
Experiment report for 2021-08-23-cloning stuck for days on 15 min


pietroborrello:
It seems that the report for the experiment [2021-08-23-cloning](https://www.fuzzbench.com/reports/experimental/2021-08-23-cloning/index.html) is stuck on 15m for something like 3 days.
We have two hypotheses on what could have gone wrong:

1. https://github.com/google/oss-fuzz/issues/6180 seems to have upgraded the Ubuntu version for gcr.io/oss-fuzz-base/base-builder images, and we were merged just before #1225 which was preventing the upgrade. This race may have caused some inconsistencies or broke our builds.
We merged the fix in our fork now, should we resubmit the PR?

2. The high memory requirements for our experiment somehow broke or frozen the containers that should run our fuzzers. In that case, we would resubmit the PR with lower memory requirements.

Let us know how should we proceed, and thank you for all your work

inferno-chromium:
@jonathanmetzman thoughts ?

inferno-chromium:
i see weird error

```
I 2021-08-31T02:23:53.840493021Z startup-script: Exception: Yaml file /work/config/experiment.yaml/experiment.yaml does not exist. 
I 2021-08-31T02:23:53.840474077Z startup-script:     raise Exception('Yaml file %s does not exist.' % yaml_filename) 
I 2021-08-31T02:23:53.840456130Z startup-script:   File "/work/src/common/yaml_utils.py", line 22, in read 
I 2021-08-31T02:23:53.840422724Z startup-script:     experiment_config = yaml_utils.read(experiment_config_filename) 
I 2021-08-31T02:23:53.840405365Z startup-script:   File "/work/src/experiment/stop_experiment.py", line 30, in stop_experiment 
I 2021-08-31T02:23:53.840387351Z startup-script:     if stop_experiment.stop_experiment(experiment_utils.get_experiment_name(), 
I 2021-08-31T02:23:53.840368071Z startup-script:   File "/work/src/experiment/dispatcher.py", line 214, in main 
I 2021-08-31T02:23:53.840345893Z startup-script:     sys.exit(main()) 
I 2021-08-31T02:23:53.840305144Z startup-script:   File "/work/src/experiment/dispatcher.py", line 222, in <module> 
I 2021-08-31T02:23:53.840187735Z startup-script: Traceback (most recent call last): 
```

inferno-chromium:
requesting new experiment in 2021-08-31-cloning

[...]

PR #1243, opened 2021-09-08
Experiment for AFL++ cloning strategies using different (smaller) map sizes

pietroborrello:
We'd like to request an experiment to study the impact of the AFL++'s edge map size on context sensitivity achieved with cloning strategies.

#1231 seems to have failed to run, even after #1237  fix, so we suspect this may be due to the high memory requirements of the experiment. We reduced it by a factor of 8x, so we hope this will run correctly.

As with PR #1202 and #1231 it is unlikely that this PR will pass the CI since the builds require quite a lot of memory and Travis seems to provide limited RAM. However, we already tested the bug benchmarks locally and excluded the ones that do not build.


andreafioraldi:
@googlebot I consent.

inferno-chromium (approved):

PR #1244, opened 2021-09-13
Fix AFL++ no_cmplog bug and ffmpeg container hash

### LINKED ISSUE #1240: Dockerfile error in ffmpeg_ffmpeg_demuxer_fuzzer
There is an issue in the Dockerfile to build ffmpeg_ffmpeg_demuxer_fuzzer at this line:

https://github.com/google/fuzzbench/blob/master/benchmarks/ffmpeg_ffmpeg_demuxer_fuzzer/Dockerfile#L17

Consequently you cannot build this benchmark. I tried to fix it by using one of the two hashes, but the compilation does not work in any case (even if with two different errors).

andreafioraldi:
This PR fixes #1240 and a bug in the AFL++ fuzzer.py

This bug, introduced to evaluate the concolic executor that uses multiple instances of AFL++, disables CmpLog even when enabled in the build function, so from ~3 months AFL++ was not using CmpLog on Fuzzbench.

Now, my experiments were made inconsistent by this mistake, even the bigger one (https://www.fuzzbench.com/reports/experimental/2021-09-08-cloning/index.html) that caused the problems with the memory requirement and it is finishing to run while I'm writing. Dunno what I should do, in theory I should be able to request an experiment running just AFL++ over all the bug benchs and merging it generating a local report with all the previous fuzzers that are not affected by the bug right?

### DIFF
--- a/benchmarks/ffmpeg_ffmpeg_demuxer_fuzzer/Dockerfile
+++ b/benchmarks/ffmpeg_ffmpeg_demuxer_fuzzer/Dockerfile
@@ -14,7 +14,7 @@
 #
 ################################################################################
 
-FROM gcr.io/oss-fuzz-base/base-builder@sha256:1b6a6993690fa947df74ceabbf6a1f89a46d7e4277492addcd45a8525e34be5a@sha256:c0eeba3437a2173c6a7115cf43062b351ed48cc2b54f54f895423d6a5af1dc3e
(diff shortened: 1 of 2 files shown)


andreafioraldi:
The bug was introduced probably for this experiment in June https://www.fuzzbench.com/reports/experimental/2021-06-27-symbolic/index.html
Please @inferno-chromium this fix is urgent

jonathanmetzman (approved):

andreafioraldi:
I'm runnign a local exp with this patch and ffmpeg seems to work http://5e41-193-55-114-4.ngrok.io/experimental/ffmpeg2/

PR #1259, opened 2021-10-14
Zafl changes

jdhiser:
Hoping to include support for Zafl in FuzzBench!

https://git.zephyr-software.com/opensrc/zafl


jdhiser:
@googlebot I signed it!

jonathanmetzman:
> @jonathanmetzman I see you added the `apt-get upgrade` fix for aspell. FYI, my commit also hits a few other benchmarks that needed the same fix.

I can't change the timeout.
If youre fine with it, I can just merge this right now.
Let me know what you want.

jonathanmetzman:
> @jonathanmetzman Will the timeouts affect the running of an experiment? If not, we can merge any time. Alternately, we could look for ways to limit which benchmarks run in workflows for Zafl?

The timeout in experiments is 12 hours per fuzzer-benchmark build. On github actions it is 6 hours for basically 1/3rd of the benchmarks (we split them up into 3). So I think we can merge.

PR #1267, opened 2021-10-22
Fix rendering of default.html template

EliaGeretto:
When running a local experiment, the rendering of the `default.html` report template fails when generating the paths to the coverage reports. This PR should fix the issue, which was probably introduced in #1209. I have not tested these changes on the cloud setup.


EliaGeretto:
@jonathanmetzman I think this is still broken on the reports shown at [www.fuzzbench.com/reports/](https://www.fuzzbench.com/reports/index.html). I don't know if this is an issue with my fix or simply a Docker image that needs to be rebuilt.

PR #1271, opened 2021-10-29
Update to make zafl failures obvious.

jdhiser:
1) When zafl_bins.sh fails, detect this in fuzzer.py and force an rm -rf
/out to force the benchmark build to fail.  Prior shortcoming caused
benchmarks to report success when zafl failed, thus no tracing
information being created resulting in black-box fuzzing instead
of grey-box fuzzing.

2) VIM swap file in .gitignore for convenience.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,6 @@ docker/generated.mk
 \#*\#
 
 .vscode
+
+# Vim backup files.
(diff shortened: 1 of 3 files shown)


jonathanmetzman (approved):
LGTM

PR #1276, opened 2021-11-07
Remove all variants, afl++ before/after test run

vanhauser-thc:
This 
 1) removes all (my) afl++ variants, huge cleanup to the CI
 2) test the new afl++ pre-release version vs honggfuzz/entropic/afl++ 3.14/3.13


vanhauser-thc:
if this test run is as expected I will update the afl++ hash in oss-fuzz and clusterfuzz

inferno-chromium (approved):

PR #1277, opened 2021-11-09
afl++ binary-only experiment

vanhauser-thc:
checks if a regression in qemu is fixed, plus if the new frida enhancements are effective.
thank you!


vanhauser-thc:
IMHO this can be merged. if some targets dont build its not an issue (although I think its a CI issue as for me all variants build just fine for the targets I tested).

PR #1283, opened 2021-11-16
Adding a new fuzzer fa-fuzz

14isnot40:
An improved version of afl fuzz waiting to be inherited and tested.
Thx


jonathanmetzman:
Superseeded by  #1297

PR #1303, opened 2021-12-07
Always define FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION

### LINKED ISSUE #1300: `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` discarded
It seems that the value of `CFLAGS` set in the builders is discarded while building the benchmarks using the appropriate `fuzzer.py` file. The discard happens here, it would seem:
https://github.com/google/fuzzbench/blob/d156b3579fe882b46fb65ab3f2bc9a5c5aeb3308/fuzzers/utils.py#L179-L180

The value of `CFLAGS` set in the builders contains `-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`, which is thus discarded while building. Discarding this value makes at least `libxslt_xpath` and `libxml2_libxml2_xml_reader_for_file_fuzzer` compile with features that make executions non-deterministic. Other benchmarks may be affected.

EliaGeretto:
This PR makes sure that `-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` is always present in `CFLAGS` and `CXXFLAGS` when building benchmarks. At least two benchmarks (`libxslt_xpath`, `libxml2_libxml2_xml_reader_for_file_fuzzer`) rely on this behavior, which is present in OSS-Fuzz.

This PR fixes #1300.

### DIFF
--- a/fuzzers/test_utils.py
+++ b/fuzzers/test_utils.py
@@ -83,9 +83,11 @@ def test_initialize_env_in_environment_without_sanitizer(fs, environ):
     utils.initialize_env()
     assert os.getenv('FUZZ_TARGET') == 'fuzzer'
     assert os.getenv('CFLAGS') == (
+        '-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION '
         '-pthread -Wl,--no-as-needed -Wl,-ldl -Wl,-lm '
         '-Wno-unused-command-line-argument -O3')
     assert os.getenv('CXXFLAGS') == (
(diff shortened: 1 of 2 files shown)


EliaGeretto:
Ah, I broke a couple of tests. I will fix them immediately.

EliaGeretto:
The `presubmit` tests are now passing, the other failures seem to be unrelated to my change.

jonathanmetzman (approved):
LGTM

jonathanmetzman:
Thanks!

PR #1308, opened 2021-12-13
libafl experiment

vanhauser-thc:
tests previous libafl vs libafl with afl-map + unstable edge detection.


vanhauser-thc:
weird that the same code parts succeed in one and fail in another.
but that is not important for the tests, this can be merged :)

jonathanmetzman (approved):
LGTM

jonathanmetzman:
Github's interface doesn't seem like it will let me merge this without CI finishing. I need to figure out a way around this.
bug

vanhauser-thc:
> Github's interface doesn't seem like it will let me merge this without CI finishing. I need to figure out a way around this. bug

for me it always works by ticket the box (like it is in the picture) and hitting the merge button. the text says otherwise but the tick text says it would work.

but doesnt matter, it is not urgent.

PR #1310, opened 2021-12-15
new afl++ and libafl experiment

vanhauser-thc:
Exciting experiment, libafl_marc will be among the top three fuzzers (entropic will be rank 4), but maybe it will even be no.1 ? :)


vanhauser-thc:
added a new coverage measurement testcase, based on google afl.
also fixed afl fuzzer to use the newest commit.

vanhauser-thc:
@jonathanmetzman this can be merged.
I have to check what is wrong the the bug benchmark after I fix the oss-fuzz issue with cmplog.

PR #1311, opened 2021-12-16
Benchmark AFL's novelty search vs. fitness max

andreafioraldi:
In this bug experiment, I want to measure how AFL performs if we use a fitness maximization algorithm based always on coverage instead of a novelty search.
Trivially, we know that the novelty search is better, but how much better? What we can expect if we combine the two metrics a la Vuzzer (that never evaluated that in the paper btw)?


jonathanmetzman (approved):
LGTM

PR #1312, opened 2021-12-17
Benchmark AFL with only edge coverage as feedback

andreafioraldi:
AFL introduced the notion of hit counts to increase the sensitivity of edge coverage as feedback and bypass roadblocks that depends for instance on loop iterations.
While we know that it increases the coverage discovery ability of a fuzzer, I want to measure the delta of such optimization in AFL.


fuzzers/afl_edges/fuzzer.py
```diff
@@ -0,0 +1,138 @@
...
+
+import json
+import os
+import shutil
+import subprocess
+
+from fuzzers import utils
+
+
+def prepare_build_environment():
```
inferno-chromium, line 24: In future, please just import the afl module and call the functions directly from there rather than duplicating. for now, i will merge, since this variant will probably get removed in some time.

inferno-chromium (approved):

PR #1323, opened 2022-01-17
AFL no culling bug experiment

andreafioraldi:
In this experiment, following the series of #1311 #1312 #1314, I want to measure the impact of minimizing the set of interesting corpus entries in AFL, removing its culling algorithm.


jonathanmetzman (approved):
LGTM

Issue #1304, opened 2021-12-07
Lots of broken projects


jonathanmetzman:
freetype, bloaty, arrow_parquet-arrow-fuzz probably more

vanhauser-thc:
@jonathanmetzman 
I am also noticing that in a fuzzbench run for aflplusplus_frida the systemd_fuzz-link-parser target fails and for aflplusplus_qemu libxslt_xpath (frida is also shown to fail, this is true though) - although both run fine when I test them locally and they find coverage. any idea what is the problem there?

jonathanmetzman:
Lemme take a look.

andreafioraldi:
I got a lof of errors in #1330 as adding benchmarks trigger the CI for every fuzzer and I spotted several issues:

 + lafintel is broken because it uses llvm 3.8 that does not support '-fsanitize=builtin' (confirmed to fail locally with `make build-lafintel-matio_matio_fuzzer`). 
 + every qemu based fuzzer (afl_qemu, weizz_qemu, eclipser, ...) will fail the test-run on bug benchmarks. This is due to the sanitizers, qemu users cannot run binaries compiled with ASan. It can run UBSan partially.
+ the afl++ commit used by fuzzolic, symcc_aflplusplus_single, symqemu_aflplusplus and eclipser is not in tree anymore (https://github.com/AFLplusplus/AFLplusplus/commit/8475cadc6307f94951e616aeea4402224d71a981) cc @vanhauser-thc 
+  the klee build is dying with `#10 4.211 ERROR: cannot verify www.uclibc.org's certificate, issued by 'CN=R3,O=Let\'s Encrypt,C=US':`
+ symcc_aflplusplus fails to link targets because -ldl is missing (a simple addition in fuzzer.py should solve it)

andreafioraldi:
I confirm that symcc_aflplusplus fails on matio too, this time because -pthread is missing. I guess this simply happens because clang is linking while the symcc runtime is in C++, so it needs dl and pthread (and probably libc++)

laurentsimon:
uclibc's cert seems fine right now: 
```
Validity
Not Before: Jan  8 03:30:08 2022 GMT
Not After : Apr  8 03:30:07 2022 GMT
```
I've not checked further than that.

jonathanmetzman:
I'll try to take a look tomorrow. This issue seems to happen when an image doesn't call apt-get upgrade for a long time.

PR #1334, opened 2022-02-15
Grammar fuzzers test on new benchmarks

andreafioraldi:
Linked to #1330 I added 4 grammar fuzzers + an experiment request on just the 3 grammar-friendly targets


andreafioraldi:
In this PR I'm also adding an optional fuzzer.yaml file into the fuzzers folders. In this optional file an user can specify the `unsupported_benchmarks` as the list `unsopported_fuzzers` inside each benchmark.yaml is becoming a mess with every temporary variants, in this way a temporary variant can just specify the unsupported benchmarks in fuzzer.yaml and a rm -rf of the fuzzer dir is enough to cleanup.
Inside fuzzer.yaml an user can specify also `allowed_benchmarks` that is an allowlist. This is useful for the grammar fuzzers for instance because they can work only with 3 targets.

andreafioraldi:
> I need to start running some more of my own experiments so this may need to wait a few days.

Np

andreafioraldi:
I'm also fixing some issues that i listed in https://github.com/google/fuzzbench/issues/1304

andreafioraldi:
symqemu is dying with a git error, I cannot fix it without fixing the code prob, fuzzolic a strange pip error.

andreafioraldi:
The aspell ubsan violation that prevents any fuzzer to run can be fixed backporting https://github.com/GNUAspell/aspell/commit/a2cd7ffd25e6213f36139cda4a911e2e03ed417c.patch
I will push it soon

andreafioraldi:
Many fails with git errors like `fatal: unable to access 'https://gitlab.gnome.org/GNOME/libxml2.git/': server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none`, i don't know how to fix them

andreafioraldi:
> Many fails with git errors like `fatal: unable to access 'https://gitlab.gnome.org/GNOME/libxml2.git/': server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none`, i don't know how to fix them but I guess this is linked to the usage of ubuntu 16 that is a dead distro

Disabling SSL for git trivially solve the problem, `RUN git config --global http.sslverify false`, dunno if acceptable for you but I don't think it can hurt

andreafioraldi:
presubmit is failing becasue markupsafe changes its API few hours ago https://markupsafe.palletsprojects.com/en/2.1.x/changes/#version-2-1-0

PR #1345, opened 2022-03-02
Remove tpm2 benchmark.

jonathanmetzman:
Under UBSAN it crashes way too often (likely very shallow bugs).
This makes it not useful as a benchmark. It also makes it bad to
use with good benchmarks since it basically steals all the measurer's
time running crashing testcases.


jonathanmetzman:
TBR

PR #1348, opened 2022-03-04
Repeat grammar experiment after fixes in benchmarks and fuzzers

andreafioraldi:
The first grammar experiment failed (https://www.fuzzbench.com/reports/experimental/2022-02-15-grammar/index.html) and I investigated why.
- The measurer is always using libfuzzer and not the binary provided by each fuzzer, so it cannot parse the serialized ASTs, that's why there are nonsense graphs in the report. I added a script to dump the serialized AST to textual files in another directory that libfuzzer can read
- The quickjs harness is broken, most of the executions are early exists, for a grammar fuzzer every execution was always the case. You should look at my patch here maybe for OSS-Fuzz too.
- The grammar files in RAM for some fuzzers may occupy too much RAM for the VM, I reduced the quality of the grammar for gramatron to fit the memory constraint

I'm requesting also to repeat the experiment with the fixes


andreafioraldi:
I confirm also that the new 2 benchs, mruby and quickjs, have many known bugs

jonathanmetzman:
Sorry for the delay in merging. I need to get rid of some of these variants, CI getting too many failures.

jonathanmetzman:
Note, there might be an issue with report generation: #1351  if the report isn't generated, ping me and I'll make it after this bug is fixed.

andreafioraldi:
@jonathanmetzman so far so good with the grammar benchmarks https://www.fuzzbench.com/reports/experimental/2022-03-04-grammar/index.html IMO you can keep quickjs and mruby in the normal bug experiments.
There are enough bugs, I removed the too shallow ones in quickjs. The experiment is meaningful, apart grimoire that probably there is a bug in my side that I cannot figure out.

PR #1364, opened 2022-03-25
Fix arrow benchmark

jonathanmetzman:
Remove arrow benchmark because it no longer builds. It no longer
builds because the build process tries to download a file from
apache.org that the website no longer hosts.


inferno-chromium (approved):
Is this benchmark used in a lot of research papers ? if yes, maybe we should try to find the file from some cache somewhere ?

jonathanmetzman:
I can fix it this time, but we will upgrade the ubuntu images soon and we will probably need to get rid of some benchmarks. We should probably do a release then and then not worry about backwards compatibility, it's not a good use of time to maintain this stuff IMO

PR #1366, opened 2022-03-28
Tortoise

waddewaa:
Integration of TortoiseFuzz to Fuzzbench


jonathanmetzman (approved):
lgtm

waddewaa:
@jonathanmetzman I have checked the error log, and it seems like the test "*curl_curl_fuzzer_http*" is broken unrelated to our commits.

However, the (tortoisefuzz, standard|bug) are reaching a timeout, this is most likely due to TortoiseFuzz requiring Clang-6.0.0 and the building process being very time consuming. Is there any way to cache the building of the fuzzer for all benchmarks? I can alter the dockerfile to parallelize the building process so that it uses all available cores, but I am unsure if that would be the most effective approach.

jonathanmetzman:
Also, yeah if we keep building clang from source, then please don't specify "-j" so we can get full parallelization in production (we use 32 core builders).

jonathanmetzman:
also fixing curl issue here: #1370 just pull and push again after merge to retrigger CI.

waddewaa:
It was not possible to download the binaries directly, it would not compile then. Presumably due to LLVM_ENABLE_ASSERTIONS=On defaults to off, which is a requirement for TortoiseFuzz. I instead only removed the -j flag. Before merging I would like to see that the fuzzer compiles in this environment, our local environment becomes resource starved when not specifying the -j flag.

waddewaa:
@jonathanmetzman If this is fine with you, I am okay with merging it.

jonathanmetzman:
Do you want to run an experiment with this?

waddewaa:
Yes, but we will be adding a couple of modified benchmarks that employ anti-fuzzing first, will be a new PR soon.

jonathanmetzman:
Exciting!

PR #1373, opened 2022-04-04
Add variations of LibFuzzer to trace data flow signals

Alan32Liu:
This PR allows private experiments on four new variations of `LibFuzzer` related to dataflow signal tracing:

1. `LibFuzzer_dataflow` traces both `store` and `load` signals;
2. `LibFuzzer_dataflow_store` traces `store` signals only;
3. `LibFuzzer_dataflow_load` traces `load` signals only;
4. `LibFuzzer_dataflow_pre` is the version of `LibFuzzer` (from `llvm-project`) right before applying the patch (i.e. a control group).

All patch files (`fuzzers/libfuzzer_dataflow*/Trace-store-and-load-commands.patch`) are identical: They are the changes from the [`trace_data_flow` PR in `llvm-project`](https://github.com/Alan32Liu/llvm-llvm-project/pull/1).
The only difference between these variations is the `-fsanitize-coverage` param they use (e.g. `-fsanitize-coverage=trace-loads,trace-stores`)

Note that all of them only tracks dataflow of global variables for now.

### DIFF
--- a/fuzzers/libfuzzer_dataflow_load/Trace-store-and-load-commands.patch
+++ b/fuzzers/libfuzzer_dataflow_load/Trace-store-and-load-commands.patch
@@ -1,8 +1,9 @@
-From 89cf0427a4b6195f7366a51e9e1beb92e50bb38d Mon Sep 17 00:00:00 2001
+From e7b02154a90480ed6bca8ba7203c70e7b5a27203 Mon Sep 17 00:00:00 2001
 From: Alan32Liu <donggeliu@google.com>
 Date: Wed, 23 Feb 2022 15:55:17 +1100
-Subject: [PATCH] Trace store and load commands
+Subject: [PATCH 1/2] Trace store and load commands
 
+Signed-off-by: Alan32Liu <donggeliu@google.com>
 ---
  compiler-rt/lib/fuzzer/FuzzerMain.cpp    |   2 +
  compiler-rt/lib/fuzzer/FuzzerTracePC.cpp | 108 +++++++++++++++++++++++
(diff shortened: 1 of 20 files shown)


Alan32Liu:
I am not sure if the CI failure is caused by the new commits, since other benchmarks (e.g. the old `libfuzzer`) also fails for the same error on the same benchmark (i.e. `arrow_parquet-arrow-fuzz`).
The failure can be reproduced by `cd /work/thrift_ep-prefix/src && /usr/local/bin/cmake -P /work/thrift_ep-prefix/src/thrift_ep-stamp/thrift_ep-download-RELEASE.cmake` in `make debug-builder-libfuzzer-arrow_parquet-arrow-fuzz`

fuzzers/libfuzzer_dataflow_load/weak.c
```diff
@@ -0,0 +1,28 @@
+// Copyright 2022 Google LLC
```
jonathanmetzman, line 1: Don't duplicate this file, please copy it from the other directories.

fuzzers/libfuzzer_dataflow_load/Trace-store-and-load-commands.patch
```diff
@@ -0,0 +1,371 @@
+From e7b02154a90480ed6bca8ba7203c70e7b5a27203 Mon Sep 17 00:00:00 2001
```
jonathanmetzman, line 1: Do we have to duplicate this file?

jonathanmetzman (commented):
Why do we want to land this PR?
I don't really like having one-off fuzzers staying in this repo permanently. Do we want to run an experiment here?

jonathanmetzman:
Can you pull again from master and push here. I think I've fixed the bug in CI causing AFL to fail.

oliverchang:
> Why do we want to land this PR? I don't really like having one-off fuzzers staying in this repo permanently. Do we want to run an experiment here?

We'll want to compare this to other fuzzing engines doing similar things in the future. Is there a better place to store these semi-permanently?

Alan32Liu:
Failed 7 checks due to incompatibility with the following two benchmark programs:
1. All 3 new `LibFuzzer`s (with the dataflow patch) failed on `curl_curl_fuzzer_http`;
2. All 4 `LibFuzzer`s (3 new + one without the dataflow patch) failed on `ffmpeg_ffmpeg_demuxer_fuzzer`.

Since `LibFuzzer_dataflow` is only experimental at this point, it might be a wiser use of our time to leave out these two benchmarks if they are not easy to fix?
More details of each benchmark are explained below.

Alan32Liu:
## `curl_curl_fuzzer_http` 
It failed because the benchmark's build script (`/src/openssl/config`) cannot take an object file parameter (i.e. `/weak.o`) with a pre-defined architecture.

### Fix
It can be fixed by building with a different script (i.e. `/src/openssl/Configure`) and specifying the architecture of `/weak.o` (i.e. `os/compiler:linux-x86_64`):

```bash
./Configure --prefix=/src/curl_install --debug enable-fuzz-libfuzzer -DPEDANTIC -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION no-shared enable-tls1_3 enable-rc5 enable-md2 enable-ssl3 enable-ec_nistp_64_gcc_128 enable-ssl3-method enable-nextprotoneg enable-weak-ssl-ciphers -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -pthread -Wl,--no-as-needed -Wl,-ldl -Wl,-lm -Wno-unused-command-line-argument -O3 -fsanitize=fuzzer-no-link -fsanitize-coverage=trace-loads,trace-stores /weak.o -fno-sanitize=alignment os/compiler:linux-x86_64
```
This is because the old script `config` first guesses the architecture and then calls the new script `Configure` with its guess. In this case, its guess was `x86_64-whatever-linux2`. 
Manually building `/weak.o` with this arch will not prevent this error either.

### Mitigation
I am not sure if there is any simple mitigation, as I presume `config` is necessary for many other fuzzers.
Should we omit this benchmark from our experiments with the dataflow patch for now?

### Error message
```bash
./config --prefix=/src/curl_install --debug enable-fuzz-libfuzzer -DPEDANTIC -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION no-shared enable-tls1_3 enable-rc5 enable-md2 enable-ssl3 enable-ec_nistp_64_gcc_128 enable-ssl3-method enable-nextprotoneg enable-weak-ssl-ciphers -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -pthread -Wl,--no-as-needed -Wl,-ldl -Wl,-lm -Wno-unused-command-line-argument -O3 -fsanitize=fuzzer-no-link -fsanitize-coverage=trace-loads,trace-stores /weak.o -fno-sanitize=alignment
Operating system: x86_64-whatever-linux2
Configuring for linux-x86_64
target already defined - linux-x86_64 (offending arg: /weak.o)
```

Alan32Liu:
## ffmpeg_ffmpeg_demuxer_fuzzer
It is quite strange because it does not fail when building with the original `libfuzzer` but fails when building with `libfuzzer_dataflow_pre`, which only upgrades `clang` to version 15 and updates `llvm-project` to a newer version.

## Mitigation
It does not seem that this error is caused by the patch (i.e. unrelated to `/weak.o`).
I am not sure how to fix this build error, maybe also omit this benchmark from our experiments with the dataflow patch for now?

## Error message
```bash
.libs/pushbuf.o:(.bss+0x0): multiple definition of `nouveau_debug'
.libs/nouveau.o:(.bss+0x0): first defined here
.libs/bufctx.o:(.bss+0x0): multiple definition of `nouveau_debug'
.libs/nouveau.o:(.bss+0x0): first defined here
.libs/abi16.o:(.bss+0x0): multiple definition of `nouveau_debug'
.libs/nouveau.o:(.bss+0x0): first defined here
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

jonathanmetzman:
> ## `curl_curl_fuzzer_http`
> It failed because the benchmark's build script (`/src/openssl/config`) cannot take an object file parameter (i.e. `/weak.o`) with a pre-defined architecture.
> 
> ### Fix
> It can be fixed by building with a different script (i.e. `/src/openssl/Configure`) and specifying the architecture of `/weak.o` (i.e. `os/compiler:linux-x86_64`):
> 
> ```shell
> ./Configure --prefix=/src/curl_install --debug enable-fuzz-libfuzzer -DPEDANTIC -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION no-shared enable-tls1_3 enable-rc5 enable-md2 enable-ssl3 enable-ec_nistp_64_gcc_128 enable-ssl3-method enable-nextprotoneg enable-weak-ssl-ciphers -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -pthread -Wl,--no-as-needed -Wl,-ldl -Wl,-lm -Wno-unused-command-line-argument -O3 -fsanitize=fuzzer-no-link -fsanitize-coverage=trace-loads,trace-stores /weak.o -fno-sanitize=alignment os/compiler:linux-x86_64
> ```
> 
> This is because the old script `config` first guesses the architecture and then calls the new script `Configure` with its guess. In this case, its guess was `x86_64-whatever-linux2`. Manually building `/weak.o` with this arch will not prevent this error either.
> 
> ### Mitigation
> I am not sure if there is any simple mitigation, as I presume `config` is necessary for many other fuzzers. Should we omit this benchmark from our experiments with the dataflow patch for now?
> 
> ### Error message
> ```shell
> ./config --prefix=/src/curl_install --debug enable-fuzz-libfuzzer -DPEDANTIC -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION no-shared enable-tls1_3 enable-rc5 enable-md2 enable-ssl3 enable-ec_nistp_64_gcc_128 enable-ssl3-method enable-nextprotoneg enable-weak-ssl-ciphers -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -pthread -Wl,--no-as-needed -Wl,-ldl -Wl,-lm -Wno-unused-command-line-argument -O3 -fsanitize=fuzzer-no-link -fsanitize-coverage=trace-loads,trace-stores /weak.o -fno-sanitize=alignment
> Operating system: x86_64-whatever-linux2
> Configuring for linux-x86_64
> target already defined - linux-x86_64 (offending arg: /weak.o)
> ```

We probably can handle curl differently in the build function in fuzzer.py

fuzzers/libfuzzer_dataflow_pre/builder.Dockerfile
```diff
@@ -0,0 +1,29 @@
+# Copyright 2022 Google LLC
```
jonathanmetzman, line 1: Instead of adding this, can we just update the current version of libFuzzer?

Alan32Liu: Sure! That would be much easier.
I add this only because I don't want to let my experiment affect other experiments that might rely on the specific version of `LibFuzzer`.

oliverchang:
@jonathanmetzman @Alan32Liu is there anything left here before we can merge?

jonathanmetzman:
> @jonathanmetzman @Alan32Liu is there anything left here before we can merge?

Yeah, I want to get rid of libfuzzer_dataflow_pre before I merge this.

jonathanmetzman:
Sorry I didn't realize clang would need to be updated in libfuzzer's image. We probably shouldn't do that. Can you revert your last two commits to get libfuzzer_data_pre back, undo the changes to libfuzzer, and I'll land it.

PR #1374, opened 2022-04-06
new afl++ experiment

vanhauser-thc:
Upgrade all afl++ plugins to the new llvm pass manager so this is an important test to see if everything still works.


jonathanmetzman:
Was the bug you mentioned on discord noticeable in past fuzzbench reports? I wonder why not

PR #1376, opened 2022-04-07
afl++ experiment: keep timeouts

vanhauser-thc:
I found a weird thing in afl++ (inherited from afl I think): if a timeout has new coverage but is not a hang it is thrown away.
In this experiment I want to see if it is actually beneficial to save those timeout-new-coverage-inputs into the queue.
(hijacking my _llvm15 variant for this)


vanhauser-thc:
@jonathanmetzman this can be merged :) thank you!

PR #1356, opened 2022-03-17
Enhance local experiments

andreafioraldi:
Followup to my `--concurrent-builds` PR to fix the local experiments.
Here there are several contributions:
+ Improve syscalls speed in the containers with `--security-opt seccomp=unconfined` (see the AFL++ system config script) when running runners containers
+ Pin runners and measures to specific cores with `--cpuset-cpus` for runners and psutil for measures.
+ Allow the user to specify how many cores the measures must use with `--measurers-cpus` in run_experiment.py
+ Allow the user to specify how many cores the runners must use with `--runners-cpus` in run_experiment.py

Now, for instance, I can easily run a local experiment even on my desktop without any explosion of docker or OOM with

```
PYTHONPATH=. python3 experiment/run_experiment.py \
--experiment-config foo.yaml \
--benchmarks libpng-1.2.56 jsoncpp_jsoncpp_fuzzer \
--experiment-name foo \
--concurrent-builds 2 \
--measurers-cpus 4 \
--runners-cpus 4 \
--fuzzers afl aflfast \
```
The outcome is

Screenshot at 2022-03-17 16-00-40

Before was a mess with cores not used at 100% and overlapped with measurers.

I tested it with and without the new arguments and it works, but before merging you should try to start a remote experiment even if in theory I didn't touch the remote part.

PR #1384, opened 2022-04-15
Fix runners core scheduling for local experimetns

apach301:
Trying to fix cpu managing for runners in local experiments (#1356). Scheduler creates threadpool with N threads, every thread have its own unique cpuset preference (obtained in `_process_init`). Then all trials are scheduled over this threadpool with `_start_trial` function: it launches a trial with binding it to current thread' cpuset and immediately exits. The problem is that the trial launching happens very fast, the thread in threadpool is freed and ready to launch another trial. In my local experiment the launching of 100 trials have been scheduled over first 12-15 threads (https://github.com/google/fuzzbench/pull/1356#issuecomment-1098136992).

This PR tries to fix this problem by picking available cpusets explicitly for every `_start_trial` job. If number of cores specified for runners is lesser than number of trials then it would scheduled around with overcommit.

### DIFF
--- a/experiment/scheduler.py
+++ b/experiment/scheduler.py
@@ -43,8 +43,6 @@ FAIL_WAIT_SECONDS = 10 * 60
 
 logger = logs.Logger('scheduler')  # pylint: disable=invalid-name
 
-CPUSET = None
-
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Can you sign the CLA please?

jonathanmetzman:
My bad

PR #1390, opened 2022-04-20
blocklist dataflow fuzzers on benchmarks they fail on

jonathanmetzman:
CC @Alan32Liu


Alan32Liu (commented):
Ah, I did not release there is a way to block them.
Thanks!

PR #1395, opened 2022-05-10
New option for custom seed corpus

jiradeto:
I'd like to add new experiment option `custom_seed_corpus_dir` to fuzzbench. This option enables users to run an experiment with custom seed corpus. To use this option, one has to set --custom-seed-corpus-dir to path that has input files for the running benchmarks.

For example:
```
PYTHONPATH=. python3 experiment/run_experiment.py  \
--experiment-config /tmp/local-experiment-config.yaml \
--experiment-name demo-exp \
--fuzzers  afl libfuzzer \
--benchmarks libjpeg-turbo-07-2017 sqlite3_ossfuzz \
--custom-seed-corpus-dir /local/my-custom-corpora 
```

and the `/local/my-custom-corpora` must have the following content:
```
|— my-custom-corpora 
|    |— sqlite3_ossfuzz  
|    |    |— fbeb1de0be72d83ed8dff924fd9df8cedb0a7497
|    |    |— fba1b82f6612e8e1d1786b757dfeecbbfabeeca5
|    |— libjpeg-turbo-07-2017
|    |    |— f78602c270b787d3e7e940f50f78365552d5a20b
|    |    |— ebc89a181d637f6bdfdcaa9d647cfd1f88d62675
|    |    |— ed0fbdb5c9f7d2fbb1c6c8e0665e80b110ba1e14
```

I'm not sure if there is already an option that does the similar thing. But If not, it think this new option would be useful to have as sometime we need to run experiment with some sort of special seed inputs.

### DIFF
--- a/common/experiment_utils.py
+++ b/common/experiment_utils.py
@@ -72,6 +72,12 @@ def get_oss_fuzz_corpora_filestore_path():
     return posixpath.join(get_experiment_filestore_path(), 'oss_fuzz_corpora')
 
 
+def get_custom_seed_corpora_filestore_path():
+    """Returns path containing the user-provided seed corpora."""
+    return posixpath.join(get_experiment_filestore_path(),
+                          'custom_seed_corpora')
+
+
 def get_dispatcher_instance_name(experiment: str) -> str:
     """Returns a dispatcher instance name for an experiment."""
     return 'd-%s' % experiment
--- a/experiment/resources/runner-startup-script-template.sh
+++ b/experiment/resources/runner-startup-script-template.sh
@@ -46,6 +46,7 @@ docker run \
 -e NO_SEEDS={{no_seeds}} \
 -e NO_DICTIONARIES={{no_dictionaries}} \
 -e OSS_FUZZ_CORPUS={{oss_fuzz_corpus}} \
+-e CUSTOM_SEED_CORPUS_DIR={{custom_seed_corpus_dir}} \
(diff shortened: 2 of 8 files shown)


jonathanmetzman:
How much do you need to upstream this? I would rather people just committed the seed corpus or something rather than add new complexity like this.

jiradeto:
I agree that this is a bit too complicated. But is there any available option in fuzzbench to have dynamic set of seed corpus? I find it's not that convenient to commit new corpus every time we want to change to the new ones.

jonathanmetzman:
> I agree that this is a bit too complicated. But is there any available option in fuzzbench to have dynamic set of seed corpus? I find it's not that convenient to commit new corpus every time we want to change to the new ones.

We do have an option in run_experiment that doesn't force you to commit all changes, I think it's `-a`

jiradeto:
@jonathanmetzman thanks for the suggestion. I think that instead of doing all the things manually, having an option to automate this will help a lot. I may come up with a simpler approach for request.

jiradeto:
@jonathanmetzman We tried to simplify the PR, and we managed to reduce the complexity quite a bit. However, it seems like some functionality is needed for our use case. In case you don't remember, we're trying to use the custom seed directory to start individual campaigns with different sets of seeds from a large pool of seeds. We don't necessarily want to commit the large pool of seeds to the repo (for instance, seeds from clusterfuzz or existing fuzzing campaigns).

jonathanmetzman (approved):
LGTM

PR #1399, opened 2022-05-25
Prevent overcommit when specifying CPU count for runners

EliaGeretto:
This PR fixes core scheduling in local experiments so that only one experiment is scheduled on a given core at any given time when specifying `--runners-cpus`. The changes in #1384 broke the previous solution that was meant to prevent overcommit, which was introduced in #1356. The changes in this PR should still avoid the scheduling issue described in #1384 while also avoiding overcommitting the available cores.

### DIFF
--- a/experiment/scheduler.py
+++ b/experiment/scheduler.py
@@ -116,21 +116,29 @@ def delete_instances(instances, experiment_config):
                                    experiment_config['cloud_compute_zone'])
 
 
-def end_expired_trials(experiment_config: dict):
+def end_expired_trials(experiment_config: dict, core_allocation: dict):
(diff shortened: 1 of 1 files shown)


EliaGeretto:
I have tested the changes on my local setup and it seems to work. It should not modify the behavior of the cloud one.

jonathanmetzman:
> I have tested the changes on my local setup and it seems to work. It should not modify the behavior of the cloud one.

Thanks. Lemme try to review this tomorrow. May end up being merged tuesday since monday is a public holiday

jonathanmetzman:
Sorry I dropped the ball on this. I'm going to review this Tuesday when I get back in office.

PR #1405, opened 2022-06-02
Option to customize snapshot period

jiradeto:
I'd like to request an option to modify the interval time to measure coverage. I couldn't find how to do so within the current implementation. This new parameter can be set via the configuration file along with other configuration parameters.
```
trials: 5
max_total_time: 86400
snapshot_period: 600 // 10 minutes
```

Could you please help to review @jonathanmetzman ?


jonathanmetzman:
This LGTM. have you tested this? It works? Can you write a unittest?

jiradeto:
Yes, I've tested and it worked. I will try to add more tests as you suggested.

PR #1412, opened 2022-06-09
AFL splicing mutation experiment

andreafioraldi:
Benchmark AFL vs AFL with splicing as mutation.

This is another rerun from https://www.fuzzbench.com/reports/experimental/2022-05-06-dissecting/index.html
It seems that many fuzzers in the big experiment have less than 20 trials while small experiments like the one I requested about scheduling run with a subset of the fuzzers used in the big one and it is all ok.

Ty as usual.


jonathanmetzman (approved):
lgtm

PR #1413, opened 2022-06-13
Pass snapshot period to runner

jiradeto:
In https://github.com/google/fuzzbench/pull/1405 I requested an option to set snapshot period so the interval time to measure coverage could be adjustable, however I just noticed that the runner also [uses this value](https://github.com/google/fuzzbench/blob/master/experiment/runner.py#L305).

This small PR passes environment variable SNAPSHOT_PERIOD to the runner so that we have the same value throughout an experiment.


jonathanmetzman (approved):
lgtm

PR #1418, opened 2022-07-04
New fuzzer integration: `Centipede`

Alan32Liu:
Adding [`Centipede`](https://github.com/google/centipede/tree/1.0.0) as a new fuzzer.


fuzzers/centipede/fuzzer.py
```diff
@@ -0,0 +1,79 @@
...
+    # TODO(Dongge): Check if they support trace-dataflow.
+    # TODO(Dongge): Build targets with sanitizers.
+    cflags = ['-fsanitize-coverage=trace-pc-guard,pc-table,trace-cmp']
+    utils.append_flags('CFLAGS', cflags)
+    utils.append_flags('CXXFLAGS', cflags)
+
+    os.environ['CC'] = 'clang'
+    os.environ['CXX'] = 'clang++'
+    os.environ['FUZZER_LIB'] = (
+        '/src/centipede/bazel-bin/libfuzz_target_runner.a '
```
oliverchang, line 33: This is a bit brittle -- if some refactoring occurs and things are added/removed, then the build may break. Can you add a TODO to fix this? 

We may need to request this from the Centipede team.

Alan32Liu: Yep, this also caused build failure in other benchmarks (e.g. `bloaty_fuzz_target`).

Alan32Liu:
There are 5 CI failures at this point, 2 from `afl` and 3 from `centipede`. 
They are either fixed in other commits or awaiting others:
1. (`AFL`, `libxml2-v2.9.2`) Failure [7206690302](https://github.com/google/fuzzbench/runs/7206690302?check_suite_focus=true) is caused by outdated `ca-certificates` and is fixed by Commit [7de83f8](https://github.com/google/fuzzbench/pull/1419/commits/7de83f8f33bca16b375a40a26e06df55179e15ce);
2. (`AFL`, `ffmpeg_ffmpeg_demuxer_fuzzer`) [7206690340](https://github.com/google/fuzzbench/runs/7206690340?check_suite_focus=true) is caused by an error in `libXext` and is fixed by [9e2a46e](https://github.com/google/fuzzbench/pull/1419/commits/9e2a46e6c4344b9d3a56814f55e0b76a122fc221).
3. (`Centipede`, `bloaty_fuzz_target`) [7206704117](https://github.com/google/fuzzbench/runs/7206704117?check_suite_focus=true) fails because `Centipede` requires multiple `FUZZER_LIB`. We are waiting for them to merge those libs to one;
4. (`Centipede`, `libxml2-v2.9.2`) [7206704176](https://github.com/google/fuzzbench/runs/7206704176?check_suite_focus=true) is the same as (`AFL`, `libxml2-v2.9.2`);
5. (`Centipede`, `arrow_parquet-arrow-fuzz`) [7206704250](https://github.com/google/fuzzbench/runs/7206704250?check_suite_focus=true) is the same as (`Centipede`, `bloaty_fuzz_target`).

fuzzers/centipede/builder.Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+  --cxxopt=-std=c++17 \
+  --cxxopt=-stdlib=libc++ \
+  --linkopt=-lc++'
+
+# Add C++15
+ADD https://commondatastorage.googleapis.com/chromium-browser-clang/Linux_x64/clang-llvmorg-15-init-1995-g5bec1ea7-1.tgz /
+RUN mkdir /bin/clang && \
+    tar zxvf /clang-llvmorg-15-init-1995-g5bec1ea7-1.tgz -C /bin/clang
+
+# Assist building targets with dataflow sanitizers
```
jonathanmetzman, line 35: End all comments with a period pleae.

fuzzers/centipede/fuzzer.py
```diff
@@ -0,0 +1,79 @@
...
+import subprocess
+import os
+
+from fuzzers import utils
+
+
+def build():
+    """Build benchmark."""
+    # TODO(Dongge): Check if they support trace-dataflow.
+    # TODO(Dongge): Build targets with sanitizers.
```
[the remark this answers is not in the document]
jonathanmetzman: I don't think it causes issues. I think eclipser used to do this.
As a feature for a fuzzer I think this is pretty user-unfriendly.

fuzzers/centipede/fuzzer.py
```diff
@@ -0,0 +1,89 @@
...
+        '/src/centipede/bazel-bin/libshared_memory_blob_sequence.a '
+        '/src/centipede/bazel-bin/libexecution_request.a '
+        '/src/centipede/bazel-bin/libexecution_result.a '
+        '/src/centipede/bazel-bin/libbyte_array_mutator.a')
+
+    utils.build_benchmark()
+
+
+def fuzz(input_corpus, output_corpus, target_binary):
+    """Run fuzzer. Wrapper that uses the defaults when calling
```
jonathanmetzman, line 52: Can the last part of the comment fit in 80 chars?

fuzzers/centipede/fuzzer.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def run_fuzzer(input_corpus, output_corpus, target_binary, extra_flags=None):
+    """Run fuzzer."""
+    if extra_flags is None:
+        extra_flags = []
+
+    # Seperate out corpus and crash directories as sub-directories of
+    # |output_corpus| to avoid conflicts when corpus directory is reloaded.
+    work_dir = os.path.join(output_corpus, 'WD')
```
jonathanmetzman, line 64: nit: "work-dir" instead of "WD"

fuzzers/centipede/fuzzer.py
```diff
@@ -0,0 +1,89 @@
...
+    os.makedirs(work_dir)
+    os.symlink(crashes_dir, work_dir_crash)
+    os.makedirs(crashes_dir)
+    os.makedirs(output_corpus)
+
+    flags = [
+        f'--workdir={work_dir}',
+        f'--corpus_dir={output_corpus},{input_corpus}',
+        f'--binary={target_binary}',
+        '--num_runs=100',
```
jonathanmetzman, line 77: What does num_runs do?

fuzzers/centipede/fuzzer.yaml
```diff
@@ -0,0 +1,2 @@
+languages:
```
jonathanmetzman, line 1: I don't think this config file is needed.

fuzzers/centipede/builder.Dockerfile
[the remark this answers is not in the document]
Alan32Liu: I failed to find `clang-15` in the container (the default is `clang-12`, which does not support dataflow tracing).
Happy to remove it if there is `clang-15`.

fuzzers/centipede/fuzzer.py
Alan32Liu: Yeah, it should not be here. Removed : )

fuzzers/centipede/weak.c
```diff
@@ -0,0 +1,28 @@
+// Copyright 2022 Google LLC
```
jonathanmetzman, line 1: Not sure if I asked already but why is weak.c needed by centipede? It was just a hack I made for dataflow.

Alan32Liu: Because we [enabled dataflow tracing](https://github.com/google/fuzzbench/blob/86f3b2447a88d2492b6922800f6ffd6018f66b07/fuzzers/centipede/fuzzer.py#L30) in `Centipede` as well : )
I recall it does work without this hack when tested locally earlier.

fuzzers/centipede/builder.Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+ARG parent_image
+FROM $parent_image
+
+ENV BAZEL_GPG_LINK='https://bazel.build/bazel-release.pub.gpg'
```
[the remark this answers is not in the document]
Alan32Liu: I see, I will fix this tmr.

fuzzers/centipede/weak.c
jonathanmetzman: Right I would be very surprised if Centipede needs this.

Alan32Liu: Oh sorry, I just realised I mistyped: I meant to say it does *not* work without this hack.
But I will double-check tmr.

Alan32Liu: You are right: we don't need `weak.c` here.
I did a bit more experiment today and can confirm that now.
Sorry about the confusion: I must have confused other error messages with it earlier.

I've fixed that locally and will push it tmr after rebasing to master.
Thanks for pointing this out!

Alan32Liu: I remember why I added that `weak.c` now:
Benchmark `arrow_parquet-arrow-fuzz` needs it, otherwise it will fail because of undefined references to `__sanitizer_cov_store*` and `__sanitizer_cov_load*` functions.
See [this CI failure](https://github.com/google/fuzzbench/runs/7329782565?check_suite_focus=true).
Apparently, it is the only one needs it. I will ask Kostya.

Alan32Liu:
The last benchmark failure is because of #1427.
Let's fix it in a separate PR and don't hold this one for it.

PR #1421, opened 2022-07-12
Avoid hardcoding the FUZZER_LIB

Alan32Liu:
The Dockerfile of benchmark `openssl` hardcodes `--with-fuzzer-lib=/usr/lib/libFuzzingEngine`.
This PR replaces `/usr/lib/libFuzzingEngine` with `$FUZZER_LIB`, as not all fuzzers save their lib to that path.

### DIFF
--- a/benchmarks/openssl_x509/build.sh
+++ b/benchmarks/openssl_x509/build.sh
@@ -21,7 +21,14 @@ then
   CONFIGURE_FLAGS="no-asm"
 fi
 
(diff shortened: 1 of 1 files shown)


Alan32Liu:
Hmm... It seems this works for `Centipede` but not for many other fuzzers.
For example, `symcc_afl_single` does not set `FUZZER_LIB`.
Would there be a better solution than making the value of `--with-fuzzer-lib=` conditional on the fuzzer's name?

jonathanmetzman (approved):
lgtm

PR #1422, opened 2022-07-12
Fix a bug in libfuzzer_dataflow

Alan32Liu:
Enable dataflow tracing without setting `UseValueProfileMask`.
This is an old bug that was fixed locally a long time ago but was not updated here.


jonathanmetzman (approved):
stamp
It's not really feasible for me to review a diff file like this.

Alan32Liu:
> stamp It's not really feasible for me to review a diff file like this.

Sorry : )
I generated the patch file in a different way to make it more succinct, but that also made it barely readable.
It's only a small change: I removed an `if` condition of a code block.
Here is [the corresponding commit](https://github.com/Alan32Liu/llvm-llvm-project/commit/9141d410f2b16964967702b07cf83ae63a73148a).

PR #1424, opened 2022-07-13
Upgrade `ca-certificates` in `honggfuzz_qemu`

Alan32Liu:
Upgrade `ca-certificates` in `honggfuzz_qemu` to fix this [CI failure](https://github.com/google/fuzzbench/runs/7314918364?check_suite_focus=true).


jonathanmetzman (approved):
LGTM

jonathanmetzman:
> This also occurred in #1419 and fixed with the same way in [7de83f8](https://github.com/google/fuzzbench/commit/7de83f8f33bca16b375a40a26e06df55179e15ce), is there a way to fix both in one location? I suspect there might be more later.

I don't think so. I think this variety of error is happening because we are inheriting from super old images. It will go away when we upgrade them.

PR #1431, opened 2022-07-21
A new dependency OpenSSL for `Centipede`

Alan32Liu:
Adding a new dependency OpenSSL.


jonathanmetzman (approved):
LGTM.
Why does it require this?

Alan32Liu:
> LGTM. Why does it require this?

From Kostya: `bazel` used to pull `boringssl` sources, but it appeared to be too noisy (fresh `clang` would reject `boringssl` sources)

They also plan to have a dependency installation script, I will leave this PR open and adapt to that when that script is ready.

PR #1433, opened 2022-07-22
Set LIB_FUZZING_ENGINE for arrow_parquet-arrow-fuzz

Alan32Liu:
Link env variable `LIB_FUZZING_ENGINE` to Centipede's lib file, as required by benchmark `arrow_parquet-arrow-fuzz`.
Unfortunately, this does not solve the `undefined reference to '__sanitizer_cov_load8'` error.


Alan32Liu:
No need to merge: This PR is just to show that fix does not work.
We probable should consider removing this benchmark (or at least from the CI), given it was not used in experiments anyway.

With that said, if there is an idea about how to fix it, I am happy to give it a shot.

PR #1437, opened 2022-07-22
[centipede] Fix compilation flags

jonathanmetzman:
Add compiler flags to instrument edges and comparisions.
Also do centipede experiment.


Navidem (approved):
Thanks!

PR #1444, opened 2022-08-03
Fail on failed checkout

jonathanmetzman:


jonathanmetzman:
Need to test this to see what it breaks.

PR #1446, opened 2022-08-04
bump libfuzzer

Navidem:


jonathanmetzman (commented):
Do we need to do this?

Navidem:
Oh, I thought you asked for it!

jonathanmetzman:
Only if it's needed. If it's not needed, I'd rather not do this.

Navidem:
well, my use-case is making a libfuzzer request for one benchmark to use as reference point for my own experiment.

Navidem:
Actually, I wonder if this merged does `patch.diff` still applies correctly?!

jonathanmetzman:
> Actually, I wonder if this merged does `patch.diff` still applies correctly?!

patch.diff isn't applied. You can make a PR to delte it.

PR #1436, opened 2022-07-22
Fix `Centipede` integration

Alan32Liu:
Improve our adaptability to future changes in `Centipede` by using scripts/flags predefined by it:
* Install `Centipede`'s dependencies with its predefined script;
* Collect the `CFLAGS` needed to build `Centipede` from its predefined file.

PR #1450, opened 2022-08-10
Increase `shm` to mitigate bus error

Alan32Liu:
To fix `Bus error (core dumped)` discovered in #1436.

### DIFF
--- a/experiment/resources/dispatcher-startup-script-template.sh
+++ b/experiment/resources/dispatcher-startup-script-template.sh
@@ -26,7 +26,7 @@ docker run --rm \
   -e POSTGRES_PASSWORD={{postgres_password}} \
   -e CLOUD_SQL_INSTANCE_CONNECTION_NAME={{cloud_sql_instance_connection_name}} \
   -e DOCKER_REGISTRY={{docker_registry}} \
-  --shm-size=1g \
+  --shm-size=2g \
   --cap-add=SYS_PTRACE --cap-add=SYS_NICE \
   -v /var/run/docker.sock:/var/run/docker.sock --name=dispatcher-container \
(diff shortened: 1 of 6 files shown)


oliverchang (approved):
LGTM, but I'll let @jonathanmetzman review too if there's any places were this is either not necessary to be added or missing completely.

experiment/resources/dispatcher-startup-script-template.sh
```diff
@@ -26,6 +26,7 @@ docker run --rm \
   -e POSTGRES_PASSWORD={{postgres_password}} \
   -e CLOUD_SQL_INSTANCE_CONNECTION_NAME={{cloud_sql_instance_connection_name}} \
   -e DOCKER_REGISTRY={{docker_registry}} \
+  --shm-size=1g \
```
jonathanmetzman, line 29: Fuzzing isn't done in this image btw. What's the default here? Because this instance usually has hundreds of GB of RAM

jonathanmetzman (approved):
lgtm

jonathanmetzman:
Awesome!

oliverchang (approved):

experiment/resources/dispatcher-startup-script-template.sh
Alan32Liu: Thanks! Yep, I can confirm that removing this won't matter.
I am not sure how to check the default instance, though. Where is the value of `docker_registry` defined?

oliverchang: I think the default value is 64mb on my local machine. If this doesn't any fuzzing and hasn't caused any problems, we can just skip this one.

jonathanmetzman:
Merge this when ready dongge

PR #1451, opened 2022-08-11
update afl++ commit id and cleanup variants

vanhauser-thc:


jonathanmetzman:
There are some CI failures. Just ignore them?

vanhauser-thc:
this failures were in before.
I looked at the error and I do not understand it - an afl++ instrumentation function is called that should be inserted at all ... will have a look when I have time.

PR #1452, opened 2022-08-11
Fix saving corpus for a cycle

apach301:
This PR fixes a problem with generating coverage reports for some libFuzzer-based experiments. The feature of the libFuzzer is that it quickly removes files from corpus, so it may be deleted at the time we access it in `_set_corpus_dir_contents`. As a result, the corpus for a specific cycle will not be archived and saved. Apparently, the coverage report generator awaits a corpus from every cycle in direct order, so HTML reports will stop updating even if the next cycle corpuses would be saved.


jonathanmetzman (approved):
LGTM

PR #1462, opened 2022-08-18
Request new focus experiment

Navidem:


service/experiment-requests.yaml
```diff
@@ -20,6 +20,14 @@
 # Please add new experiment requests towards the top of this file.
 #
 
+- experiment: 2022-08-18-libfuzzer-focus
```
jonathanmetzman, line 23: This has the same bug as the last request right?
https://github.com/google/fuzzbench/pull/1462/files#diff-c89bd86bcf4447f199973358896b7a69d69d92f258653f28e77583ee8967dd13L63

Navidem: you mean `type=bug`, I misunderstood your comment then, I thought it fails CI not the actual experiment! will fix it.

PR #1467, opened 2022-08-23
Uploading focus functions map

Navidem:


fuzzers/introspector_driven_focus/focus_map.yaml
```diff
@@ -0,0 +1,357 @@
...
+  std::__1::vector, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>
+  >, bool, long, unsigned long, double, std::__1::allocator, proj_nlohmann::adl_serializer,
+  std::__1::vector<unsigned char, std::__1::allocator<unsigned char> > > const&)
+- osgeo::proj::io::createFromUserInput(std::__1::basic_string<char, std::__1::char_traits<char>,
+  std::__1::allocator<char> > const&, std::__1::shared_ptr<osgeo::proj::io::DatabaseContext>
+  const&, bool, pj_ctx*)
+- proj_create_crs_to_crs_from_pj
+- inv_finalize(PJconsts*, PJ_COORD&)
+quickjs_eval-2020-01-05:
+- JS_CallInternal
```
oliverchang, line 277: These are all the same. Can we de-dupe better?

Navidem: These are reported repeatedly because there are multiple blockers in those functions. Here I am just extracting function names from branch blockers. I can de-dup either when requesting experiment or the step before both works.

oliverchang: Why not just reduce some complexity and do this here?

Navidem: sure! will prune it tomorrow.

jonathanmetzman (commented):
+1 for pruning

fuzzers/introspector_driven_focus/focus_map.yaml
Navidem: Done.

jonathanmetzman (approved):
LGTM

PR #1469, opened 2022-08-24
Reuse the weak references defined in Centipede

Alan32Liu:
Our `weak.c` duplicates Centipede's `weak_sancov_stubs.cc` file, which already has the weak symbols defined.
* Remove our redundant `weak.c`;
* Reuse their `weak_sancov_stubs.cc`.


oliverchang (approved):
nice!

fuzzers/centipede/builder.Dockerfile
```diff
@@ -28,5 +28,4 @@ RUN git clone -n \
     bazel build -c opt :all) && \
   cp "$CENTIPEDE_SRC/bazel-bin/centipede" '/out/centipede'
 
-COPY weak.c /src
-RUN /clang/bin/clang /src/weak.c -c -o /lib/weak.o
+RUN /clang/bin/clang "$CENTIPEDE_SRC/weak_sancov_stubs.cc" -c -o /lib/weak.o
```
oliverchang, line 31: Actually, can we just use bazel to build this? https://github.com/google/centipede/blob/96fff1a32b54802510178e50b67a14ca6779a071/BUILD#L327

fuzzers/centipede/builder.Dockerfile
```diff
@@ -28,4 +28,4 @@ RUN git clone -n \
     bazel build -c opt :all) && \
   cp "$CENTIPEDE_SRC/bazel-bin/centipede" '/out/centipede'
 
-RUN /clang/bin/clang "$CENTIPEDE_SRC/weak_sancov_stubs.cc" -c -o /lib/weak.o
+RUN cp "$CENTIPEDE_SRC/bazel-bin/libweak_sancov_stubs.so" /out/weak.o
```
Alan32Liu, line 31: Here is something I don't fully understand:
In the past, we used to output the `weak.o` generated (by `clang`) to `/lib/` and it was never lost in the container that runs fuzzing.
But here I had to copy `libweak_sancov_stubs.so` to `/out/` instead of `/lib/`, because it will be lost otherwise.
What has changed to cause this difference?

jonathanmetzman: Shared object vs static object.
.so is shared object and needs to be used at runtime. Running is done in another image.

fuzzers/centipede/builder.Dockerfile
```diff
@@ -28,5 +28,4 @@ RUN git clone -n \
     bazel build -c opt :all) && \
   cp "$CENTIPEDE_SRC/bazel-bin/centipede" '/out/centipede'
 
-COPY weak.c /src
-RUN /clang/bin/clang /src/weak.c -c -o /lib/weak.o
+RUN cp "$CENTIPEDE_SRC/bazel-bin/libweak_sancov_stubs.so" /out/weak.o
```
jonathanmetzman, line 31: No please don't change the extension here, this is very wrong (-:

fuzzers/centipede/fuzzer.py
```diff
@@ -27,7 +27,7 @@ def build():
         '-ldl',
         '-lrt',
         '-lpthread',
-        '/lib/weak.o',
+        '/out/weak.o',
```
jonathanmetzman, line 30: .so

fuzzers/centipede/builder.Dockerfile
Alan32Liu: I see, thanks : )

Alan32Liu:
Revert back to building static lib (i.e., `weak.o`) by ourselves, as we prefer static lib and Centipede only builds shared ones with bazel.

jonathanmetzman:
> Revert back to building static lib (i.e., `weak.o`) by ourselves, as we prefer static lib and Centipede only builds shared ones with bazel.

Wait why? I think we should just do whatever centipede does by default.

Alan32Liu:
> > Revert back to building static lib (i.e., `weak.o`) by ourselves, as we prefer static lib and Centipede only builds shared ones with bazel.
> 
> Wait why? I think we should just do whatever centipede does by default, we should not be doing anything custom. Anything needed for centipede should be done upstream.

Because Centipede does not build a *static* equivalent of `weak.o` library via `Bazel` and @oliverchang suggested avoiding using the shared library (`weak.so`) so that it can be encapsulated into the binary.
We are still using `libweak_sancov_stubs.cc` from Centipede, just building the static object by ourselves.

Alan32Liu:
Hope there is no objection to merging this PR?

jonathanmetzman:
> Hope there is no objection to merging this PR?

I would rather just do what centipede does upstream instead of doing our own things downstream.

jonathanmetzman:
Merge it if you want though

Alan32Liu:
> > Hope there is no objection to merging this PR?
> 
> I would rather just do what centipede does upstream instead of doing our own things downstream.

Yeah, I agree that would be more elegant.
But I am not sure if that is possible because of [the way we use `CFLAGS` and `CXXFLAGS`](https://github.com/google/oss-fuzz/pull/8046#discussion_r954393403).
I will merge this for now so that the redundant `weak.c` won't mislead others,  but I am more than happy to come back and fix it if we could find a solution to the issue caused by `C(XX)FLAGS`.

oliverchang:
Yes let's merge this. The reason why we can't use Centipede's way of doing things is more a limitation of the build system, rather than any inherent reason why it's better.

jonathanmetzman:
> Yes let's merge this. The reason why we can't use Centipede's way of doing things is more a limitation of the build system, rather than any inherent reason why it's better.

What do you mean limitation? I'm not really sure what this is. But I think if using shared objects is problematic centipede should stop doing so upstream or offer a compiler wrapper to link against it easily. I don't think it makes sense for centipede to ship something broken

Alan32Liu:
> > Yes let's merge this. The reason why we can't use Centipede's way of doing things is more a limitation of the build system, rather than any inherent reason why it's better.
> 
> What do you mean limitation? I'm not really sure what this is. But I think if using shared objects is problematic centipede should stop doing so upstream or offer a compiler wrapper to link against it easily. I don't think it makes sense for centipede to ship something broken

I reckon `Centipede` builds that shared object for its own purpose, but it happens to be useful for us.
Not sure if it is too much to request them to build a static one just for us, but I am happy to submit a CL to Centipede for that, if necessary.

jonathanmetzman:
>I reckon `Centipede` builds that shared object for its own purpose, but it happens to be useful for us. 

What I'm saying is if we need to do anything extra to build a target beyond the commandline flags centipede says it needs, then centipede is broken and should be fixed.
I assume this counts, because without this our build breaks.

PR #1471, opened 2022-08-24
Check for env before setting it

Navidem:


jonathanmetzman:
I think it makes more sense to do this fix in the PR we are fixing?

jonathanmetzman (approved):
LGTM

PR #1478, opened 2022-08-30
Remove debug print

Navidem:


jonathanmetzman (commented):
I wouldn't even bother with this change. Is this print not helpful because it's redundant?

Navidem:
> I wouldn't even bother with this change. Is this print not helpful because it's redundant?

Yeah, just remembered this was left in the code since I was testing it yesterday!

PR #1479, opened 2022-08-30
Update curl's unsupported fuzzers

Navidem:
Add `centipede_function_filter` to the list of unsupported fuzzers.

### DIFF
--- a/benchmarks/curl_curl_fuzzer_http/benchmark.yaml
+++ b/benchmarks/curl_curl_fuzzer_http/benchmark.yaml
@@ -21,3 +21,4 @@ unsupported_fuzzers:
   - libfuzzer_dataflow
   - libfuzzer_dataflow_load
   - libfuzzer_dataflow_store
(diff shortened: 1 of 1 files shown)


benchmarks/curl_curl_fuzzer_http/benchmark.yaml
```diff
@@ -21,3 +21,4 @@ unsupported_fuzzers:
   - libfuzzer_dataflow
   - libfuzzer_dataflow_load
   - libfuzzer_dataflow_store
```
jonathanmetzman, line 23: Can you also add centipede please?

Navidem: ah! yes, that's the original failing one.

Alan32Liu: Thanks for adding `Centipede` : )

PR #1482, opened 2022-08-31
Clean up CI

jonathanmetzman:

PR #1483, opened 2022-08-31
Adding fuzzers from First, Fuzz the Mutants paper for afl, honggfuzz, eclipser, libfuzzer

kjain14:
Adding all remaining fuzzers first mutate the source code and then run fuzzing as normal for afl, honggfuzz, eclipser, libfuzzer. These fuzzers are from our paper: https://www.ndss-symposium.org/wp-content/uploads/fuzzing2022_23002_paper.pdf.

The fuzzers added are in two categories: 

prioritize - first run FPF prioritization on mutants and use those mutants as part of the fuzzing loop
random - randomly select mutants to be a part of the fuzzing loop

Also added experiment and removed our fuzzers from certain benchmarks, as we only mutate C and C++ source files.


jonathanmetzman:
Please pull again so the build jobs will run (see #1482)

jonathanmetzman:
You still have failures.

jonathanmetzman:
Ping

kjain14:
Sorry, have been sick with COVID. Will address this in the next week or so, depending on recovery time.

jonathanmetzman:
Ooof. Feel better!

kjain14:
Of the four checks that failed in the recent PR, 2 are not our fuzzers (afl_virginmap and afl_maxmap) and the other 2 are failing in the first build of the bug benchmark for aflplusplus (before any of our specific mutation logic is taking place). Wondering if you have any insights into why this may be happening with our aflplusplus mutation variants specifically (does the benchmark work for ordinary aflplusplus?)?

jonathanmetzman:
Sorry for the delay I'm traveling for work

jonathanmetzman:
> Planning to push a version today which will deal with the out of memory errors occurring, by capping the number of files mutated to 200. However, the other errors are occurring before our specific mutant generation/prioritization even begins -- all we are doing here is building the benchmark for the first time to confirm it builds properly and extract the output binary (for example: https://github.com/google/fuzzbench/runs/8118740438?check_suite_focus=true in this run the error occurs on line 89, when all we do is call the build method of aflplusplus). Wondering if you could provide any insight into these cases?

Looks like this is a bug in AFL++. You can always blacklist your fuzzer for a particular benchmark https://github.com/google/fuzzbench/blob/master/benchmarks/arrow_parquet-arrow-fuzz/benchmark.yaml

jonathanmetzman:
> Of the four checks that failed in the recent PR, 2 are not our fuzzers (afl_virginmap and afl_maxmap) and the other 2 are failing in the first build of the bug benchmark for aflplusplus (before any of our specific mutation logic is taking place). Wondering if you have any insights into why this may be happening with our aflplusplus mutation variants specifically, since the benchmark is failing just importing aflplusplus's build method and calling build once?

If the error is not yours, lets just skip the benchmark

PR #1487, opened 2022-09-14
Centipede modes

Alan32Liu:
Add different modes of `Centipede` for coverage experiments:
* [Default](https://github.com/google/fuzzbench/blob/master/fuzzers/centipede/fuzzer.py#L71) (no extra flags)
* `--use_counter_features=1`
* `--use_corpus_weights=0`
* `--use_coverage_frontier=1`
* `--max_corpus_size=1000`
* `--max_corpus_size=10000`
* `--use_cmp_features=0`
* `--path_level=5`
* `--path_level=10`
* `--use_pcpair_features=1`
* `--use_dataflow_features=0`

Probably unnecessary to merge, as they are for temporary experiments and this is only for us to check the settings/flags.

### DIFF
--- a/fuzzers/centipede/runner.Dockerfile
+++ b/fuzzers/centipede/runner.Dockerfile
@@ -13,3 +13,5 @@
 # limitations under the License.
 
 FROM gcr.io/fuzzbench/base-image
+
+COPY --from=base-clang /usr/local/bin/llvm-symbolizer /usr/local/bin/
--- a/fuzzers/centipede_corpus_1000/builder.Dockerfile
+++ b/fuzzers/centipede_corpus_1000/builder.Dockerfile
@@ -1,0 +1,31 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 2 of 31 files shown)


Alan32Liu:
Here is the list of modes of `Centipede` as discussed : ) @kcc @ussuri

jonathanmetzman (approved):
lgtm. if we merge make sure to delete some of these quickly. No need to test this in CI all the time.

fuzzers/centipede/runner.Dockerfile
```diff
@@ -13,3 +13,5 @@
 # limitations under the License.
 
 FROM gcr.io/fuzzbench/base-image
+
```
jonathanmetzman, line 16: Youre not doing this correctly. Need to add a from ... As base-clang.
See https://github.com/google/fuzzbench/blob/d35a6951e2f05b7bab34f9a8c94e381c9512f903/fuzzers/libfuzzer_focus_idx0/runner.Dockerfile#L15

Alan32Liu: Done, thanks!

Alan32Liu:
/gcbrun trial_build.py all --sanitizer address --fuzzing-engine centipede

Issue #1490, opened 2022-09-19
Port to Python 3.10


andreafioraldi:
Python 3.8 is not anymore available on recent Linux distros like Ubuntu 22 and FuzzBench is broken when using Python 3.10 or later versions.

jonathanmetzman:
This is now complete

PR #1493, opened 2022-09-21
Upgrade to python3.10

jonathanmetzman:


jonathanmetzman:
CC @andreafioraldi Fixes https://github.com/google/fuzzbench/issues/1490

presubmit.py
```diff
@@ -241,14 +241,12 @@ def pytype(paths: List[Path]) -> bool:
     """Run pytype on |path| if it is a python file. Return False if it fails
     type checking."""
     paths = [path for path in paths if is_python(path)]
-    if not paths:
```
oliverchang, line 244: Is this (and the print below) an intentional change?

oliverchang (approved):
LGTM, but should we also updated https://google.github.io/fuzzbench/getting-started/prerequisites/#python-programming-language ?

presubmit.py
jonathanmetzman: Yeah. We don't need this early bail out. The print wasn't intended though.

PR #1499, opened 2022-09-27
Request new saturated experiment for centipede vs centipede filter

oliverchang:


Alan32Liu (approved):

service/experiment-requests.yaml
```diff
@@ -19,6 +19,30 @@
 # You can run "make presubmit" to do basic validation on this file.
 # Please add new experiment requests towards the top of this file.
 #
+- experiment: 2022-09-27-centipede-code-saturated
```
Navidem, line 22: Not sure if this long experiment name causes failure.

Navidem (commented):
Thanks!

service/experiment-requests.yaml
oliverchang: yep, thanks :) i already fixed this in https://github.com/google/fuzzbench/pull/1500

PR #1441, opened 2022-08-02
upgrade images

jonathanmetzman:

Issue #1353, opened 2022-03-16
OSS-Fuzz integration does not work for the majority for the projects


florian-university:
I have been trying to identify new interesting benchmarks to evaluate a fuzzer. For this I wrote [some scripts](https://github.com/florian-university/format_fuzzer_experiments) that integrate all OSS-Fuzz experiments to fuzzbench. It basically runs the oss_fuzz_benchmark_integration.py and tests that integration by running `make test-run-afl-{benchmark_name}`. 

After integrating all, only about 40 out of >300 projects succeed.

I'm not quite sure, whether this is because of wrong integration on my part or broken experiments.

Thank you very much.

jonathanmetzman:
Yeah I tested things out and I'm pretty sure the timestamps you've picked won't work because they point to a version of OSS-Fuzz that uses a different Ubuntu version than FuzzBench. I'll try to add a warning about this.
This is blocked on https://github.com/google/fuzzbench/issues/1275

florian-university:
Thanks for the quick answer. I'm currently running the benchmarks on Ubuntu 20.04, so I guess the problem was on my part. I did a test run, where for each experiment I use 3 different commits (the newest, the oldest and one in the middle). In this test run I didn't see much more success in older commits but I guess the problem then is that I'm currently using Ubuntu 20.04.

I can rerun my scripts on Ubuntu 16.04 and only use commits older than 25/8/2021 and update you on the issue in the next days.

jonathanmetzman:
Can you clarify what you mean by "I can rerun my scripts on Ubuntu 16.04", it doesn't matter what your machine is, it matters what docker images are being used. Commits after August 25 2021 use 20.04 and therefore won't work.
You said you didn't see more success with older commits, can you please provide the data on this so I can investigate.
Maybe there's a problem similar to the one above, but at the end of the day the integration script is best effort and devs may need to do some work on their own to get benchmarks working.

florian-university:
By "FuzzBench still only supports 16.04", I thought Fuzzbench needs to be run on 16.04 and did not think about the docker images. Thus I spend some time on getting this to run and didn't check github. (didn't get an email notification for some reason) ¯\_( '-' )_/¯ So please excuse the late answer. 

[...]

PR #1504, opened 2022-09-28
Fix benchmark integration script

jonathanmetzman:
The previous model used a submodule that was never updated and thus the stuff copied from oss-fuzz repo was very old and generally not working. Change this to use the latest oss-fuzz clone.


jonathanmetzman:
CC @maxammann 
Related: https://github.com/google/fuzzbench/issues/1353

jonathanmetzman:
This is blocking #1441

maxammann (approved):
That looks good to me! I was actually quite puzzled why there is a submodule :)

There is ofc never a guarantee that integration works. Most of the time only minior tweaks will be necessary I think.

Did not test this yet though.

maxammann:
Like you said in Discord: It might be a good idea to remove this script. Most of the time this will not easily work. And frankly, without understanding what is going on, how benchmarks are built, which tools they use, just copying it is very dangerous.

I think it would be good to "force" users going the more complicated way of manually creating a new benchmark.

oliverchang (approved):

Issue #1417, opened 2022-07-01
Number of benchmarks in critical difference plot


jiradeto:
I tried to understand how plotting critical difference in fuzzbench works and I see that you use [number of benchmarks](https://github.com/google/fuzzbench/blob/master/analysis/plotting.py#L479) when generating a plot. From what I understood, if N is meant to be the number of datasets then I think this should instead be the _number of trials_ in our case? 

For example given two experiments (trials=10 and trials=100) having the same ranking, should the experiment with 100 trials have smaller critical difference? 

Trials = 10
![10](https://user-images.githubusercontent.com/16756648/176852171-fe593bf2-71b2-43b3-bc3d-e9c014ccbb8f.png)

Trials = 100
![100](https://user-images.githubusercontent.com/16756648/176852250-16968376-fcaf-4af8-970e-d5162e00eaae.png)

wuestholz:
@jonathanmetzman Let me rephrase Jiradet's questions slightly. It seems like the number of trials does not affect the critical difference diagrams while the number of benchmarks does. Is this intentional? Intuitively, it seems like more trials should reduce the critical difference since it increases the confidence in the assigned ranks.

jonathanmetzman:
I found this surprising too, but increasing the trials shouldn't affect the critical difference diagram. The diagram shows the avg rank of the median trial for each benchmark. So while adding trials to a benchmark will not change the result, increasing the number of benchmarks will.
@lszekeres may be able to explain better.

lszekeres:
Thanks for looking into this! Would this change be consistent with the definition of "critical difference" in https://www.jmlr.org/papers/volume7/demsar06a/demsar06a.pdf ?

wuestholz:
@lszekeres Thanks for taking a look! I'm not an expert in statistics, but to me it seems consistent with the definition in the paper you're referencing. This part is from their intro:

![Screenshot 2022-10-05 093726](https://user-images.githubusercontent.com/51156/194006263-89977580-48a1-4e36-925a-06af0a8e0e2f.png)

I think there's some room for interpretation when considering the last two sentences. One might ask the question if the sub-benchmarks we propose (in https://github.com/google/fuzzbench/pull/1509) are "reliable" given that the performance of the fuzzer is only evaluated for a single campaign and there is nondeterminism (due to the source of randomness).

[...]

PR #1509, opened 2022-10-03
Improve how the critical difference diagram is computed

jiradeto:
This PR fixes an issue with how the criticial difference diagram is computed (see https://github.com/google/fuzzbench/issues/1417). Namely, increasing the number of trials per benchmarks (T) does not reduce the critical difference since the computation is based on the average rank across all trials. Increasing the number of trials does increase the confidence in the average rank, but this is not reflected in the critical difference diagram.

This PR proposes the following alternative: instead of computing the average rank per benchmark we group trials for a given benchmark such that each sub-benchmark only contains a single trial per fuzzer. In other words, for each benchmark we create T sub-benchmarks with exactly one trial per fuzzer; i.e., given B benchmarks with T trials each, we end up with BxT sub-benchmarks with 1 trial each. Finally, the critical difference diagram is compute on the sub-benchmarks.

As a result, any increase in the number of trials T will reduce the critical difference.

To demonstrate the effect, let us compare the two approaches for a past fuzzbench run with 20 trials (see [Main experiment](https://www.fuzzbench.com/reports/paper/Main%20Experiment/index.html)). This is the critical difference diagram with the existing approach:

image

This is the critical difference diagram when using the alternative approach:

image

As we can see, the alternative approach actually reduces the critical difference significantly. For instance, there is now a significant difference between fuzzers aflplusplus and honggfuzz whereas the difference was not significant with the existing approach.

### DIFF
--- a/analysis/experiment_results.py
+++ b/analysis/experiment_results.py
@@ -277,6 +277,23 @@ class ExperimentResults:  # pylint: disable=too-many-instance-attributes
                               key=self._relevant_column),
             experiment_level_ranking_function)
 
+    def _ranking_not_aggregated(self, benchmark_level_ranking_function,
+                                experiment_level_ranking_function):
+        df = self._experiment_snapshots_df
+        # group trials by assigning a trial group that ranges from 1 to the maximum
+        # number of trials for each combination of fuzzer and benchmark
+        df['trials_group'] = df.groupby(['fuzzer', 'benchmark']).cumcount() + 1
+        df['trials_group'] = df['trials_group'].astype(str)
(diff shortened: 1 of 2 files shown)


jonathanmetzman:
@lszekeres @laurentsimon WDYT?

lszekeres:
Thanks @jiradeto! This is interesting! We should make sure that we're consistent with the definition of "critical difference" though. I don't remember from the top of my head what it was, but linked the source in https://github.com/google/fuzzbench/issues/1417 --  let's continue discussing there. Thanks!

PR #1510, opened 2022-10-04
Fixing bugs with UM fuzzers

kjain14:
PR to fix build issues and runtime issues along with starting a new experiment for the UM fuzzer fixed.


service/experiment-requests.yaml
```diff
@@ -20,6 +20,12 @@
 # Please add new experiment requests towards the top of this file.
 #
 
+- experiment: 2022-10-04-um
+  description: "UM fuzzer experiment"
+  fuzzers:
+    - aflplusplus
+    - aflplusplus_um_random
```
jonathanmetzman, line 27: Just request all of them here. Let's debug as much as possible, the resources aren't as important.

jonathanmetzman (approved):

service/experiment-requests.yaml
kjain14: oh, I only fixed one of the fuzzers currently, since everything else is just basically copy paste of this with one minor change, everything should work if this works.

PR #1517, opened 2022-10-07
Fix build issues for UM fuzzers

kjain14:
Removing the final build after fuzzing mutants entirely, since this is the point of failure on all UM fuzzers.

Instead on initial build copying output directory and copying it back after final UM build.


jonathanmetzman (approved):
lgtm

PR #1521, opened 2022-10-12
Update ubuntu to `jammy` and python to `3.10`

Alan32Liu:
1. Updates `Python` to `3.10`,
2. Updates `Pip` packages that require a newer version to work under `Python-3.10`,
3. Updates `Ubuntu` to `22.04 (Jammy Jellyfish)`,
4. Fix all errors/warnings caused by using outdated packages.


Alan32Liu:
Not sure if this is a good idea, but I updated Ubuntu to `22.04 (Jammy Jellyfish)` instead of `20.04 (Focal Fossa)` (which is planned), because:
1. We want `Python-3.10`, which is the default in `22.04` but not in `20.04`;
2. The default `Python-3.10` in `20.04` works smoothly with the current code, while the `Python-3.10` I built manually in `20.04` caused a few errors (which should be solvable).
3. Both `22.04` and. `20.04` are fully supported, `22.04` will last longer.

jonathanmetzman:
Sorry Dongge, I don't think we want to upgrade to Jammy. We should upgrade to focal.

jonathanmetzman:
We should just build Python from source if we need a certain version.
Maybe you can take over this PR https://github.com/google/fuzzbench/pull/1441/files btw

Alan32Liu:
> Sorry Dongge, I don't think we want to upgrade to Jammy. We should upgrade to focal.

OK, I will stick to focal, then : )
Would you mind explaining a bit more to me about why would we prefer focal over jammy?
Is it because focal has been released for a while and is more widely supported, or because we want to make it consistent with OSS-Fuzz?
Thanks!

jonathanmetzman:
> > Sorry Dongge, I don't think we want to upgrade to Jammy. We should upgrade to focal.
> 
> OK, I will stick to focal, then : ) Would you mind explaining a bit more to me about why would we prefer focal over jammy? Is it because focal has been released for a while and is more widely supported, or because we want to make it consistent with OSS-Fuzz? Thanks!

It's to be consistent with oss-fuzz yeah. That's where most of our benchmarks come from. That's also the platform we care most about (e.g. if centipede worked on jammy but not focal, it would be useless to us).

Alan32Liu:
> > > Sorry Dongge, I don't think we want to upgrade to Jammy. We should upgrade to focal.
> > 
> > 
> > OK, I will stick to focal, then : ) Would you mind explaining a bit more to me about why would we prefer focal over jammy? Is it because focal has been released for a while and is more widely supported, or because we want to make it consistent with OSS-Fuzz? Thanks!
> 
> It's to be consistent with oss-fuzz yeah. That's where most of our benchmarks come from. That's also the platform we care most about (e.g. if centipede worked on jammy but not focal, it would be useless to us).

I see : ) Thanks!
I will make use of the PR you shared and (hopefully) make it ready by today : )

jonathanmetzman:
> > > > Sorry Dongge, I don't think we want to upgrade to Jammy. We should upgrade to focal.
> > > 
> > > 
> > > OK, I will stick to focal, then : ) Would you mind explaining a bit more to me about why would we prefer focal over jammy? Is it because focal has been released for a while and is more widely supported, or because we want to make it consistent with OSS-Fuzz? Thanks!
> > 
> > 
> > It's to be consistent with oss-fuzz yeah. That's where most of our benchmarks come from. That's also the platform we care most about (e.g. if centipede worked on jammy but not focal, it would be useless to us).
> 
> I see : ) Thanks! I will make use of the PR you shared and (hopefully) make it ready by today : )

I think we will need to just junk some bug benchmarks.

Alan32Liu:
> > > > > Sorry Dongge, I don't think we want to upgrade to Jammy. We should upgrade to focal.
> > > > 
> > > > 
> > > > OK, I will stick to focal, then : ) Would you mind explaining a bit more to me about why would we prefer focal over jammy? Is it because focal has been released for a while and is more widely supported, or because we want to make it consistent with OSS-Fuzz? Thanks!
> > > 
> > > 
> > > It's to be consistent with oss-fuzz yeah. That's where most of our benchmarks come from. That's also the platform we care most about (e.g. if centipede worked on jammy but not focal, it would be useless to us).
> > 
> > 
> > I see : ) Thanks! I will make use of the PR you shared and (hopefully) make it ready by today : )
> 
> I think we will need to just junk some bug benchmarks.

Yep! I recall you mentioned having another FuzzBench2.0 and stop supporting some benchmarks.
I will pay attention to that and list the benchmarks that appear to be buggy.

jonathanmetzman:
> > > > > > Sorry Dongge, I don't think we want to upgrade to Jammy. We should upgrade to focal.
> > > > > 
> > > > > 
> > > > > OK, I will stick to focal, then : ) Would you mind explaining a bit more to me about why would we prefer focal over jammy? Is it because focal has been released for a while and is more widely supported, or because we want to make it consistent with OSS-Fuzz? Thanks!
> > > > 
> > > > 
> > > > It's to be consistent with oss-fuzz yeah. That's where most of our benchmarks come from. That's also the platform we care most about (e.g. if centipede worked on jammy but not focal, it would be useless to us).
> > > 
> > > 
> > > I see : ) Thanks! I will make use of the PR you shared and (hopefully) make it ready by today : )
> > 
> > 
> > I think we will need to just junk some bug benchmarks.
> 
> Yep! I recall you mentioned having another FuzzBench2.0 and stop supporting some benchmarks. I will pay attention to that and list the benchmarks that appear to be buggy.

to be clear, I mean the benchmarks that are used to benchmark bug finding ability.

jonathanmetzman:
but yeah these will be buggy too :-)

PR #1522, opened 2022-10-12
Creating UM Experiments for First, Fuzz the Mutants Paper

kjain14:
In order to avoid fuzzbench rate limits and timing issues, creating 10 experiments each with two fuzzers to be merged at the end.

This should be the full set of experiments that we need for our paper if everything builds and runs successfully. 

Question: Can we set the retries to 3 in order to mitigate some of the build flakiness we have seen?

### DIFF
--- a/service/experiment-requests.yaml
+++ b/service/experiment-requests.yaml
@@ -20,61 +20,61 @@
 # Please add new experiment requests towards the top of this file.
 #
 
-- experiment: 2022-10-12-um-full_10
+- experiment: 2022-10-12-um-full-10
   description: "UM fuzzer experiment (eclipser_um_prioritize, eclipser_um_parallel)"
   fuzzers:
     - eclipser_um_prioritize
     - eclipser_um_parallel
 
-- experiment: 2022-10-12-um-full_9
+- experiment: 2022-10-12-um-full-9
   description: "UM fuzzer experiment (eclipser, eclipser_um_random)"
   fuzzers:
     - eclipser
     - eclipser_um_random
 
-- experiment: 2022-10-12-um-full_8
+- experiment: 2022-10-12-um-full-8
   description: "UM fuzzer experiment (libfuzzer_um_prioritize, libfuzzer_um_parallel)"
   fuzzers:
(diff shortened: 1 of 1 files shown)


jonathanmetzman (approved):
lgtm

jonathanmetzman:
Yeah I can add back retries. But you should really check on the builds sooner rather than later. I'll post here once they are started

jonathanmetzman:
Frankly running 10 experiments at the same time is more than we are used to (and we're actually running some of your older ones) so I'm really not sure how I will proceed here. I might kick off half of these right away and hope for the best

kjain14:
We should be good to kill other experiments since these 10 are the full set needed for the paper.

jonathanmetzman:
Are you sure. Do we have reason to think this will succeed where https://www.fuzzbench.com/reports/experimental/2022-10-07-um-full/index.html failed? Note that it may be missing trials simply because the trials failed to complete.

kjain14:
Yeah, there have been some PRs since that experiment fixing build issues.

jonathanmetzman:
I added back retries. https://github.com/google/fuzzbench/commit/55754f765eb58d1a8061a939f404e69f1dea4841

jonathanmetzman:
Yeah I'm gonna kick these off soon

jonathanmetzman:
I'm pretty sure there are more benchmarks that are totally broken. woff2 is one. You should test these locally

kjain14:
I don't see a build file for woff2, however I am seeing errors with non aflplusplus fuzzers (we only were testing aflplusplus). Will look into this tonight and first thing tomorrow.

kjain14:
Did we start anything other than just the first experiment? Would be helpful if I can get all the failing builds at once tomorrow so I can take a look through and do analysis.

jonathanmetzman:
Yeah gsutil ls gs://fuzzbench-data/*um-full-*/build-logs/

jonathanmetzman:
Sorry github's markdown screwed up my previous comment: `gsutil ls gs://fuzzbench-data/*um-full-*/build-logs/`

PR #1525, opened 2022-10-14
Update libFuzzer (and its variations)

Alan32Liu:
1. Updated `libFuzzer` to the latest commit,
2. Updated `libFuzzer`'s dataflow variations to the latest commit.

### DIFF
--- a/.github/workflows/fuzzers.yml
+++ b/.github/workflows/fuzzers.yml
@@ -107,6 +107,15 @@ jobs:
           - afl_maxmap
           - introspector_driven_focus
           - libfuzzer_focus_idx0
+          - libfuzzer_focus_idx1
+          - libfuzzer_focus_idx2
+          - libfuzzer_focus_idx3
+          - libfuzzer_focus_idx4
+          - libfuzzer_focus_idx5
+          - libfuzzer_focus_idx6
+          - libfuzzer_focus_idx7
+          - libfuzzer_focus_idx8
+          - libfuzzer_fork_parallel
           - centipede_function_filter
 
         benchmark_type:
--- a/.github/workflows/fuzzers.yml
+++ b/.github/workflows/fuzzers.yml
@@ -106,15 +106,6 @@ jobs:
           - afl_virginmap
           - afl_maxmap
           - introspector_driven_focus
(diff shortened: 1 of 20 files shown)


Alan32Liu:
There are other variations of `libFuzzer`(e.g., `libfuzzer_focus_*`, `libfuzzer_um_*`), would it be a good idea to update them as well?
Thanks!

Alan32Liu:
TODOs (Please let me know if I missed anything):
1. Wait for 'fuzz the mutants' experiments to finish.
2. Compare the performance of `libFuzzer` and `Entropic` (maybe also `Centipede`?),
3. Replace `Entropic` with `Centipede` in CI tests.

jonathanmetzman (approved):
LGTM.
Thanks Dongge!
Can you add an experiment request here and try to remember to merge this next Thursday?
After that experiment finishes we can remove entropic and add centipede to core-fuzzers

jonathanmetzman:
> There are other variations of `libFuzzer`(e.g., `libfuzzer_focus_*`, `libfuzzer_um_*`), would it be a good idea to update them as well? Thanks!

Yes definitely! Thank you for proactively taking care of this!

.github/workflows/fuzzers.yml
```diff
@@ -107,6 +107,15 @@ jobs:
           - afl_maxmap
           - introspector_driven_focus
           - libfuzzer_focus_idx0
+          - libfuzzer_focus_idx1
```
jonathanmetzman, line 110: We don't want this. It will clog up CI.
actually @Navidem can we get rid of this or will you need more experiments?

jonathanmetzman (approved):
LGTM

Alan32Liu:
There are some CI failures with `libfuzzer_dataflow*` on `bug` benchmarks, I give up fixing them after some attempts, as they are only fixable if we separate flags for building and linking.
I reckon we will probably stop supporting `libFuzzer_dataflow` in the long-term?
The main problem is that `Centipede` will also fail on them for the same reason.

Alan32Liu:
`Centipede` did not fail on them in CIs because it is not built with them in CIs.

Issue #1512, opened 2022-10-04
Use private pools for faster builds


jonathanmetzman:
FuzzBench is currently bottlenecked on 30 concurrent builds.
We should aim to do all builds in parallel by using private pools like we do in oss-fuzz.
https://cloud.google.com/build/docs/private-pools/run-builds-in-private-pool

I think this would be an easy early issue as experiment/builder.py can be run locally by passing the right env vars.
@Navidem @Alan32Liu do you want to try this?
CC @oliverchang

jonathanmetzman:
I did this in https://github.com/google/fuzzbench/pull/1523 but still need to document it/make configurable for open source users.

Alan32Liu:
I can probably help with it today : )

jonathanmetzman:
Awesome thanks! (this isn't urgent btw).
I'd stop hardcoding this https://github.com/google/fuzzbench/blob/master/experiment/build/gcb_build.py#L34 and instead use a value passed in the yaml config.
Actually, that config can be optional and if they don't provide one dont pass the worker pool arg: https://github.com/google/fuzzbench/blob/master/experiment/build/gcb_build.py#L87
Also we should update the number of concurrent builds to be configurable (default to 30 maybe) https://github.com/google/fuzzbench/blob/master/experiment/build/builder.py#L44 since this will be extremely project dependant
Finally, we should update our experiment-config to support these new options https://github.com/google/fuzzbench/blob/master/service/experiment-config.yaml

Alan32Liu:
Working on this now, the main blocker is that I am not sure how to pass values from the config file to `gcb_build.py`.
I guess I will let [`read_and_validate_experiment_config()`](https://github.com/google/fuzzbench/blob/master/experiment/run_experiment.py#L64) parse that value first, but not sure what is the best way to access that config in `gcb_build.py`.
Will continue on this later/next week.

jonathanmetzman:
This is done now.

PR #1535, opened 2022-10-23
More configurable builds

Alan32Liu:
Address #1512:
1. Make the worker pool configurable;
2. Make the number of  concurrent builds configurable.


experiment/build/gcb_build.py
```diff
@@ -84,7 +84,9 @@ def _build(
 
         # Use "s" suffix to denote seconds.
         timeout_arg = '--timeout=%ds' % timeout_seconds
-        worker_pool_arg = f'--worker-pool={WORKER_POOL_NAME}'
+        worker_pool_name = yaml_utils.read(experiment_config_path()).get(
+            'pool_name', DEFAULT_WORKER_POOL_NAME)
```
Alan32Liu, line 88: I guess we can get `pool_name` in this way during experiments, but not in unit tests?
IIUC, unit tests do not copy the `experiment-config.yaml` to this directory like in experiments, hence CI fails on this line because it cannot find that file.
An alternative solution that will work for both is to carry that information all the way here. I am working on that now, but please let me know if there is a better solution.

experiment/dispatcher.py
```diff
@@ -156,7 +160,8 @@ def dispatcher_main():
     trials = build_images_for_trials(experiment.fuzzers, experiment.benchmarks,
                                      experiment.num_trials,
                                      experiment.preemptible,
-                                     experiment.config['concurrent_builds'])
+                                     experiment.config['concurrent_builds'],
+                                     experiment.config['worker_pool_name'])
```
Alan32Liu, line 164: 1. Won't this line and the line above fail if these two keys do not exist?
2. I reckon `concurrent_builds` was not a mandatory field in the configuration YAML, why won't the old code break?

Alan32Liu: Also, the current implementation passes `concurrent_builds` and `worker_pool_name` to all functions along the way to code that uses them, which makes those functions bloated.
Alternatively, I can prepare them in [experiment/run_experiment.py](https://github.com/google/fuzzbench/pull/1535/files#diff-f29fa4fe4e55f233de1357ff219fd6df4fc408fcbdd0406f3fb5a022e11c670b) and  define them as ENV variables in [`dispatcher-startup-script-template.sh`](https://github.com/google/fuzzbench/blob/master/experiment/resources/dispatcher-startup-script-template.sh).
However, that will globally expose those two variables.
3. Which one would you prefer?

experiment/run_experiment.py
```diff
@@ -249,7 +261,8 @@ def start_experiment(  # pylint: disable=too-many-arguments
     validate_experiment_name(experiment_name)
     validate_benchmarks(benchmarks)
 
-    config = read_and_validate_experiment_config(config_filename)
+    config = read_and_validate_experiment_config(config_filename,
+                                                 concurrent_builds)
```
Alan32Liu, line 265: I presume this `concurrent_builds` (from the arguments of `start_experiment()`) is the variable we want to configure?

experiment/run_experiment.py
```diff
@@ -128,6 +130,16 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
...
 
+    # Concurrent build can be defined either via parameters or in a config file.
+    # If both define it, they need to be consistent.
+    concurrent_builds_yaml = config.get('concurrent_builds')
+    if concurrent_builds_yaml and concurrent_builds_arg and (
+            concurrent_builds_yaml != concurrent_builds_arg):
+        valid = False
+        logs.error('Inconsistent requirements of number of concurrent builds: '
+                   f'File {config_filename} requests {concurrent_builds_yaml}, '
+                   f'but argument requests {concurrent_builds_arg}.')
```
Alan32Liu, line 141: Is this check necessary?

experiment/run_experiment.py
```diff
@@ -128,6 +130,16 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
...
+
```
Alan32Liu, line 142: Do we also need a check for `worker_pool_name`? For example, should it be in the same zone as `cloud_compute_zone`?

experiment/dispatcher.py
Alan32Liu: 4. I thought the old code already ensures we can config `concurrent_builds`, as it parses `concurrent_builds` from YAML, and when it is `None`, [`builder.py` uses the default value](https://github.com/google/fuzzbench/pull/1535/files#diff-30d6caa95ef383e7b9176f3f0d64ee98f6dd5b1c466ab03b44492509763661beL114).

Alan32Liu (commented):
Although this passes the tests, there are several things that I am not certain.

experiment/run_experiment.py
jonathanmetzman: Nope. Because it should only be a config.

experiment/run_experiment.py
jonathanmetzman: No dont bother with that.

experiment/run_experiment.py
jonathanmetzman: We should only get this from the config file. Sorry for not being clear about this.

experiment/build/builder.py
```diff
@@ -111,13 +111,15 @@ def build_measurer(benchmark: str) -> bool:
 
 def build_all_measurers(
         benchmarks: List[str],
-        num_concurrent_builds: int = DEFAULT_MAX_CONCURRENT_BUILDS
-) -> List[str]:
+        num_concurrent_builds: int = DEFAULT_MAX_CONCURRENT_BUILDS,
```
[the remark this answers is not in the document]
Alan32Liu: Yes! That is actually what I had before seeing these default arguments.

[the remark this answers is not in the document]
Alan32Liu: Just to double-check:
1. We check if these values are defined in the `experiment-config.yaml` file in `run_experiment`,
2. When these values are needed by other code in a different docker container, we pass them as `ENV` variables, and always access them via `os.getenv()`.

experiment/dispatcher.py
Alan32Liu: > I think you introduced it right? 

Nah, the credit belongs to [8692e](https://github.com/google/fuzzbench/commit/8692edd60879f64fc010199676a6cd7cba5f7180) and [66980](https://github.com/google/fuzzbench/commit/669805b3b107c1b13c27c636a3e964ad929df0ba).

I was curious why it has not broken because 
1. This `experiment.config` is simply [a dictionary constructed with the content of `experiment-config.yaml`](https://github.com/google/fuzzbench/blob/5624dfea608c1ff55a11f82ac8f724877b7c85fa/experiment/dispatcher.py#L96). It does not set the default value.
2. It uses a direct key query (`experiment.config['concurrent_builds']`) instead of the `get` method (`experiment.get('concurrent_builds')`). If that key does not exist, the key query will break instead of returning `None`.
3. Many of our `experiment-config.yaml` does not have `concurrent_builds`, for example, [this one](https://github.com/google/fuzzbench/blob/5624dfea608c1ff55a11f82ac8f724877b7c85fa/service/experiment-config.yaml).
So I was curious why it has not broken so far.

Alan32Liu:
Not directly related to this PR, but I reckon [the necessity check](https://github.com/google/fuzzbench/blob/5624dfea608c1ff55a11f82ac8f724877b7c85fa/experiment/run_experiment.py#L88) can be done automatically with [`argparser`'s `required`option](https://docs.python.org/3/library/argparse.html#required), while [these type checks](https://github.com/google/fuzzbench/blob/5624dfea608c1ff55a11f82ac8f724877b7c85fa/experiment/run_experiment.py#L94) can be simplified and replaced with [`Python`'s `argparser`](https://docs.python.org/3/library/argparse.html#type).

experiment/build/gcb_build.py
```diff
@@ -84,7 +82,8 @@ def _build(
 
         # Use "s" suffix to denote seconds.
         timeout_arg = '--timeout=%ds' % timeout_seconds
-        worker_pool_arg = f'--worker-pool={WORKER_POOL_NAME}'
+        worker_pool_arg = (
+            f'--worker-pool={worker_pool_name or DEFAULT_WORKER_POOL_NAME}')
```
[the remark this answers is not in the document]
Alan32Liu: I see... Does it mean it is ok to have a cloud build without specifying worker pool?

Alan32Liu:
CI tests passed, but shall we add some testcases?

experiment/run_experiment.py
```diff
@@ -59,36 +59,54 @@
 _OSS_FUZZ_CORPUS_BACKUP_URL_FORMAT = (
     'gs://{project}-backup.clusterfuzz-external.appspot.com/corpus/'
     'libFuzzer/{fuzz_target}/public.zip')
+DEFAULT_MAX_CONCURRENT_BUILDS = 30
```
jonathanmetzman, line 62: rename: DEFAULT_CONCURRENT_BUILDS

experiment/run_experiment.py
```diff
@@ -59,36 +59,54 @@
...
+    optional_params = worker_config
     required_params = int_params.union(filestore_params).union(docker_config)
     bool_params = {'private', 'merge_with_nonprivate'}
 
     local_experiment = config.get('local_experiment', False)
     snapshot_period = config.get('snapshot_period',
                                  experiment_utils.DEFAULT_SNAPSHOT_SECONDS)
-    if not local_experiment:
+    if local_experiment:
+        optional_params = optional_params.union(cloud_config)
```
jonathanmetzman, line 92: ? I don't think this is right? cloud_config isn't supposed to be used in local_experiments.

experiment/run_experiment.py
```diff
@@ -59,36 +59,54 @@
...
         required_params = required_params.union(cloud_config)
 
     valid = True
     if 'cloud_experiment_bucket' in config or 'cloud_web_bucket' in config:
         logs.error('"cloud_experiment_bucket" and "cloud_web_bucket" are now '
                    '"experiment_filestore" and "report_filestore".')
-
-    for param in required_params:
-        if param not in config:
+    missing_params = required_params - config.keys()
```
jonathanmetzman, line 100: Do two loops instead of this.
Go through missing_params and set valid = False and log
then go through optional_params - config.keys() and log that hey are missing.

experiment/run_experiment.py
```diff
@@ -59,36 +59,54 @@
...
     """Reads |config_filename|, validates it, finds as many errors as possible,
     and returns it."""
     config = yaml_utils.read(config_filename)
+    set_default_config_values(config)
     filestore_params = {'experiment_filestore', 'report_filestore'}
     cloud_config = {'cloud_compute_zone', 'cloud_project'}
     docker_config = {'docker_registry'}
-    string_params = cloud_config.union(filestore_params).union(docker_config)
-    int_params = {'trials', 'max_total_time'}
+    worker_config = {'worker_pool_name'}
```
jonathanmetzman, line 80: worker_pool_name should be part of cloud_config.

experiment/run_experiment.py
```diff
@@ -59,36 +59,54 @@
...
+    string_params = cloud_config.union(filestore_params).union(
+        docker_config).union(worker_config)
+    int_params = {'trials', 'max_total_time', 'concurrent_builds'}
```
jonathanmetzman, line 83: Can you change the name of concurrent_builds to concurrent_cloud_builds since it's not repected for local builds.

jonathanmetzman (commented):
Thanks,
This validation code is really ugly.
Yeah tests would be good.

experiment/run_experiment.py
Alan32Liu: Done, but I am a bit confused here.
Do we still have a parameter to set [concurrent build value for local builds](https://github.com/google/fuzzbench/blob/da05da0dcb3680deee9d894d1bfa61edbbf4a90c/docs/running-a-local-experiment/running_a_local_experiment.md?plain=1#L95) or shall I remove that line from our doc?
Allowing concurrent local builds sounds like a useful feature to me, maybe we could use that value for both local and cloud builds?

experiment/run_experiment.py
Alan32Liu: Done, I also split the function to make it look better.

experiment/build/gcb_build.py
jonathanmetzman: I didn't design config very well in fuzzbench. some config is put in the env other is read into memory and passed around. I think an env var is good for something like this, definitely don't read that file again please.

experiment/run_experiment.py
jonathanmetzman: I think I was confused sorry, I thought it wasn't respected for local. Ignore this suggestion please. This is definitely a useful feature for local.

Alan32Liu:
Added a testcase for `worker_pool`.
I noticed that there is no existing test for command line parameters (e.g. `concurrent_build`), so it is probably better to add them together in a different PR.
In addition, we are [validating the argument parameters manually](https://github.com/google/fuzzbench/blob/da05da0dcb3680deee9d894d1bfa61edbbf4a90c/experiment/run_experiment.py#L616), which is probably better done with [`argparser`'s type](https://docs.python.org/3/library/argparse.html#type).
I will do that in the new PR.

experiment/build/gcb_build.py
```diff
@@ -30,10 +31,6 @@
 # Maximum time to wait for a GCB config to finish build.
 GCB_BUILD_TIMEOUT = 13 * 60 * 60  # 4 hours.
 
-# TODO(metzman): Make configurable.
-WORKER_POOL_NAME = (
-    'projects/fuzzbench/locations/us-central1/workerPools/buildpool')
```
Alan32Liu, line 35: We require all cloud_params to be [in lower cases](https://github.com/google/fuzzbench/blob/da05da0dcb3680deee9d894d1bfa61edbbf4a90c/experiment/run_experiment.py#L100).
Can we expect `work_pool_name` will be in lower cases?
For example, our previous default value contains an upper-case letter.
Now that this default value is replaced by a field in the config file, do we know for sure that that field will not use upper-case letters?

Alan32Liu:
I reckon this PR is ready, please have a final look at it and let me know if there is anything else I can do.
Thanks!

experiment/build/gcb_build.py
jonathanmetzman: Yeah we can't assume theyre all lowercase.

experiment/build/gcb_build.py
```diff
@@ -93,9 +88,12 @@ def _build(
             str(utils.ROOT_DIR),
             config_arg,
             timeout_arg,
-            worker_pool_arg,
         ]
 
+        if os.getenv('WORKER_POOL_NAME'):
```
jonathanmetzman, line 93: Just do
```
worker_pool_name = os.getenv('WORKER_POOL_NAME)
if worker_pool_name:
...
```

experiment/run_experiment.py
```diff
@@ -59,52 +59,61 @@
...
+                                bool_params: Set[str],
+                                filestore_params: Set[str],
+                                local_experiment: bool) -> bool:
+    """Validates if |params| types and formats in |config| are correct."""
+
+    valid = True
+    for param, value in config.items():
         if param in string_params and (not isinstance(value, str) or
-                                       value != value.lower()):
+                                       not value.islower()):
```
jonathanmetzman, line 104: yeah i think we jus tneed to get rid of this.

Alan32Liu:
I reckon this is the same as we discussed.
In the next PR, I will:
1. Merge the config `YAML` values and the command line parameters to the intermediate config file.
2. Create a separate module to set `ENV` values with the intermediate config file.
3. Use the module to access the settings from `ENV` values.

jonathanmetzman:
> I reckon this is the same as we discussed. In the next PR, I will:
> 
> 1. Merge the config `YAML` values and the command line parameters to the intermediate config file.
> 2. Create a separate module to set `ENV` values with the intermediate config file.
> 3. Use the module to access the settings from `ENV` values.
> 
> Please let me know if I missed anything, thanks! @jonathanmetzman

I think this is it.

experiment/run_experiment.py
```diff
@@ -59,80 +59,122 @@
...
+# pylint: disable=too-many-locals
+def read_and_validate_experiment_config(config_filename: str) -> Dict:
+    """Reads |config_filename|, validates it, finds as many errors as possible,
+    and returns it."""
+    # Reads config from file.
+    config = yaml_utils.read(config_filename)
+
+    # Validates config contains all the required parameters.
+    local_experiment = config.get('local_experiment', False)
+    # param_name: (mandatory, type, lowercase, startswith).
```
jonathanmetzman, line 153: nit: used namedtuple for this.

jonathanmetzman (commented):
I think the validation code is kind of ugly but it was always like that and it's relatively unimportant

jonathanmetzman (approved):
lgtm with nits

Alan32Liu:
It just occur to me that I have not updated the config YAML accordingly.
I think that is why our experiment with the new version of `libFuzzer` did not launch.
Let me fix that in a new PR later.

PR #1537, opened 2022-10-24
Remove a bunch of fuzzers from CI testing

jonathanmetzman:


Alan32Liu (approved):
LGTM, but will this cause `presubmit` errors?
For example, I noticed that `afl_no_favored` is still in `fuzzers/`, will `presubmit` complain about it is not in the fuzzers list?

Alan32Liu:
Although the fuzzers' build tests are not shown in CI tests here, they seem to be running [in the background](https://github.com/google/fuzzbench/actions/runs/3316198118).

jonathanmetzman:
> LGTM, but will this cause `presubmit` errors? For example, I noticed that `afl_no_favored` is still in `fuzzers/`, will `presubmit` complain about it is not in the fuzzers list?

It causes presubmit errors when those fuzzers are modified. I frankly don't know what to do about this. Maybe just remove these fuzzers.

jonathanmetzman:
> Although the fuzzers' build tests are not shown in CI tests here, they seem to be running [in the background](https://github.com/google/fuzzbench/actions/runs/3316198118).

They were run on an earlier commit because they are only run when fuzzers/ or benchmarks/ is changed.

PR #1542, opened 2022-11-06
Fix Bug in UM Prioritize and add Muttfuzz

kjain14:
Making the following changes:
1. Adding a version of our "First, Fuzz the Mutants" paper fuzzers that works at the bytecode level. 
2. Also fixing bug in um_prioritize, whereby copying does not occur when moving between mutants.
3. Add two new experiments, one for um_prioritize and one for our new muttfuzz fuzzer


kjain14:
Failed build jobs are either not the changed fuzzers or failing on the build of the benchmark in aflplusplus (before any of our code takes place).

PR #1541, opened 2022-11-04
Fix & update benchmarks and fuzzers

Alan32Liu:
1. Fix and update (to the latest tag/release) the (nontrivial) benchmarks and fuzzers to adapt to `Ubuntu:20.04` and `Python3.10.8`.
2. Delete the trivial/buggy benchmarks and fuzzers.
Base: #1526

# Benchmarks
'[x]' means it can build and run under the new environment.

## Support
* [x] All benchmarks in #1441.
  * [x] [`systemd_fuzz-link-parser`](https://github.com/google/fuzzbench/pull/1541/commits/4ec585808f90d2ce0f6a348815a1af0761ef5c90).
  * [x] [`systemd_fuzz-varlink`](https://github.com/google/fuzzbench/pull/1541/commits/4ec585808f90d2ce0f6a348815a1af0761ef5c90).
  * [x] [`njs_njs_process_script_fuzzer`](https://github.com/google/fuzzbench/pull/1541/commits/5720ea0e26db8459c755b3f9e771b63e2bb090ea).
  * [x] [`libxslt_xpath`](https://github.com/google/fuzzbench/pull/1526/commits/cd60b7efcb2b8281031d5a62ae1471a8b05bd6e5).
  * [x] [`openh264_decoder_fuzzer`](https://github.com/google/fuzzbench/commit/07008f194b7794c5a30b21bc480b3c4c7a79e07a).
  * [x] [`libarchive_libarchive_fuzzer`](https://github.com/google/fuzzbench/commit/da41386fafb75f13c421dbb591e34dee99e2a529).
  * [x] [`ffmpeg_ffmpeg_demuxer_fuzzer`](https://github.com/google/fuzzbench/commit/91d3b2b05694ee492771976b447102c68b9c83de).
    * I `apt install` some packages that were previously built manually, to make its `build.sh` consistent with `OSS-Fuzz`, hope that would be alright.
  * [x] [`mruby-2018-05-23`](https://github.com/google/fuzzbench/commit/53aa92b5f5affaab11ab729f8f0387f8028fe72e).
  * [x] [`openssl_x509`](https://github.com/google/fuzzbench/commit/23e39f1eafa3672a4d822c8c39923e3361baab3f).
  * [x] [`php_php-fuzz-parser-2020-07-25`](https://github.com/google/fuzzbench/commit/52fc22b1e4c8317de5d0bd559141b833c4dd01a6).
[...]

PR #1543, opened 2022-11-08
symcc: fix build

adrianherrera:
Use the latest Rust. Now SymCC builds.


Alan32Liu:
Thanks, @adrianherrera!
I am also working on supporting `AFL++-SymCC` on `Ubuntu:20.04` (#1541 ), but there are failures like `undefined reference 'size_t'` when building it with benchmark programs.
Do you happen to know how to fix them? Thanks : )

adrianherrera:
Hmm, I did not encounter that. But I was building with the current Docker images, which use Ubuntu 16.04

fuzzers/symcc_aflplusplus/builder.Dockerfile
```diff
@@ -35,7 +35,14 @@ RUN cd /afl && unset CFLAGS && unset CXXFLAGS && \
     cp utils/aflpp_driver/libAFLDriver.a /
 
 # Install the packages we need.
-RUN apt-get install -y ninja-build flex bison python zlib1g-dev cargo 
+RUN apt-get install -y ninja-build flex bison python zlib1g-dev
```
jonathanmetzman, line 38: Do an apt-get update first or there can  be issues.

adrianherrera: Done.

PR #1548, opened 2022-11-11
Add worker pool name to config

Alan32Liu:
A follow-up of #1535: Add `worker_pool_name` to config file.


Alan32Liu:
TDR. Merged for experiments.

PR #1551, opened 2022-11-11
Make 2 optional config parameters recognizable

Alan32Liu:
Fix #1535.


Alan32Liu:
TBR. Merged now to unblock experiments.

PR #1552, opened 2022-11-11
Re-relaunch these two experiments

Alan32Liu:
#1549 relaunched 3 experiments but was unsuccessful because of a code bug: two valid parameters in the config file were not recognized.
#1551 fixed that bug.

I re-relaunched the `libFuzzer` experiment manually to confirm it works, and re-relaunch the other two here.


Alan32Liu:
TBR. Merged now to kickoff experiments.

jonathanmetzman:
lgtm

PR #1554, opened 2022-11-15
Log statistics about restart/omit trials

Alan32Liu:
As discussed, logging these statistics has two benefits:
1. Help us confirm that all missing data are intended.
2. Give us the statistics needed to find a better solution to missing data.

### DIFF
--- a/experiment/scheduler.py
+++ b/experiment/scheduler.py
@@ -219,6 +219,8 @@ class TrialInstanceManager:  # pylint: disable=too-many-instance-attributes
     def __init__(self, num_trials, experiment_config):
         self.experiment_config = experiment_config
         self.num_trials = num_trials
+        self.num_omitted_trials = 0
+        self.num_converted_trials = 0
 
         # Bound for the number of nonpreemptibles we can start if the experiment
         # specified preemptible_runners.
@@ -354,6 +356,15 @@ class TrialInstanceManager:  # pylint: disable=too-many-instance-attributes
             if self.can_start_preemptible():
                 # See if we can replace with a preemptible.
                 replacements.append(replace_trial(trial, preemptible=True))
+
+                self.num_converted_trials += 1
+                logs.info(
+                    'Restarting a preemptible trial as a preemptible one: '
+                    'Benchmark-Fuzzer pair: %s-%s.'
+                    'Trial ID %d, restarted %3.2f%% trials (%d/%d) as '
+                    'preemptilbes so far.', trial.benchmark, trial.fuzzer,
+                    trial, self.num_converted_trials / self.num_trials * 100,
+                    self.num_converted_trials, self.num_trials)
                 continue
 
             if self.can_start_nonpreemptible(nonpreemptible_starts):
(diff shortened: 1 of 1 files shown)


experiment/scheduler.py
```diff
@@ -354,15 +356,41 @@ def _get_preempted_replacements(self,
             if self.can_start_preemptible():
                 # See if we can replace with a preemptible.
                 replacements.append(replace_trial(trial, preemptible=True))
+
+                self.num_converted_trials += 1
+                logs.info(
+                    'Restarting a preemptible trial as a preemptible one: '
+                    'Benchmark-Fuzzer pair: %s-%s.'
+                    'Trial ID %d, restarted %3.2f%% trials (%d/%d) as '
+                    'preemptilbes so far.', trial.benchmark, trial.fuzzer,
```
oliverchang, line 365: nit: fix spelling of preemptibles.

experiment/scheduler.py
```diff
@@ -354,15 +356,41 @@ def _get_preempted_replacements(self,
...
                 nonpreemptible_starts += 1
                 replacements.append(replace_trial(trial, preemptible=False))
+
+                logs.info(
+                    'Restarting a preemptible trial as a nonpreemptible one: '
+                    'Benchmark-Fuzzer pair: %s-%s.'
+                    'Trial ID %d, restarted %3.2f%% trials (%d/%d) as '
+                    'nonpreemptibles so far.', trial.benchmark, trial.fuzzer,
+                    trial, nonpreemptible_starts / self.num_trials * 100,
+                    nonpreemptible_starts, self.num_trials)
```
oliverchang, line 382: nit: both this and the log above are very similar. Can we create a helper function and pass in only the parameters that change?

oliverchang (approved):

experiment/scheduler.py
```diff
@@ -353,16 +370,26 @@ def _get_preempted_replacements(self,
...
                 # replace it with a nonpreemptible.
                 nonpreemptible_starts += 1
                 replacements.append(replace_trial(trial, preemptible=False))
+
+                self._log_restart(False, trial, nonpreemptible_starts)
                 continue
 
+            self.num_preemptible_omits += 1
+            logs.warning(
+                'Omitting a trial due to time and budget constraints: %s',
```
jonathanmetzman, line 390: "time and budget constraints" is unnecessary.

experiment/scheduler.py
```diff
@@ -334,6 +336,21 @@ def get_nonpreemptible_starts(self) -> int:
         return get_started_trials(self.experiment_config['experiment']).filter(
             models.Trial.preemptible.is_(False)).count()
 
+    def _format_statistic_info(self, trial: models.Trial,
```
jonathanmetzman, line 339: This seems pretty overengineered for what we are trying to do. I would consider replacing this with something simpler.

experiment/scheduler.py
```diff
@@ -334,6 +336,21 @@ def get_nonpreemptible_starts(self) -> int:
...
+                               statistic: int) -> str:
+        """Formats a trial's statistic and information for logging."""
+        return (f'Trial ID: {trial.id}. '
+                f'Benchmark-Fuzzer pair: {trial.benchmark}-{trial.fuzzer}. '
+                f'Accumulating to {statistic/self.num_trials*100:3.2f}% '
+                f'({statistic} / {self.num_trials}) of all trials.')
+
+    def _log_restart(self, preemptible: bool, trial: models.Trial,
+                     statistic: int) -> None:
```
jonathanmetzman, line 348: nit: stastic can just be called "count"

jonathanmetzman:
LGTM

jonathanmetzman:
I'll point out that this is based on what I think is an incorrect theory on why fuzzbench experiments do not complete. That theory blames restarting trials. But in reality, it isn't only preemptible experiments that fail to terminate but non-preemptible ones as well.

Alan32Liu:
> I'll point out that this is based on what I think is an incorrect theory on why fuzzbench experiments do not complete. That theory blames restarting trials. But in reality, it isn't only preemptible experiments that fail to terminate but non-preemptible ones as well.

Right, I think there are two problems with experiments:
1. Missing results: Some trial results are missing in experiments. 
    * We know this because the final report has red banners saying 'The following fuzzers do not have enough samples: <fuzzer_name>' on some benchmarks.
    * Here is [an example for this case](https://storage.googleapis.com/www.fuzzbench.com/reports/experimental/2022-10-11-14-54-23h-dongge/index.html#libpcap_fuzz_both-summary). You can see the red banner above the statistical summary of benchmark `libpcap_fuzz_both-summary`.
    * Note that this example is separate from the second one below because it does not have a subtitle suggesting the experiment is incomplete.
2. Nonhaulting experiments.
    * We know this because the final report has a subtitle saying '(experiment incomplete/still running...)'.
    * Here is [an example for this case](https://storage.googleapis.com/www.fuzzbench.com/reports/experimental/2022-10-17-update-libfuzzer/index.html#experiment-incomplete-still-running).
    * Note that this example is also separate from the first one above because it does not have any red banner suggesting some results are missing.

experiment/scheduler.py
oliverchang: Let's just get this in and address any further feedback in a subsequent PR :)

PR #1559, opened 2022-11-23
Do not report unused args as errors

Alan32Liu:
Some projects use `-Werror` to turn all warnings into errors.
This affects `Centipede` as we do not separate build and linking flags as it expects, which leads to `unused-command-line-argument` warnings.
This PR disables turning that specific warning into errors and keeps the rest the same.

See [the same PR from OSS-Fuzz](https://github.com/google/oss-fuzz/pull/9030) for more info and [the error in this PR](https://github.com/google/oss-fuzz/pull/8990) for its use case.

### DIFF
--- a/benchmarks/systemd_fuzz-varlink/benchmark.yaml
+++ b/benchmarks/systemd_fuzz-varlink/benchmark.yaml
@@ -1,7 +1,6 @@
 fuzz_target: fuzz-varlink
 project: systemd
 unsupported_fuzzers:
-  - centipede
   - aflcc
   - afl_qemu
   - aflplusplus_qemu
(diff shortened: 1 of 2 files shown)


Alan32Liu:
This PR adds a new flag (i.e., `-Wno-error=unused-command-line-argument`) to make benchmark`systemd_fuzz-varlink` compatible with `Centipede`, yet `systemd_fuzz-link-parser` still fails for the same old reason.
This is because the latter adds another`-Werror=unused-command-line-argument` at the end of our flags, which specifically overwrites the flag added in this PR.
We could use a wrapper to remove that or reorder the flags (e.g., put ours at the end), but I don't think that is necessary in this case because we already have `systemd_fuzz-varlink`.

Alan32Liu:
TBR.

jonathanmetzman (commented):
lgtm

PR #1560, opened 2022-11-23
new afl++ experiment

vanhauser-thc:
the last experiment had a debug feature that crashed the instances :(


vanhauser-thc:
@jonathanmetzman 
this can be merged. thank you!

jonathanmetzman:
I don't like this presubmit failure

vanhauser-thc:
yes the log shows its not the commit updates for afl++, but some python stuff/modules of fuzzbench features.

as I would love to have the results (two PR submitters waiting for the results so we know what to merge and how (default/optional, etc.)), if we could try to see if it runs through with the risk it fails... :)

PR #1562, opened 2022-11-29
Fix failure when merging previous experiments

Alan32Liu:
1. Exclude benchmarks if they no longer exist (or do not have a `benchmark.yaml`).
2. Exclude experiments if their types are different from the main experiment.

### DIFF
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -15,6 +15,9 @@
 from analysis import stat_tests
 from common import benchmark_utils
 from common import environment
+from common import logs
+
+logger = logs.Logger('data_utils')
 
 
 class EmptyDataError(ValueError):
@@ -99,9 +102,22 @@ def filter_benchmarks(experiment_df, included_benchmarks):
     """Returns table with only rows where benchmark is in
     |included_benchmarks|."""
     valid_benchmarks = [
-        benchmarks for benchmarks in included_benchmarks
-        if benchmark_utils.validate(benchmarks)
+        benchmark for benchmark in included_benchmarks
+        if benchmark_utils.validate(benchmark)
     ]
+    logger.info('Included benchmarks: %s', included_benchmarks)
+    logger.warning('Invalid included benchmarks: %s',
+                   set(included_benchmarks) - set(valid_benchmarks))
+    experiment_df = experiment_df[experiment_df['benchmark'].isin(
+        valid_benchmarks)]
+
+    valid_benchmarks = [
+        benchmark for benchmark in experiment_df['benchmark']
+        if benchmark_utils.validate(benchmark)
+    ]
+    logger.info('experiment_df benchmarks: %s', experiment_df['benchmark'])
+    logger.warning('Invalid experiment_df Includedbenchmarks: %s',
+                   set(included_benchmarks) - set(experiment_df['benchmark']))
     return experiment_df[experiment_df['benchmark'].isin(valid_benchmarks)]
 
 
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -117,7 +117,7 @@ def filter_benchmarks(experiment_df, included_benchmarks):
     ]
     logger.info('experiment_df benchmarks: %s', experiment_df['benchmark'])
     logger.warning('Invalid experiment_df Includedbenchmarks: %s',
-                   set(included_benchmarks) - set(experiment_df['benchmark']))
+                   set(experiment_df['benchmark']) - set(valid_benchmarks))
     return experiment_df[experiment_df['benchmark'].isin(valid_benchmarks)]
 
 
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -107,7 +107,7 @@ def filter_benchmarks(experiment_df, included_benchmarks):
     ]
     logger.warning('Filtered out invalid benchmarks: %s.',
                    set(included_benchmarks) - set(valid_benchmarks))
-    logger.debug('Benchmarks passed: %s.', valid_benchmarks)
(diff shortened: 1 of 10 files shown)


Alan32Liu:
[A test run](https://storage.googleapis.com/www.fuzzbench.com/reports/2022-11-29-15-56-15m-ldg/index.html) to show:
1. It can successfully merge with prev results.
2. It does not have the 'not completed' subtitle.

analysis/data_utils.py
```diff
@@ -97,7 +101,14 @@ def filter_fuzzers(experiment_df, included_fuzzers):
...
     """Returns table with only rows where benchmark is in
     |included_benchmarks|."""
-    return experiment_df[experiment_df['benchmark'].isin(included_benchmarks)]
+    valid_benchmarks = [
+        benchmark for benchmark in included_benchmarks
+        if benchmark_utils.validate(benchmark)
+    ]
+    logger.warning('Filtered out invalid benchmarks: %s.',
+                   set(included_benchmarks) - set(valid_benchmarks))
+    logger.debug('Benchmarks passed: %s.', valid_benchmarks)
```
jonathanmetzman, line 110: change to `Valid benchmarks: %s.`

analysis/generate_report.py
```diff
@@ -143,6 +144,40 @@ def get_experiment_data(experiment_names, main_experiment_name,
     return experiment_df, description
 
 
+def filter_experiments(all_experiment_names):
+    """Helper function that reads data from disk or from the database. Returns
+    a list of old experiments of the same type as the main experiment."""
+    main_experiment_name, prev_experiment_names = all_experiment_names[
```
jonathanmetzman, line 150: Set these variables on different lines.
Also set them closer to where they are used.

analysis/queries.py
```diff
@@ -21,6 +21,18 @@
...
+def get_experiment_benchmarks(experiment_names):
+    """Get benchmarks of |experiment_names| from the database."""
+    with db_utils.session_scope() as session:
+        snapshots_query = session.query(Trial.benchmark)\
+            .select_from(Experiment)\
+            .join(Trial)\
+            .filter(Experiment.name.in_(experiment_names))\
+            .filter(Trial.preempted.is_(False))
+
+    return pd.read_sql_query(snapshots_query.statement, db_utils.engine)
```
jonathanmetzman, line 33: make this return `result['benchmark'].unique().tolist()` so that it returns a list of unique benchmark names, which is what the name of this function implies

analysis/generate_report.py
```diff
@@ -143,6 +144,40 @@ def get_experiment_data(experiment_names, main_experiment_name,
...
+        0], all_experiment_names[1:]
+    logger.debug('Verifying benchmark type of the main experiment...')
+    experiment_benchmark_df = queries.get_experiment_benchmarks(
+        [main_experiment_name])
+    main_experiment_type = experiment_utils.get_experiment_type(
+        experiment_benchmark_df['benchmark'].unique().tolist())
+    logger.info('Main experiment benchmark type is %s.', main_experiment_type)
+
+    same_type_experiments = [main_experiment_name]
+    for experiment_name in prev_experiment_names:
```
jonathanmetzman, line 160: This does a lot of unnecessary SQL queries. Can we rethink this so that we get the data for all experiments at once?
In SQL the query looks something like `SELECT DISTINCT benchmark, experiment FROM trial WHERE ...` just convert this to the right SQL alchemy instead of doing a new query per trial.

jonathanmetzman:
Thanks a ton for doing this!

analysis/generate_report.py
Alan32Liu: Yep, that is a good point. I worried about the same issue when writing the `for` loop, but I did not think of a better solution at that time.
How about this:
I guess we only care about the benchmarks used in the `main_experiment`, so instead of filtering experiment names here, I will add another filter to only include those benchmarks when [querying the experiment data](https://github.com/google/fuzzbench/blob/cb748014dc3f1da4b4ac01c4c435b09ff6858ea5/analysis/generate_report.py#L140).

Alan32Liu: Here is [the latest sample](https://storage.googleapis.com/www.fuzzbench.com/reports/2022-11-30-16-33-15m-ldg/index.html), which includes old results of the same benchmarks as in the current experiments only.
I reckon this is also more reasonable, because if we did not run the new fuzzer on a benchmark, then we cannot compare the new fuzzer with old results on that benchmark.

Alan32Liu: @jonathanmetzman : A friendly ping on checking the new logic above to filter benchmarks. 
Here is the implementation: [Only merge with the same benchmark programs](https://github.com/google/fuzzbench/pull/1562/commits/73478a9c4d099b56aae58a2cf314b190278f8f6a).
The other comments are addressed in [Minor fixes on log and code style](https://github.com/google/fuzzbench/pull/1562/commits/4226bed505775be1b027b740ba3a742c3c36030b).
Thanks!

experiment/reporter.py
```diff
@@ -64,9 +64,12 @@ def output_report(experiment_config: dict,
     # Don't merge with nonprivate experiments until the very end as doing it
     # while the experiment is in progress will produce unusable realtime
     # results.
+    logger.info('In progress: %s.', in_progress)
     merge_with_nonprivate = (not in_progress and experiment_config.get(
         'merge_with_nonprivate', False))
+    logger.info('Is merging with nonprivate:: %s.', merge_with_nonprivate)
```
jonathanmetzman, line 70: drop extra colon

jonathanmetzman (approved):
LGTM. Thanks!

PR #1526, opened 2022-10-14
Upgrade base images, benchmark images, and Python.

Alan32Liu:
I guess this branch has what we need? Not sure why it does not has a PR though.
I make it up-to-date with the master branch so that we can discuss here.

PR #1570, opened 2022-12-05
Deprecate fuzzzers and ignore fuzzers from CI

Alan32Liu:
1. Deprecate the list of fuzzers as discussed.
2. Let CI ignore fuzzers that we want to keep but are not yet compatible with Ubuntu20 or Python3.10.


jonathanmetzman:
Are we deleting much more than is needed here?
For example, wingfuzz doesn't need to be modified? https://github.com/google/fuzzbench/pull/1568

Alan32Liu:
> Are we deleting much more than is needed here? For example, wingfuzz doesn't need to be modified? #1568

I don't think we have deleted [`wingfuzz`](https://github.com/google/fuzzbench/pull/1570/files#diff-f7784c9acea0c152e9ad76d92fcded79c342009768bc4d3800c8bd467318baedL33).
I noticed that it is [compatible with the new version](https://github.com/google/fuzzbench/actions/runs/3519079031/jobs/5898647948) and is [under active development](https://github.com/WingTecherTHU/wingfuzz), so I did not remove it although we did not select it in [the previous list](https://github.com/google/fuzzbench/pull/1526#issuecomment-1334641934).

Alan32Liu:
I added back `aflplusplus_cmplog` and `aflplusplus_cmplog_r`, as there is a new experiment request of it in the master branch.
Is there any other fuzzers like them (e.g., not in the previous list but should be kept for experiments)?
Thanks! I will add them back.

Alan32Liu:
@jonathanmetzman : Do you happen to know which part of the code affects the CI tests' names?
For example, the first CI test below is `Build fuzzers / build (afl, klee, oss-fuzz) (pull_request)`, which has an extra `klee` between the fuzzer `afl` and benchmark type `oss-ufzz`.
I reckon this may relate to [the new category](https://github.com/google/fuzzbench/pull/1570/files#diff-f7784c9acea0c152e9ad76d92fcded79c342009768bc4d3800c8bd467318baedR44) I added in ` fuzzers.yaml`, but I was not sure how that affects the name of the CI tests.
I double-checked that the actual command run in each test looks normal. Taking the same example, the first CI is actually running `afl` on `oss-fuzz` test cases as expected, and does not involve `klee` at all.

Alan32Liu:
@jonathanmetzman : A friendly ping on this, to continue our discussion this morning : )
The problem is stated above.

PR #1581, opened 2022-12-10
Revive bug benchmarks

Alan32Liu:
Bring back 5 bug-based benchmarks:
1. `arrow_parquet-arrow-fuzz`
2. `aspell_aspell_fuzzer`
3. `ffmpeg_ffmpeg_demuxer_fuzzer`
4. `file_magic_fuzzer`
5. `grok_grk_decompress_fuzzer`


benchmarks/arrow_parquet-arrow-fuzz/benchmark.yaml
```diff
@@ -1,5 +1,7 @@
+commit: e9222ae00b515a0f57d8c7d54bdb19457d68582f
```
jonathanmetzman, line 1: are these the old commits?

benchmarks/aspell_aspell_fuzzer/build.sh
```diff
@@ -15,12 +15,7 @@
 #
 ################################################################################
 
-# Suppress an immediate UBSan violation that prevents fuzzing
```
jonathanmetzman, line 18: Why is this not necessary anymore?

service/test_automatic_run_experiment.py
```diff
@@ -83,12 +83,7 @@ def test_run_requested_experiment(mocked_get_requested_experiments,
         'sqlite3_ossfuzz',
         'systemd_fuzz-link-parser',
         'zlib_zlib_uncompress_fuzzer',
-        'arrow_parquet-arrow-fuzz',
```
jonathanmetzman, line 86: Why are these no longer expected because they are now bug based?

jonathanmetzman (approved):
lgtm

benchmarks/aspell_aspell_fuzzer/build.sh
Alan32Liu: I changed its build file to be [the same as in OSS-Fuzz](https://github.com/google/oss-fuzz/blob/master/projects/aspell/build.sh).

service/test_automatic_run_experiment.py
Alan32Liu: Yes, I added these benchmarks here temporarily [when converting bug-based benchmarks to coverage-based ones](https://github.com/google/fuzzbench/pull/1526/commits/ce36cc916e1f46686995066f4025bab99a13c5de).
I plan to gradually remove them while reviving more of them later, hope you would not mind : )

Alan32Liu:
@oliverchang Now all bug-based benchmarks are using the earliest buggy commits since June.

benchmarks/arrow_parquet-arrow-fuzz/benchmark.yaml
jonathanmetzman: >They are old enough to have multiple known bugs to be discovered by fuzzers.

Ack. thanks
Are these bugs from a single fuzz target? Otherwise they aren't useful.

benchmarks/aspell_aspell_fuzzer/build.sh
jonathanmetzman: ACK

service/test_automatic_run_experiment.py
jonathanmetzman: It's problematic that these were made into coverage benchmarks because now coverage experiments have to build these broken benchmarks.

benchmarks/arrow_parquet-arrow-fuzz/benchmark.yaml
Alan32Liu: Ah, that is a good point, I will double-check when adding new bug-based benchmarks.

For these 5 benchmarks:
1. `arrow`: one UBSAN error from `parquet-arrow-fuzz`.
2. `aspell`: one UBSAN error from `aspell_fuzzer `.
3. `ffmpeg`: one ASAN error from `ffmpeg_DEMUXER_fuzzer`.
4. `file`: 20+ ASAN&UBSAN errors from `magic_fuzzer`.
5. `grok`: 14 ASAN&UBSAN errors from `grk_decompress_fuzzer`.

I wonder if the first 2 are useful as I reckon we don't have `UBSAN` in FuzzBench, right?

jonathanmetzman: We do have UBSAN but one bug is probably not enough

jonathanmetzman: I think the `file` fuzzers do crazy things with processes not sure if it's good to use them.

Alan32Liu: I see, I will add some alternatives, then : )

jonathanmetzman: It might be fine. I really am not sure.

Alan32Liu: No problem, I think we do plan to gradually revive more in 1-2 months so I am more than happy to do that earlier.
BTW, would you mind me asking how we are using UBSAN?
I thought [it requires `-fsanitize=undefined`](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#usage), but I did not find that flag in our code...

jonathanmetzman: we pick the individual parts of UBSAN ourselves. e.g. integer-overflow

jonathanmetzman: i would make sure to do a bunch of experiments with a benchmark before including it in a public competition

Alan32Liu: Good point, I will start one to double-check the current 5, then : )

Alan32Liu: Launched in #1589.

PR #1582, opened 2022-12-11
Fix eclipser

Alan32Liu:
1. Update `Eclipser` to a later commit that supports `Ubuntu20.04`.
2. Install and set `python2` as the default `python` to build `Eclipser`.


oliverchang (approved):

Alan32Liu:
@jonathanmetzman:
Some `Eclipser` CI tests failed with exit code 137, which indicates running out of memory.
I was not able to reproduce them locally.
But I presume this is OK, as we are expecting a new CI mechanism? Thanks.

jonathanmetzman (approved):
lgtm

jonathanmetzman:
> @jonathanmetzman: Some `Eclipser` CI tests failed with exit code 137, which indicates running out of memory. I was not able to reproduce them locally. But I presume this is OK, as we are expecting a new CI mechanism? Thanks.

It's Ok because the Ci isn't the same as the prod environment.

PR #1583, opened 2022-12-13
Expect mock test to run with 150 concurrent builds

Alan32Liu:
Our earlier commit [fix issue with service](https://github.com/google/fuzzbench/commit/5ec25663710c9dc20d6623f82e01e79a7630f94c) did not edit the corresponding test, hence introducing a CI failure.
This PR fixes that failure by adding the new parameter to the expected call.


Alan32Liu:
TBR, I will merge this now as it is required to pass CI in #1581 : )

jonathanmetzman:
LGTM

Issue #1584, opened 2022-12-13
Mruby is broken


andreafioraldi:
I'm using mruby to test things in my local checkout of fuzzbench and it I noticed that it is currently broken as it changed from the version that I uploaded months ago.

+ The build script is ignoring the compiler wrappers overriding CC and CXX https://github.com/google/fuzzbench/blob/5d7362a6fc6647740abdebd0c7db13beb57809ea/benchmarks/mruby-2018-05-23/build.sh#L24 this make any fuzzer with a compiler wrapper like AFL++ and libafl incompatible

+ The commit was updated and it is not anymore the version 2018-05-23 (so the name has no sense now) and thus there are no known bugs anymore

Alan32Liu:
Thanks, @andreafioraldi,

> The build script is ignoring the compiler wrappers overriding CC and CXX

Yes, good catch. I am more than happy to switch to that build script later.
Feel free to submit a PR if you are interested in fixing it, I would really appreciate that as well : )

> The commit was updated and it is not anymore the version 2018-05-23 

Yep, long story short, we recently upgraded the version of Ubuntu used by FuzzBench, and that requires us to update many benchmarks to build them. Eventually, we will update all benchmarks to their newer versions.
As you said, bug-based benchmarks are a bit special, as their latest version might not have any known bugs, so we will need a bit of time to identify the best version that is compatible with the new Ubuntu but still has as many known bugs as possible. At this moment, we only revived [5 bug benchmarks](https://github.com/google/fuzzbench/pull/1581), but I promise I will gradually bring more back later.

I reckon this will benefit fuzzer evaluation as well: New bugs will help fuzzers identify what they missed in the past and avoid overfitting on finding the old ones.

PR #1585, opened 2022-12-13
Libafl forkserver test

andreafioraldi:
Recently, we added the support to run AFL++ binaries in LibAFL with CmpLog and shmem inputs.
With this PR, I'm requesting a bug-based benchmark (as the advantage of the forkserver is reliability even when the target is crashing often) to test if the Rust implementation matches the original C version of AFL++ in code and bug coverage.

Related to this, https://github.com/google/fuzzbench/issues/1584


jonathanmetzman:
Bug based benchmarks is extremely broken now.

andreafioraldi:
> Bug based benchmarks is extremely broken now.

Every benchmark? Just 2 or 3 would be fine.

presubmit doesn't fail locally and it seems to fail a test that is (probably) unrelated to my PR.

andreafioraldi:
Do you know the last working commit? At least I can rebase on it and run on a subset of the programs locally

jonathanmetzman:
Do you want to run this now? I think we're in better shape

andreafioraldi:
> Do you want to run this now? I think we're in better shape

No for a strange reason some benchmarks don't build with afl++ under libafl_forkserver, I'm trying to figure out why, I'll remove the draft mark when ready

PR #1586, opened 2022-12-13
Fix bug-based benchmark labelling

jonathanmetzman:
During upgrade some benchmarks were erroneously converted from
bug to coverage. Make these bug-based again so that they aren't run during bug-based experiments.


Alan32Liu (commented):
Thanks! I will look into the error logs to see if there is more I can do with the bug-based benchmarks, e.g. at least not let them break CIs.

PR #1587, opened 2022-12-14
Fix report generation by replacing all `NaN`s in experiment dataframe with `0`s.

Alan32Liu:
Report generation breaks with error `pandas.errors.IntCastingNaNError: Cannot convert non-finite values (NA or inf) to integer`, after we upgrade `pandas` to a newer version.
This PR fixes this error by converting all `NaN` values in the result data frame to `0`.


Alan32Liu:
`presubmit` failed as we tag some benchmarks as bug but did not remove them from the list in `test_automatic_run_experiment.py`.
I will fix that in a different PR.

Alan32Liu:
Here is a [sample experiment](https://storage.googleapis.com/www.fuzzbench.com/reports/experimental/2022-12-14-21-45-15m-dongge/index.html) to confirm that we have fixed this error.
It can successfully generate the final report without any error.

oliverchang (commented):
do we know why this started being a problem all of a sudden?

jonathanmetzman:
> do we know why this started being a problem all of a sudden?

Maybe it's because we had a bunch of benchmarks that completely failed?

Alan32Liu:
> do we know why this started being a problem all of a sudden?

Because we upgrade `pandas` to a newer version while upgrading `Python3.10`.
I recall that requires us to add `.astype(int)`, which caused this error.

Alan32Liu:
> maybe it's worth kicking off the experiments again (sicne they wont run all the bug benchmarks this time)

I think it is, my earlier experiment (with bug-based ones, though) can run successfully and produce [the report as expected](https://storage.googleapis.com/www.fuzzbench.com/reports/experimental/2022-12-14-21-45-15m-dongge/index.html).

jonathanmetzman (approved):
lgtm

PR #1592, opened 2022-12-27
Fix bug

jonathanmetzman:


Alan32Liu (approved):
Ah right, we only need one target : )
Thanks!

jonathanmetzman:
Please take another look.

Alan32Liu (approved):
Yes! Thanks for the clear naming.
It reminds me that I updated some of the bug-based benchmarks but did not update the version/date in their names. I should come back to fix that later.

PR #1600, opened 2023-01-06
Update experiment-requests.yaml

LIIT0215:
New test on fuzzbench between 9 fuzzers

### DIFF
--- a/service/experiment-requests.yaml
+++ b/service/experiment-requests.yaml
@@ -19,7 +19,32 @@
 # You can run "make presubmit" to do basic validation on this file.
 # Please add new experiment requests towards the top of this file.
 #
-
+- experiment: 2023-01-06-scheduler-test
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Why do you want this experiment?

LIIT0215:
Sorry, I used fuzzbench for the first time. I have integrated K-scheduler into libafl locally, so I want to experience the periodic service before add it to the fuzzer

vanhauser-thc:
@LIIT0215 and where does fuzzbench magically get your local modifications? you must add a libafl variant to the fuzzers on fuzzbench that checks out your specific versions. see the docs one the steps to do that to add to the config, copy the libafl folder to a new name and add the git checkout information for your fuzzer.
then you would only need libafl and your variant to run to compare.

PR #1529, opened 2022-10-19
Fix lint

Alan32Liu:
Based on and required by #1526.

This PR fixes new lint errors from the latest `pylint`, which is required as #1526 upgrades `Python` to `3.10`.
I reckon I have fixed all meaningful errors in the *`FuzzBench` framework*, not sure if the rest is worth fixing or could be ignored.

The rest lint errors are:
1. From individual fuzzers (e.g., `fuzzers/*`);
2. In the database directory (i.e., `database/*`);
3. trivial errors:
    * conventional naming (e.g., `df` doesn't conform to snake_case naming style, but is common for `pandas.dataframe`).
    * too-few-public-methods (e.g., `Too few public methods`).
    * `Using the global statement`.
    * Improper `Consider using 'with' for resource-allocating operations`, e.g., with `Popen`, `urlopen`.
    * Improper ` Using open without explicitly specifying an encoding`, e.g., when opening a binary file.

Could you please take a look at the rest errors and let me know which one is worth fixing?
Otherwise, I will find a way to let pylint ignore them in the future.

PR #1604, opened 2023-01-09
Revert prev lint fix

### LINKED ISSUE #1594: trials not launching in local experiment
I'm running into an issue running local experiments: the dispatcher fails to launch any of the individual trials. Even waiting for hours, no trials are launched, though the system appears to still be checking for coverage data in the measurement loop from the logs.

This happens every time I try to run a local experiment.

I've checked `/tmp` in the dispatcher container during one of these failed runs, and there are no logs from any of the runner startup scripts, nor do the runner containers appear to ever start. However, if I then manually run the startup script in the dispatcher container (e.g. `docker exec -it dispatcher-container /bin/bash; /tmp/r-startup-script3.sh`) the trial / runner container starts up with no issues.

I've tried adding some logging to the python code that starts the runners but everything looked fine from that end and I was not able to figure out what the cause of the issue is.

I've been able to reproduce this issue on multiple machines running recent versions of Ubuntu. The only changes I made to master were to add `rsync` to the relevant dockerfiles to work around: https://github.com/google/fuzzbench/issues/1593

OS: Ubuntu (e.g. 22.04)
Docker version: e.g. 20.10.22
Commit: aa6ddd05
Reproduction: 
1. add rsync to Dockerfiles; 
2. make; 
3. docker build the dispatcher image; 
4. run an experiment with basic config e.g. [experiment-config.yaml.txt](https://github.com/google/fuzzbench/files/10322923/experiment-config.yaml.txt) using command such as
[exp.sh.txt](https://github.com/google/fuzzbench/files/10322924/exp.sh.txt)

Any insight would be much appreciated! And I'd be happy to provide logs / additional information as needed.

Thanks!

Alan32Liu:
Fixes #1594:
1. Revert a previous lint fix (in #1529) to avoid using `with` on subprocesses, as the context manager made local experiments fail to launch runner instances.
2. Simplify the previous return statement, which was true by definition.


Alan32Liu:
@dliyanage @kdsjZh could you please test if this works? Thanks!
I am also running a new experiment on my local machine with this change, which seems to work so far.

Alan32Liu:
CI failed due to an expected lint error, as this PR reverts that lint fix.
I will handle that later.

kdsjZh:
Now the runners container started, which seems work from my side! Thanks for the quick fix!

Alan32Liu:
Revert all lint fixes like `with subprocess` in #1529.
`presubmit` fails due to the following temporary fuzzers not being in fuzzer list:
1. `eclipser_aflplusplus`
2. `neuzz`
3. `fuzzolic_aflplusplus_z3`
4. `fuzzolic_aflplusplus_fuzzy`

Alan32Liu:
@jonathanmetzman A friendly gentle ping on this. Local experiments rely on this (fixes #1594).
I reckon the `presubmit` error is acceptable, because of the following reason:

> Revert all lint fixes like `with subprocess` in #1529. `presubmit` fails due to the following temporary fuzzers not being in fuzzer list:
> 
> 1. `eclipser_aflplusplus`
> 2. `neuzz`
> 3. `fuzzolic_aflplusplus_z3`
> 4. `fuzzolic_aflplusplus_fuzzy`

common/gcloud.py
```diff
@@ -130,10 +130,9 @@ def run_local_instance(startup_script: Optional[str] = None) -> bool:
     """Does the equivalent of "create_instance" for local experiments, runs
     |startup_script| in the background."""
     command = ['/bin/bash', startup_script]
-    with subprocess.Popen(command,
-                          stdout=subprocess.PIPE,
-                          stderr=subprocess.STDOUT):
-        return new_process.ProcessResult(0, '', False)
+    # pylint: disable=consider-using-with
+    subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    return new_process.ProcessResult(0, '', False).retcode == 0
```
jonathanmetzman, line 135: This will always be true wont it?

common/gcloud.py
```diff
@@ -130,10 +130,9 @@ def run_local_instance(startup_script: Optional[str] = None) -> bool:
     """Does the equivalent of "create_instance" for local experiments, runs
     |startup_script| in the background."""
     command = ['/bin/bash', startup_script]
-    with subprocess.Popen(command,
-                          stdout=subprocess.PIPE,
-                          stderr=subprocess.STDOUT):
-        return new_process.ProcessResult(0, '', False)
+    # pylint: disable=consider-using-with
+    subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
jonathanmetzman, line 134: I don't think this does anything to check if the command succeeded.

jonathanmetzman (changes requested):
Sorry, but could you explain more?
what is this reverting and how does this fix a problem? THe changed code seems broken IMO

common/gcloud.py
Alan32Liu: This reverts to https://github.com/google/fuzzbench/pull/1529/files#diff-f61c598f4891541e693644bbc9016e5841fcddc986ea967eb3e192586a8c57deL134

Alan32Liu: Hmm, maybe I can also fix it by adding `==0` to `return new_process.ProcessResult(0, '', False)`.

Alan32Liu: I don't think the new code (after revert) will always be true.
For example, it will be false if the process fails and returns non-zero value, right?

common/gcloud.py
Alan32Liu: It checks the return status of the command in the next line.

common/gcloud.py
aeflores: The way it was before using a context manager was:
```
subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
return new_process.ProcessResult(0, '', False)
```
That indeed does not check anything about Popen.  `new_process.ProcessResult(0, '', False)` is just creating a named tuple (see definition here https://github.com/google/fuzzbench/blob/master/common/new_process.py#L67). There is no connection between that line and the call to `Popen`.

The important fix here though is to stop using the context manager. As per the documentation (https://docs.python.org/3/library/subprocess.html?highlight=popen#subprocess.Popen)

>Popen objects are supported as context managers via the [with](https://docs.python.org/3/reference/compound_stmts.html#with) statement: on exit, standard file descriptors are closed, and the process is waited for.

Using a context manager closes file descriptors and waits for the process to terminate, which is probably what is causing problems. This code probably needs to start the process, but not wait for its termination.

My suggestion would be:
```
subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
return True
```
If we want to be extra careful, we should check that the subprocess is created successfully, but I think we will get an exception if that were not the case anyway.

Alan32Liu: Thanks, @aeflores!
Yeah, your suggestion is cleaner as the old return statement is trivially true.
I've fixed it accordingly : )

fuzzers/neuzz/fuzzer.py
```diff
@@ -91,8 +91,8 @@ def run_neuzz(input_corpus,
         'python2', './nn.py', '--output-folder', afl_output_dir, target_binary
     ]
     print('[run_neuzz] Running command: ' + ' '.join(command))
-    with subprocess.Popen(command, stdout=output_stream, stderr=output_stream):
-        pass
+    # pylint: disable=consider-using-with
```
jonathanmetzman, line 94: I think this patch changes too much. I think with usage didn't make sense in the first location, but I don't see why you would change these other ones. I think it makes sense here.

Alan32Liu: Right, I think I should have considered this case by case:
1. Should use context manager (`with`) to wait for the process :
    * Here, to enable `asan` (probably can remove `sleep`, but I suppose it is better to leave it in this PR?).
2. Should not use context manager (`with`):
    * `gcloud`, to fix the bug in local experiments.
3. Neutral:
    * All others, using context manager has a neutral effect.

I am inclined to not use `with` in 3, because it looks cleaner, and it was that way before I [attempt to use `with` to fix linter errors](https://github.com/google/fuzzbench/pull/1529/files#r1003870199).
@jonathanmetzman what would you think? Thanks!

jonathanmetzman: I think you should get rid of it in gcloud.py.
I think the other cases should be unchanged. but I'm not sure about them, maybe try a make test-run-... to check them. Either way these files are unimportant so just fix gcloud for now and ignore these please.
Thanks!

Alan32Liu: I see!
Done, thanks!

common/gcloud.py
oliverchang: This seems like the same behaviour as before: https://github.com/google/fuzzbench/pull/1529/files#diff-f61c598f4891541e693644bbc9016e5841fcddc986ea967eb3e192586a8c57deL133

The previous return was a no-op that would've always returned True (side note: is this behaviour actually correct?) 

@Alan32Liu would be good to make this more explicit in the PR description in the future to make it clearer to reviewers :)

oliverchang (approved):

Navidem:
With this PR, I do not get `No snapshot data` warning, but am getting this:
`WARNING:root:Corpus not found for cycle: 3. Extras: {'fuzzer': 'centipede', 'benchmark': 'libpng-1.6.38', 'trial_id': '35', 'cycle': '3'}`

Navidem (approved):
LGTM

Navidem:
I think we can merge this.

PR #1607, opened 2023-01-11
Update dependencies of aflplusplus-qemu and aflplusplus-frida

aeflores:
This follows how dependencies are specified in aflplusplus https://github.com/google/fuzzbench/blob/master/fuzzers/aflplusplus/builder.Dockerfile#L18 but only specifies what is really needed for building.

It also updates the ALF++ commit hash to coincide with the one in aflplusplus https://github.com/google/fuzzbench/blob/master/fuzzers/aflplusplus/builder.Dockerfile#L41

This was motivated because `libstdc++-5-dev` was no longer available to be installed.


vanhauser-thc:
you need to add aflplusplus_frida and _qemu to .github/workflows/fuzzers.yml (just add it below aflplusplus_cmplog). then it can be merged.

vanhauser-thc:
The CI fails for libpcap because:
```
clang-15: error: no such file or directory: '/usr/lib/x86_64-linux-gnu/libdbus-1.a'
```
I do not understand why the libpcap container does not install the necessary libraries, and why libfuzzer and honggfuzz etc. do not fail too.
can you please add to install `libdbus-1-dev` to the build.Dockerfile?

and then you can test in your local fuzzbench checkout by `make test-run-aflplusplus_frida-libpcap_fuzz_both` if this is enough or if another lib is also needed.

.github/workflows/fuzzers.yml
```diff
@@ -95,8 +95,8 @@ jobs:
           # - afl_qemu
           # - honggfuzz_qemu
           # - weizz_qemu
-          # - aflplusplus_qemu
-          # - aflplusplus_frida
+          - aflplusplus_qemu
+          - aflplusplus_frida
```
jonathanmetzman, line 99: You don't need to do this anymore.

vanhauser-thc: Not? I get a CI error 3 weeks ago where I forgot that if I remember correctly

benchmarks/libpcap_fuzz_both/Dockerfile
```diff
@@ -16,7 +16,7 @@
 
 FROM gcr.io/oss-fuzz-base/base-builder@sha256:fb1a9a49752c9e504687448d1f1a048ec1e062e2e40f7e8a23e86b63ff3dad7c
 RUN apt-get update && \
-    apt-get install -y make cmake flex bison
+    apt-get install -y make cmake flex bison libdbus-1-dev
```
aeflores, line 19: @vanhauser-thc  I added the missing dependency here.

It is weird that this dependency is specified here though https://github.com/google/fuzzbench/blob/master/fuzzers/aflplusplus/fuzzer.py#L40 I don't know why that would be the case

vanhauser-thc:
now you have to fix the conflict :)

vanhauser-thc:
done, thanks!

PR #1608, opened 2023-01-12
new afl++ token/libafl experiment

vanhauser-thc:
hopefully fuzzbench is not throwing up again :)


jonathanmetzman:
Weird that CI is actually in a good state now

vanhauser-thc:
hmm it is again spurious checkout failures, at least for one. @Alan32Liu

Alan32Liu:
Hi @vanhauser-thc, I was investigating this, but I really did not find a possible explanation:
It seems more like a strange issue in `git`/`docker` than in your code or FuzzBench.

But I guess there are several more things we can try together, although they might not matter:
1. Specify and only pull the belonging branch of the commit. Given the commit `e4ff0eb` is in `dev` instead of the default (`stable`), maybe do `git clone -b dev https://github.com/AFLplusplus/AFLplusplus /afl`.
2. Reorganise the docker commands, e.g. splitting `cd /afl` and `git checkout` to two `RUN`s.

Again, I don't think these will somehow **_solve_** the issue, but it might **_avoid_** triggering them.
Please let me know what you think.

PR #1610, opened 2023-01-18
Fix centipede linker flags

Alan32Liu:
Mirrors [the fixes from OSS-Fuzz](https://github.com/google/oss-fuzz/pull/9427):
1. Use [`-Wl` on the linker flags](https://github.com/google/oss-fuzz/pull/9427#issuecomment-1385205488).
2. Use [`LDFLAGS`](https://github.com/google/oss-fuzz/pull/9427#issuecomment-1385375441).

### DIFF
--- a/fuzzers/centipede/fuzzer.py
+++ b/fuzzers/centipede/fuzzer.py
@@ -24,11 +24,8 @@ def build():
     san_cflags = ['-fsanitize-coverage=trace-loads']
 
     link_cflags = [
-        '-Wno-error=unused-command-line-argument',
-        '-ldl',
-        '-lrt',
-        '-lpthread',
-        '/lib/weak.o',
+        '-Wno-unused-command-line-argument',
+        '-Wl,-ldl,-lrt,-lpthread,/lib/weak.o'
     ]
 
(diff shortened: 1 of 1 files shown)


oliverchang (approved):

Alan32Liu:
> @Alan32Liu thanks! `systemd` can still fail on FuzzBench due to [systemd/systemd#20542 (comment)](https://github.com/systemd/systemd/issues/20542#issuecomment-1361547403) but [systemd/systemd#25842](https://github.com/systemd/systemd/pull/25842) should address that in the foreseeable future hopefully.

Terrific, cheers!
I noticed that #25842 is still a draft and failed a CI test, I presume it is still WIP?
Happy to test it out on FuzzBench and OSS-Fuzz when it is ready : )

evverx:
That PR is a draft and I'm a bit reluctant to bring it to systemd: https://github.com/systemd/systemd/pull/25844#issuecomment-1365494235. Other than that it isn't compatible with `centipede` yet because it doesn't handle `SANITIZER=none`. Then again I think sooner or later it should be merged one way or another.

Issue #1636, opened 2023-01-25
Make logs publicly available.


jonathanmetzman:
So users can help debug experiments

jonathanmetzman:
This should have logs in an hour assuming there's no service account issues: https://pantheon.corp.google.com/storage/browser/fuzzbench-logs;tab=objects?project=fuzzbench&prefix=&forceOnObjectsSortingFiltering=false
CC @Alan32Liu

jonathanmetzman:
If this works we need to document how to read logs for experiments (and include links/instructions in reports).

jonathanmetzman:
It works!

oliverchang:
woot!!! can we close this?

jonathanmetzman:
Need #1638 to land and me to update the router and then test it in an experiment and then we can close.

jonathanmetzman:
Testing in an experiment isn't stricly necessary. I updated the router. When I come online tomorrow if I see junk in there like audit logs then I didn't do what I wanted and i'll reopen this. But I think we can close yeah.

jonathanmetzman:
I think this needs more thought. We can't expose details from private experiments

PR #1638, opened 2023-01-25
Make it easier to route logs from fuzzbench code to a public bucket

jonathanmetzman:
By getting rid of this we can match
logName=projects/fuzzbench/logs/fuzzbench

Related: https://github.com/google/fuzzbench/issues/1636


jonathanmetzman:
Take a look at gs://fuzzbench-logs/ right now you will see things like audit logs in there. With this change we will be able to include just the logs we want and not logs we don't!

oliverchang (approved):

PR #1639, opened 2023-01-25
Make changes for oss-fuzz on demand

jonathanmetzman:


jonathanmetzman:
These are a bit hacky. I can leave this stuff on a branch until it's more production ready.

fuzzers/utils.py
```diff
@@ -170,6 +171,11 @@ def get_dictionary_path(target_binary):
 
 def set_fuzz_target(env=None):
     """Set |FUZZ_TARGET| env flag."""
+
+    # Don't call get_config_value which depends on things that will break in
```
oliverchang, line 175: Do we have an idea for how to make this less hacky? 

Can we refactor this to make `get_config_value` more pluggable (and mockable in the OSS-Fuzz case) ?

jonathanmetzman: I don't really know what this would look like and I suspect the answer is no. This is a function that assumes that the benchmark being built is in the fuzzbench repo. If it's from oss-fuzz it won't be in the repo and it won't have a benchmark.yaml file.

oliverchang: I don't know how this is planned to be called, but is it possible to refactor the `get_config_value` function into some kind of `Config` object that gets passed to `set_fuzz_target` instead? 

on FuzzBench, the `Config` object would just do what the current `get_config_value` does. When called from OSS-Fuzz on demand, we instead supply an object with our desired settings.

jonathanmetzman: I'll do this later. I need to unblock myself on oss-fuzz.

Issue #1489, opened 2022-09-16
Sampling Initial Seed Corpus and Analysis


dylanjwolff:
TO @jonathanmetzman @lszekeres
CC @mboehme @inferno-chromium 

We have two related features which we've implemented on a private fork that we'd like to integrate into Fuzzbench. The first is the ability to sample from a larger pool of seeds to provide a unique corpus to each fuzzer per trial during a benchmarking run. The second consists of additional data-analysis to give some insight into how various aspects of the initial corpora and programs under test might be affecting benchmarking outcomes.

The purpose of this issue it to establish the following:
1. [Sampling] We currently have a script we've been using for local experiments that samples from e.g. a project's OSS-Fuzz corpus to generate random initial corpora. We then mount those in the docker containers of the runners. We also kick off the first measurer cycle before launching the fuzzer process to grab the initial coverage of the corpus. Are there other considerations or another approach we should take for adding this feature?
2. [Properties] Which properties would you consider to be interesting? We currently have 
    * seed-corpus: initial coverage, number of seeds, average seed exec time, average seed size
    * program: size (and others). Anything else that you would like to look at?
3. [UI/UX] What is the interface that you want to present to the users? For the seed sampling, probably additional field(s) in the YAML configuration file to select a sampling level / strategy? For the data presentation, we have produced several visualizations which show the relative impact of a particular property on the final ranking of a fuzzer or its coverage. We are happy to share these separately and would welcome any feedback you might have on where and how to present this data in a Fuzzbench report.

Thanks!

mboehme:
The key idea is essentially: Instead of saying fuzzer A is the top fuzzer, we say that Fuzzer A is the top fuzzer under these circumstances while fuzzer B is the top fuzzer under those other circumstances. For any given benchmark run, a user could essentially use a slider on those benchmark properties to see how the fuzzer ranking changes.

jonathanmetzman:
Sorry for the delay, I've had a bit of a crazy schedule with my holidays.
I personally think the second might be more interesting and seems less of a maintenance burden (the analysis just gets done at the end right?) But I'm interested in seeing both.

jonathanmetzman:
The UI/UX question is tricky, let me think about it more.

Alan32Liu:
> 2. [Properties] Which properties would you consider to be interesting? We currently have
>    
>    * seed-corpus: initial coverage, number of seeds, average seed exec time, average seed size
>    * program: size (and others). Anything else that you would like to look at?

[...]

PR #1642, opened 2023-01-25
Start Recording Coverage at Cycle 0

dylanjwolff:
Hello @Alan32Liu and Fuzzbench maintainers!

The goal of this PR is to start recording the coverage statistics at cycle 0, before the fuzzers are started up. To do so, I'm copying over the input corpus to the output directory immediately and calling do_sync before starting the fuzzer thread. Additionally, I adjusted both the logscale and linear time plots to start at t=0.

Overall, I believe this makes the linear plots quite a bit better, as we can now see the starting coverage for a given run. There is some distortion around zero for the logscale plots, happy to discuss how best to handle this if needed. 

These changes also relate to https://github.com/google/fuzzbench/issues/1489 in that we have observed the initial coverage to have an impact on the relative performance of fuzzers in Fuzzbench. Archiving the initial corpus along with the other snapshots of coverage for the run and recording the initial edges covered is a first step towards accounting for the impacts of additional variables on benchmarking.

I've attached some example plots below:
linear_time
log_time


vanhauser-thc:
IMHO that is a good thing to have (the functionality, no clue on the code changes :)

vanhauser-thc:
Note that you will have to fix these two items to satify the CI:

```
 analysis/plotting.py:23:0: W0611: Unused experiment_utils imported from common (unused-import)
```

```
File "experiment/measurer/measure_manager.py", line 125: attribute-error disabled from here to the end of the file [late-directive]
  Consider limiting this directive's scope or moving it to the top of the file.

```

experiment/runner.py
```diff
@@ -278,6 +278,18 @@ def conduct_trial(self):
 
         max_total_time = environment.get('MAX_TOTAL_TIME')
         args = (max_total_time, self.log_file)
+
+        input_corpus = environment.get('SEED_CORPUS_DIR')
+        output_corpus = environment.get('OUTPUT_CORPUS_DIR')
+
+        # Ensure seeds are in output corpus
```
jonathanmetzman, line 285: nit end with period

experiment/runner.py
```diff
@@ -278,6 +278,18 @@ def conduct_trial(self):
...
+        os.rmdir(output_corpus)
+        os.makedirs(input_corpus, exist_ok=True)
+        shutil.copytree(input_corpus, output_corpus)
+
+        # sync initial corpus before fuzzing begins
```
jonathanmetzman, line 290: nit: start with capital and end with period.

analysis/plotting.py
```diff
@@ -184,21 +186,24 @@ def coverage_growth_plot(self,
         axes.set(xlabel='Time (hour:minute)')
 
         if self._logscale or logscale:
```
jonathanmetzman, line 188: could you better explain the graph changes?

jonathanmetzman (commented):
PR mostly looks good. Please fix the nits and the issues Marc/CI pointed out.

analysis/plotting.py
dylanjwolff: For the logscale graphs the chages are to:
1. switch to 'symlog' -- i.e. linear scaling around zero to slightly reduce distortion, seemed slightly better than the default 'log' behavior
2. set the xtick range to start at zero rather than the first snapshot time
3. manually insert a tick for zero itself
4. clear the ticks set by the lineplot default plot

For linear graphs:
1. again set the minimum xtick range to zero
2. ensure that there is always at least one tick (as snapshot time can now be zero)

Both:
ensure the y-axis is at zero

The logscale plots were the best I could come up with playing around with the parameters, but there is an inherent issue with starting them at zero as `lim x-->0 log(x) = -inf`. The main visible result is that the space between 0 and the first snapshot will take up a large portion of the graph. You can see from the logscale graph attached to the PR. I'm open to suggestions to change this, or we can just revert to the old behavior and drop data from t=0 to get the same logscale graphs as before.

jonathanmetzman (approved):
lgtm

PR #1651, opened 2023-01-30
Remove bug-draft

jonathanmetzman:


Alan32Liu:
Yeah, maybe it is a good idea to remove them.
I can find more bug-based benchmarks from OSS-Fuzz, not necessarily reusing and updating these.

Alan32Liu (approved):

jonathanmetzman:
> Yeah, maybe it is a good idea to remove them. I can find more bug-based benchmarks from OSS-Fuzz, not necessarily reusing and updating these.

These are probably good to start with, unless you've tried fixing them and they are hard to fix. Have you?

Alan32Liu:
> > Yeah, maybe it is a good idea to remove them. I can find more bug-based benchmarks from OSS-Fuzz, not necessarily reusing and updating these.
> 
> These are probably good to start with, unless you've tried fixing them and they are hard to fix. Have you?

Not at this moment.
I do plan to add more bug-based benchmarks (not necessarily these ones), but it's probably cleaner to delete them now and add them back later if needed.

jonathanmetzman:
I think these CI jobs must not be considered mandatory...

PR #1653, opened 2023-01-30
Don't specify minor version of python

jonathanmetzman:


jonathanmetzman:
TBR

Alan32Liu (commented):
Thanks!
Yep, I agree it is better this way.

PR #1662, opened 2023-01-31
Don't build python anymore

jonathanmetzman:


jonathanmetzman:
Fixes https://github.com/google/fuzzbench/issues/1659
Depends on https://github.com/google/oss-fuzz/pull/9532

jonathanmetzman:
We should have a canary project in CI so PRs like this don't actually pass.

PR #1669, opened 2023-01-31
Fix jsoncpp by correcting its lib path

Alan32Liu:


Alan32Liu:
Found the correct path from [OSS-Fuzz](https://github.com/google/oss-fuzz/blob/master/projects/jsoncpp/build.sh#L27).

oliverchang (approved):

jonathanmetzman (commented):
lgtm

PR #1671, opened 2023-01-31
don't log spam

jonathanmetzman:


jonathanmetzman:
Also get rid of experiment as an optional param.
Dongge why was this added in the first place? I can't remember

experiment/run_experiment.py
```diff
@@ -96,10 +96,6 @@ def _validate_config_parameters(
     for param in missing_params:
         logs.error('Config does not contain required parameter "%s".', param)
 
-    # Notify if any optional parameters are missing in config.
-    for param in optional_params:
-        logs.info('Config does not contain optional parameter "%s".', param)
```
Alan32Liu, line 101: I used this to double-check that all my parameters are passed correctly.
It's ok to remove it.

experiment/run_experiment.py
```diff
@@ -173,8 +169,6 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
             Requirement(not local_experiment, str, True, ''),
         'worker_pool_name':
             Requirement(not local_experiment, str, False, ''),
-        'experiment':
-            Requirement(False, str, False, ''),
```
Alan32Liu, line 177: This is the experiment name, in case anyone set it in the config.
It's probably OK to remove it as well, as we expect users to pass the name from the cmdline?

Alan32Liu:
CI failed because we need to edit the testcase accordingly.

experiment/run_experiment.py
jonathanmetzman: Yeah it should be set from the command line since it changes very often (literally every use)

PR #1649, opened 2023-01-30
Anchor benchmark on commit

Alan32Liu:
Background: #1632

Fix coverage-based benchmarks:
1. Remove `--depth 1` when cloning the benchmark repo.
2. Add the commit hash to the benchmark's `benchmark.yaml`.
3. Add the commit date to the benchmark's `benchmark.yaml`.

PR #1674, opened 2023-01-31
Fix libpcap

Alan32Liu:
By reusing some of [the steps from OSS-Fuzz](https://github.com/google/oss-fuzz/tree/master/projects/libpcap).


Alan32Liu:
I changed my mind: It's cleaner to make it part of #1649 because of the nature of these modifications.

PR #1632, opened 2023-01-24
Fail on failed checkout

jonathanmetzman:

PR #1688, opened 2023-02-07
Anchor benchmark on commit 2

Alan32Liu:
Fix the CI failures in #1632:
1. Remove `--depth 1` when cloning the benchmark repo.
2. Add the **commit hash** to the benchmark's `benchmark.yaml`.
3. Add the **commit date** to the benchmark's `benchmark.yaml`.

A follow-up of #1649.


jonathanmetzman:
Thanks! Sorry for the late review

> @jonathanmetzman Could you please take a look at benchmark `freetype2`? It [builds successfully in OSS-Fuzz](https://oss-fuzz-build-logs.storage.googleapis.com/index.html#freetype2) and I don't think I missed anything from its [Dockerfile](https://github.com/google/oss-fuzz/blob/master/projects/freetype2/Dockerfile) and [build script](https://github.com/google/oss-fuzz/blob/master/projects/freetype2/build.sh) in OSS-Fuzz, yet it still failed to build in FuzzBench.
> 
> I also tried building it with the latest base image from OSS-Fuzz, but that does not work either. Thanks!

Did you figure out freetype? Also see my comment about failing gracefully, https://github.com/google/fuzzbench/commit/72038a4669280ee3b73dc93cb3ca16a79c572335#r99883219 I think we need to rethink this a bit.

Alan32Liu:
> Did you figure out freetype? 

I reverted it to its earlier commit that can build after I found it failed to build in OSS-Fuzz with a local test.
This is still strange to me, because our OSS-Fuzz cloud build log did not report any build error.
Please take a look if you have time, it's not very urgent as at least we have a working (old) version for benchmarking.

> Also see my comment about failing gracefully, [72038a4#r99883219](https://github.com/google/fuzzbench/commit/72038a4669280ee3b73dc93cb3ca16a79c572335#r99883219) I think we need to rethink this a bit.

Ah, I somehow missed that, I will reply to you over there.

PR #1690, opened 2023-02-07
Don't build python for no reason. Copy it

jonathanmetzman:


jonathanmetzman:
Not sure why even switched to building it.

jonathanmetzman:
> Not sure why even switched to building it.

One thing this may exacerbate: we may mess with the state of Python as it was left by the benchmark's Dockerfile

jonathanmetzman:
Fixes https://github.com/google/fuzzbench/issues/1659

jonathanmetzman:
TBR

Alan32Liu (commented):
LGTM

PR #1693, opened 2023-02-08
Refactor runner

jonathanmetzman:
Explicitly use OUTPUT_CORPUS_DIR as the corpus, don't assume it's at /out/corpus


jonathanmetzman:
This is useful for oss-fuzz on demand in addition to making the code more understandable.

Alan32Liu (approved):

PR #1709, opened 2023-02-15
Update LibAFL to have more compatible benchmarks (like arrow)

andreafioraldi:
LibAFL is now just more robust to poorly designed build systems and can compile more things on fuzzbench


jonathanmetzman:
> LibAFL is now just more robust to poorly designed build systems and can compile more things on fuzzbench

Could you share what you think is bad about these? If the cases are egregiously bad we can remove them. If they are bad in a normal way I think they should stay.

andreafioraldi:
No just minor things about linker args and the fact that fuzzers must not link runtimes that contain a main function. We were using a weak reference to main to compile libfuzzer harnesses but some build systems use compilation errors due to missing main functions to infer if some preprocessor variables are defined.

For instance, in arrow:
```
#if defined(__EMSCRIPTEN__)
int main() {return 0;}
#endif
```
Cmake use this program as oracle to know if __EMSCRIPTEN__ is defined but compiling it with a custom compiler linking a fuzzer runtime was causing the check to succeed (due to the weak main in the fuzzer runtime) even if __EMSCRIPTEN__ is not defined.

jonathanmetzman:
> No just minor things about linker args and the fact that fuzzers must not link runtimes that contain a main function. We were using a weak reference to main to compile libfuzzer harnesses but some build systems use compilation errors due to missing main functions to infer if some preprocessor variables are defined.
> 
> For instance, in arrow:
> 
> ```
> #if defined(__EMSCRIPTEN__)
> int main() {return 0;}
> #endif
> ```
> 
> Cmake uses this program as an oracle to know if `__EMSCRIPTEN__` is defined but compiling it with a custom compiler linking a fuzzer runtime was causing the check to succeed (due to the weak main in the fuzzer runtime) even if `__EMSCRIPTEN__` is not defined.

Ah, build troubles :-)

jonathanmetzman:
Do you know why libjpeg-turbo is failing?

Issue #1648, opened 2023-01-30
Build process is killed unexpectedly.


Alan32Liu:
A process that builds `Centipede` with benchmark `freetype2-2017` was killed unexpectedly.
The [build log file](https://storage.googleapis.com/fuzzbench-data/2023-01-27-test-local-launch/build-logs/benchmark-freetype2-2017-fuzzer-centipede.txt) did not show any error before cut off.
The [gcloud log](https://pantheon.corp.google.com/logs/query;query=2023-01-27-test-local-launch%0Aseverity%3E%3DWARNING%0Acentipede%0Afreetype2-2017;timeRange=2023-01-25T22:29:52.080Z%2F2023-01-29T23:29:52.080Z;cursorTimestamp=2023-01-29T00:17:53.044294409Z?project=fuzzbench) did not show useful info expect that it `died with <Signals.SIGKILL: 9>`.

The same error also happened on other benchmarks, e.g. [`harfbuzz`](https://pantheon.corp.google.com/logs/query;query=2023-01-27-test-local-launch%0Aseverity%3E%3DWARNING%0Acentipede%0Aharfbuzz%0A;timeRange=2023-01-25T22:29:52.080Z%2F2023-01-29T23:29:52.080Z;cursorTimestamp=2023-01-29T00:17:53.044294409Z?project=fuzzbench).

`make test-run-centipede-freetype2-2017` works perfectly.

Other info:
These processes were part of a locally-launched cloud experiment `2023-01-27-test-local-launch`, with which we want to test if the latest framework changes will affect experiments launched from locally or requested by `requests.yaml`.

Alan32Liu:
> Could this be due to the OOM killer?

I do not recall any change in memory limit or related code since  `centipede`'s last successful experiment on the same benchmark `freetype2-2017` on [18-01](https://storage.googleapis.com/www.fuzzbench.com/reports/experimental/2023-01-18-centipede-test/index.html#freetype2-2017-summary), though.

[...]

PR #1711, opened 2023-02-15
Reduce concurrent builds down to 30.

jonathanmetzman:
We only need more for um experiments, it's just nice to have for others. But it breaks things.

Related: https://github.com/google/fuzzbench/issues/1648

### DIFF
--- a/experiment/build/builder.py
+++ b/experiment/build/builder.py
@@ -34,7 +34,7 @@ from common import utils
 from common import logs
 
 from experiment.build import build_utils
-from experiment.run_experiment import DEFAULT_CONCURRENT_BUILDS
+from experiment import run_experiment
 
 if not experiment_utils.is_local_experiment():
     import experiment.build.gcb_build as buildlib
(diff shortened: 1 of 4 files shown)


jonathanmetzman:
TBR

jonathanmetzman:
CC @oliverchang @Alan32Liu

oliverchang (commented):
LGTM

experiment/build/builder.py
```diff
@@ -34,7 +34,7 @@
 from common import logs
 
 from experiment.build import build_utils
-from experiment.run_experiment import DEFAULT_CONCURRENT_BUILDS
+from experiment import run_experiment
```
Alan32Liu, line 37: Out of curiosity: Why is importing `run_experiment` preferred over `DEFAULT_CONCURRENT_BUILDS ` here, given we only use the latter?

Alan32Liu (commented):
LGTM

experiment/build/builder.py
jonathanmetzman: Youre only supposed to import packages/modules, not classes, variables or functions: https://google.github.io/styleguide/pyguide.html#22-imports

Alan32Liu: Thanks!

PR #1716, opened 2023-02-15
Fix build breakage in harfbuzz.

jonathanmetzman:
Harfbuzz is broken by using python3.10 because it clobbers python3.8. So make sure not to depend on the python3.8 installation. Fixes: https://github.com/google/fuzzbench/issues/1708


jonathanmetzman:
We probably should come up with a better fix where we don't destroy the python3.8 install.

Issue #1715, opened 2023-02-15
honggfuzz broken on proj4


jonathanmetzman:
https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/22a9614c-3927-464c-af0a-1c864b8f1f3d;step=4?project=fuzzbench

jonathanmetzman:
My fix broke the coverage build

PR #1718, opened 2023-02-15
Use correct name of liblzma when linking

jonathanmetzman:


jonathanmetzman:
Fixes https://github.com/google/fuzzbench/issues/1715

Alan32Liu (approved):

PR #1720, opened 2023-02-16
fix harfbuzz benchmark

vanhauser-thc:
without removing the freetype package before building the fuzzer can be linked to the lib which is not in the runner.


jonathanmetzman:
> without removing the freetype package before building the fuzzer can be linked to the lib which is not in the runner.

I'm surprised it's still broken because it passed CI before this change: https://github.com/google/fuzzbench/actions/runs/4187324723/jobs/7257005953

vanhauser-thc:
also fixes proj4

vanhauser-thc:
> > without removing the freetype package before building the fuzzer can be linked to the lib which is not in the runner.
> 
> I'm surprised it's still broken because it passed CI before this change: https://github.com/google/fuzzbench/actions/runs/4187324723/jobs/7257005953

it works for libfuzzer, but does not for aflplusplus variants.
maybe it is because of packages we install for aflplusplus and then are picked up by the harfbuzz configure script? something link that I guess.

jonathanmetzman (approved):
lgtm

PR #1725, opened 2023-02-16
Remove stutter in benchmark names

jonathanmetzman:


Alan32Liu (commented):
I thought the convention is `<project_name>_<binary_name>`?
Happy to change this, though, as the current names can be confusing when project/binary names have `_`.

jonathanmetzman:
> I thought the convention is `<project_name>_<binary_name>`? Happy to change this, though, as the current names can be confusing when project/binary names have `_`.

It is the convention but some of the names we have are horrible as a result of it.

PR #1732, opened 2023-02-17
Fix autodict linkage on libject for libafl

andreafioraldi:
Followup of #1709 to fix the build for libjpeg-turbo


jonathanmetzman (approved):
lgtm

PR #1741, opened 2023-02-20
[SBFT23] Update builder.Dockerfile for AFLSmart++

thuanpv:
With this, the newest version of AFLSmart++ will be used for experiments. Thanks.


fuzzers/aflsmart_plusplus/builder.Dockerfile
```diff
@@ -47,7 +47,6 @@ RUN dpkg --add-architecture i386 && \
 # Download and compile AFLSmart.
 RUN git clone https://github.com/thuanpv/aflsmart /afl && \
     cd /afl && \
-    git checkout 4286ae47e0e5d8c412f91aae94ef9d11fb97dfd8 && \
```
vanhauser-thc, line 50: I am not sure this is a good idea. this way the CI cannot see anymore when you updated your fuzzer and no test will be performed.
Not sure that it would even build a new docker and might use the previous one?
@jonathanmetzman or @Alan32Liu ?

jonathanmetzman: Yeah I don't really support doing this.

thuanpv: Thanks @vanhauser-thc for your comment. It makes sense to me. I have just update the docker file to use a specific version of the fuzzer.

PR #1748, opened 2023-02-23
Lower disk size on runners

jonathanmetzman:


Alan32Liu:
Thanks! Do we need to test if 30 GB is enough on the dev branch?

jonathanmetzman:
> Thanks! Do we need to test if 30 GB is enough on the dev branch?

It was what we used before the mutant paper.

Alan32Liu (approved):

PR #1751, opened 2023-02-24
Add a new fuzzer and experiments.

Ricardo-609:


jonathanmetzman:
Were gonna hold off on experiments for a few days because of the competition were running now

PR #1760, opened 2023-02-27
[SBFT23] libafl_libfuzzer updates

VTCAKAVSMoACE:
Further updates for the second round of evaluation of SBFT.


VTCAKAVSMoACE:
Fun fact: double-linking C++ seems to cause the exception handler to no longer catch anything :grimacing: Might affect other fuzzers for frequently-crashing cases like proj4.

jonathanmetzman:
> Fun fact: double-linking C++ seems to cause the exception handler to no longer catch anything  Might affect other fuzzers for frequently-excepting cases like proj4.

Can you elaborate on this with an example. Is it only if you link in libstdc++?

jonathanmetzman:
> > > Fun fact: double-linking C++ seems to cause the exception handler to no longer catch anything  Might affect other fuzzers for frequently-excepting cases like proj4.
> > 
> > 
> > Can you elaborate on this with an example. Is it only if you link in libstdc++?
> 
> Sure; I noticed this when debugging [this version](https://github.com/google/fuzzbench/pull/1760/commits/3b2bec7475e8de2d9af9bc6e61ef627662c074d2#diff-462f1eea701bffec2d05c8b4f9d3d314449cb437e45f0d54515994fbef2ea006R46) of libafl_libfuzzer, that we had many spurious unhandled exception errors in proj4 (for proj.db not existing). None of the other fuzzers had this issue, and it appeared during this change, so I looked further and saw that libcxx's libc++abi was being used, and that libc++ was already linked. I dropped my C++-dependent dependency and removed the linker flag, and we're back in good order now in my local fuzzbench.
> 
> My assumption is that stdc++ was being used for the libraries, and libcxx's c++abi was being used elsewhere. I didn't spend the time to fully debug this (as I was only really looking for a solution), but double-linking C++ seems erroneous, so I suspect this is the cause.

Oy this is an annoying issue. libc++ is being used by default though (we add it to CFLAGS) so I wonder what's going on here.

VTCAKAVSMoACE:
Another fun issue: without exporting some of the ASAN flags (enabled in the previous commit), the fuzzer will not be able to restore and dies, but *only* in the `run-libafl_libfuzzer-benchmark` target. It will not die in debug targets. I believe this is caused by some interaction between the signal handling of ASAN and the interactive, but I didn't really have the time to debug.

PR #1762, opened 2023-02-27
Make corpus archiving more robust

jonathanmetzman:


jonathanmetzman:
This could have some unforseen consequences. Normally if a file gets deleted for a normal fuzzer while we are archiving then we would retry. However, this would probably prevent that retry from happening.

PR #1765, opened 2023-03-02
Fix ranking

Alan32Liu:
1. Fix average rank ranking:
    * Need to replace `NaN` in the `experiment_df` with `0`; otherwise, it may cause counterintuitive ranking in rare cases (e.g., two fuzzer both found zero bug on one benchmark and `NaN` on the rest. They are ranked differently because one got 0 on benchmark `A` and the other got 0 on `B`, where some other fuzzer found some bug in `A` but no one found any bug in `B`.)
    * Use `min` as the ranking method.
    * Do not need `na_option` as `NaN` is now `0`. Otherwise, it should at least be changed to `bottom`.
2. Fix average score ranking:
    * Replace `NaN` in the `experiment_df` with `0`.

### DIFF
--- a/analysis/data_utils.py
+++ b/analysis/data_utils.py
@@ -388,7 +388,7 @@ def experiment_rank_by_average_normalized_score(experiment_pivot_df):
     # Normalize coverage values.
     benchmark_maximum = experiment_pivot_df.fillna(0).max(axis='columns')
     normalized_score = experiment_pivot_df.fillna(0).div(benchmark_maximum,
-                                               axis='index').mul(100)
+                                                         axis='index').mul(100)
 
     average_score = normalized_score.mean().sort_values(ascending=False)
     return average_score.rename('average normalized score')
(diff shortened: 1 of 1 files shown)


Alan32Liu:
Tested locally and looks alright, merged now to run a test experiment.
TBR.

oliverchang:
Nice!

Alan32Liu:
New ranking on public [cov-based benchmarks](https://storage.googleapis.com/www.fuzzbench.com/reports/experimental/2023-03-03-cov-pub-test/index.html) and [bug-based benchmarks](https://storage.googleapis.com/www.fuzzbench.com/reports/experimental/2023-03-03-bug-all-test/index.html).

Alan32Liu:
BTW, we need a more sophisticated ranking method for bug-based benchmarks.
This is because, most of the time, each benchmark only has very few reproducible bugs (e.g., 1-2), so the performances of fuzzers are very close.
Currently, we rank fuzzers by the median number of bugs found, making their performance even closer.

Alternatively, we can use tuple comparison to include more statistics, e.g., `(medium, mean, time_to_find_1st_bug)`.

PR #1766, opened 2023-03-02
Remove future & decprecation warnings

Alan32Liu:
1. Suppress a future warning on `seaborn.lineplot`'s `ci`.
2. Suppress a deprecation warning on `Orange`.
3. Adjust a warning on invalid benchmarks.

### DIFF
--- a/analysis/plotting.py
+++ b/analysis/plotting.py
@@ -16,6 +16,7 @@
 import numpy as np
 import Orange
 import seaborn as sns
+import warnings
 
 from matplotlib import colors
 from matplotlib import pyplot as plt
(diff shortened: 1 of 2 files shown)


analysis/plotting.py
```diff
@@ -16,13 +16,16 @@
...
 import seaborn as sns
+import warnings
 
 from matplotlib import colors
 from matplotlib import pyplot as plt
 from analysis import data_utils
 
 _DEFAULT_TICKS_COUNT = 12
 _DEFAULT_LABEL_ROTATION = 30
+warnings.simplefilter(action='ignore', category=FutureWarning)
```
jonathanmetzman, line 27: I think the correct move is to fix what these things are complaining about instead of ignoring them.

PR #1771, opened 2023-03-03
Fix validation of runner config options

EliaGeretto:
This PR modifies the config file validation to allow to specify `runner_machine_type` and `runner_num_cpu_cores`.


EliaGeretto:
I am proposing this change since there seems to be no way to specify `runner_num_cpu_cores`. Both the config file validation and the command line option validation refuse it. `runner_machine_type` simply seems to have the same problem.

jonathanmetzman (approved):
lgtm

PR #1774, opened 2023-03-04
[SBFT23] Update AFLSmart++

thuanpv:


thuanpv:
AFLSmart++ failed the checks for harfbuzz even though I forced it to use normal AFL. Looking at the error reported during the checks (as copied below), it shows some issues with the libmono which is quite strange to me because if it is the root cause, other benchmarks could fail as well. Looking forward to receiving your suggestions. Thanks.

'''
#22 12.99 The following packages have unmet dependencies:
#22 13.08  libmono-system-web4.0-cil : Depends: libmono-system-drawing4.0-cil (>= 5.12.0.309) but it is not going to be installed
#22 13.08                              Depends: libmono-system-web-services4.0-cil (>= 4.2.0) but it is not going to be installed
#22 13.08                              Depends: libgdiplus but it is not going to be installed
#22 13.08  monodoc-http : Depends: mono-xsp4 but it is not going to be installed or
#22 13.08                          mono-apache-server4 but it is not going to be installed or
#22 13.08                          mono-fastcgi-server4 but it is not going to be installed
#22 13.10 E: Error, pkgProblemResolver::Resolve generated breaks, this may be caused by held packages.
#22 13.10 Traceback (most recent call last):
#22 13.10   File "<string>", line 1, in <module>
#22 13.10   File "/src/fuzzers/aflsmart_plusplus/fuzzer.py", line 25, in build
#22 13.10     afl_fuzzer.build()
#22 13.10   File "/src/fuzzers/afl/fuzzer.py", line 40, in build
#22 13.10     utils.build_benchmark()
#22 13.10   File "/src/fuzzers/utils.py", line 81, in build_benchmark
#22 13.10     subprocess.check_call(['/bin/bash', '-ex', build_script], env=env)
#22 13.10   File "/usr/local/lib/python3.10/subprocess.py", line 369, in check_call
#22 13.10     raise CalledProcessError(retcode, cmd)
#22 13.10 subprocess.CalledProcessError: Command '['/bin/bash', '-ex', '/src/build.sh']' returned non-zero exit status 100.
#22 ERROR: executor failed running [/bin/sh -c echo "Run fuzzer_build to build the target" && if [ -z "$debug_builder" ] ; then fuzzer_build; fi]: exit code: 1
'''

Alan32Liu:
Thanks for reporting this, @thuanpv.
I tested building `afl` with both `harfbuzz` projects locally, and they work for me.

The log message indicates a conflict of dependencies. 
Could you please check which dependency caused this in `aflsmart_plusplus`?

thuanpv:
Thanks @Alan32Liu for your quick support. I believe the issue is related to libmono, as included in the error report. However, when I look at the build instruction of harfbuzz  (https://github.com/harfbuzz/harfbuzz/blob/main/BUILD.md), I don't see any direct dependence of harfbuzz on mono, which is the case for AFLSmart/AFLSmart++. I will see if I find more information.

PR #1779, opened 2023-03-09
Fix building libjpeg-turbo benchmark

apach301:
Hi! During my local experiments I found that libjpeg-turbo benchmark building is incorrect:

1. It ignores commit-hash specified in benchmark.yaml and always build from the latest commit in repo.
2. It downloads and builds 3 branches of libjpeg-turbo while using only fuzztarget from main-branch. Moreover, it uses fuzztarget from libjpeg repo instead of one provided in fuzzbench.

The fact that libjpeg is built with unintended version is due to `docker/benchmark-builder/checkout_commit.py` logic: it visits every directory of $SRC (which contains aflplusplus, hongfuzz etc) and tries to checkout on commit from yaml. It stops on succeed, but in case of libjpeg-turbo there are 3 versions of libjpeg repo in $SRC. I think it would be more reliable if `checkout_commit.py` will explicitly perform checkout only for `project` dir (obtained from yaml). But this solution required all benchmarks cloned from git with this defined name.

This PR proposes changes that enforce building of provided `libjpeg_turbo_fuzz.cc` target instead of using default target from remote libjpeg-turbo repo. However, I don't know how appropriate is that. Maybe it should be always build the default target.

### DIFF
--- a/benchmarks/libjpeg-turbo_libjpeg_turbo_fuzzer/build.sh
+++ b/benchmarks/libjpeg-turbo_libjpeg_turbo_fuzzer/build.sh
@@ -16,12 +16,10 @@
 set -e
 set -u
 
-cat fuzz/branches.txt | while read branch; do
-    pushd libjpeg-turbo.$branch
-    if [ "$branch" = "main" ]; then
-        sh fuzz/build.sh
-    else
-        sh fuzz/build.sh _$branch
-    fi
-    popd
-done
+cd libjpeg-turbo
+
+# Substitute original fuzz target
+rm fuzz/decompress.cc
+cp ../decompress.cc fuzz/decompress.cc
+
+sh fuzz/build.sh
--- a/benchmarks/libjpeg-turbo_libjpeg_turbo_fuzzer/Dockerfile
+++ b/benchmarks/libjpeg-turbo_libjpeg_turbo_fuzzer/Dockerfile
@@ -19,11 +19,7 @@ FROM gcr.io/oss-fuzz-base/base-builder@sha256:87ca1e9e19235e731fac8de8d1892ebe8d
 RUN apt-get update && \
     apt-get install -y make yasm cmake
 
-RUN git clone https://github.com/libjpeg-turbo/fuzz && \
-    cat fuzz/branches.txt | while read branch; do \
-      git clone https://github.com/libjpeg-turbo/libjpeg-turbo \
-          -b $branch libjpeg-turbo.$branch; \
-    done
+RUN git clone https://github.com/libjpeg-turbo/libjpeg-turbo libjpeg-turbo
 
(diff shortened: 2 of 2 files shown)


benchmarks/libjpeg-turbo_libjpeg_turbo_fuzzer/build.sh
```diff
@@ -16,12 +16,10 @@
...
-    fi
-    popd
-done
+cd libjpeg-turbo
+
+# Substitute original fuzz target
+rm fuzz/decompress.cc
+cp ../decompress.cc fuzz/decompress.cc
+
+sh fuzz/build.sh
```
SweetVishnya, line 25: Just encountered the same libjpeg-turbo benchmark build failure during my local FuzzBench experiments.

We should also add `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` definition in order to overcome the following build failure:

```
#22 16.78 In file included from /src/libjpeg-turbo/fuzz/cjpeg.cc:35:
#22 16.78 /src/libjpeg-turbo/fuzz/../cjpeg.c:700:12: error: no member named 'max_pixels' in 'cjpeg_source_struct'
#22 16.78   src_mgr->max_pixels = 1048576;
#22 16.78   ~~~~~~~  ^
```

The error appears because `max_pixels` field is enabled via the mentioned [define](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/3b19db4e6e7493a748369974819b4c5fa84c7614/cdjpeg.h#L39).

It may be fixed with the following `sed`:

```
sed -i 's/-DNDEBUG/-DNDEBUG -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION/' fuzz/build.sh
```

SweetVishnya (changes requested):

benchmarks/libjpeg-turbo_libjpeg_turbo_fuzzer/build.sh
apach301: Flag `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` should be set by default in fuzzers/utils.py when it initializes other standard compilation flags.

SweetVishnya: Yeah, you are right!

SweetVishnya (approved):
@jonathanmetzman, can we merge this? It successfully fixes libjpeg-turbo benchmark.

apach301:
> @jonathanmetzman, can we merge this? It successfully fixes libjpeg-turbo benchmark.

Which fuzztarget should be build: the one which provided in libjpeg-turbo repo or the one provided in fuzzbench?

PR #1781, opened 2023-03-10
new afl++ experiment

vanhauser-thc:


vanhauser-thc:
@Alan32Liu please give me a heads-up once this experiment can be run, thank you :-)

vanhauser-thc:
btw the CI throws a weird error:

```
=================================== FAILURES ===================================
________________________ test_run_requested_experiment _________________________

mocked_get_requested_experiments = <MagicMock name='_get_requested_experiments' id='139637487384272'>
mocked_start_experiment = <MagicMock name='start_experiment' id='139637493715712'>
[...]
        mocked_get_requested_experiments.return_value = EXPERIMENT_REQUESTS
        expected_experiment_name = '2020-06-05'
```

vanhauser-thc:
another issue with the CI is that it tests fuzzers that are not touched in the PR and outdated ... e.g. it runs `test-run-afl_qemu-libxml2_xml` and that has builder.Dockerfile that is for the old ubuntu.

Alan32Liu:
> @Alan32Liu please give me a heads-up once this experiment can be run, thank you :-)

Both reports are ready now, free feel to start at your convenience.

Alan32Liu:
> another issue with the CI is that it tests fuzzers that are not touched in the PR and outdated ... e.g. it runs `test-run-afl_qemu-libxml2_xml` and that has builder.Dockerfile that is for the old ubuntu.

Strange...
I will look into them on Monday.

vanhauser-thc:
> > @Alan32Liu please give me a heads-up once this experiment can be run, thank you :-)
> 
> Both reports are ready now, free feel to start at your convenience.

Can’t start until the two ci issues are resolved :(

Alan32Liu:
There are two errors:
1. [x] `presubmit`: Failed due to #1782.
    * Fixed by #1783.
3. [ ] Extra build test: The extra fuzzers are included because they depend on the fuzzers you changed. For example, [`afl_qemu` depends on `aflplusplus`](https://github.com/google/fuzzbench/blob/master/fuzzers/afl_qemu/fuzzer.py#L17). We have at least two options:
     1. Remove [the requirement on building all fuzzers that depends on the ones changed](https://github.com/google/fuzzbench/blob/2fb2e121d5412913f1f5bd2041c623572c150dc8/src_analysis/change_utils.py#L40).
     2. Remove the redundant/outdated fuzzers (e.g., `afl_qemu) as they do not work anyway.
     
 @jonathanmetzman, which one would you prefer?
 I reckon ii) is better, but I may miss something.
 By the way, we could also solve this by proceeding with the 'aggressive changes' I proposed about having a dev branch.

vanhauser-thc:
hmm this is twofold.
a) there are fuzzers like afl_qemu that are for the old ubuntu setup and are not working until they are migrated. so either they should be migrated or moved away until they are.
b) if a fuzzer depends on aflplusplus and fails for whatever reason, that will block updating aflplusplus.

I understand why if updating aflplusplus fuzzer depending on it should also be tested, but on the other hand this can create issues as well as we see.

IMHO fuzzer dependencies should not run in the CI.

Alan32Liu:
> a) there are fuzzers like afl_qemu that are for the old ubuntu setup and are not working until they are migrated. so either they should be migrated or moved away until they are.

I should have removed them.
I recall there was a PR to remove them after upgrading FB to Ubuntu20, but we did not merge it (for some reason I forgot).

vanhauser-thc:
Then let's move them :-)

Alan32Liu:
> Then let's move them :-)

Let's wait for @jonathanmetzman in case there is something I missed.

jonathanmetzman:
> > Then let's move them :-)
> 
> Let's wait for @jonathanmetzman in case there is something I missed.

Which ones, I think I'm generally fine with removing them, maybe there's some i think we should fix but, generally they should be removed.

PR #1787, opened 2023-03-17
Sampling corpora

dylanjwolff:
Hi!

The purpose of this PR is to allow users to easily "sample" initial corpora per trial from a larger pool of seeds. This should help to mitigate some possible issues with overfitting / overtraining to the static seed corpus of a particular benchmark program. Additionally, it allows us to begin to quantify the impact of various changes in the composition of the starting corpus on a fuzzer / campaign (https://github.com/google/fuzzbench/issues/1489).

All of these changes are opt-in and gated behind a dict in the experiment configuration YAML file. Happy to discuss if the configuration options should be restructured or reworded. I've attached an example config file for a local experiment:

[example-config.yaml.txt](https://github.com/google/fuzzbench/files/11000791/example-config.yaml.txt)

In terms of the changes themselves, the sampling is done in each runner container / process deterministically based on a random seed. For each runner, the unsampled files are removed from the $SEED_CORPUS_DIR before starting the fuzzer. 
The corpora are sampled in a "matched pairs" style design, so the 1st trial of AFL on a given target has the same corpus as the 1st trial of libfuzzer on that target and so on.


mboehme:
Hi @Alan32Liu, this is the PR related to the https://arxiv.org/abs/2212.09519 we talked about. Can you have a look?

experiment/runner.py
```diff
@@ -26,6 +27,9 @@
 import threading
 import time
 import zipfile
+from random import Random
```
jonathanmetzman, line 30: Google's style is to only import modules. So please change this to `import random` and then use `random.Random`.

Please do this for numpy as well, and please keep imports in alphabetical order.

experiment/scheduler.py
```diff
@@ -772,11 +800,25 @@ def render_startup_script_template(  # pylint: disable=too-many-arguments
...
+        'corpus_variant_id': corpus_variant_id,
         'private': experiment_config['private'],
         'cpuset': cpuset,
         'custom_seed_corpus_dir': experiment_config['custom_seed_corpus_dir'],
     }
 
+    if 'seed_sampling' in experiment_config:
+        kwargs['use_seed_sampling'] = True
+
+        kwargs['randomness_seed'] = experiment_config['seed_sampling'] \
```
jonathanmetzman, line 812: We generally use paren to break up long expressionsthat don't fit, instead of backslashes.

experiment/scheduler.py
```diff
@@ -652,12 +652,34 @@ def update_started_trials(trial_proxies, trial_id_mapping, core_allocation):
     return started_trials
 
 
+def get_corpus_variant_id_mapping(trials):
+    """Generate a shared ID for the Nth trial of each fuzzer on each benchmark
+    This lets us align e.g. the '1st' trial of AFL on bloaty-fuzz-target and
+    the '1st' trial of LibFuzzer on bloaty-fuzz-target. When the seed corpora
```
jonathanmetzman, line 658: nit: first

experiment/scheduler.py
```diff
@@ -652,12 +652,34 @@ def update_started_trials(trial_proxies, trial_id_mapping, core_allocation):
...
+    are sampled across trials, this ID tells us which trials have identical
+    starting corpora"""
```
jonathanmetzman, line 660: nit: end this with a period.

Alan32Liu:
> Hi @Alan32Liu, this is the PR related to the https://arxiv.org/abs/2212.09519 we talked about. Can you have a look?

Sure! I will read the code today.

experiment/run_experiment.py
```diff
@@ -185,6 +185,8 @@ def read_and_validate_experiment_config(config_filename: str) -> Dict:
             Requirement(False, str, True, ''),
         'runner_num_cpu_cores':
             Requirement(False, int, False, ''),
+        'seed_sampling':
+            Requirement(False, dict, False, ''),
```
Alan32Liu, line 189: I guess we only expect the `seed_sampling` dictionary to define two items in `experiment_config.yaml`:
1. `distribution` (a str, either `'UNIFORM'` or `'EXP'`).
1. `mean_seed_utilization` (a float, within `[0,1]]`).

Would it be better to define them as two individual items in the YAML file and validate them here?
This allows us to capture any mistakes of them (e.g., `'Unimplemented sampling algorithm'`) immediately when the experiment launches, before waiting for a long time to build benchmarks & fuzzers.

Alan32Liu (commented):
Looks good in general! Left two nit comments and a question out of curiosity.

experiment/scheduler.py
```diff
@@ -652,12 +652,34 @@ def update_started_trials(trial_proxies, trial_id_mapping, core_allocation):
     return started_trials
 
 
+def get_corpus_variant_id_mapping(trials):
```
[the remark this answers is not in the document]
dylanjwolff: Not sure if I understand... maybe you can clarify?

It sounds like you are proposing that groups of M trials for each fuzzer share a starting corpus, rather than just a single Nth trial sharing the corpus? So instead of

```
AFL  +  corpus 1
libfuzzer + corpus 1
AFL  +  corpus 2
libfuzzer + corpus 2
AFL  +  corpus 3
libfuzzer + corpus 3
...
AFL  +  corpus 50
libfuzzer + corpus 50
```

you might have this for e.g. 50%2:

```
AFL  +  corpus 1
libfuzzer + corpus 1
AFL  +  corpus 1
libfuzzer + corpus 1
AFL  +  corpus 2
libfuzzer + corpus 2
...
AFL  +  corpus 25
libfuzzer + corpus 25
```

If so, this seems reasonable to me, but maybe we add it in a future set of changes?

experiment/run_experiment.py
dylanjwolff: Yeah, I don't have any strong preferences for the YAML file, this sounds reasonable to me

experiment/scheduler.py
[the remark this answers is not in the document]
Alan32Liu: Ah sure! I was just curious : )

experiment/run_experiment.py
Alan32Liu: Thanks!

dylanjwolff: Configuration is now flattened out, as below:
[example-config.yaml.txt](https://github.com/google/fuzzbench/files/11397420/example-config.yaml.txt)

experiment/runner.py
dylanjwolff: Should be all set on this, but I did need to modify the `.pylintrc` file to workaround this issue:

https://github.com/pylint-dev/pylint/issues/1975

experiment/run_experiment.py
Alan32Liu: Looks amazing, thanks!

Alan32Liu:
@dylanjwolff Would you like me to launch a test experiment if this PR is ready?
If so, could you please: 
1. List the fuzzers, benchmarks, and the new flags for the experiment.
2. Add the new configuration file to  `service/` under a different name (e.g., `experiment-config-sampling.yaml`).
2. Make a trivial modification to [service/gcbrun_experiment.py](https://github.com/google/fuzzbench/blob/master/service/gcbrun_experiment.py). This will allow me to launch experiments in this PR before merging. Here is [an example](https://github.com/google/fuzzbench/pull/1818/files#diff-e870d8214beac56eaaeda2edc69de842f8382c428329dd8cf826a9cfa5edf7e9).

Thanks!

dylanjwolff:
Thanks @Alan32Liu!

I've added the YAML file and modified gcbrun. I wasn't sure what the default options should be for various parameters, so I just used one of the other YAML files in `/service`.

I don't have a preference for which fuzzers, any set of two or more should be fine. No other configuration flags etc. should be needed.

I've run a number of local experiments and everything works fine so I don't expect any issues.

Alan32Liu:
/gcbrun run_experiment.py -a --experiment-config /opt/fuzzbench/service/seed-sampling-example-config.yaml --experiment-name 2023-05-05-sample --fuzzers libfuzzer aflplusplus afl centipede honggfuzz --benchmarks bloaty_fuzz_target curl_curl_fuzzer_http freetype2_ftfuzzer harfbuzz_hb-shape-fuzzer jsoncpp_jsoncpp_fuzzer lcms_cms_transform_fuzzer libjpeg-turbo_libjpeg_turbo_fuzzer libpcap_fuzz_both libpng_libpng_read_fuzzer libxml2_xml libxslt_xpath openh264_decoder_fuzzer openssl_x509 openthread_ot-ip6-send-fuzzer proj4_proj_crs_to_crs_fuzzer re2_fuzzer sqlite3_ossfuzz stb_stbi_read_fuzzer systemd_fuzz-link-parser vorbis_decode_fuzzer woff2_convert_woff2ttf_fuzzer

Alan32Liu:
Thanks for the prompt response @dylanjwolff !
The experiment has been launched, I will come back later to check if everything goes well and add the link to the experiment data and report when they will have been created.

experiment/runner.py
```diff
@@ -157,6 +161,56 @@ def _unpack_clusterfuzz_seed_corpus(fuzz_target_path, corpus_directory):
...
+    ]
+    corpus_paths.sort()  # need to be ordered for deterministic sampling
+    logs.info('Sampling from %d files in seed corpus dir %s.',
+              len(corpus_paths), corpus_dir)
+
+    num_seeds = len(corpus_paths)
+
+    if distribution == 'UNIFORM':
+        trial_num_seeds = gen.randint(
+            1, int(numpy.round(mean_seed_util * num_seeds)))  # inclusive []
```
Alan32Liu, line 192: Here is a question that I wanted to ask earlier but was unsure if I missed anything:
If the benchmark does not come with a seed corpus, or `mean_seed_util * num_seeds` is less than 0.5 (i.e., evaluates to `randint(1, 0)`), would this cause a `ValueError` because the upper bound (`0`) is smaller than the lower bound (`1`)?

Alan32Liu: Previously I thought we would only expect to run experiments on preselected benchmarks that have enough seeds, but I guess my assumption was wrong?

Alan32Liu:
[Experiment data](https://storage.googleapis.com/fuzzbench-data/index.html?prefix=2023-05-05-sample/).
Experiment report URL unready yet.

service/seed-sampling-example-config.yaml
```diff
@@ -15,7 +15,7 @@ preemptible_runners: true
 
 # This experiment should generate a report that is combined with other public
 # "production" experiments.
-merge_with_nonprivate: true
+merge_with_nonprivate: false
```
Alan32Liu, line 18: I suspect that the experiment did not use the new sampling strategy: Its report shows the results of 20 fuzzer trials running on each benchmark for 23 hours, which 1) disagrees with the experiment config, and 2) should not complete in 16 hours.
The strange behavior may be caused by this line, which allows re-using previous results. As a result, the experiment merges previous 23-hour 20-trial results (without seed sampling)with the new result (with sampling).

I pushed this commit to fix it and re-run the experiment.

Alan32Liu:
/gcbrun run_experiment.py -a --experiment-config /opt/fuzzbench/service/seed-sampling-example-config.yaml --experiment-name 2023-05-06-sample --fuzzers libfuzzer aflplusplus afl centipede honggfuzz --benchmarks bloaty_fuzz_target curl_curl_fuzzer_http freetype2_ftfuzzer harfbuzz_hb-shape-fuzzer jsoncpp_jsoncpp_fuzzer lcms_cms_transform_fuzzer libjpeg-turbo_libjpeg_turbo_fuzzer libpcap_fuzz_both libpng_libpng_read_fuzzer libxml2_xml libxslt_xpath openh264_decoder_fuzzer openssl_x509 openthread_ot-ip6-send-fuzzer proj4_proj_crs_to_crs_fuzzer re2_fuzzer sqlite3_ossfuzz stb_stbi_read_fuzzer systemd_fuzz-link-parser vorbis_decode_fuzzer woff2_convert_woff2ttf_fuzzer

Alan32Liu:
05-06:
[Experiment data](https://storage.googleapis.com/fuzzbench-data/index.html?prefix=2023-05-06-sample/).
[Experiment report](https://storage.googleapis.com/www.fuzzbench.com/reports/2023-05-06-sample/index.html) unready.

Alan32Liu:
The 45-minute test run looks fine.
Shall we merge this (to the master branch or an individual branch)? @dylanjwolff @jonathanmetzman

experiment/runner.py
dylanjwolff: Yeah, this is a good point. I'm not sure restricting the benchmarks makes sense, since people may use their own custom seed corpora and its possible the default corpora might change over time.

Also, does Fuzzbench automatically insert a file for the "empty" corpus? I seem to remember seeing a "hi.txt" file floating around on "empty" runs, maybe because AFL won't run with an empty input folder?

If not can add an assertion that the size of the seed corpus should be non-zero to make things more clear when it fails, but I expect it won't come up very frequently since the feature is opt-in for now.

dylanjwolff: Looking at the results from the run now and it seems look like `lcms_cms_transform_fuzzer` has no coverage to start, so I guess there is no issue

dylanjwolff:
> Shall we merge this (to the master branch or an individual branch)?

I vote master!

I think it's an important feature and it's opt-in, so it shouldn't disrupt anyone's current workflow. At some point maybe there is even a discussion to be had about whether or not it should be the default, but I think at least making it an option for everyone is a good idea

dylanjwolff:
@jonathanmetzman Any thoughts on getting this merged into master?

PR #1789, opened 2023-03-20
Revert addition of benchmarks

jonathanmetzman:


jonathanmetzman:
@Alan32Liu if you want to reland this can you do so in a way that doesn't break experiments.

Alan32Liu:
Yep, sorry that I should have checked with you before adding so many benchmarks.

PR #1790, opened 2023-03-20
Redo ecofuzz experiment

jonathanmetzman:
@Alan32Liu @andreafioraldi

### DIFF
--- a/service/experiment-requests.yaml
+++ b/service/experiment-requests.yaml
@@ -20,6 +20,13 @@
 # Please add new experiment requests towards the top of this file.
 #
 
+- experiment: 2023-03-20-ecofuzz
+  description: "Test EcoFuzz."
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
TBR

Alan32Liu (commented):
LGTM, thanks!

andreafioraldi:
Ty, it looks AFL didn't start yet on some targets like woff2, it will be scheduled later or is it a bug?
It is reasonable for Ecofuzz to fail to run some benchmarks but AFL should work with any benchmark in theory.
Maybe try to reduce the number of trials to 10 or even 5 if the issue is resource exhaustion.

jonathanmetzman:
I think AFL is working. I don't think there's resource exhaustion. Measurement is just bad. sorry

PR #1798, opened 2023-03-25
afl++ version comparison

vanhauser-thc:
This run https://www.fuzzbench.com/reports/experimental/2023-03-23-various/index.html
showed that v3.14c has a better performance than the current 4.06a which is baffling. The difference is huge!

so this experiment tests all afl++ versions starting with 3.13 up to current with the same llvm version (it needs version 13 because 3.14 is obviously not updated to current llvm changes).

It might be the llvm version it might be something else ... no clue, I need more data.
Curous that this did not come up earlier as I test this rigorously before a release.

### DIFF
--- a/fuzzers/aflplusplus_313/builder.Dockerfile
+++ b/fuzzers/aflplusplus_313/builder.Dockerfile
@@ -1,0 +1,56 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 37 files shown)


vanhauser-thc:
@Alan32Liu if you can merge this latest on monday ... I need to check this out, and it will take some time I fear...
the failed compilations are nothing I can do about as these are older versions.

Alan32Liu (approved):

Alan32Liu:
Done.
Note that there are some CI failures.

> @Alan32Liu if you can merge this latest on monday ... I need to check this out, and it will take some time I fear... the failed compilations are nothing I can do about as these are older versions.

PR #1804, opened 2023-03-29
Fix bad link in documentation.

mvanotti:
The last link in the ["Setting up a cloud project"](https://google.github.io/fuzzbench/running-a-cloud-experiment/setting-up-a-google-cloud-project/) docs was wrong.


Alan32Liu (approved):
Good catch!
Thanks!

Issue #1801, opened 2023-03-29
Cloud Config: Passwords with non-alphanumeric characters might not work.


mvanotti:
I am following the documentation for [setting up a cloud project](https://google.github.io/fuzzbench/running-a-cloud-experiment/setting-up-a-google-cloud-project/) and they mention to set up a password for the sql database.
I used a chrome auto-suggested password that contained non-alphabetic characters ('@', '.', and ':'), and that caused the `alembic upgrade` command to fail:

```shell
PYTHONPATH=. alembic upgrade head
```

```
  File "/data/fuzzbench/database/alembic/env.py", line 90, in <module>                                                                     
    run_migrations_online()                                                                                                                
  File "/data/fuzzbench/database/alembic/env.py", line 73, in run_migrations_online                                                        
    connectable = engine_from_config(                                                                                                      
  File "/data/fuzzbench/.venv/lib/python3.10/site-packages/sqlalchemy/engine/create.py", line 743, in engine_from_config                   
    return create_engine(url, **options)                                                                                                   
  File "<string>", line 2, in create_engine                                                                                                
  File "/data/fuzzbench/.venv/lib/python3.10/site-packages/sqlalchemy/util/deprecations.py", line 309, in warned                           
    return fn(*args, **kwargs)                                                                                                             
  File "/data/fuzzbench/.venv/lib/python3.10/site-packages/sqlalchemy/engine/create.py", line 518, in create_engine                        
    u = _url.make_url(url)                                                                                                                 
  File "/data/fuzzbench/.venv/lib/python3.10/site-packages/sqlalchemy/engine/url.py", line 725, in make_url                                
    return _parse_rfc1738_args(name_or_url)                                                                                                
  File "/data/fuzzbench/.venv/lib/python3.10/site-packages/sqlalchemy/engine/url.py", line 781, in _parse_rfc1738_args                     
    components["port"] = int(components["port"])                                                                                           
ValueError: invalid literal for int() with base 10: '@127.0.0.1:5432' 
```

It seems like the regexp used to parse the database url does not support a password with strange characters.
A workaround is to remove the offending characters from the password.

Alan32Liu:
Sorry, I am not sure if I understood you correctly.
Would removing the offending characters affect the correctness of the password?

Alternatively, we can emphasize that non-alphabetic characters are not supported for now.

mvanotti:
The password is correct and valid (with the non-alphanumeric characters).

[...]

PR #1812, opened 2023-04-03
Support non-alphanumeric passwords in alembic.

mvanotti:
This commit modifies alembic's `env.py`, so that it supports non-alphanumeric passwords.

The script creates an url to connect to the database, but the passworld was not url-escaped. Passwords with characters like `@`, `:` and `.` were causing failures.

The password is set as a config in sqlalchemy, which uses python's string interpolation. Now we replace instances of `%` with `%%` so that passwords that contain `%` can be used.

Fixes issue #1801

### DIFF
--- a/database/alembic/env.py
+++ b/database/alembic/env.py
@@ -15,6 +15,7 @@
 
 from logging.config import fileConfig
 import os
+import urllib.parse
 
 from sqlalchemy import engine_from_config
 from sqlalchemy import pool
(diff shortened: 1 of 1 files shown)


Alan32Liu:
> This commit needs more work, I barely tested it.

I see, I will convert it to a draft for now, but feel free to convert it back once you reckon it is ready.
Thanks!

PR #1813, opened 2023-04-03
Fix error message in stop-experiment.py

mvanotti:
It now prints the name of the script instead of `{0}`


jonathanmetzman (approved):
lgtm

PR #1825, opened 2023-04-19
[common] Use logs.error to report errors in create_instance

mvanotti:


mvanotti:
I haven't tested it but I hope it works :)

jonathanmetzman:
I don't like this PR. It's not necessarily an error when we can't create an instance. Usually it is not. Usually it is because we run into the quota.

mvanotti:
What happens if we run out of quota? Does it keep trying ?

In any case, can we at least put it as a warning? I had to parse all the logs before figuring out why it was failing.

PR #1826, opened 2023-04-19
[gcb_build] log output if command fails.

mvanotti:


mvanotti:
I haven't tested it. I hope it works :)

jonathanmetzman (changes requested):
I don't think this is a good idea. It will probably hit some limit and error out (i think we used to do this). Text containing failed build logs can be viewed in your experiment bucket.

mvanotti:
There are other reasons for a failure. For example, if you mispelt workerPools and instead workerpool, the build command will not work and you will have no idea what went wrong. Another example, is that this command *streams* the logs by default, and it *will fail* if the service account is not a Viewer for your entire project (which I think it is a problem by itself), however, both of those errors would be lost if we don't log the error.

Running this build command locally to repro the issue is not easy, but even if you did, if there are permission errors wrt the service account, you won't be able to repro those locally.

mvanotti:
> Where? in the config? We should add an earlier check.

Yes, I had that issue before and I didn't get the error logged anywhere, because the program was not able to communicate with the worker pool. What I mean is that if there's a problem *with the cli command*, it is not logged anywhere.

> Could you explain why this is problematic?

It was another example of the CLI failing without getting any logs out. 

> I think my point about the limit of log length applies. This will break things.

I see what you mean. Some ideas:

* Print the first X bytes.
* Print *only* `stderr`

PR #1838, opened 2023-05-06
fishfuzz benchmark

vanhauser-thc:
test the new fishfuzz fuzzer


vanhauser-thc:
@Alan32Liu 
this can be merged. only specific bug benchmarks are enabled that work.
It can happen that php and bloaty take too long to build and are then disabled.
but as this is a first test run for fishfuzz this is fine.

Alan32Liu (approved):

PR #1845, opened 2023-05-21
afl++ experiment

vanhauser-thc:
this tests a new variant of symqemu concolic within afl++ and a different selection mechanism for splicing


vanhauser-thc:
@Alan32Liu can you merge this please? thank you! :)

Alan32Liu (approved):

Alan32Liu:
> the changes to the base aflplusplus here do not impact performance positively or negatively, it is enhancements and fixes for qemu, llvm 17, custom mutators, etc. the change that could make an impact is the variant I have in here for weighted splicing, that is why I am testing it.
> 
> I understand that on oss-fuzz it should only be updated if everything is right and needs thorough testing. for fuzzbench it needs to be more up to date though.
> 
> if you want I can add a variant aflplusplus_fixed and maybe you could add something like a cron job to test this vs the current aflplusplus set once a month to ensure nothing is going wrong?
> 
> tldr - I can add that line, but I would want rather handle this differently.

Thanks for the detailed explanation! I've approved it. 
If it does not affect performance, then it is totally fine by me : )

PR #1846, opened 2023-05-22
Add RUSTFLAGS to LibAFL fuzzers

tokatoka:
This will make rust compiler build more optimized code

### DIFF
--- a/fuzzers/libafl/builder.Dockerfile
+++ b/fuzzers/libafl/builder.Dockerfile
@@ -45,6 +45,7 @@ RUN cd /libafl && git checkout 8ff8ae41f1ed2956bb1e906c5c7bd0505ca110c0 || true
 RUN cd /libafl && \
     unset CFLAGS CXXFLAGS && \
     export LIBAFL_EDGES_MAP_SIZE=2621440 && \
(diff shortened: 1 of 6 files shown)


jonathanmetzman:
CC @andreafioraldi

tokatoka:
Can we merge this?

jonathanmetzman:
CC @vanhauser-thc

jonathanmetzman (approved):
lgtm

vanhauser-thc:
I am not sure this is a good idea.
How oss-fuzz works is that it builds the package for fuzzing and then clusterfuzz fuzzes that package. however these are usually different machines.
so if oss-fuzz optimizes for the CPU oss-fuzz is building the packages, this an break when run on clusterfuzz - or just have a bit worse performance if lucky.

jonathanmetzman:
This is in fuzzbench. But yeah, good point.

jonathanmetzman:
I'm going to have to revert this.

vanhauser-thc:
Ah did this on my phone and thought it is oss-fuzz :-D
We ran into this issue when ppl downloaded our docker container and the chip there was older than where it was built.

tokatoka:
no no, this is fuzzbench 😢

PR #1847, opened 2023-05-25
remove outdated instances, rename entropic to libfuzzer

vanhauser-thc:
This is a proposal :)

I would propose to rename entropic to libfuzzer, it is unclear to people that this is the current state of libfuzzer.
And deleting the outdated libfuzzer.

Additionally I removed any variant (not main fuzzer) that was not modified in the last 6 months, so basically cleanup.
If someone needs it it can be easily restored.

WDYT? 

@jonathanmetzman  @Alan32Liu


tokatoka:
you can also delete libafl_tokens_alt/neu
`fuzzers/fuzzbench_tokens` are already deleted from libafl
cc @andreafioraldi

Alan32Liu:
> Additionally I removed any variant (not main fuzzer) that was not modified in the last 6 months, so basically cleanup. If someone needs it it can be easily restored.

Thanks! Yes, you must have read my mind : )
I was always hoping we could keep&maintain the core fuzzers & benchmarks in the main branch, and experiment with new fuzzers/variations & benchmarks in individual branches.
It will clearly show people which version of fuzzers/benchmarks/fuzzbench were used in their experiment, and make reproduction easier.

Alan32Liu:
@kjain14: Just to double-check, are there any UM fuzzers that you'd like to keep?

kjain14:
can we keep the following UM fuzzers: aflplusplus_um_random, aflplusplus_um_random_75, aflplusplus_um_prioritize, aflplusplus_um_prioritize_75, honggfuzz_um_random, honggfuzz_um_random_75, honggfuzz_um_prioritize, honggfuzz_um_prioritize_75?

we need to run some more eval for the paper on these ones, but otherwise we should be good to remove the others.

PR #1860, opened 2023-06-12
new afl++ experiment

vanhauser-thc:
This tests the performance of 4.06c with 4.07c and the new mutations engine (binary and text) in both a coverage and a bug benchmark.

@Alan32Liu happy hunting why the experiments are not starting and I hope you enjoyed your public holiday :-)


Alan32Liu (approved):

vanhauser-thc:
hmm again neither of the two experiments did start ... :(

Alan32Liu:
I think I found out the reason: The VM instance that runs the auto experiment service got 'connection refused' when it attempted to query our database.
I am discussing with Jonathan on that.

PR #1869, opened 2023-07-04
new bug benchmarks

vanhauser-thc:
adds 7 bug benchmarks (2 text, 5 binary) so altogether now 12 which is makes the bug benchmark way more useful (4 text, 8 binary).  in only of these afl++ was the best in sbft23, so I fear I am hurting myself here ;-)

### DIFF
--- a/benchmarks/openh264_decoder_fuzzer/build.sh
+++ b/benchmarks/openh264_decoder_fuzzer/build.sh
@@ -15,11 +15,6 @@
 #
 ################################################################################
 
-# prepare corpus
-svn export https://github.com/mozillasecurity/fuzzdata.git/trunk/samples/h264 corpus/
-mv ./res/*.264 ./corpus/
-zip -j0r ${OUT}/decoder_fuzzer_seed_corpus.zip ./corpus/
-
 # build
 if [[ $CXXFLAGS = *sanitize=memory* ]]; then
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
@Alan32Liu 
https://www.fuzzbench.com/reports/experimental/2023-07-14-aflpp-bug/index.html shows that the re-added bug benchmarks are working fine. what is failing are outdated benchmarks that try to install nonexisting libstdc++-5. 
 can this PR be merged? or is anything missing?

Alan32Liu:
@jonathanmetzman would we be interested in adding new bug-based benchmarks?
I recall that we decided not to because bug grouping is unreliable and we want to avoid having too many benchmarks.
Not sure if I remembered that correctly, or if I missed anything.

vanhauser-thc:
@jonathanmetzman @Alan32Liu ? :)

Alan32Liu:
> @jonathanmetzman @Alan32Liu ? :)

Sorry, Jonathan and I were (and probably will be) distracted by other projects a bit.
I noticed that almost all CI failed. Shall we fix them?
Also, Jonathan is working on a big change on FB; maybe we could pause this until he finishes the change?
This may also avoid some repeated work for you : )
If you'd like use them in your experiments, we can still do that in a PR.

Thanks!

PR #1880, opened 2023-08-04
sfuzz 

zerokay:
compare against core fuzzers

PR #1886, opened 2023-08-21
Fix a case where `fuzz_target_path` is mysteriously None

DonggeLiu:
This was observed in [this experiment](https://github.com/google/fuzzbench/pull/1880#issuecomment-1681602328). Here is [the error log](https://pantheon.corp.google.com/logs/query;cursorTimestamp=2023-08-17T12:43:22.568776528Z;endTime=2023-08-21T01:38:11.678Z;query=jsonPayload.experiment%3D%222023-08-04-sfuzz%22%0AjsonPayload.benchmark%3D%22woff2_convert_woff2ttf_fuzzer%22%0Aseverity%3E%3DWARNING%0AjsonPayload.message!%3D%22Corpus%20not%20found%20for%20cycle:%200.%22%0Atimestamp%3D%222023-08-17T12:43:18.463572869Z%22%0AinsertId%3D%2218yfmbdffhas18%22;startTime=2023-08-16T01:38:11.678Z?project=fuzzbench):
```
  File "/src/experiment/runner.py", line 454, in experiment_main
    runner.conduct_trial()
  File "/src/experiment/runner.py", line 276, in conduct_trial
    self.set_up_corpus_directories()
  File "/src/experiment/runner.py", line 261, in set_up_corpus_directories
    _unpack_clusterfuzz_seed_corpus(target_binary, input_corpus)
  File "/src/experiment/runner.py", line 132, in _unpack_clusterfuzz_seed_corpus
    seed_corpus_archive_path = get_clusterfuzz_seed_corpus_path(
  File "/src/experiment/runner.py", line 98, in get_clusterfuzz_seed_corpus_path
    fuzz_target_without_extension = os.path.splitext(fuzz_target_path)[0]
  File "/usr/local/lib/python3.10/posixpath.py", line 118, in splitext
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

I could not reproduce this error locally, but this PR should fix the error.

### DIFF
--- a/experiment/runner.py
+++ b/experiment/runner.py
@@ -95,6 +95,8 @@ def _clean_seed_corpus(seed_corpus_dir):
 def get_clusterfuzz_seed_corpus_path(fuzz_target_path):
     """Returns the path of the clusterfuzz seed corpus archive if one exists.
     Otherwise returns None."""
+    if fuzz_target_path is None:
+        return None
     fuzz_target_without_extension = os.path.splitext(fuzz_target_path)[0]
     seed_corpus_path = (fuzz_target_without_extension +
                         SEED_CORPUS_ARCHIVE_SUFFIX)


jonathanmetzman:
Not sure this is a good idea. I'm more interested in why this fails. I think this fix could potentially mess up results. If one fuzzer is messing with paths and we can't use the seed corpus it's better it fails then runs without using the seed corpus when others do use it (skewing the results)

jonathanmetzman:
And these other fuzzers do appear to find the seed corpus: https://www.fuzzbench.com/reports/experimental/2023-08-04-sfuzz/index.html

oliverchang:
> Not sure this is a good idea. I'm more interested in why this fails. I think this fix could potentially mess up results. If one fuzzer is messing with paths and we can't use the seed corpus it's better it fails then runs without using the seed corpus when others do use it (skewing the results)

+1 we should fix the root cause here, rather than papering over issues.

Alan32Liu:
> Not sure this is a good idea. I'm more interested in why this fails. I think this fix could potentially mess up results. If one fuzzer is messing with paths and we can't use the seed corpus it's better it fails then runs without using the seed corpus when others do use it (skewing the results)

I agree that it's better to let it run without using the seeds: I thought that's the intention of this PR?
Without this PR, the program will crash and won't be able to run.

But finding the root cause is a better idea. I will look further into that.

jonathanmetzman:
> > Not sure this is a good idea. I'm more interested in why this fails. I think this fix could potentially mess up results. If one fuzzer is messing with paths and we can't use the seed corpus it's better it fails then runs without using the seed corpus when others do use it (skewing the results)
> 
> I agree that it's better to let it run without using the seeds: I thought that's the intention of this PR? Without this PR, the program will crash and won't be able to run.

We don't agree :-) I think it's better to produce a noticeable failure than a subtly incorrect failure.

> 
> But finding the root cause is a better idea. I will look further into that.

Alan32Liu:
> We don't agree :-) I think it's better to produce a noticeable failure than a subtly incorrect failure.

Oh, I see!

DonggeLiu:
Closing this as we will address it in new PRs.

PR #1889, opened 2023-08-24
update afl++ commit id

vanhauser-thc:
@Alan32Liu please merge :)
so new comparisons use a current AFL++ state


Alan32Liu:
Thanks!
It seems CI tests failed due to [`[-] PROGRAM ABORT : -a input mode needs to be "text" or "binary".
         Location : main(), src/afl-fuzz.c:540`](https://github.com/google/fuzzbench/actions/runs/5965023393/job/16181516241?pr=1889#step:8:15332).

Is this expected?

vanhauser-thc:
thanks for checking. yes that is to be expected as that was just an outdated variant. I removed all outdated variants now :)

Issue #1883, opened 2023-08-14
Invalid OSS-Fuzz corpus download url


jiradeto:
Hi I found an issue that for php-fuzz-parser and stbi_read_fuzzer benchmarks, the url to download oss-fuzz are invalid due to the condition: https://github.com/google/fuzzbench/blob/a516f8889f44b9804c23f8c5c7807eb8a0b55d6e/experiment/run_experiment.py#L382-L385

For example, the download link for php is

_php-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/php-fuzz-parser/public.zip_ 

However the valid one that I can download from is 

_php-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/php_php-fuzz-parser/public.zip_

The same also applies for stbi_read_fuzzer. 

I can create PR to fix but what would be the ideal solution for this? Should we have new attribute `oss_fuzz_corpus_url` in benchmark.yaml for these certain benchmarks?

jonathanmetzman:
This is fixed right?

PR #1895, opened 2023-09-05
Fix invalid oss-fuzz corpus download url

jiradeto:
To solve invalid oss-fuzz corpus download url problem described in https://github.com/google/fuzzbench/issues/1883


jonathanmetzman (approved):
LGTM

PR #1932, opened 2023-12-14
LibAFL: scoped coverage experiment

addisoncrump:
Attempting to address or partially resolve a coverage information gap described [by Marc Heuse regarding libwebp](https://www.srlabs.de/blog-post/advanced-fuzzing-unmasks-elusive-vulnerabilities) (cc @vanhauser-thc, you might be interested).

Namely, this implements "scoped" coverage -- the coverage of edges constrained to an observable scope (in this case, stack depth). This allows us to effectively measure the number of times in which a block is executed during a function call, only for that exact frame (i.e. uses of that block in callees do not contribute to this count).

The current implementation does not scope for nested loops, only for function calls; this could potentially be implemented using e.g. sancov's CFG pass to identify cycles, but this will get expensive fast.

Please run this experiment with just `libafl` and `libafl_scoped_cov`, provided in this PR.


vanhauser-thc:
/gcbrun run_experiment.py -a --experiment-config /opt/fuzzbench/service/experiment-config.yaml --experiment-name 2023-12-14-libafl --fuzzers libafl libafl_scoped_cov

PR #1940, opened 2023-12-23
Update pip package to be consistent with the code and each other

Alan32Liu:
So that it can run on my workstation : )


Alan32Liu:
They are false positives according to:
https://stackoverflow.com/questions/75789383/pylint-func-max-is-not-callable#:~:text=The%20Pylint%20error%20you%20have,it%20has%20a%20max%20method.

PR #1968, opened 2024-04-15
Fix log in retry wrap

gustavogaldinoo:
Logs made in retries always should check for the `log_retries` parameter, to avoid the following cyclic scenario:

1. A log fails (e.g. permission issue)
2. The logging method uses a retry wrapper, so it tries again
3. The code inside the retry wrapper also calls a log method, so we go back to step 1

### DIFF
--- a/common/retry.py
+++ b/common/retry.py
@@ -67,9 +67,10 @@ def wrap(  # pylint: disable=too-many-arguments
                 sleep(get_delay(num_try, delay, backoff))
                 return True
 
-            logs.error('Retrying on %s failed with %s. Raise.',
-                       function_with_type,
-                       sys.exc_info()[1])
+            if log_retries:
+                logs.error('Retrying on %s failed with %s. Raise.',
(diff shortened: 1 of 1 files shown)


common/retry.py
```diff
@@ -67,9 +67,10 @@ def handle_retry(num_try, exception=None):
                 sleep(get_delay(num_try, delay, backoff))
                 return True
 
-            logs.error('Retrying on %s failed with %s. Raise.',
-                       function_with_type,
-                       sys.exc_info()[1])
+            if log_retries:
```
jonathanmetzman, line 70: This is a good fix.
But I think it would be better to eliminate the loop/circular import entirely, what do you think?

gustavogaldinoo: Thanks Jonathan, I agree. 

I am closing this PR as I created [another one](https://github.com/google/fuzzbench/pull/1969) trying to get rid of the cycle entirely

PR #1969, opened 2024-04-17
Remove retry circular dependency

gustavogaldinoo:
Removing usage of `@retry.wrap` in logs methods.

This removal is necessary because retries also use logs inside of their implementation so we want to avoid the following cyclic scenario:

1. Log method is called
2. A log fails (e.g. permission issue)
3. The logging method uses a retry wrapper, so it tries again
4. The code inside the retry wrapper also calls a log method, so we go back to step 1

Before this PR, this scenario was avoided by using a `log_retries` parameter, that would disable loggings on retries functions when the retry was called from a log function. 

This PR remove this parameter and get rid of the circular dependency by making a custom retry logic inside of the logs methods.

### DIFF
--- a/common/logs.py
+++ b/common/logs.py
@@ -36,7 +36,7 @@ _default_extras = {}
 
 LOG_LENGTH_LIMIT = 250 * 1000
 
-NUM_ATTEMPTS = 5
+NUM_ATTEMPTS = 4
 RETRY_DELAY = 1
 BACKOFF = 2
 
--- a/common/logs.py
+++ b/common/logs.py
@@ -13,10 +13,12 @@
 # limitations under the License.
 """Set up for logging."""
 from enum import Enum
+from common import retry
 import logging
 import os
 import sys
 import traceback
+import time
 
 import google.cloud.logging
 from google.cloud.logging_v2.handlers.handlers import CloudLoggingHandler
@@ -25,7 +27,6 @@ from google.cloud import error_reporting
 # Disable this check since we have a bunch of non-constant globals in this file.
 # pylint: disable=invalid-name
 
-from common import retry
 from common import utils
 
 _default_logger = None
@@ -37,6 +38,7 @@ LOG_LENGTH_LIMIT = 250 * 1000
 
 NUM_RETRIES = 5
 RETRY_DELAY = 1
+BACKOFF = 2
 
 
 def _initialize_cloud_clients():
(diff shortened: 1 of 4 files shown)


common/logs.py
```diff
@@ -13,10 +13,12 @@
 # limitations under the License.
 """Set up for logging."""
 from enum import Enum
+
 import logging
 import os
 import sys
 import traceback
+import time
```
jonathanmetzman, line 21: Annoying google nit, can you move this line up so the imports are alphabetical order

common/logs.py
```diff
@@ -25,8 +27,8 @@
 # Disable this check since we have a bunch of non-constant globals in this file.
 # pylint: disable=invalid-name
 
-from common import retry
 from common import utils
+from common.retry import get_delay
```
jonathanmetzman, line 31: Annother annoying Google nit:we don't import functions, only modules:
(see https://google.github.io/styleguide/pyguide.html)

gustavogaldinoo: Oh, got it!

I did this way to avoid logs importing retry and retry importing logs, not sure if it is a problem since log is not using retry.wrap decorator though, but I wanted to reuse the get_delay function

common/logs.py
```diff
@@ -153,47 +156,61 @@ class LogSeverity(Enum):
...
-    }
-    all_extras = _default_extras.copy()
-    extras = extras or {}
-    all_extras.update(extras)
-    struct_message.update(all_extras)
-    severity = LogSeverity(severity).name
-    logger.log_struct(struct_message, severity=severity)
+    # Custom retry logic to avoid circular dependency
+    # as retry from retry.py uses log
+    for num_try in range(1, NUM_RETRIES + 1):
```
jonathanmetzman, line 164: Can you do this from 0? Or change NUM_RETRIES to NUM_ATTEMPTS?
As far as I can tell this will run the loop a max of 3 times. but num_retries=3 technically implies a max of 4.

common/logs.py
jonathanmetzman: THis is a good point. I think we should just reimplement get_delay here instead of having circular imports.

jonathanmetzman: Or actually, can you move it to a third module that both modules can import?

common/logs.py
gustavogaldinoo: Nice, I decided to change the nomenclature to NUM_ATTEMPTS, as the get_retry function already was 1-index based

common/test_common_utils.py
```diff
@@ -0,0 +1,31 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for utils.py"""
+
+from common import utils
+
+
+def test_get_retry_delay():
```
jonathanmetzman, line 19: Can you mock sleep so that the test doesn't take more time than it needs to.

PR #1970, opened 2024-04-17
Fix LibAFL's builder.Dockerfile and remove `libafl_text`

tokatoka:
- we have duplicate rust installation in builder.Dockerfile
- libafl_text is outdated, so I removed it


tokatoka:
fwiw; harfbuzz and php-fuzz-parser is broken cuz it is not compatible with the latest llvm version that we use. (and it is not used in the standard experiment either

PR #1971, opened 2024-04-18
Improving log message

gustavogaldinoo:
Improving log message by displaying only function name, instead of trying to print function itself.

Also took the opportunity to replace `Callable` with `FunctionType` when declaring argument type, as Callable is more generic.

Before:
Screenshot 2024-04-18 at 11 43 37

After:
Screenshot 2024-04-18 at 12 06 15

### DIFF
--- a/experiment/build/builder.py
+++ b/experiment/build/builder.py
@@ -23,7 +23,7 @@ import subprocess
 import sys
 import time
 import types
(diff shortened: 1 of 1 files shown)


jonathanmetzman:
Add me as a reviewer to make sure I see it!

PR #1977, opened 2024-05-06
Add method to check if benchmarks are mixed

gustavogaldinoo:
Currently, Fuzzbench fails to generate a report if benchmarks are a mix between bug and coverage benchmarks.

This PR adds a method to check if benchmarks are a mix between bug and coverage benchmarks before starting the experiment, to spot the problem early and avoid wasting time.

### DIFF
--- a/common/benchmark_utils.py
+++ b/common/benchmark_utils.py
@@ -165,18 +165,13 @@ def get_language(benchmark):
     return config.get('language', 'c++')
 
 
-def are_benchmarks_mixed(benchmarks):
-    """"Returns True if benchmarks list
-    is a mix of bugs and coverage benchmarks"""
-    if not benchmarks:
-        return False
+def are_benchmarks_mixed(benchmarks=None):
+    """Returns True if benchmarks list
+    is a mix of bugs and coverage benchmarks."""
+    if benchmarks is None:
+        benchmarks = get_all_benchmarks()
 
-    selected_coverage_benchmarks = get_coverage_benchmarks(benchmarks)
-    only_coverage_benchmarks = len(selected_coverage_benchmarks) == len(
-        benchmarks)
-    only_bug_benchmarks = len(selected_coverage_benchmarks) == 0
+    unique_benchmarks_types = set(
+        get_type(benchmark) for benchmark in benchmarks)
 
-    if only_bug_benchmarks or only_coverage_benchmarks:
-        return False
-
-    return True
+    return len(unique_benchmarks_types) > 1
--- a/common/benchmark_utils.py
+++ b/common/benchmark_utils.py
@@ -166,8 +166,8 @@ def get_language(benchmark):
 
 
 def are_benchmarks_mixed(benchmarks=None):
-    """Returns True if benchmarks list
-    is a mix of bugs and coverage benchmarks."""
+    """Returns True if benchmarks list is a mix of bugs and coverage
+    benchmarks."""
     if benchmarks is None:
         benchmarks = get_all_benchmarks()
 
--- a/common/benchmark_utils.py
+++ b/common/benchmark_utils.py
@@ -133,18 +133,18 @@ def get_all_benchmarks():
     return sorted(all_benchmarks)
 
 
-def get_coverage_benchmarks():
+def get_coverage_benchmarks(benchmarks=get_all_benchmarks()):
     """Returns the list of all coverage benchmarks."""
     return [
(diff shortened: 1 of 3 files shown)


common/benchmark_utils.py
```diff
@@ -133,18 +133,18 @@ def get_all_benchmarks():
     return sorted(all_benchmarks)
 
 
-def get_coverage_benchmarks():
+def get_coverage_benchmarks(benchmarks=get_all_benchmarks()):
```
jonathanmetzman, line 136: I don't think we should make this change.
There's weird things that happen when you call things in default arguments in python.
https://betterprogramming.pub/the-unexpected-workings-of-pythons-default-arguments-fe1a4addb1dc?gi=32b4d3d6d61f

jonathanmetzman: Instead you can do
```
def ...(benchmarks=None):
  if benchmarks is None:
    benchmarks = get_all_benchmarks()
```

common/benchmark_utils.py
```diff
@@ -163,3 +163,20 @@ def get_language(benchmark):
     """Returns the prorgamming language the benchmark was written in."""
     config = benchmark_config.get_config(benchmark)
     return config.get('language', 'c++')
+
+
+def are_benchmarks_mixed(benchmarks):
+    """"Returns True if benchmarks list
```
jonathanmetzman, line 169: Please fix the formatting of this docstring.

jonathanmetzman (commented):
Why isn't the existing check sufficient? https://github.com/google/fuzzbench/blob/dbcc1dfd8c466f71b6d67e7fd0eca962598182be/experiment/run_experiment.py#L747 
Why not improve it?

DonggeLiu:
> Why isn't the existing check sufficient?

I recall the current code only checks benchmark-type consistency at the end of experiments and cannot generate a report if inconsistent.

This warns users earlier and helps save time and computation power.

common/benchmark_utils.py
gustavogaldinoo: Just realized there's an extra apostrophe there.

Is there something else to fix, like the line break? I did the line break due to a lint error

experiment/run_experiment.py
```diff
@@ -743,6 +744,11 @@ def run_experiment_main(args=None):
             parser.error('Cannot enable options "custom_seed_corpus_dir" and '
                          '"oss_fuzz_corpus" at the same time')
 
+    if benchmark_utils.are_benchmarks_mixed(args.benchmarks):
+        raise ValidationError(
+            'Selected benchmarks are a mix between coverage '
+            'and bug benchmarks. This is currently not supported.')
```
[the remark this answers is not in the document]
gustavogaldinoo: Good point, thanks

common/benchmark_utils.py
jonathanmetzman: Right I mean the line break. It's formatted a bit weirdly. You should try to put as many chars as possible on a docstring line before breaking if youre continuing the same sentence. Only "benchmarks" needs to be on the second.
My editor has a shortcut for this btw: https://www.gnu.org/software/emacs/manual/html_node/emacs/Fill-Commands.html (fill-paragraph)
You probably shouldn't switch to my editor, but yours probably has something like this.
Also please end all sentences with a period.

jonathanmetzman:
> > Why isn't the existing check sufficient?
> 
> I recall the current code only checks benchmark-type consistency at the end of experiments and cannot generate a report if inconsistent.
> 
> This warns users earlier and helps save time and computation power.

For sure checking ahead of the experiment is important. But if we aren't doing this currently, there's a bug to track down here that this PR hasn't fixed right?

jonathanmetzman:
> > > Why isn't the existing check sufficient?
> > 
> > 
> > I recall the current code only checks benchmark-type consistency at the end of experiments and cannot generate a report if inconsistent.
> > This warns users earlier and helps save time and computation power.
> 
> For sure checking ahead of the experiment is important. But if we aren't doing this currently, there's a bug to track down here that this PR hasn't fixed right?

Ah actually I see where this might be fixed. Gustavo do you know why this didn't work previously?

DonggeLiu (approved):

common/benchmark_utils.py
gustavogaldinoo: Thanks, I am using Rewrap on vscode and its working now

jonathanmetzman (approved):
LGTM

Feel free to merge this.

PR #1993, opened 2024-05-29
Fix saturated corpus experiment requests

ndrewh:
The parser for experiment-requests.yml checks `oss_fuzz_corpus`:

https://github.com/google/fuzzbench/blob/421187b515cfaf4d7ad8bfe9bfa9512372dceaae/service/automatic_run_experiment.py#L64-L67

Experiment request docs say to use `oss-fuzz-corpus`, not `oss_fuzz_corpus`

https://github.com/google/fuzzbench/blob/421187b515cfaf4d7ad8bfe9bfa9512372dceaae/service/experiment-requests.yaml#L9

and all the past experiment requests are using `oss-fuzz-corpus`
https://github.com/google/fuzzbench/blob/421187b515cfaf4d7ad8bfe9bfa9512372dceaae/service/experiment-requests.yaml#L2273-L2276

PR #1995, opened 2024-06-10
Fix broken test for corpus experiment requests

ndrewh:


ndrewh:
Follow up to #1993

ndrewh:
@DonggeLiu

DonggeLiu:
Thanks!

DonggeLiu (approved):

DonggeLiu:
In addition to the testcases, did you happen to have a chance to check if there are other inconsistent usages elsewhere?
E.g., would these be related by any chance?
https://github.com/google/fuzzbench/blob/0f85b4a1843e022f323f8af0db7443b8d79544d0/experiment/scheduler.py#L777
https://github.com/google/fuzzbench/blob/0f85b4a1843e022f323f8af0db7443b8d79544d0/experiment/test_data/experiment-config.yaml#L34

Thanks!

ndrewh:
I think everything else is "fine". experiment_requests.yml is not consistent with the experiment config itself but that is not a "problem" necessarily but is certainly a future source of confusion.

If you prefer: instead of #1993 we can switch experiment_requests.yml to use the underscore instead, which would be consistent with the experiment configs. We would revert #1993 and instead just fix docs and probably re-write the old experiment requests to use the underscore name (to avoid errors where someone copy/pastes them). But as was pointed out earlier, I don't think experiment_requests.yml is actively used anymore, so it may be a moot point.

DonggeLiu:
Thanks @ndrewh!
It would be amazing if you could help us make the name consistent to avoid future confusion.
We don't hold a strong opinion on using `-` or `_`, the one that's the most convenient for you is the best for us : )

There is no need to rewrite the old experiment requests YAML: it will be not used/parsed/revived anytime soon.

Thanks!

ndrewh:
@DonggeLiu #1997

PR #2003, opened 2024-07-23
Update doc in running_a_local_experiment.md

tokatoka:
there is no more freetype2-2017 only freetype2_ftfuzzer


DonggeLiu (approved):
Thanks @tokatoka!

PR #2020, opened 2024-08-08
Archive coverage data alongside corpus archives

addisoncrump:
Currently, only corpora are saved in the archive and the summaries of coverage are provided at the end of the experiment. This change simply incorporates the saving of the coverage data snapshots next to the trial corpus snapshots.

PR #2021, opened 2024-08-12
Update libafl-based fuzzers

addisoncrump:
As requested in #2020.


addisoncrump:
@tokatoka Seems that aflrustrust is broken. Should we remove it?

addisoncrump:
Hm, wait, something strange is happening here. I tested `stb_stbi_read_fuzzer` with `make test-run-...` for all of the fuzzers I changed and it worked.

tokatoka:
> @tokatoka Seems that aflrustrust is broken. Should we remove it?

you should remove it. we have alternative from the GSoC student

addisoncrump:
Seems to be an llvm version compatibility issue...

addisoncrump:
Annoyingly, no, this works fine locally.

addisoncrump:
Just confirmed, this builds just fine locally. @DonggeLiu I think there is some bad cache associated with CI here, I have tested these locally and most are fine. `harfbuzz_hb-shape-fuzzer_17863b` still does not build, but I believe this bug benchmark to be too old to be compatible with modern clang due to an invalid function pointer cast. I'm investigating how to fix that.

tokatoka:
> harfbuzz_hb-shape-fuzzer_17863b still does not build

yes from what i saw, usually the projects with that commit hash does not build (but it's not used for cov benchmark either

DonggeLiu:
Thanks both!

DonggeLiu:
If you think the `Fuzzers+Benchmarks` CIs here are unreliable and blocking you, we can verify this via an experiment.
Use minimum values for these 2 lines will help speed up the experiment and save resource:

https://github.com/google/fuzzbench/blob/d8a77233f3919477648c98d2c8228b530b7036ba/service/experiment-config.yaml#L5-L6

E.g.,
```yaml
trials: 2
max_total_time: 1800  # 30 min to show two timestamps
```

PR #2034, opened 2024-08-19
Add PCRE2 as a fuzzing benchmark

addisoncrump:
PCRE2 is a complicated target which is prone to timeouts. This timeout-proneness is caused, in part, by the input representing *instructions* rather than flat data. Additionally, the state space of the program is significantly larger than the code coverage suggests. These two issues together make this a very hard target to fuzz, which has led to interesting problems in OSS-Fuzz and in local testing.

I think this target is interesting from a benchmarking perspective as it introduces complications which are not present in other targets in Fuzzbench and would like to add it to the standard benchmark suite. I have used the standard OSS-Fuzz integration script.

CC @PhilipHazel @alexdowad for input.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -36,6 +36,7 @@ jobs:
           - openh264_decoder_fuzzer
           - openssl_x509
           - openthread_ot-ip6-send-fuzzer
+          - pcre2_pcre2_fuzzer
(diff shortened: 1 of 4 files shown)


addisoncrump:
Seems to be breaking with a few standard fuzzers. Resolving now.

addisoncrump:
See #2035.

PR #2048, opened 2024-09-20
Fix the setup of Grimoire

renatahodovan:
The previous version of fuzzer.py was possibly copy-pasted from Nautilus and applied unexpected CLI setup. The patch fixes these issues.


renatahodovan:
Honestly, I'm not an expert on Grimoire, but I encountered this issue while trying to execute it. As far as I know, Grimoire doesn't require a grammar, and looking into the implementation of its CLI supports this [here](https://github.com/AFLplusplus/libafl_fuzzbench/blob/main/grimoire/src/lib.rs#L71).

@DonggeLiu @addisoncrump You seem familiar with LibAFL-based fuzzers. Could you please take a look at this?

renatahodovan:
@addisoncrump I'm not sure that my patch above is correct. Even if neither `-r`, nor `-g` are supported by the Grimoire implementation of libafl_fuzzbench, but while `-g` is reasonable since Grimoire doesn't use a grammar, but `-r` might be needed to avoid putting the internal representation of the test cases to the output corpus. But the `-r` support should be implemented in libafl_fuzzbench. Or am I wrong?

addisoncrump:
Hey, sorry for the delay on response. The repo you're pulling from is somewhat outdated, and I've not used that fuzzer. Instead, you might consider looking at (what used to be called) `libafl_text` from some revisions ago. We removed it because we found that grimoire was not relevant for the targets being benchmarked for the most part.

PR #2051, opened 2024-10-18
Add FOX fuzzer

addisoncrump:
This PR adds the FOX fuzzer, which showed [impressive performance over typical benchmarking periods](https://arxiv.org/pdf/2406.04517). Given this success, it should be added as one of the baselines for future experiments.

This will need an experiment run to establish the baseline. I've made a corresponding dummy change.


addisoncrump:
Oops, there are some fox-specific flags I missed. Doing a quick test run and patch.

addisoncrump:
@prashast Is there something I'm doing obviously wrong? Getting the error: `BUGG: inline table wrong, try to fix...`

Issue #1108, opened 2021-03-09
How fuzzbench detect bugs?


andreafioraldi:
I'm running some local bug experiments and I noticed that fuzzbench reports some bugs while AFL++ didn't find any crashes (looking at results/fuzzer-log.txt).

What is the explanation for that? Aren't you looking at the output folder of AFL++ for crashes?

inferno-chromium:
It does a libfuzzer merge run on entire corpus dir (with crashes subdir) from cloud storage (from that trial run)- https://github.com/google/fuzzbench/blob/78947df34aff867e97a64c89879f3fb37724e506/experiment/measurer/run_coverage.py#L52 and then it captures crashes from that. then it processes them one-by-one for signature/crash params - https://github.com/google/fuzzbench/blob/master/experiment/measurer/measure_manager.py#L512

andreafioraldi:
Ok so it is possible that a fuzzer doesn't detect a bug but it is reported in the report?
You should check crashes only in the crashes/ folder IMHO, but good to know this behavior because a was observing an empty crashes folder for a fuzzer without ASan but bugs reported in the report and I was confused.

andreafioraldi:
I still don't get why in the report I have 0 bugs for an AFL++ based fuzzer but when I look at results/fuzzer-log.txt there are at least 30 crashes.
Now it is possible that all these 30 crashes are duplicated, but then FuzzBench should report at least one bug for that fuzzer.

andreafioraldi:
AFL++ stire crashes in OUTPUT/default/crashes, is it possible that you are not executing the testcases in this dir when measuring crashes, but only from the queue?

andreafioraldi:
From what I'm understanding (correct me please) you extract the corpus tar.gz, the for AFL++ contains a directory struct of type corpus/default/{queue, crashes, hangs}, [here](https://github.com/google/fuzzbench/blob/master/experiment/measurer/measure_manager.py#L495), then you run libfuzzer merge [here](https://github.com/google/fuzzbench/blob/d8d1a982463057b446c9dd9f1e6ecd8853e62f44/experiment/measurer/run_coverage.py#L52).
This is ok as you said it executes the crashes subdir too, but I'm perplex about the mem limit and the timeout.

I'm hitting this problem on aspell, that seems a slow target (there are a lot of hanging testcases in AFL++) and so there is the possibility that libfuzzer is considering the crashing inputs found by AFL++ as timeouts. I'm right?

[...]

PR #2107, opened 2026-01-24
Catch fuzzbench measurer worker exception.

am009:
The measurer worker is launched using multiprocessing.Pool's apply_async. If there is any uncaught exception, the worker thread will die and stop to do any more work. Probably related to #1108

https://github.com/google/fuzzbench/blob/fe8c15e1edbf796121cd3558305a1e7473a8c749/experiment/measurer/measure_manager.py#L813


vanhauser-thc:
@jonathanmetzman ping :-)

PR #2117, opened 2026-05-17
Fix corpus archiving

q9982:
Fix corpus archiving to avoid `tarfile.ReadError: unexpected end of data` during experiments when files change while the tar archive is being created.

Corpus files can still change while the runner is creating a tar archive. Adding them directly from disk can therefore write a tar header for one size and then read a different amount of data, producing a truncated tar member.

Stage regular files in a temporary file before writing them to the tar archive. Non-regular entries only need their tar header, so avoid reading file contents for them.

### DIFF
--- a/experiment/runner.py
+++ b/experiment/runner.py
@@ -23,6 +23,7 @@ import shutil
 import subprocess
 import sys
 import tarfile
+import tempfile
 import threading
 import time
 import zipfile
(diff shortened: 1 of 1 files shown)


q9982:
We have reproduced the issue using a deliberately designed fuzzer; see [https://github.com/q9982/fuzzbench/commit/f889f7a3f41d3d452ea30aacd25dd432817aaad6](https://github.com/q9982/fuzzbench/commit/f889f7a3f41d3d452ea30aacd25dd432817aaad6)

